PR #116, opened 2020-07-20
WIP: Install should check for istio port access

tbeerbower:
The 3-install-verrazzano script now checks if the LB can reach the nginx ports.  We need the same test for the istio 80 & 443 ports.

This test currently fails!

For the Nginx ingress port check we do a curl which succeeds (with a 404) if things are okay or the curl fails if the rule for the port is missing from the seclist.

If I try to do a curl through the Istio ingress during install, the curl fails with 'curl: (56) Recv failure: Connection reset by peer' whether or not the seclist rule is set for the port.  

After I deploy the demo, obviously the curl succeeds.

I've looked into deploying a simple app during the install to do a kind of health check and then cleaning it up after the install.  Istio sample app may work ...
https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/

For the simple case (http) this only requires a few steps.

- create service
- create deployment
- create gateway

For https it also requires 

- generate cert and key
- create secret
- create gateway
- create virtual service

### DIFF
--- a/install/3-install-verrazzano.sh
+++ b/install/3-install-verrazzano.sh
@@ -45,9 +45,11 @@ function check_ingress_ports() {
 
       # Attempt to access the port on the $INGRESS_IP
       if [ $TARGETPORT == "https" ]; then
-        curl -k https://$INGRESS_IP:$PORT
+        ARGS=(-k https://$INGRESS_IP:$PORT)
+        call_curl 0 response http_code ARGS
       else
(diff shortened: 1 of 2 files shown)


tbeerbower:
No longer relevant.  Since this goes beyond adding a simple check, it's been decided that we shouldn't add the additional complexity to the install process.

PR #215, opened 2020-08-21
Validate Load Balancer External EndPoints exist prior to exiting Install

ddrechse:
The Validation that the gateways, nginx and istio, have a valid ip is at the end of the install process is because, sometimes, it can take some time for the LB to be constructed. If either gateway does not have a valid external endpoint, the install fails


ddrechse:
Need to push to new develop branch

PR #240, opened 2020-08-28
Update Rancher grep to be more namespace specific

davis-gomes:
Goals to fix:
- Rancher namespace grep was grabbing too many items
- delete finalizers created by rancher in all namespace deletions
- delete rancher within API before deleting resources

### DIFF
--- a/uninstall/uninstall-steps/1-uninstall-istio.sh
+++ b/uninstall/uninstall-steps/1-uninstall-istio.sh
@@ -111,6 +111,14 @@ function delete_secrets() {
 }
 
 function delete_istio_namepsace() {
+  local istio_ns_fin_res=("$(kubectl get namespaces --no-headers -o custom-columns=":metadata.name" \
+    | grep -E 'istio-system' || true)")
+
+  printf "%s\n" "${istio_ns_fin_res[@]}" \
+    | awk '{print $1}' \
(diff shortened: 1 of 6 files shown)


mgianatagh:
You will need to change the target of this merge to be the develop branch

uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -111,10 +111,43 @@ function delete_secrets() {
...
   log "Deleting istio-system namespace"
   kubectl delete namespace istio-system --ignore-not-found=true || return $?
 }
 
+function finalize() {
+  # Grab all leftover Helm repos and delete resources
+  log "Deleting Helm repos"
+  helm repo ls || true \
+    | awk 'NR>1 {print $1}' \
+    | xargs -I name helm repo remove name \
```
mgianatagh, line 131: Could we be deleting helm repos that we didn't add?

davis-gomes: Yea this could happen. I think it was brought up that we don't necessarily have to delete help repos because they aren't a cluster resource. It might just be easier to leave all helm repo's as is. Alternatively, we could look for our specific repos and delete them individually.

davis-gomes: I made a task to fix this.

mgianatagh (approved):

PR #245, opened 2020-09-01
Add Helm version check to installer

desagar:


install/1-install-istio.sh
```diff
@@ -214,11 +215,14 @@ function verify_ocr_secret()
...
 
 function check_kube_version {
-    kubeVer=$(kubectl version -o json)
-    servVer=$(echo $kubeVer | jq -r '.serverVersion.gitVersion')
-    major=$(echo $kubeVer | jq -r '.serverVersion.major')
-    minor=$(echo $kubeVer | jq -r '.serverVersion.minor')
-    patch=$(echo $servVer | cut -d'.' -f 3)
+    local kubeVer=$(kubectl version -o json)
+    log "------Begin Kubernetes Version Info----"
+    log $kubeVer
```
gilbode, line 220: kubectl version -o json prints out a nicely formatted multi-line json.  saving it to a variable and echoing it results in a single line unreadable blob of json.  would be nice if you could preserve the pretty printing (and perhaps write yaml to the console instead of json).

install/1-install-istio.sh
```diff
@@ -73,6 +73,7 @@ function install_istio()
     helm repo add istio.io https://storage.googleapis.com/istio-release/releases/${ISTIO_VERSION}/charts || return $?
 
     log "Fetch istio charts for istio and istio-init"
+    log "Using TMP_DIR $TMP_DIR"
```
gilbode, line 76: Of what use is this message given TMP_DIR is deleted at the end of the script?

install/1-install-istio.sh
```diff
@@ -234,6 +238,22 @@ function check_kube_version {
     fi
 }
 
+function check_helm_version {
+    local helmVer=$(helm version --short | cut -d':' -f2 | tr -d " ")
+    log "Helm Version is $helmVer"
```
gilbode, line 243: Minor comment but Version should not be capitalized, similar with the words Major & Minor below, just looks strange.

bobdonat:
Should the install readme be updated to reflect the required versions of Helm.

install/1-install-istio.sh
```diff
@@ -214,11 +215,14 @@ function verify_ocr_secret()
 }
 
 function check_kube_version {
-    kubeVer=$(kubectl version -o json)
-    servVer=$(echo $kubeVer | jq -r '.serverVersion.gitVersion')
-    major=$(echo $kubeVer | jq -r '.serverVersion.major')
-    minor=$(echo $kubeVer | jq -r '.serverVersion.minor')
-    patch=$(echo $servVer | cut -d'.' -f 3)
+    local kubeVer=$(kubectl version -o json)
```
gilbode, line 218: This doesn't work if your kubeconfig is connected to an unreachable server:

```
kubectl version -o json
{
  "clientVersion": {
    "major": "1",
    "minor": "17",
    "gitVersion": "v1.17.2",
    "gitCommit": "59603c6e503c87169aea6106f57b9f242f64df89",
    "gitTreeState": "clean",
    "buildDate": "2020-01-23T14:22:01Z",
    "goVersion": "go1.13.6",
    "compiler": "gc",
    "platform": "darwin/amd64"
  }
}
The connection to the server localhost:8080 was refused - did you specify the right host or port?
echo $?
1
```

install/1-install-istio.sh
desagar: TMP_DIR is not deleted if the helm fetch /helm template operations fail.

install/1-install-istio.sh
desagar: done

install/1-install-istio.sh
desagar: Yep I knew that but had forgotten about it. Just fixed that issue.

desagar:
> Should the install readme be updated to reflect the required versions of Helm.

done.

install/1-install-istio.sh
kminder: Could consider using debug instead of log but that probably means we will never have the info when things fail.

desagar: Yeah - this came out of the fact that the user reported a failure, and there was practically no information in the log. If we use debug we will have to turn on debug all the time, which defeats the purpose.

gilbode: How is it you're saying TMP_DIR is not deleted?  it's done in an exit trap.

https://github.com/verrazzano/verrazzano/blob/fc7e5d5f1a208b8b15b7e8ac5b2e974c71789efd/install/1-install-istio.sh#L19

install/1-install-istio.sh
```diff
@@ -214,11 +215,19 @@ function verify_ocr_secret()
...
+    log "------Begin Kubernetes Version Info----"
+    log "$kubeVer"
+    log "------End Kubernetes Version Info----"
+    local servVer=$(echo $kubeVer | jq -r '.serverVersion.gitVersion')
```
gilbode, line 222: the exit code will alway be 0 here b/c you'll get the exit code of local not the command substitution.  You need to declare the variable separate from the assignment for the exit code check to work.

install/1-install-istio.sh
desagar: I did not read closely - you're right, it's of no use. Removed the log line.

install/1-install-istio.sh
desagar: I tried declaring the variable separate from assignment, and the exit code is still 0 when there is a failure. 
I think the exit code must be the result of assignment. The only way I could get a non-zero exit code on failure is to re-run the command without an assignment.

I removed the exit code check altogether. The other conditions are sufficient to fail when the server version is not available....jq returns the string "null" in that case.

PR #247, opened 2020-09-02
Verify that Service LBs have been successfully created prior to exiti…

ddrechse:
Adds code to the installer that verifies both the nginx and istio-ingress gateways have valid load balancers associated with them. This is based upon an acceptance test failure due to no lb associated with the istio-ingress gateway.

### DIFF
--- a/install/4-install-keycloak.sh
+++ b/install/4-install-keycloak.sh
@@ -252,7 +252,7 @@ consoleout "Password: kubectl get secret --namespace keycloak keycloak-http -o j
 if [ ${CLUSTER_TYPE} == "OKE" ] -a -z "$(kubectl get services -n istio-system istio-ingressgateway -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ] 2> /dev/null
 then
   consoleout
-  consoleout "WARNING: Istio Ingressgateway does not have a valid external ip assigned yet. Public access to deployed applications will not work."
+  consoleout "WARNING: istio-ingressgateway service does not have a valid external IP assigned yet. Public access to deployed applications will not work."
   consoleout "Use the following command to check if an External IP has been assigned to the gateway."
   consoleout "kubectl get svc istio-ingressgateway -n istio-system"
 fi
--- a/install/4-install-keycloak.sh
+++ b/install/4-install-keycloak.sh
@@ -249,10 +249,10 @@ consoleout
 consoleout "Keycloak - https://keycloak.${ENV_NAME}.${DNS_SUFFIX}"
 consoleout "User: keycloakadmin"
 consoleout "Password: kubectl get secret --namespace keycloak keycloak-http -o jsonpath={.data.password} | base64 --decode; echo"
-if  [ -z "$(kubectl get svc istio-ingressgateway -n istio-system -o json | jq -r '.status.loadBalancer.ingress[].ip')" ]
+if [ ${CLUSTER_TYPE} == "OKE" ] -a -z "$(kubectl get services -n istio-system istio-ingressgateway -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ] 2> /dev/null
 then
   consoleout
-  consoleout "WARNING: Istio Ingressgateway did not have a valid external ip assigned yet. Public access to deployed applications will not work."
-  consoleout "Use the following command to check if an Exteranl IP has been assigned to the gateway."
+  consoleout "WARNING: Istio Ingressgateway does not have a valid external ip assigned yet. Public access to deployed applications will not work."
+  consoleout "Use the following command to check if an External IP has been assigned to the gateway."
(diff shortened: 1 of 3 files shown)


bobdonat:
Will this work with OLCNE?  I thought I saw in some readme or doc that a NodePort could be used instead of LoadBalancer.

install/4-install-keycloak.sh
```diff
@@ -224,6 +233,7 @@ action "Installing MySQL" install_mysql
   fi
 
 action "Installing Keycloak" install_keycloak || exit 1
+action "Verifying Service IPs" verify_service_ips || exit 1
```
mgianatagh, line 236: I feel these checks should occur sooner.  Could we check the istio-ingressgateway after the install of istio?  And check the ingress-nginx controller after the install of ingress-nginx?

ddrechse: Here's the thing, you aren't guaranteed a LB immediately and that would make the install appear to hang

install/4-install-keycloak.sh
```diff
@@ -249,3 +249,10 @@ consoleout
 consoleout "Keycloak - https://keycloak.${ENV_NAME}.${DNS_SUFFIX}"
 consoleout "User: keycloakadmin"
 consoleout "Password: kubectl get secret --namespace keycloak keycloak-http -o jsonpath={.data.password} | base64 --decode; echo"
+if  [ -z "$(kubectl get svc istio-ingressgateway -n istio-system -o json | jq -r '.status.loadBalancer.ingress[].ip')" ]
+then
+  consoleout
+  consoleout "WARNING: Istio Ingressgateway did not have a valid external ip assigned yet. Public access to deployed applications will not work."
```
mgianatagh, line 255: "does not" ?

jmaron99, line 255: *does* not...

bobdonat: Should the text say "istio-ingressgateway service ..." to be more specific
 "ip" should be capitalized.

install/common.sh
```diff
@@ -42,7 +42,7 @@ function wait_for_ingress_ip() {
   done
   if [ "$retries" -ge 10 ] ; then
     log "An error occurred - ingress $ingress_name in namespace $namespace did not have an IP address"
-    exit 1
+    return 1
```
desagar, line 45: If you're changing this to return 1, then there is a call to this function in 3-install-verrazzano that must be changed to check the return status and exit if failed (i.e.) `wait_for_ingress_ip ...args... || exit 1`

install/4-install-keycloak.sh
desagar: Should we add a "This could be because the creation of an OCI load balancer has not succeeded." ?

install/4-install-keycloak.sh
```diff
@@ -249,3 +249,10 @@ consoleout
 consoleout "Keycloak - https://keycloak.${ENV_NAME}.${DNS_SUFFIX}"
 consoleout "User: keycloakadmin"
 consoleout "Password: kubectl get secret --namespace keycloak keycloak-http -o jsonpath={.data.password} | base64 --decode; echo"
+if [ ${CLUSTER_TYPE} == "OKE" ] -a -z "$(kubectl get services -n istio-system istio-ingressgateway -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ] 2> /dev/null
```
desagar, line 252: Looks like a syntax error unless I'm not reading this right. You seem to have an extra closing `]`. Does this run?
You should be able to do:
```
if [ "${CLUSTER_TYPE} == "OKE" ] && [ -z "$(kubectl .....etc...)" ]; then
```

install/common.sh
ddrechse: Check it now

desagar (approved):

install/4-install-keycloak.sh
```diff
@@ -249,3 +249,10 @@ consoleout
 consoleout "Keycloak - https://keycloak.${ENV_NAME}.${DNS_SUFFIX}"
 consoleout "User: keycloakadmin"
 consoleout "Password: kubectl get secret --namespace keycloak keycloak-http -o jsonpath={.data.password} | base64 --decode; echo"
+if [ ${CLUSTER_TYPE} == "OKE" -a -z "$(kubectl get services -n istio-system istio-ingressgateway -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ] 2> /dev/null
+then
+  consoleout
+  consoleout "WARNING: istio-ingressgateway service does not have a valid external IP assigned yet. Public access to deployed applications will not work."
+  consoleout "Use the following command to check if an External IP has been assigned to the gateway."
```
bobdonat, line 256: I think the text should be consistent - either "external IP" or "External IP" - "external IP" I think is correct.

PR #249, opened 2020-09-03
Docs: OLCNE requires load balancers

mark-au:
A manual DNS mode requires load balancers, the removed text is not longer true.
Signed-off-by: Mark Cram <mark.cram@oracle.com>


desagar (approved):

desagar:
FYI switched the target branch to `develop`

PR #250, opened 2020-09-03
Find alternative to grep to find resources

tbeerbower:
Currently, some of the resources in the uninstall are grabbed by grepping for an aspect of their descriptions. Unfortunately, grep returns an error status of 1 when nothing is found. This is a functionality we want when performing the uninstall. Thus, we need a command that will grab these items but still return 0 when nothing is found to improve the error handling. then, the two part delete commands can be merged into one command again.

As an example, in 1-uninstall-istio.sh function delete_secrets() the following code block has to be written in 2 parts because `grep "istio.io" || true` will always be true.

 ```
 # delete secrets left over in kube-system
  local secret_res=("$(kubectl get secrets -n kube-system --no-headers -o custom-columns=":metadata.name,:metadata.annotations" \
  | grep "istio.io" || true)")

  printf "%s\n" "${secret_res[@]}" \
  | awk '{print $1}' \
  | xargs kubectl delete secret -n kube-system \
  || return $? # return on pipefail
```
By performing the grep functionality in the awk call we should be able to replace the above with the following code block (awk will return 0 if no match is found).
```
  # delete secrets left over in kube-system
  kubectl get secrets -n kube-system --no-headers -o custom-columns=":metadata.name,:metadata.annotations" 
  | awk '/istio.io/ {print $1}' \
  | xargs kubectl delete secret -n kube-system \
  || return $? # return on pipefail
```

### DIFF
--- a/uninstall/uninstall-steps/0-uninstall-applications.sh
+++ b/uninstall/uninstall-steps/0-uninstall-applications.sh
@@ -43,23 +43,21 @@ function initializing_uninstall {
 }
 
 function delete_bindings {
-  binding_crd=$(kubectl get crd | grep "verrazzanobinding" || true)
-  if [ -z "$binding_crd" ] ; then
-    return
+  crd=$(kubectl get crd) || return $?
+  if echo $crd | grep -q "verrazzanobinding" ; then
+    kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
+      | xargs kubectl delete VerrazzanoBindings \
+      || return $? # return on pipefail
   fi
-  kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
-    | xargs kubectl delete VerrazzanoBindings \
-    || return $? # return on pipefail
 }
 
 function delete_models {
-  model_crd=$(kubectl get crd | grep "verrazzanomodel" || true)
-  if [ -z "$binding_crd" ] ; then
-    return
+  crd=$(kubectl get crd) || return $?
+  if echo $crd | grep -q "verrazzanomodel" ; then
+    kubectl get VerrazzanoModels --no-headers -o custom-columns=":metadata.name" \
+      | xargs kubectl delete VerrazzanoModels \
+      || return $? # return on pipefail
   fi
-  kubectl get VerrazzanoModels --no-headers -o custom-columns=":metadata.name" \
-    | xargs kubectl delete VerrazzanoModels \
-    || return $? # return on pipefail
 }
 
 action "Initializing Uninstall" initializing_uninstall || exit 1
--- a/uninstall/uninstall-steps/0-uninstall-applications.sh
+++ b/uninstall/uninstall-steps/0-uninstall-applications.sh
@@ -43,21 +43,17 @@ function initializing_uninstall {
 }
 
 function delete_bindings {
-  crd=$(kubectl get crd) || return $?
-  if echo $crd | grep -q "verrazzanobinding" ; then
-    kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
-      | xargs kubectl delete VerrazzanoBindings \
-      || return $? # return on pipefail
-  fi
+  kubectl get crd verrazzanobindings.verrazzano.io || return 0
+  kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
+    | xargs kubectl delete VerrazzanoBindings \
+    || return $? # return on pipefail
 }
 
 function delete_models {
-  crd=$(kubectl get crd) || return $?
-  if echo $crd | grep -q "verrazzanomodel" ; then
-    kubectl get VerrazzanoModels --no-headers -o custom-columns=":metadata.name" \
-      | xargs kubectl delete VerrazzanoModels \
-      || return $? # return on pipefail
-  fi
+  kubectl get crd verrazzanomodels.verrazzano.io || return 0
+  kubectl get VerrazzanoModels --no-headers -o custom-columns=":metadata.name" \
+    | xargs kubectl delete VerrazzanoModels \
+    || return $? # return on pipefail
 }
 
 action "Initializing Uninstall" initializing_uninstall || exit 1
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: uninstall/uninstall-steps/0-uninstall-applications.sh [hunk 1]
```diff
@@ -43,23 +43,21 @@ function initializing_uninstall {
 }
 
 function delete_bindings {
-  binding_crd=$(kubectl get crd | grep "verrazzanobinding" || true)
-  if [ -z "$binding_crd" ] ; then
-    return
+  crd=$(kubectl get crd) || return $?
```
desagar, line 46: Shouldn't this be `kubectl get crd verrazzanobindings.verrazzano.io`? That way we verify the binding crd is there before we proceed.

uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -43,23 +43,21 @@ function initializing_uninstall {
...
+      | xargs kubectl delete VerrazzanoBindings \
+      || return $? # return on pipefail
   fi
-  kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
-    | xargs kubectl delete VerrazzanoBindings \
-    || return $? # return on pipefail
 }
 
 function delete_models {
-  model_crd=$(kubectl get crd | grep "verrazzanomodel" || true)
```
desagar, line 56: same q. as for binding

davis-gomes:
I made some changes to the way that Istio uninstalls, and I iust merged it to the develop branch. When you resolve the conflicts, please make sure to update the `grep`s for that file as well.

uninstall/uninstall-steps/0-uninstall-applications.sh
tbeerbower: @desagar , thanks for reviewing.  This was just a refactoring of the existing logic that was trying to achieve what you suggest... `verify the binding crd is there before we proceed.`
I think the idea is that if the `kubectl` command fails (returns non-zero) for some reason then the function should exit and return that code.  So, it tries to get all of the CRDs and then greps for the verrazzano binding.  It's a subtle difference but with what you suggest ('kubectl get crd verrazzanobindings.verrazzano.io') if a non-zero code is returned we don't know if it happened because the v8o binding CRD doesn't exist or some other reason.  I don't really know if it's a real concern and your way looks cleaner, I think.  @davis-gomes what do you think?

davis-gomes: I believe we can say `binding_crd=$(kubectl get crd verrazzanobindings.verrazzano.io) || return 0` as this checks to see if the crd exists for verrazzano binding and returns cleanly if it does not.

tbeerbower: @davis-gomes I guess we don't need `binding_crd`, so this should do I think...
```
  kubectl get crd verrazzanobindings.verrazzano.io || return 0
  kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
    | xargs kubectl delete VerrazzanoBindings \
    || return $? # return on pipefail
```

desagar: I agree with the suggestions above @tbeerbower and @davis-gomes - this definitely looks cleaner Tom.

tbeerbower: @desagar made the changes as shown above.

uninstall/uninstall-steps/0-uninstall-applications.sh
tbeerbower: @desagar made the changes as shown above.

tbeerbower:
@davis-gomes I've rebased with the latest develop and updated the greps.

uninstall/uninstall-steps/0-uninstall-applications.sh
desagar: An afterthought - theoretically if there is some other issue (network blip, etc) at this point in the uninstall, the `kubectl get crd verrazzanobindings.verrazzano.io` could fail for a reason other than the absence of the VZ CRD. In that unlikely event, we might leave the CRDs lying around. 
But I'm not sure how you'd guard against that here - I guess you could do something a bit clunkier like `kubectl get crd && kubectl get crd verrazzanobindings.verrazzano.io || return 0`, but it seems like overkill. Thoughts?

davis-gomes: The chance that there will be a blip in the network for such a short period of time without the script failing on another command seems super slim to me. I think any errors with verifying vb's will most likely be addressed by previous and subsequent calls to kubectl. It seems a little bit overkill to me to add another command.

tbeerbower: Yeah that's what I meant when I said it's a subtle difference but I didn't know if it's a real concern.  I think that what I had originally is better than the two kubectl commands.  The single kubectl and grep would be faster and clearer.  Anyway, I think that `kubectl get crd verrazzanobindings.verrazzano.io || return 0` is the cleanest solution and the others are a bit of overkill.

uninstall/uninstall-steps/2-uninstall-system-components.sh
```diff
@@ -165,32 +133,23 @@ function delete_rancher() {
...
     || return $? # return on pipefail
 
   # delete cattle namespaces
-  local rancher_ns_res=("$(kubectl get namespaces --no-headers -o custom-columns=":metadata.name" \
-    | grep -E 'cattle-|local|p-|user-' || true)")
-
-  printf "%s\n" "${rancher_ns_res[@]}" \
-    | awk '{print $1}' \
+  kubectl get namespaces --no-headers -o custom-columns=":metadata.name" \
+    | awk '/cattle-|local|p-|user-/ {print $1}' \
```
desagar, line 143: I thought at some point we had made a change to get rid of this matching by `p-` etc. by checking the labels on the rancher namespaces. Did that not work and get removed?

davis-gomes: That had been fixed. Unfortunately the patch was inconsistent and would still leave these resources lying around in some cases. I had to revert back to the previous way of collecting them.

desagar: Ok I guess we should revisit that - matching by a name like `p-` is a bit risky IMO. But that's got nothing to do with this PR :)

desagar (approved):

PR #252, opened 2020-09-03
Clean up error and logging output

davis-gomes:


uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -13,13 +13,13 @@ set -o pipefail
 function initializing_uninstall {
   # Deleting rancher through API
   log "Deleting Rancher through API"
-  rancher_host_name="$(kubectl get ingress -n cattle-system --no-headers -o custom-columns=":spec.rules[0].host")" || return $?
```
desagar, line 16: are you sure the semicolon will be parsed by Mac and all Unix flavors as `(kubectl command) || (print error; return $?)`, and not `(kubectl command || print error) ; return $?`

davis-gomes: I am not 100% sure. Do you know a way to ensure that this happens?

desagar: On my Mac, I see this behavior - the second case is problematic and indicates the stuff after the semicolon will happen even if the command succeeds.

```
First Command does not succeed:

ls hello || echo one;echo two
ls: hello: No such file or directory
one
two
```

```
First command succeeds: 
ls server.js || echo one;echo two

server.js
two
```

desagar: parentheses around the stuff to the right of the `||` seems to work but I am not sure if it's ok to use. Needs a bit of research into how Unix treats parens (I know in some cases it is treated as a subshell)

davis-gomes: Okay, thanks. I will look into this.

desagar: Also, I just noticed you are returning $?, which will be the exit status of the `error` function call, not the original kubectl command

davis-gomes: added a function to fix these errors.

uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -46,17 +46,19 @@ function initializing_uninstall {
 }
 
 function delete_bindings {
+  log "Deleting VerrazzanoBindings"
   kubectl get crd verrazzanobindings.verrazzano.io || return 0
   kubectl get VerrazzanoBindings --no-headers -o custom-columns=":metadata.name" \
     | xargsr kubectl delete VerrazzanoBindings \
```
desagar, line 52: is there an xargsr command or is this a typo?

davis-gomes: There is a new command from another PR to handle how xargs works on Mac vs Linux.

uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -20,73 +20,75 @@ CONFIG_DIR=$INSTALL_DIR/config
...
+    || err_return $? "Could not delete ClusterRoles from Istio" || return $? # return on pipefail
 }
 
 function delete_secrets() {
   # Delete istio.default in all namespaces
-  log "Collecting istio secrets for deletion"
-  kubectl delete secret istio.default --ignore-not-found=true || return $?
-  kubectl delete secret istio.default -n kube-public --ignore-not-found=true || return $?
-  kubectl delete secret istio.default -n kube-node-lease --ignore-not-found=true || return $?
+  log "Retrieveing istio secrets for deletion"
```
desagar, line 57: Spelling :-) 
"Retrieving"

uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -20,73 +20,75 @@ CONFIG_DIR=$INSTALL_DIR/config
...
+  kubectl delete secret istio.default --ignore-not-found=true || err_return $? "Could not delete secret from Istio in namespace default" || return $?
+  kubectl delete secret istio.default -n kube-public --ignore-not-found=true || err_return $? "Could not delete secret from Istio in namespace kube-public" || return $?
+  kubectl delete secret istio.default -n kube-node-lease --ignore-not-found=true || err_return $? "Could not delete secret from Istio in namespace kuce-node-lease" || return $?
```
desagar, line 60: typo => message says "kuce-node-lease" instead of "kube-node-lease"

davis-gomes: This has been fixed.

uninstall/uninstall-steps/1-uninstall-istio.sh
davis-gomes: Find and replace failed me! I just updated this.

desagar (dismissed):

desagar (approved):

PR #253, opened 2020-09-04
WIP:  Change Robert Books to use Coherence Extend

pfmackin:
Change the model file so that Robert Helidon is no longer a member of the Coherence cluster, but now accesses the cluster using extend.  This required changes to the model CRD as follows:

1. Allow Helidon ENV vars in the model. This is needed to specify coherence cache config
2. Specify Coherence extend port

The Verrazzano operator was changed to process the new model fields and also to disable TLS for Helidon to Coherence extend TCP connection.


install/chart/values.yaml
```diff
@@ -13,7 +13,7 @@ docker:
 verrazzanoOperator:
   name: verrazzano-operator
   imageName: verrazzano-operator
-  imageVersion: v0.0.88-71a55ef-4
+  imageVersion: v0.0.90
```
mgianatagh, line 16: The docker tag generated by the v-o build appends a short-sha and build number.  This tag will not work.  Should be `v0.0.90-de60bf8-1`

mgianatagh (approved):

PR #255, opened 2020-09-04
Find clean way to delete helm repo elements.

tbeerbower:
In finalize() ...

  ```helm repo ls \
    | awk '/istio.io|stable|jetstack|rancher-stable|codecentric/ {print $1}' \
    | xargsr -I name helm repo remove name \
    || return $?
```

```
helm repo ls
Error: no repositories to show

echo $?
1
```
gives an error output of 1 when no elements are present. This causes the pipeline to fail. This task will be completed when the uninstall avoids non-zero output with no helm repo elements, and deletes elements when they are present in the repo.


davis-gomes:
lgtm

davis-gomes:
lgtm

PR #256, opened 2020-09-08
Fix the uninstall to delete cluster roles

jmaron99:
Simple fix to remove the intended resource (cluster role)

### DIFF
--- a/uninstall/uninstall-steps/1-uninstall-istio.sh
+++ b/uninstall/uninstall-steps/1-uninstall-istio.sh
@@ -46,7 +46,7 @@ function uninstall_istio() {
 
   # delete istio cluster roles
   log "Deleting Istio Cluster Roles"
-  kubectl get clusterrolebinding --no-headers -o custom-columns=":metadata.name" \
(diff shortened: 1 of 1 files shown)


uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -103,4 +103,4 @@ function finalize() {
 action "Deleting Istio Components" uninstall_istio || exit 1
 action "Deleting Istio Secrets" delete_secrets || exit 1
 action "Deleting Istio Namespace" delete_istio_namepsace || exit 1
-action "Finalizing Uninstall" finalize || exit 1
\ No newline at end of file
+action "Finalizing Uninstall" finalize || exit 1
```
davis-gomes: What is this change?

jmaron99: I'm not sure how it was picked up, but it looks like when I pulled from develop the finalize action was re-added?  Again, not certain how that happened?

davis-gomes (approved):

PR #261, opened 2020-09-09
Timeout for Rancher API deletion

davis-gomes:


uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -32,7 +32,9 @@ function initializing_uninstall {
     if [ "$status" != 200 ] && [ "$status" != 404 ] ; then
       return 1
     fi
-    while [ true ] ; do
+    local time_out=30
```
desagar, line 35: this is a nitpick, but technically this isn't a timeout, since timeouts are about waiting a specific amount of time. It's a "max_retries"

davis-gomes: I agree, fixed!

desagar (approved):

PR #263, opened 2020-09-09
Disable KinD support in installer

desagar:
Installer will fail fast if CLUSTER_TYPE is not OKE or OLCNE. KIND is no longer an allowed option.
Removed install-kind.sh and scripts to create/delete KinD cluster.


mgianatagh:
Should we also remove `install/config/images`?

bobdonat:
remove load-images.sh too?

desagar:
Good point, @mgianatagh and @bobdonat - removed install/config/images and load-images.sh as well

bobdonat:
There are still refs to KIND in install-hello-world.sh and install-sock-shop.sh

desagar:
Also removed all special handling for KinD in the install scripts.

desagar:
> There are still refs to KIND in install-hello-world.sh and install-sock-shop.sh

Thanks Bob.

desagar:
Removed KIND references from application install scripts install-hello-world.sh and install-sock-shop.sh

markxnelson (approved):

PR #273, opened 2020-09-11
WIP: Initial groundwork to transition docker images to ghcr.io

mgianatagh:
Initial changes to lay groundwork for moving the Verrazzano docker images into ghcr.io.

- Support an optional imagePullSecret named `verrazzano-container-registry`.  If this secret is detected then all deployments will use it.  The use case for this is when the docker images are private.

- Pull the verrazzano-operator image from ghcr.io.  The remainder of the images will be transitioned in subsequent pull requests.


install/common.sh
```diff
@@ -133,6 +133,21 @@ function validate_environment_name {
   return 0
 }
 
+# Check if the optional global registry secret exists
+function check_registry_secret_exists() {
+  kubectl get secret ${GLOBAL_REGISTRY_SECRET} -n default > /dev/null 2>&1
+  echo $?
```
markxnelson, line 139: not 100% sure that this will do what you want?

markxnelson (dismissed):

install/common.sh
mgianatagh: it does, after many iterations of testing :-(

mgianatagh: I wanted to get this to work, which it now does.
`REGISTRY_SECRET_EXISTS=$(check_registry_secret_exists)`

gilbode:
Looks OK but I find the name GLOBAL_REGISTRY_SECRET confusing as to what it is used for.

install/1-install-istio.sh
```diff
@@ -72,23 +72,32 @@ function install_istio()
     helm fetch istio.io/istio --untar=true --untardir=$TMP_DIR || return $?
     helm fetch istio.io/istio-init --untar=true --untardir=$TMP_DIR || return $?
 
+    EXTRA_ISTIO_ARGUMENTS=""
+    if [ ${REGISTRY_SECRET_EXISTS} == "0" ]; then
```
bobdonat, line 76: I would expect this value to equal "1" or "true" instead of "0" to set the extra arguments.  Is there something I am missing here?

mgianatagh: I was basing this on the response from the kubectl command - which returns "0" for success.  I could change the method to return "true" for "0" and "false" for "1".

mgianatagh: I am changing the code to test for "TRUE", that will be more readable

bobdonat (approved):

PR #276, opened 2020-09-11
Use Coherence extend for bobby

pfmackin:
Change the model file to use bobbys helidon and bobbys coherence to communicate via coherence extend.  This means that bobby helidon will no longer be a member of the coherence cluster, but rather will access the cache data over the extend protocol.


examples/bobs-books/bobs-books-model.yaml
```diff
@@ -81,9 +81,15 @@ spec:
               datasourceName: books
   helidonApplications:
     - name: "bobbys-helidon-stock-application"
-      image: "container-registry.oracle.com/verrazzano/example-bobbys-helidon-stock-application:0.1.10-3-e5ae893-124"
+      image: "ghcr.io/verrazzano/example-bobbys-helidon-stock-application:0.1.10-3-6e5f030-129"
       imagePullSecrets:
         - name: ocr
```
mgianatagh, line 86: the ocr secret can be removed

examples/bobs-books/bobs-books-model.yaml
```diff
@@ -81,9 +81,15 @@ spec:
...
+        - name: github-packages
+      env:
```
mgianatagh, line 88: we may want a more generic name than `github-packages`.  And the images are actually in GitHub container registry.

I have suggested the name be `verrazzano-container-registry`

mgianatagh:
The comments I left could be handled later

mgianatagh (approved):

PR #279, opened 2020-09-14
Remove dependency of OCR secret from the Verrazzano installation

mgianatagh:
Remove the dependency of create a secret named `ocr` from the verrazzano installation.  A secret of that name is only required if you install one of the sample applications.  The sample includes separate instructions for creating dependent secrets.


install/README.md
```diff
@@ -37,17 +37,7 @@ The following software must be installed on your system.
...
-
-```
-   kubectl create secret docker-registry ocr \
-       --docker-username=<username> \
-       --docker-password=<password> \
-       --docker-server=container-registry.oracle.com
-```
-
-### 3. Do the install
+### 2. Do the install
```
bobdonat, line 40: Do steps need to be renumbered after this step since the original step 2 has been removed.

mgianatagh: Yes, not sure how I missed those.

mgianatagh: I pushed the fix

markxnelson (approved):
LGTM

bobdonat (approved):

PR #283, opened 2020-09-17
test v8o-operator change VZ-1250

tbeerbower:


tbeerbower:
Acceptance test check.

PR #284, opened 2020-09-17
JIRA:VZ-1214 github action to generate pagerduty incident when an issue is opened

jmaron99:
When an issue is opened or reopened in the verrazzano repo the action will be triggered and generate a pagerduty incident to alert the on-call engineer

### DIFF
--- a/.github/workflows/pagerduty.yml
+++ b/.github/workflows/pagerduty.yml
@@ -1,0 +1,15 @@
+on:
+  issues:
+    types: [opened, reopened]
+jobs:
+  pagerduty:
+    runs-on: ubuntu-latest
+    steps:
+      - name: web-request
+        uses: satak/webrequest-action@master
+        with:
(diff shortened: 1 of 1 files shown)


desagar (approved):

.github/workflows/pagerduty.yml
```diff
@@ -0,0 +1,15 @@
+on:
+  issues:
+    types: [opened, reopened]
+jobs:
+  pagerduty:
+    runs-on: ubuntu-latest
+    steps:
+      - name: web-request
+        uses: satak/webrequest-action@master
```
desagar, line 9: Looks good to a pagerduty novice like me, but what is this satak/webrequest-action@master?

jmaron99: It's a workflow action that can be "used".  In this case it's a web request action to which we can pass the payload, headers, url etc in order to invoke the REST request to trigger the pagerduty incident

PR #308, opened 2020-10-01
Upgrade WLS Operator to 3.0.2

ddrechse:


install/chart/values.yaml
```diff
@@ -15,12 +15,12 @@ verrazzanoOperator:
   imageVersion: v0.0.99-07df6c6-1
   cohMicroImage: ghcr.io/verrazzano/verrazzano-coh-cluster-operator:v0.0.10
   helidonMicroImage: ghcr.io/verrazzano/verrazzano-helidon-app-operator:v0.0.9
-  wlsMicroImage: ghcr.io/verrazzano/verrazzano-wko-operator:v0.0.12
+  wlsMicroImage: ghcr.io/verrazzano/verrazzano-wko-operator-jenkins:ddd10abbae8886a49cb70b36e1834b661c67950b
```
mgianatagh, line 18: We cannot merge -jenkins references into the develop or master branches, those variants are private images and would require everyone to set an impagePullSecret to access.

ddrechse: Oh yeah, thats right, I have another out there, let me go look
This is from the test it all before you merge thread

mgianatagh (approved):

PR #313, opened 2020-10-06
Promote verrazzano operator v.0.0.102-24f5934-2

jmaron99:
Build:  https://build.verrazzano.io/job/verrazzano/job/jmaron%252FVZ-1395/

### DIFF
--- a/install/chart/values.yaml
+++ b/install/chart/values.yaml
@@ -12,7 +12,7 @@ image:
 verrazzanoOperator:
   name: verrazzano-operator
   imageName: ghcr.io/verrazzano/verrazzano-operator
-  imageVersion: v0.0.100-915d4e9-1
+  imageVersion: v.0.0.102-24f5934-2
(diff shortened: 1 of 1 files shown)


install/chart/values.yaml
```diff
@@ -12,7 +12,7 @@ image:
 verrazzanoOperator:
   name: verrazzano-operator
   imageName: ghcr.io/verrazzano/verrazzano-operator
-  imageVersion: v0.0.100-915d4e9-1
+  imageVersion: v.0.0.102-24f5934-2
```
mgianatagh, line 15: For format of the tags has been `v0.0.n` where this tag is `v.0.0.n`.  That is probably ok to leave for now, rather than rerun a full test suite to change the tag back to the format we have been using.  On the next tag we could return to regularly scheduled programming?

mgianatagh (approved):

PR #330, opened 2020-10-13
Consistent Docker Image Tag Formatting

shihchang:
change keycloak-oracle-theme image tag formatting: "${VERRAZZANO_DEV_VERSION}-${SHORT_COMMIT_HASH}-${BUILD_NUMBER}"


keycloak-oracle-theme/Jenkinsfile
```diff
@@ -57,13 +61,13 @@ pipeline {
             when { not { buildingTag() } }
```
desagar, line 61: we should be able to remove the when { not { buildingTag() } } as well

shihchang: we will have to either disallow adding a new git-tag or in jenkins pipelines describe "do nothing if bindingTag"

desagar (approved):

PR #341, opened 2020-10-22
[WIP] release image retrieval and image publication scripts

jmaron99:
Work in progress....

These scripts can be utilized to query an existing cluster for the ghcr.io related images utilized.  These images can subsequently can be pulled, tagged, and pushed to OCR (the source and destination repositories can be specified/overridden).  These scripts will be utilized to automate the image compilation and publication of release-associated images via Jenkins job currently being composed.

Additional details:
- the image retrieval script creates a file (verrazzano_img_list.txt) containing the paths of the cluster associated images
- the image pull/push script will generate both a file indicating images to push as well as a file indicating any images that were not found in the source repository (skipped_images.txt).  The former can be used to assist the PM in submitting the approvals for release images publication.

This PR is a work in progress due to the following:
- multiple issues have prevented successful build completion on Jenkins over the last few days
- It is unclear whether this is the appropriate repository for these scripts or whether they should exist in another existing or new repository


kminder:
As you mention this repo doesn't seem like the right home but I'm unable to suggest something better.  Does seem like something that should be in gitlab not github though.

kminder:
Do you envision these being executed by a "release" jenkins pipeline?

jmaron99:
> Do you envision these being executed by a "release" jenkins pipeline?

Yes - that's the idea.  working on one now.

pfmackin:
Tom has the helper code with a CLI which does some github related activity.  Does it make sense to have this functionality included in that single tool?

jmaron99:
> Tom has the helper code with a CLI which does some github related activity. Does it make sense to have this functionality included in that single tool?

Some of the other functionality in the release would make sense.  In this case the functionality isn't related to GitHub but rather to Kubernetes (kubectl) and docker.  I imagine there may be a go approach to doing these functions but I believe the scripting of these functions is appropriate.

util/pull_and_push_imgs.sh
```diff
@@ -0,0 +1,76 @@
...
+    consoleerr "  -h             		Help"
+    consoleerr
+    exit 0 
+}
+
+IMG_LIST_FILE=verrazzano_img_list.txt
+SOURCE_REPO=ghcr.io
+SOURCE_REPO_USER=foo
+SOURCE_REPO_PWD=bar
+DEST_REPO=container-registry-admin.oraclecorp.com
```
markxnelson, line 30: cannot put the name of an internal system into a public repo

markxnelson (changes requested):

PR #348, opened 2020-10-29
VZ-1426: Add error checking for openssl commands in install script

speidelj:
Added error checking for openssl commands in v8o install script.
Now if an openssl command fails, the script will fail and provide an error message.
Also, made a change to generate temporary artifacts into a temp directory instead of the previous directory under build.
The temp directory is now cleaned up via a signal trap instead of on the last line of the script to guarantee that the cleanup will occur even in the event of a failure.

### DIFF
--- a/install/3-install-verrazzano.sh
+++ b/install/3-install-verrazzano.sh
@@ -7,7 +7,8 @@ SCRIPT_DIR=$(cd $(dirname "$0"); pwd -P)
 . $SCRIPT_DIR/common.sh
 
 CONFIG_DIR=$SCRIPT_DIR/config
-CERTS_OUT=$SCRIPT_DIR/build/admin-control-cert
+TMP_DIR=$(mktemp -d)
+trap 'rc=$?; rm -rf ${TMP_DIR} || true; _logging_exit_handler $rc' EXIT
 
 function set_INGRESS_IP() {
   if [ ${CLUSTER_TYPE} == "OKE" ]; then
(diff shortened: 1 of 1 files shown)


markxnelson (changes requested):
target branch must be develop, not master

install/3-install-verrazzano.sh
```diff
@@ -76,36 +77,54 @@ set -eu
...
   # Prepare verrazzano_admission_controller_ca_config.txt and verrazzano_admission_controller_cert_config.txt
-  sed "s/VERRAZZANO_NS/${VERRAZZANO_NS}/g" $CONFIG_DIR/verrazzano_admission_controller_ca_config.txt > $CERTS_OUT/verrazzano_admission_controller_ca_config.txt
-  sed "s/VERRAZZANO_NS/${VERRAZZANO_NS}/g" $CONFIG_DIR/verrazzano_admission_controller_cert_config.txt > $CERTS_OUT/verrazzano_admission_controller_cert_config.txt
+  sed "s/VERRAZZANO_NS/${VERRAZZANO_NS}/g" $CONFIG_DIR/verrazzano_admission_controller_ca_config.txt > $TMP_DIR/verrazzano_admission_controller_ca_config.txt
+  sed "s/VERRAZZANO_NS/${VERRAZZANO_NS}/g" $CONFIG_DIR/verrazzano_admission_controller_cert_config.txt > $TMP_DIR/verrazzano_admission_controller_cert_config.txt
 
   # Create the private key for our custom CA
-  openssl genrsa -out $CERTS_OUT/ca.key 2048
+  openssl genrsa -out $TMP_DIR/ca.key 2048
+  if ! openssl genrsa -out $TMP_DIR/ca.key 2048 ; then
```
jmaron99, line 89: why can't lines 88 and 89 be combined?

speidelj: good catch Jon, thanks

speidelj:
> target branch must be develop, not master

Thanks Mark, fixed.

jmaron99 (approved):

markxnelson (approved):

PR #350, opened 2020-11-02
JIRA:VZ-1601 modify workflow to use curl directly

jmaron99:
Given new security constraints that only allow local, oracle/*, or docker/* actions the workflow that trigger pagerduty alerts for new, re-opened, or commented issues has been modified to use curl directly.

### DIFF
--- a/.github/workflows/pagerduty.yml
+++ b/.github/workflows/pagerduty.yml
@@ -1,5 +1,6 @@
 # Copyright (c) 2020, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+name: pagerduty-trigger
 on:
   issues:
     types: [opened, reopened]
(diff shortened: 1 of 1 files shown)


.github/workflows/pagerduty.yml
```diff
@@ -1,19 +1,18 @@
...
       - name: web-request
-        uses: satak/webrequest-action@master
-        with:
-          url: https://events.pagerduty.com/v2/enqueue
-          method: POST
-          payload: '{"routing_key":"${{ secrets.PAGERDUTY_ROUTING_KEY }}","event_action":"trigger","payload":{"summary":"Issue https://github.com/verrazzano/verrazzano/issues/${{ github.event.issue.number }} has been opened, commented on, or re-opened","source":"build.verrazzano.io","severity":"critical","component":"verrazzano"}}'
-          headers: '{"content-type": "application/json"}'
-
+        run: |
+          curl -v -k -d '{"routing_key":'"\"$ROUTING_KEY\""',"event_action":"trigger","payload":{"summary":"Issue https://github.com/verrazzano/verrazzano/issues/'"$ISSUE_NUMBER"' has been opened, commented on, or re-opened","source":"build.verrazzano.io","severity":"critical","component":"verrazzano"}}' -H 'content-type: application/json' -X POST https://events.pagerduty.com/v2/enqueue
```
mgianatagh, line 18: Is exposing `build.verrazzano.io` a security issue?  Should that value be passed in as an environment variable?

jmaron99: It wasn't noted as such in previous reviews but, if it is, I'll need assistance in creating a GitHub secret from @markxnelson or @gilbode

jmaron99: So the source field is arbitrary and "build.verrazzano.io" is probably not really accurate in this context so I changed it to "user" (indicating the change is user initiated)

mgianatagh: thank you, sounds good to me

mgianatagh (approved):

PR #357, opened 2020-11-10
Uptake new verrazzano-monitoring-operator to fix dev profile install

ddrechse:


install/chart/templates/03-verrazzano-monitoring-operator.yaml
```diff
@@ -394,7 +394,6 @@ spec:
             successThreshold: 1
             timeoutSeconds: 5
           args:
-            - --v=4
```
kminder, line 397: What was this required before but not now?

ddrechse: See Chen's logging changes, are checkins have become inermingled

ddrechse: without that, monitoring-operator fails to start

kminder (approved):
Approved.

PR #362, opened 2020-11-12
Update istio helm chart version to work with newer versions of helm 3.x.

gilbode:
Update istio helm chart version to work with newer versions of helm 3.x.


install/common.sh
```diff
@@ -255,6 +255,7 @@ ISTIO_CORE_DNS_PLUGIN_TAG=0.2-20201016204812-23723dcb
 ISTIO_CORE_DNS_IMAGE=container-registry.oracle.com/olcne/coredns
 ISTIO_CORE_DNS_TAG=1.6.2
 ISTIO_VERSION=1.4.6
+ISTIO_HELM_CHART_VERSION=1.4.10
```
desagar, line 258: I can't argue with it if it works, and it seems reasonable. Only concern I have is whether this could come back to bite in some other way since the chart is from a different version

install/common.sh
```diff
@@ -255,6 +255,7 @@ ISTIO_CORE_DNS_PLUGIN_TAG=0.2-20201016204812-23723dcb
 ISTIO_CORE_DNS_IMAGE=container-registry.oracle.com/olcne/coredns
 ISTIO_CORE_DNS_TAG=1.6.2
 ISTIO_VERSION=1.4.6
```
desagar, line 257: what is ISTIO_VERSION now used for?

gilbode: It's still used for all the image versions

desagar (approved):

PR #363, opened 2020-11-13
Update to latest docker images

mgianatagh:
Pick up latest verrazzano-operator, verrazzano-cluster-operator and verrazzano-admission-controller images.

Also an updated CRD file for models.


bobdonat (dismissed):
Nice, I would have forgotten the crd.

bobdonat (approved):

PR #364, opened 2020-11-16
Scchen/fix check kube version

shihchang:
* fix check_kube_version to v1.16.8 ~ v1.17.9

### DIFF
--- a/install/1-install-istio.sh
+++ b/install/1-install-istio.sh
@@ -173,12 +173,12 @@ function check_kube_version {
     local major=$(echo $kubeVer | jq -r '.serverVersion.major')
     local minor=$(echo $kubeVer | jq -r '.serverVersion.minor')
     local patch=$(echo $servVer | cut -d'.' -f 3)
-    VER_ERROR_MSG="Kubernetes serverVersion $servVer must be greater than or equal to v1.16.8 and less than or equal to v1.18"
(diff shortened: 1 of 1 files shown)


install/1-install-istio.sh
```diff
@@ -173,12 +173,12 @@ function check_kube_version {
     local major=$(echo $kubeVer | jq -r '.serverVersion.major')
     local minor=$(echo $kubeVer | jq -r '.serverVersion.minor')
     local patch=$(echo $servVer | cut -d'.' -f 3)
-    VER_ERROR_MSG="Kubernetes serverVersion $servVer must be greater than or equal to v1.16.8 and less than or equal to v1.18"
+    VER_ERROR_MSG="Kubernetes serverVersion $servVer must be greater than or equal to v1.16.8 and less than or equal to v1.17.*"
```
mgianatagh, line 176: I feel we should have @gilbode weigh in on this change.  Do we know if the install use to work on 1.18?  Maybe we changed something recently that broke 1.18 support.  I am a little concerned about us making a change that reduces what we use to say we supported.

gilbode: Why are we adding this constraint?

desagar: If we decide we are going to make this change, I'd prefer we also did it in the actual code that checks the kubernetes version - the README should match the behavior of our validation.

shihchang: In README, "For `KUBERNETES VERSION`, select `v1.16.8`."

desagar (approved):

PR #367, opened 2020-11-16
Tvlatas/vz 1736

agv426:
VZ-1736 related work.

This adds in basic helm value override files. These currently only have the override for the installProfile setting which will be going away with other work going on in parallel. This sets up the automation though so we can run acceptance tests with profile values which are in values.NAME.yaml (NAME being prod, dev, whatever...)

I'm creating a separate PR for the automation side. This has the temporary workaround discussed earlier for the helm issue

### DIFF
--- a/install/2b-install-system-components-ocidns.sh
+++ b/install/2b-install-system-components-ocidns.sh
@@ -148,6 +148,8 @@ spec:
 
 function install_external_dns()
 {
+    helm repo add stable https://charts.helm.sh/stable
+    helm repo update
     helm upgrade external-dns stable/external-dns \
         --install \
         --namespace cert-manager \
--- a/install/2b-install-system-components-ocidns.sh
+++ b/install/2b-install-system-components-ocidns.sh
@@ -148,8 +148,6 @@ spec:
 
 function install_external_dns()
 {
-    helm repo add stable https://charts.helm.sh/stable
-    helm repo update
     helm upgrade external-dns stable/external-dns \
         --install \
         --namespace cert-manager \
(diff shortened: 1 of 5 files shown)


fredtibbitts:
Seems fine to me, but probably worth having somebody who has been here longer than a week review as well. :)

fredtibbitts (approved):

install/2b-install-system-components-ocidns.sh
```diff
@@ -148,6 +148,8 @@ spec:
 
 function install_external_dns()
 {
+    helm repo add stable https://charts.helm.sh/stable
```
shihchang, line 151: This might have been redundant after https://github.com/verrazzano/verrazzano/pull/365

install/4-install-keycloak.sh
```diff
@@ -41,8 +41,10 @@ function install_mysql {
   sed -e "s|MYSQL_IMAGE_TAG|${MYSQL_IMAGE_TAG}|g" \
       -e "s|MYSQL_USERNAME|${MYSQL_USERNAME}|g" \
       $SCRIPT_DIR/config/mysql-values-template.yaml > ${TMP_DIR}/mysql-values-sed.yaml
-  
+
   log "Install MySQL helm chart"
+  helm repo add stable https://charts.helm.sh/stable
+  helm repo update
```
shihchang, line 47: This might have been redundant after #365

agv426: I'll try removing it from here locally to see if the install still works without it (and remove from this level if it still works).

install/2b-install-system-components-ocidns.sh
agv426: I'll try removing it from here locally to see if the install still works without it (and remove from this level if it still works).

agv426: it was fine locally and in the pipeline, but the pipeline was hitting image pull errors so I'll keep spinning runs until it gets past those

shihchang (approved):

fredtibbitts (approved):

PR #372, opened 2020-11-18
VZ-1741.  Adopt new profiles implementation in script installer

michael-cico:
Update installer and values files to use new settings for dev/prod profiles.

Depends on
https://github.com/verrazzano/verrazzano-operator/pull/205
https://github.com/verrazzano/verrazzano-monitoring-operator/pull/49

### DIFF
--- a/install/chart/values.yaml
+++ b/install/chart/values.yaml
@@ -12,8 +12,9 @@ image:
 
 verrazzanoOperator:
   name: verrazzano-operator
-  imageName: ghcr.io/verrazzano/verrazzano-operator
-  imageVersion: 0.5.0-20201113214945-6105e62
+  useSystemVMI: false
+  imageName: ghcr.io/verrazzano/verrazzano-operator-jenkins
+  imageVersion: 0.5.0-20201118013236-85be4c8
   cohMicroImage: ghcr.io/verrazzano/verrazzano-coh-cluster-operator:0.5.0-20201112155855-709d712
   helidonMicroImage: ghcr.io/verrazzano/verrazzano-helidon-app-operator:0.5.0-20201112153844-e0e4790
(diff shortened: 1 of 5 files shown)


install/chart/values.yaml
```diff
@@ -31,15 +32,20 @@ verrazzanoOperator:
   esMasterNodeRequestMemory: 1.4Gi
   esIngestNodeRequestMemory: 2.5Gi
   esDataNodeRequestMemory: 4.8Gi
+  esMasterNodeReplicas: 3
```
gilbode, line 35: I'm not sure I like the structure of this - just a big flat list of properties under Verrazzano operator.  Seems you are trying to configure 2 different things here - whether and how we create 1 or more Elasticsearch environments for the user - and whether or how we connect the application to it.  I'd like to revisit this before we make this end user facing.

agv426 (approved):
Looks good to me, I'm assuming there will be more evolving here around Mike Gilbode's comment (ie: structure may change before it is public)

install/chart/values.yaml
michael-cico: OK...the other question is how much the values approach will survive the shift to the new config and the in-cluster installer, and what will be exposed.  When I talked with @desagar and @pfmackin on Tuesday, it sounds like these values files will not be exposed.

That being said, I can take another crack at making these more hierarchical.  Did you want things grouped by say, ES vs. Prometheus vs. Kibana, for example? That's kind of how the VMI is organized also.

I think at some point we're either going to end up exposing the VMI through a binding, which might line up with that thinking, or exposing something like the CRDs for prometheus/ES operators if we end up switching to those and retiring the VMO, etc.

michael-cico: Updated per our meeting, and tested locally.

desagar (approved):
lgtm

PR #376, opened 2020-11-24
VZ-1715 move verrazzano-platform-operator 

shihchang:
VZ-1715 move verrazzano-platform-operator


desagar:
What about the README in platform operator? The instructions in this readme are still the old ones which do work as of now, but we will need the new instructions in here soon. When should we reconcile those?

desagar (dismissed):

jmaron99:
It seems strange that the Makefile etc are top-level files in the verrazzano repo.  I guess the lack of existing code to compile etc drove that decision?  Are we all comfortable with that approach?

bobdonat:
The README from the platform-operator is missing.

mgianatagh:
> It seems strange that the Makefile etc are top-level files in the verrazzano repo. I guess the lack of existing code to compile etc drove that decision? Are we all comfortable with that approach?

I am fine with this.  We also decided to minimize changes between the repos initially.  Changing the structure of folders would have been more difficult.  The repo can be reorganized later if we feel it is necessary.

shihchang:
> It seems strange that the Makefile etc are top-level files in the verrazzano repo
@jmaron99 Yes, we want to move root dir to root dir first which results in moving the Makefile from platform-operator repo root to verrazzano root dir. We may restructure the codes later if we prefer.

mgianatagh:
> The README from the platform-operator is missing.

I feel we want to handle the README in a separate PR and need to get Rosemary to review.

mgianatagh:
> It seems strange that the Makefile etc are top-level files in the verrazzano repo
> @jmaron99 Yes, we want to move root dir to root dir first which results in moving the Makefile from platform-operator repo root to verrazzano root dir. We may restructure the codes later if we prefer.

I feel we need to carefully consider if and how to restructure.  We should probably adhere to this model:
https://github.com/golang-standards/project-layout

PLATFORM-OPERATOR-README.md
```diff
@@ -0,0 +1,128 @@
+[![Go Report Card](https://goreportcard.com/badge/github.com/verrazzano/verrazzano-platform-operator)](https://goreportcard.com/report/github.com/verrazzano/verrazzano-platform-operator)
```
bobdonat, line 1: repo name is verrazzano/verrazzano now

shihchang: fixed

desagar:
> > It seems strange that the Makefile etc are top-level files in the verrazzano repo
> > @jmaron99 Yes, we want to move root dir to root dir first which results in moving the Makefile from platform-operator repo root to verrazzano root dir. We may restructure the codes later if we prefer.
> 
> I feel we need to carefully consider if and how to restructure. We should probably adhere to this model:
> https://github.com/golang-standards/project-layout

There was originally a plan to move the go code down one level, similar to the istio operator repo but we decided to limit the changes in the first round. There are also some issues with attribution helper behavior.

mgianatagh (approved):

desagar (approved):

bobdonat (approved):

PR #381, opened 2020-11-25
JIRA:VZ-1771 refactoring associated with using one oci config secret

jmaron99:
In addition to the code refactorings there is a script for generating the OCI configuration secret.


controllers/verrazzano_controller.go
```diff
@@ -240,6 +251,24 @@ func (r *VerrazzanoReconciler) createInstallJob(ctx context.Context, log logr.Lo
 	return err
 }
 
+func getDNSAuth(r *VerrazzanoReconciler, dns installv1alpha1.DNS) (*installjob.DNSAuth, error) {
+	if dns.OCI != (installv1alpha1.OCI{}) {
+		secret := &corev1.Secret{}
+		err := r.Get(context.TODO(), types.NamespacedName{Name: dns.OCI.OCIConfigSecret, Namespace: "default"}, secret)
```
bobdonat, line 257: I think the namespace should be the same namespace as the verrazzano resource and not hard coded to "default".

mgianatagh:
Should we add some temporary instructions in PLATFORM-OPERATOR-README to describe how to create the secret for OCI DNS?

controllers/verrazzano_controller.go
jmaron99: fixed

jmaron99:
> Should we add some temporary instructions in PLATFORM-OPERATOR-README to describe how to create the secret for OCI DNS?

took a stab at adding some steps

mgianatagh (approved):

bobdonat (approved):
Looks good to me.

PR #382, opened 2020-11-25
KIND support

guoyongzhang:
The acceptance test is in verrazzano-acceptance-test-suite `guzhang/vz-1646` branch.  
https://build.verrazzano.io/view/Tests/job/verrazzano-kind-acceptance-test-suite/job/guzhang%252Fvz-1646/


install/0-create-kind-cluster.sh
```diff
@@ -0,0 +1,28 @@
...
+# Copyright (c) 2020, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+SCRIPT_DIR=$(cd $(dirname $0); pwd -P)
+
+set -u
+
+. $SCRIPT_DIR/logging.sh
+
+KIND_IMAGE="kindest/node:v1.16.9@sha256:7175872357bc85847ec4b1aba46ed1d12fa054c83ac7a8a11f5c268957fd5765"
```
bobdonat, line 12: Should this be bump to 1.17.  platform-operator and application-operator are usiing 1.17

install/config/config_kind.json
```diff
@@ -0,0 +1,50 @@
...
+   "certificates": {
+      "issuerType": "ca",
+      "ca": {
+         "secretName": "tls-rancher",
+         "clusterResourceNamespace": "cattle-system"
+      }
+   },
+   "ingress": {
+      "type": "NodePort",
+      "nodePort": {
```
bobdonat, line 16: is nodePort a new field in the json config.  The platform-operator does that set this value.

install/0-create-kind-cluster.sh
mgianatagh: I do not feel this could change to 1.17 yet since we know this does not work for an OKE cluster.  The intel-tests of the other operators are probably ok with 1.17 since those individual components would work on 1.17, but not verrazzano as a whole yet.

bobdonat: Not sure I understand.  We have been testing OKE clusters with 1.17.9 for a while.

guoyongzhang: I changed it to 1.17 and it is being tested now.

install/config/config_kind.json
guoyongzhang: nodePort is a brand new field.  platform-operator needs to add it to support KIND.

install/0-create-kind-cluster.sh
```diff
@@ -0,0 +1,28 @@
...
+# Copyright (c) 2020, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+SCRIPT_DIR=$(cd $(dirname $0); pwd -P)
+
+set -u
+
+. $SCRIPT_DIR/logging.sh
+
+KIND_IMAGE="kindest/node:v1.17.0@sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62"
```
gilbode, line 12: We need to figure out how to make this work with a recent version of kind - not 0.7.0.  And I don't understand exactly why we need this script any longer b/c we no longer use the shell scripts.

install/0-create-kind-cluster.sh
```diff
@@ -0,0 +1,28 @@
...
+}
+
+env LOGFILE=$LOGFILE ${SCRIPT_DIR}/5-delete-kind-cluster.sh >&5 6>&5
+
+action "Creating kind cluster ${KIND_CLUSTER_NAME}..." \
+    kind create cluster \
+        --wait 30s \
+        --image ${KIND_IMAGE} \
+        --name ${KIND_CLUSTER_NAME} \
+        --config ${SCRIPT_DIR}/config/kind-config.yaml \
```
gilbode, line 27: This needs to work with a single-node kind cluster as per the story.  Why do we need a custom config?

desagar (approved):
Reviewed interactively with GZ

desagar (approved):

PR #387, opened 2020-11-30
Tvlatas/vz 1696

agv426:
Logging update


controllers/verrazzano_controller.go
```diff
@@ -11,10 +11,11 @@ import (
 	"sigs.k8s.io/yaml"
 	"time"
 
-	"github.com/go-logr/logr"
+	//"github.com/go-logr/logr"
```
jmaron99, line 14: remove if not used

agv426: Yup, missed that

operator/controllers/verrazzano_controller.go
```diff
@@ -75,7 +75,7 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 			// Remove the finalizer and update the verrazzano resource if the uninstall has finished.
 			for _, condition := range vz.Status.Conditions {
 				if condition.Type == installv1alpha1.UninstallComplete || condition.Type == installv1alpha1.UninstallFailed {
-					log.Info(fmt.Sprintf("Removing finalizer %s", finalizerName))
+					log.Info("Removing finalizer %s", finalizerName)
```
bobdonat, line 78: I think this should be Infof

operator/controllers/verrazzano_controller.go
```diff
@@ -46,7 +46,7 @@ const finalizerName = "install.verrazzano.io"
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;watch;list;create;update;delete
 func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.TODO()
-	log := r.Log.WithValues("resource", fmt.Sprintf("%s:%s", req.Namespace, req.Name)).WithName("reconcile")
+	log := zap.S().With("resource", fmt.Sprintf("%s:%s", req.Namespace, req.Name)).Named("reconcile")
```
bobdonat, line 49: Since we have caller being logged for these message the named value adds no value here.  All messages logged with this logger will have "name: reconcile"

operator/controllers/verrazzano_controller.go
agv426: updating

operator/controllers/verrazzano_controller.go
agv426: sounds good, will get rid of that

bobdonat (approved):

PR #388, opened 2020-11-30
Upgrade Verrazzano using helm charts

pfmackin:
Upgrade Verrazzano using the helm chart upgrade.  This will use all of the install overrides that were applied during the initial install, but apply the new chart.  This code does upgrade at the component level, but only Verrazzano itself is currently upgraded.  Upgrade is initiated by changes the Verrazzano CR version field to match the new helm Chart version.

Sample of Verrazzano CR Status.Conditions after upgraded:

    Last Transition Time:  2020-12-02T17:32:16Z
    Message:               Verrazzano upgrade to version 0.7.0 in progress
    Status:                True
    Type:                  UpgradeStarted
    Last Transition Time:  2020-12-02T17:33:19Z
    Message:               Verrazzano upgraded to version 0.7.0 successfully
    Status:                True
    Type:                  UpgradeComplete
  Version:                 0.7.0


operator/api/v1alpha1/verrazzano_types.go
```diff
@@ -161,6 +161,8 @@ type Certificate struct {
 
 // VerrazzanoSpec defines the desired state of Verrazzano
 type VerrazzanoSpec struct {
+	// Version is the Verrazzano version
+	Version string `json:"version,omitempty"`
```
bobdonat, line 165: What does this field mean when doing an install?  Can I put any string in there?

operator/controllers/verrazzano_controller.go
```diff
@@ -83,10 +83,18 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 				}
 			}
 		}
-
 		return reconcile.Result{}, nil
 	}
 
+	// If the version is specified then see if upgrade is needed
+	if len(vz.Spec.Version) > 0 {
+		if vz.Spec.Version != vz.Status.Version {
```
michael-cico, line 91: Do we want to ensure that the new version is > than the current version here, and not just check for != ?

michael-cico: and, if it's not > than the current version, then maybe log/error out, of course

michael-cico: Talked to Paul, sounds like this is deferred to another task.

mgianatagh:
Current code coverage is 88.59%, I tested this branch and the code coverage drops to 82.7%.   More tests are needed for the new code.

operator/api/v1alpha1/verrazzano_types.go
pfmackin: This is the target version for the upgrade.  Once we do validation, it will need to match the helm Chart.yaml version

michael-cico (approved):

operator/config/scripts/run.sh
```diff
@@ -44,17 +44,18 @@ function dump-uninstall-logs {
 # The same docker image is shared between the verrazzano-platform-operator and
 # the installation jobs that the operator creates.  The default mode is to run
 # the verrazzano-platform-operator.
+
+# Create a kubeconfig
+create-kubeconfig
```
bobdonat, line 49: @mgianatagh is this something we want to always do.  I am wondering why this is being changed

operator/controllers/upgrade_reconciler.go
```diff
@@ -0,0 +1,92 @@
...
+	targetVersion := cr.Spec.Version
+	err := validateVersion(targetVersion)
+	if err != nil {
+		log.Error(err, "Invalid upgrade version")
+		return ctrl.Result{}, nil
+	}
+
+	// Only allow upgrade to retry a certain amount of times during any upgrade attempt.
+	if upgradeFailureCount(cr.Status) > failedUpgradeLimit {
+		log.Info("Upgrade failure limit reached, upgrade will not be attempted")
```
bobdonat, line 28: Should this be an error

pfmackin: Maybe, I thought info was good enough.  Maybe I should add a condition UpgradeLimitRetryExceeded or something

operator/config/scripts/run.sh
pfmackin: helm is being called from the platform operator and needs this kubeconfig file.  I slacked Mike who thought it was ok to change it.  At some point we might move to the helm API and may not need this.

bobdonat (approved):

markxnelson (approved):

PR #391, opened 2020-12-01
Update platform-operator image in operator.yaml

bobdonat:
This pull request changes the Jenkins build to update the platform-operator image in operator.yaml with the newly built image.  This will only be done if the branch is 'master' or 'develop'.

Updating operator.yaml requires a commit and push.  The commit contains the string "[ci skip]" which will result in another build job running.  The build will be skipped when the commit contains "[ci skip]".


Jenkinsfile
```diff
@@ -65,7 +65,16 @@ pipeline {
             steps {
                 script {
                     checkout scm
-                }
+                    // Check if commit message contains "[ci skip]"
```
pfmackin, line 68: can we call this skipBuild to make it more obvious what it means.  This would be similar to the maven surefile skipTests

Jenkinsfile
```diff
@@ -256,13 +296,43 @@ pipeline {
...
+                    cat config/crd/bases/install.verrazzano.io_verrazzanos.yaml >> deploy/operator.yaml
+                    git add deploy/operator.yaml
+                    git commit -m "[ci skip] update operator image"
+                    git push
+                   """
+            }
+        }
     }
 
     post {
```
pfmackin, line 338: should the post have a check for skipBuild?

bobdonat: I don't think it is needed.  The post is run on failure only.

Jenkinsfile
bobdonat: It is case sensitive.  I used [ci skip] because that was a convention used by other CI/CD implementations.  gitlab, circleci, etc.   The others also allow [skip ci].

Jenkinsfile
```diff
@@ -256,13 +296,43 @@ pipeline {
...
+            when {
+                allOf {
+                    not { buildingTag() }
+                    equals expected: false, actual: skipBuild
+                    anyOf { branch 'master'; branch 'develop' }
+                }
+            }
+            steps {
+                sh """
+                    git clone -b ${env.BRANCH_NAME} https://github.com/verrazzano/verrazzano
```
pfmackin, line 323: wasn't the repo already cloned as part of the previous stages?

bobdonat: I used code from the verrazzano-operator.  I agree this is probably not needed.  I will try to fix.

pfmackin (dismissed):

operator/README.md
```diff
@@ -11,9 +11,9 @@ THIS IS A TEMPORARY README WHILE WE COMPLETE THE TRANSITION OF THE VERRAZZANO-PL
 * docker
 * kubectl
 
-## Performing a Verrazzano install
+## Performing a build and Verrazzano install
 
-To install Verrazzano with the verrazzano-platform-operator follow these steps.
+To build and then install Verrazzano with the verrazzano-platform-operator follow these steps.
```
pfmackin, line 16: maybe get rid of `then`

pfmackin (approved):

PR #396, opened 2020-12-02
add NodePort install

guoyongzhang:


operator/internal/installjob/install_config.go
```diff
@@ -173,6 +179,10 @@ func GetInstallConfig(vz *installv1alpha1.Verrazzano, auth *DNSAuth) (*InstallCo
 		return newOCIDNSInstallConfig(vz, auth)
 	}
 
+	if vz.Spec.Ingress.Type == installv1alpha1.NodePort {
```
bobdonat, line 182: I don't see why we are adding another type of install here.   The additional node port stuff should be incorporated into the 3 flavors we have now - xip.io, oci dns, and external.

guoyongzhang: good point. i will update

operator/scripts/install/config.sh
```diff
@@ -182,7 +182,7 @@ function validate_config_json {
 function get_verrazzano_ingress_ip {
   local ingress_type=$(get_config_value ".ingress.type")
   if [ ${ingress_type} == "NodePort" ]; then
-    ingress_ip=$(get_config_value ".ingress.nodePort.ingressIp")
+    ingress_ip=$(kubectl -n ingress-nginx get pods --selector app=nginx-ingress,component=controller -o jsonpath='{.items[0].status.hostIP}')
```
desagar, line 185: how does this work automatically? Don't you need some change in script 2 for this ingress IP to work? For KinD nodeport it is usually the worker nodes IP that is the ingress right?

guoyongzhang: script 2 calls this after nginx install, during which controller pod was bound to a node.  the container IP of this node will used as ingress IP.

guoyongzhang: script 2 change is not needed.

operator/api/v1alpha1/verrazzano_types.go
```diff
@@ -85,6 +85,9 @@ type InstallArgs struct {
 	// Value for named install argument
 	// +optional
 	Value string `json:"value,omitempty"`
+	// If the Value is a literal string
+	// +optional
+	SetString string `json:"setString,omitempty"`
```
mgianatagh, line 90: The readme files have been updated to describe the verrazzano custom resource and define each field.  Can you rebase you transaction to pick up the new readme files and add this new `setString` parameter?

operator/config/samples/install-kind.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2020, Oracle and/or its affiliates.
```
mgianatagh, line 1: This sample file will be visible to users, so they will want to try to use it.

@gilbode are we adding Kind support to the initial release of the in-cluster installer?  If not this file should probably be removed from the PR.

gilbode: We're not explicitly supporting it for initial release.  If this is something end users will try by its location I agree we should leave this out of this repo until it works - or put the file elsewhere that makes it clear it probably won't work yet.

guoyongzhang: removed

operator/api/v1alpha1/verrazzano_types.go
guoyongzhang: added "setString" to "General format of the Verrazzano custom resource"

pfmackin: I don't understand what this setString arg is, can you explain?  I see we already have a name/value pait

guoyongzhang: It is a helm thing.  https://github.com/roboll/helmfile/issues/119 is the feature request that was implemented.

mgianatagh:
I just ran code coverage and you are going to need to add a little more test coverage.  Current coverage is at 90.4% and your PR will bring it down to 89.1%.  This will cause the build pipeline to fail on the develop branch.

mgianatagh (approved):

operator/api/v1alpha1/verrazzano_types.go
bobdonat: Shouldn't this be a boolean

PR #397, opened 2020-12-02
VZ-1788.  Save current Verrazzano install state

michael-cico:
Save the install resource spec to an internal configmap
- Also, add support for a NOOP mode for the install jobs to allow testing operator flow in-cluster without trigger actual installs/uninstalls
- add a helper make target to generate an operator deploy YAML file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,3 +221,13 @@ push-tag:
 	docker pull "${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG}"; \
 	docker tag "${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG}" "${DOCKER_IMAGE_FULLNAME}:$$PUBLISH_TAG"; \
 	docker push "${DOCKER_IMAGE_FULLNAME}:$$PUBLISH_TAG"
+
+.PHONY: create-test-deploy
+create-test-deploy:
+	if [ -n "${VZ_DEV_IMAGE}" ]; then \
+		echo "Building local operator deployment resource file in /tmp/operator.yaml, VZ_DEV_IMAGE=${VZ_DEV_IMAGE}"; \
+		cat operator/config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|${VZ_DEV_IMAGE}|g" > /tmp/operator.yaml; \
+		cat operator/config/crd/bases/install.verrazzano.io_verrazzanos.yaml >> /tmp/operator.yaml; \
+	else \
+		echo "VZ_DEV_IMAGE not defined, please set it to a valid image name/tag"; \
+	fi
--- a/operator/config/scripts/run.sh
+++ b/operator/config/scripts/run.sh
@@ -47,7 +47,12 @@ function dump-uninstall-logs {
 
 # Create a kubeconfig
 create-kubeconfig
-if [ ${MODE} == "INSTALL" ]; then
+if [ ${MODE} == "NOOP" ]; then
+  echo "*************************************************************"
+  echo " Running in NOOP mode, exiting                               "
+  echo "*************************************************************"
+  exit 0
+elif [ ${MODE} == "INSTALL" ]; then
   # Run the installation
   ./install/1-install-istio.sh || dump-install-logs 1
   ./install/2-install-system-components.sh || dump-install-logs 1
--- a/operator/controllers/verrazzano_controller.go
+++ b/operator/controllers/verrazzano_controller.go
@@ -5,8 +5,10 @@ package controllers
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/verrazzano/verrazzano/operator/internal"
 	"os"
 	"sigs.k8s.io/yaml"
 	"time"
@@ -35,10 +37,14 @@ type VerrazzanoReconciler struct {
 	Log        *zap.SugaredLogger
 	Scheme     *runtime.Scheme
 	Controller controller.Controller
+	DryRun     bool
 }
 
-// Name of finializer
+// Name of finalizer
 const finalizerName = "install.verrazzano.io"
+
+// Key into ConfigMap data for stored install Spec, the data for which will be used for update/upgrade purposes
+const configDataKey = "spec"
 
 // Reconcile will reconcile the CR
 // +kubebuilder:rbac:groups=install.verrazzano.io,resources=verrazzanos,verbs=get;list;watch;create;update;patch;delete
@@ -110,7 +116,12 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 		return reconcile.Result{}, err
 	}
 
-	if err := r.createInstallJob(ctx, log, vz, getConfigMapName(vz.Name)); err != nil {
+	if err := r.createInstallJob(ctx, log, vz, buildConfigMapName(vz.Name)); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Create/update a configmap from spec for future comparison on update/upgrade
+	if err := r.saveVerrazzanoSpec(ctx, vz); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -120,7 +131,7 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 // createServiceAccount creates a required service account
 func (r *VerrazzanoReconciler) createServiceAccount(ctx context.Context, log *zap.SugaredLogger, vz *installv1alpha1.Verrazzano) error {
 	// Define a new service account resource
-	serviceAccount := installjob.NewServiceAccount(vz.Namespace, getServiceAccountName(vz.Name), os.Getenv("IMAGE_PULL_SECRET"), vz.Labels)
+	serviceAccount := installjob.NewServiceAccount(vz.Namespace, buildServiceAccountName(vz.Name), os.Getenv("IMAGE_PULL_SECRET"), vz.Labels)
 
 	// Set verrazzano resource as the owner and controller of the service account resource.
 	// This reference will result in the service account resource being deleted when the verrazzano CR is deleted.
@@ -130,10 +141,10 @@ func (r *VerrazzanoReconciler) createServiceAccount(ctx context.Context, log *za
 
 	// Check if the service account for running the scripts exist
 	serviceAccountFound := &corev1.ServiceAccount{}
-	log.Infof("Checking if install service account %s exist", getServiceAccountName(vz.Name))
-	err := r.Get(ctx, types.NamespacedName{Name: getServiceAccountName(vz.Name), Namespace: vz.Namespace}, serviceAccountFound)
+	log.Infof("Checking if install service account %s exist", buildServiceAccountName(vz.Name))
+	err := r.Get(ctx, types.NamespacedName{Name: buildServiceAccountName(vz.Name), Namespace: vz.Namespace}, serviceAccountFound)
 	if err != nil && errors.IsNotFound(err) {
-		log.Infof("Creating install service account %s", getServiceAccountName(vz.Name))
+		log.Infof("Creating install service account %s", buildServiceAccountName(vz.Name))
 		err = r.Create(ctx, serviceAccount)
 		if err != nil {
 			return err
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -221,3 +221,9 @@ push-tag:
 	docker pull "${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG}"; \
 	docker tag "${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG}" "${DOCKER_IMAGE_FULLNAME}:$$PUBLISH_TAG"; \
 	docker push "${DOCKER_IMAGE_FULLNAME}:$$PUBLISH_TAG"
+
+.PHONY: operator-deploy
+operator-deploy:
+	echo "Building local operator deployment resource file in /tmp/operator.yaml, VZ_DEV_IMAGE=${VZ_DEV_IMAGE}"; \
```
pfmackin, line 227: @mgianatagh are you ok with this?

operator/config/scripts/run.sh
```diff
@@ -47,7 +47,12 @@ function dump-uninstall-logs {
 
 # Create a kubeconfig
 create-kubeconfig
-if [ ${MODE} == "INSTALL" ]; then
+if [ ${MODE} == "NOOP" ]; then
```
pfmackin, line 50: I don't think we need this if the operator has a --dry-run mode since it won't start the job.

operator/controllers/verrazzano_controller.go
```diff
@@ -219,7 +228,8 @@ func (r *VerrazzanoReconciler) createConfigMap(ctx context.Context, log *zap.Sug
 // createInstallJob creates the installation job
 func (r *VerrazzanoReconciler) createInstallJob(ctx context.Context, log *zap.SugaredLogger, vz *installv1alpha1.Verrazzano, configMapName string) error {
 	// Define a new install job resource
-	job := installjob.NewJob(vz.Namespace, getInstallJobName(vz.Name), vz.Labels, configMapName, getServiceAccountName(vz.Name), os.Getenv("VZ_INSTALL_IMAGE"))
+	r.Log.Infof("Creating install job for %s, dry-run=%v", vz.Name, r.DryRun)
+	job := installjob.NewJob(vz.Namespace, getInstallJobName(vz.Name), vz.Labels, configMapName, getServiceAccountName(vz.Name), os.Getenv("VZ_INSTALL_IMAGE"), r.DryRun)
```
pfmackin, line 232: The number of input params is getting really large.  Would be better to have a JobConfig struct that gets initialized and passed to NewJob.

operator/controllers/verrazzano_controller.go
```diff
@@ -371,6 +382,11 @@ func getConfigMapName(name string) string {
 	return fmt.Sprintf("verrazzano-install-%s", name)
 }
 
+// getInternalConfigMapName returns the name of the internal configmap associated with an install resource.
+func getInternalConfigMapName(name string) string {
```
pfmackin, line 386: this is better called buildInternal...

operator/controllers/verrazzano_controller.go
```diff
@@ -468,6 +484,57 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
 	return r.updateStatus(log, vz, "Verrazzano uninstall in progress", installv1alpha1.UninstallStarted)
 }
 
+// saveInstallSpec Saves the install spec in a configmap to use with upgrade/updates later on
+func (r *VerrazzanoReconciler) saveInternalConfigMap(ctx context.Context, vz *installv1alpha1.Verrazzano) (err error) {
```
pfmackin, line 488: Should be saveVerrazzanoSpec.   It will be called for both install and upgrade

operator/controllers/verrazzano_controller.go
```diff
@@ -468,6 +484,57 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
...
+	installSpec, err := yaml.Marshal(vz.Spec)
```
pfmackin, line 489: We should base64 encode the spec.

operator/controllers/verrazzano_controller.go
```diff
@@ -468,6 +484,57 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
...
+						Kind:       vz.Kind,
+						Name:       vz.Name,
+						UID:        vz.UID,
+					}},
+				},
+				Data: configData,
+			}
+			err = r.Create(ctx, installConfig)
+			if err != nil {
+				r.Log.Errorf("Unable to create installer config map %s: %v", configMapName, err)
```
pfmackin, line 520: return error immediately after the r.log

operator/controllers/verrazzano_controller_test.go
```diff
@@ -1396,6 +1403,116 @@ func TestJobCreateError(t *testing.T) {
...
+}
+
+// TestUpdateInternalConfigMap tests the saveInternalConfigMap method to update an existing internal configmap
+// GIVEN a call so save the internal configmap resource
+// WHEN an intenral configmap already exists for the install
+// THEN ensure that update is called for the configmap
+func TestUpdateInternalConfigMap(t *testing.T) {
+	namespace := "verrazzano"
+	name := "test"
+	//labels := map[string]string{"label1": "test1"}
```
pfmackin, line 1469: remove commented code

operator/controllers/verrazzano_controller_test.go
```diff
@@ -1396,6 +1403,116 @@ func TestJobCreateError(t *testing.T) {
...
+	err := reconciler.saveInternalConfigMap(context.TODO(), vz)
+
+	// Validate the results
+	mocker.Finish()
+	asserts.NotNil(err)
+}
+
+// TestUpdateInternalConfigMap tests the saveInternalConfigMap method to update an existing internal configmap
+// GIVEN a call so save the internal configmap resource
+// WHEN an intenral configmap already exists for the install
```
pfmackin, line 1464: spelling internal

operator/internal/installjob/job.go
```diff
@@ -15,8 +15,12 @@ import (
 // labels - labels of verrazzano resource
 // saName - service account name
 // image - docker image
-func NewJob(namespace string, jobName string, labels map[string]string, configMapName string, saName string, image string) *batchv1.Job {
+func NewJob(namespace string, jobName string, labels map[string]string, configMapName string, saName string, image string, dryRun bool) *batchv1.Job {
 	var backOffLimit int32 = 0
+	mode := "INSTALL"
```
pfmackin, line 20: It looks like DryRun is actually going to start a job and pod, is that right?  If so the names should reflect dry-run or it might be confusing when looking at this.

operator/controllers/verrazzano_controller.go
michael-cico: It's returned at the end of the function, but it's probably more consistent to return here.

operator/internal/installjob/job.go
michael-cico: Yes, it does actually run the job, but the job does nothing.  I didn't want to change the name of the pod, as I just want dry run to do everything that we normally do, except the jobs do nothing except spit out some text.

Perhaps the better thing to do is indicate dry-run in the status object?

operator/config/scripts/run.sh
michael-cico: It does run the jobs, but the jobs are no-ops.

Makefile
michael-cico: I thought it might be useful to generate the operator file, save a few C&Ps from the README.  Although it does require defining the env var.

operator/controllers/verrazzano_controller.go
michael-cico: After talking with Paul, we'll do this as part of another issue.

operator/controllers/verrazzano_controller.go
michael-cico: I was following the pattern of the other name getters... should we change them all?

Makefile
mgianatagh: this target sounds like would be deploying the operator.  This target seems like it might be useful for creating a deployment file for local testing.  If we do keep the target maybe the name should be something like `create-test-deploy`.  The target would also need to be documented in `operator/README.md`

operator/controllers/verrazzano_controller_test.go
```diff
@@ -1396,6 +1403,116 @@ func TestJobCreateError(t *testing.T) {
...
+	asserts.NotNil(mockStatus)
+
+	vz := &vzapi.Verrazzano{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+		Spec: vzapi.VerrazzanoSpec{
+			Profile: "dev",
+		},
+	}
+
+	//expectedConfigMapDataBytes, err := yaml.Marshal(vz.Spec)
```
mgianatagh, line 1428: remove commented out code

michael-cico: Fixed

operator/controllers/verrazzano_controller_test.go
michael-cico: Fixed

operator/controllers/verrazzano_controller_test.go
michael-cico: Fixed

operator/controllers/verrazzano_controller.go
michael-cico: Fixed

operator/controllers/verrazzano_controller.go
michael-cico: Fixed

operator/controllers/verrazzano_controller.go
michael-cico: Fixed

operator/controllers/verrazzano_controller.go
michael-cico: Renamed all the getXXXName() internal calls to buildXXXName()

operator/controllers/verrazzano_controller.go
michael-cico: Actually, fixed this one and made some structs for job configs.

Makefile
michael-cico: OK.  Do we generally frown on these kinds of helpers?  Mainly, it was just to avoid the C&P in the README... it's probably not a whole lot better, but seemed like a good idea at the time.

michael-cico: Renamed the target to create-test-deploy and updated the operator README.

operator/internal/installjob/job.go
michael-cico: added a "dry-run" annotation to the batch jobs/pods.

operator/internal/installjob/job.go
```diff
@@ -4,42 +4,57 @@
...
-// image - docker image
-func NewJob(namespace string, jobName string, labels map[string]string, configMapName string, saName string, image string) *batchv1.Job {
+func NewJob(jobConfig *JobConfig) *batchv1.Job {
 	var backOffLimit int32 = 0
+	mode := installMode
+	if jobConfig.DryRun {
+		mode = internal.NoOpMode
+	}
+	annotations := make(map[string]string, 1)
+	annotations[internal.DryRunAnnotationName] = strconv.FormatBool(jobConfig.DryRun)
```
pfmackin, line 32: I think it would be better to only add the annotation if dryrun is true, but this is ok and not a merge blocker

pfmackin (approved):

operator/internal/installjob/job.go
michael-cico: OK, I'll make a note of it and do that in the next JIRA.

operator/controllers/verrazzano_controller.go
```diff
@@ -468,6 +508,59 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
 	return r.updateStatus(log, vz, "Verrazzano uninstall in progress", installv1alpha1.UninstallStarted)
 }
 
+// saveInstallSpec Saves the install spec in a configmap to use with upgrade/updates later on
+func (r *VerrazzanoReconciler) saveVerrazzanoSpec(ctx context.Context, vz *installv1alpha1.Verrazzano) (err error) {
+	installSpecBytes, err := yaml.Marshal(vz.Spec)
+	installSpec := base64.StdEncoding.EncodeToString(installSpecBytes)
```
bobdonat, line 514: shouldn't this line of code be in the if statement below.  If the Marshall function fails the installSpecBytes is probably null

pfmackin: i missed that, the code should check for err immediately and return

bobdonat (approved):

PR #402, opened 2020-12-03
Add OAM support to Platform Installer

ddrechse:


operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -194,6 +225,36 @@ function install_verrazzano()
   log "Verrazzano install completed"
 }
 
+function install_wls_operator {
+  log "Add Weblogic Operator helm repository"
```
pfmackin, line 229: Is this going to conflict with the WebLogic operator that is currently deployed per app?

operator/scripts/install/chart/crds/oam.verrazzano.io_ingresstraits.yaml
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) 2020, Oracle and/or its affiliates.
```
pfmackin, line 1: How are these CRDs getting into the Verrazzano repo if the OAM repo changes them.  Are folks expected to manually copy?

operator/scripts/install/chart/templates/05-verrazzano-application-operator.yaml
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2020, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: verrazzano-application
```
pfmackin, line 7: all of the OAM App Operator kubernetes resources should be in verrazzano-system

operator/scripts/install/chart/templates/05-verrazzano-application-operator.yaml
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2020, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
```
pfmackin, line 2: Instead of having all these n-chart.yaml files it might be better to have nested charts.  It certainly would be cleaner.

operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -209,4 +270,6 @@ if [ ${REGISTRY_SECRET_EXISTS} == "TRUE" ]; then
 fi
 
 action "Creating admission controller cert" create_admission_controller_cert || exit 1
```
pfmackin, line 272: I don't see where the uninstall of OAM runtime and WebLogic is called

operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -122,6 +122,37 @@ function create_admission_controller_cert()
...
+
+  log "Add OAM helm repository"
+  helm repo add crossplane-master https://charts.crossplane.io/master/
+  if [ $? -ne 0 ]; then
+    error "Failed add OAM helm repository."
+    return 1
+  fi
+
+  log "Install OAM"
+  helm install oam --namespace oam-system crossplane-master/oam-kubernetes-runtime --devel
```
kminder, line 141: We need to switch this to installing a specific version (e.g.   helm install oam --namespace "${NAMESPACE}" crossplane-master/oam-kubernetes-runtime --version 0.3.0)

operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -122,6 +122,37 @@ function create_admission_controller_cert()
   fi
 }
 
+function install_oam {
+  log "Create oam-system namespace"
+  kubectl create namespace oam-system
```
kminder, line 127: There is other work going on to consolidate things into the verrazzano-system namespace.  Therefore this should be removed and verrazzano-system used for the namespace below.

operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -122,6 +122,37 @@ function create_admission_controller_cert()
...
+
+  log "Install OAM"
+  helm install oam --namespace oam-system crossplane-master/oam-kubernetes-runtime --devel
+  if [ $? -ne 0 ]; then
+    error "Failed to OAM helm install."
+    return 1
+  fi
+
+  log "Setup OAM roles"
+  kubectl create clusterrolebinding cluster-admin-binding-oam --clusterrole cluster-admin --user system:serviceaccount:oam-system:oam-kubernetes-runtime-oam
```
kminder, line 148: I think the oam-system here is a namespace so it will need to be changed to verrazzano-system.

operator/scripts/install/3-install-verrazzano.sh
kminder: In our testing we have not seen it conflict with the "per app" WLS operator.

operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -194,6 +225,36 @@ function install_verrazzano()
...
+function install_wls_operator {
+  log "Add Weblogic Operator helm repository"
+  helm repo add weblogic-operator https://oracle.github.io/weblogic-kubernetes-operator/charts
+  if [ $? -ne 0 ]; then
+    error "Failed add OAM helm repository."
+    return 1
+  fi
+
+  log "Create weblogic-operator namespace"
+  kubectl create namespace weblogic-operator
```
kminder, line 237: This should also go into the verrazzano-system namespace.

operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
```diff
@@ -12,6 +12,20 @@ UNINSTALL_DIR=$SCRIPT_DIR/..
...
 
+# This makes an attempt to uninstall the WebLogic Kubernetes operator, ignoring errors so that this can work
+# in the case where there is a partial installation
+
+function uninstall_weblogic_operator {
+  log "Uninstalling WebLogic Kubernetes operator"
+  helm delete weblogic-operator --namespace weblogic-operator
+
+  log "Delete weblogic-operator namespace"
+  kubectl delete ns weblogic-operator
```
kminder, line 23: namespace

operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
```diff
@@ -59,4 +73,27 @@ function delete_verrazzano() {
...
\ No newline at end of file
+# This makes an attempt to uninstall OAM, ignoring errors so that this can work
+# in the case where there is a partial installation
+function uninstall_oam {
+
+  log "Uninstall OAM"
+  helm delete oam --namespace oam-system crossplane-master/oam-kubernetes-runtime
+
+  log "Delete OAM roles"
+  kubectl delete clusterrole oam-kubernetes-runtime-oam:system:aggregate-to-controller
```
kminder, line 84: Should we be doing a `helm delete oam ...` instead of or in addition to these delete?

operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
```diff
@@ -59,4 +73,27 @@ function delete_verrazzano() {
...
+  kubectl delete clusterrolebinding oam-kubernetes-runtime-oam:system:aggregate-to-controller
+  kubectl delete clusterrolebinding cluster-admin-binding-oam
+  kubectl delete ScopeDefinition healthscopes.core.oam.dev
+  kubectl delete TraitDefinition manualscalertraits.core.oam.dev
+  kubectl delete WorkloadDefinition containerizedworkloads.core.oam.dev
+
+  log "Delete oam-system namespace"
```
kminder, line 91: Should use verrazzano-system so don't delete that here.

operator/scripts/install/chart/crds/oam.verrazzano.io_ingresstraits.yaml
kminder: Copying the CRDs into `operator/scripts/install/chart/crds` seems to be the direction things are heading as the model/binding CRDs are currently being copied here.  I agree that this might need some more thought but for this change we should not break the pattern.

operator/scripts/install/chart/values.yaml
```diff
@@ -92,6 +92,13 @@ verrazzanoAdmissionController:
   caBundle:
   RequestMemory: 15Mi
 
+verrazzanoApplicationOperator:
+  name: verrazzano-application-operator
+  namespace: verrazzano-application
+  imageName: iad.ocir.io/weblogicondocker/verrazzano-application-operator-dev
+  imageVersion: local-e626d89
```
kminder, line 99: What is this imageVersion and where does it come from?

kminder (changes requested):

operator/scripts/install/chart/values.yaml
```diff
@@ -92,6 +92,13 @@ verrazzanoAdmissionController:
   caBundle:
   RequestMemory: 15Mi
 
+verrazzanoApplicationOperator:
+  name: verrazzano-application-operator
+  namespace: verrazzano-application
+  imageName: iad.ocir.io/weblogicondocker/verrazzano-application-operator-dev
```
pfmackin, line 98: this imageName should be ghcr.io.  We will also need a temp secret to access it.  Check with @mgianatagh

operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
pfmackin: helm delete is done at line 81, but it doesn't delete those other resources for some reason.

PR #403, opened 2020-12-04
VZ-1747 : Set a default value for CRD property protocol

bhatpmk:
This PR fixes the validation issue mentioned in VZ-1747, by setting a default value for protocol.  In a v1 CRD a property specified as a list-map item key must be a required property or be given a default value.

As part of this change, I have not relaxed the validation for the k8s version in operator/scripts/install/1-install-istio.sh, as supporting k8s v.1.18.10 requires some more changes, which will be taken care by VZ-1808

Acceptance test with the change, with k8s v1.17.9 for OKE Cluster : https://build.verrazzano.io/job/verrazzano-in-cluster-oke-acceptance-test-suite/job/master/125/

### DIFF
--- a/operator/scripts/install/chart/crds/verrazzano.io_verrazzanomodels_crd.yaml
+++ b/operator/scripts/install/chart/crds/verrazzano.io_verrazzanomodels_crd.yaml
@@ -1684,6 +1684,7 @@ spec:
                                       description: Protocol for port. Must be UDP,
                                         TCP, or SCTP. Defaults to "TCP".
                                       type: string
+                                      default: TCP
                                   required:
                                   - containerPort
                                   type: object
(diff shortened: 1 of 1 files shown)


agv426 (approved):

bobdonat (changes requested):
This is a generated file from the verrazzano-crd-generator repo.  The  change needs to be done in that repo.  The generated file is then copied to this repo.

agv426:
> This is a generated file from the verrazzano-crd-generator repo. The change needs to be done in that repo. The generated file is then copied to this repo.

thanks, didn't realize this was generated. will remove the approval

bhatpmk:
> This is a generated file from the verrazzano-crd-generator repo. The change needs to be done in that repo. The generated file is then copied to this repo.

Thank you @bobdonat , I will make the change as appropriate

PR #405, opened 2020-12-04
Use generated operator.yaml when running integ-tests 

bobdonat:
This pull request has a change to use the generated operator.yaml when running integ-test during a Jenkins run.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -198,6 +198,17 @@ pipeline {
             }
         }
 
+        stage('Generate operator.yaml') {
+            when { not { buildingTag() } }
+            steps {
+                sh """
+                    cd ${GO_REPO_PATH}/verrazzano/operator
+                    cat config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|${env.DOCKER_REPO}/${env.DOCKER_NAMESPACE}/${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}|g" > deploy/operator.yaml
+                    cat config/crd/bases/install.verrazzano.io_verrazzanos.yaml >> deploy/operator.yaml
+                   """
+            }
+        }
+
         stage('Integration Tests') {
             when { not { buildingTag() } }
             steps {
@@ -240,7 +251,7 @@ pipeline {
             }
         }*/
 
-        stage('Update operator.yaml') {
+        stage('Commit operator.yaml') {
             when {
                 allOf {
                     not { buildingTag() }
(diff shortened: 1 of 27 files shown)


Jenkinsfile
```diff
@@ -259,10 +271,8 @@ pipeline {
                     git config --global user.name $DOCKER_CREDS_USR
                     git config --global user.email "70212020+verrazzanobot@users.noreply.github.com"
                     git checkout -b ${env.BRANCH_NAME}
-                    cat config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|${env.DOCKER_REPO}/${env.DOCKER_NAMESPACE}/${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}|g" > deploy/operator.yaml
-                    cat config/crd/bases/install.verrazzano.io_verrazzanos.yaml >> deploy/operator.yaml
                     git add deploy/operator.yaml
-                    git commit -m "Jenkins update to operator image in operator.yaml"
+                    git commit -m "[verrazzanobot] Update verrazzano-platform-operator image version to ${DOCKER_IMAGE_TAG} in operator.yaml"
```
mgianatagh, line 275: I am not sure if we want to publish this username.  You can use ${DOCKER_CREDS_USR} instead.

Jenkinsfile
```diff
@@ -259,10 +271,8 @@ pipeline {
                     git config --global user.name $DOCKER_CREDS_USR
                     git config --global user.email "70212020+verrazzanobot@users.noreply.github.com"
```
mgianatagh, line 272: change this reference for username to ${DOCKER_CREDS_USR}

Jenkinsfile
bobdonat: This was the message suggesged by @gilbode

Jenkinsfile
bobdonat: What will that change?

Jenkinsfile
bobdonat: Wouldn't DOCKER_CREDS_USR be a different user

gilbode: I'm indifferent to whether it says verrazzanobot or not - I was modeling the message after things like maven release plugin which would say maven-release-plugin in the brackets.  All I really wanted was something that would allow us to easily filter the commit messages to get only the ones produced by the automation.  I was thinking of verrazzanobot more as a process vs. a user.  Feel free to suggest alternative - but it should not be a username it needs to be a fixed string.

mgianatagh (approved):

PR #406, opened 2020-12-04
Update verrazzano CRD to use apiextensions.k8s.io/v1

bobdonat:
This pull request changes the verrazzano CRD to use apiextensions.k8s.io/v1 instead of  apiextensions.k8s.io/v1beta1


mgianatagh:
The CRD is now a versioned document.  Is that an intended behavior we want going forward?

I also notice that more of the CRD changed than expected.  It appears a previous change should have resulted in generating a new CRD file, but that did not happen.

mgianatagh (dismissed):

mgianatagh (approved):

PR #416, opened 2020-12-07
fix examples links

desagar:


mgianatagh (approved):

mgianatagh:
Merging change before acceptance test suite runs because the README and script changes are not tested.

PR #418, opened 2020-12-07
Fix bug where the verrazzano CR Certificate.CA fields were ignored

bobdonat:
ClusterResourceNamespace and SecretName when specified in verrazzano resource were being ignored by platform-operator. This is fixed with this pull request.


guoyongzhang (approved):
Looks good

mgianatagh (approved):

PR #421, opened 2020-12-08
JIRA:VZ-1830 streamline usage of secret for OCI DNS installations

jmaron99:
- Modify installation scripts to leverage existing, generated secret
- Modify operator code to leverage existing secret and pass it to installation scripts via config property


operator/controllers/verrazzano_controller.go
```diff
@@ -117,6 +123,16 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	return ctrl.Result{}, err
 }
 
+func doesSecretExist(r *VerrazzanoReconciler, vz *installv1alpha1.Verrazzano) error {
```
desagar, line 126: since this only looks for the specific secret, maybe call this doesOCIDNSConfigSecretExist

operator/controllers/verrazzano_controller.go
```diff
@@ -117,6 +123,16 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	// ensure the secret exists before proceeding
+	secret := &corev1.Secret{}
+	err := r.Get(context.TODO(), types.NamespacedName{Name: vz.Spec.DNS.OCI.OCIConfigSecret, Namespace: vz.Namespace}, secret)
```
desagar, line 129: I find this confusing - I would expect a method called r.Get(...) to actually get and return something. I guess it's getting the secret in the output parameter.
Also, I'd expect a method called doesXXX to return a boolean, rather than just a nil error. Maybe the Golang idioms are different.

operator/controllers/verrazzano_controller.go
jmaron99: Fine but I'm not attempting another acceptance run for this change....

operator/controllers/verrazzano_controller.go
jmaron99: 1.  the reconciler provides a generic resource retrieval api that I'm making use of here.  you'll see multiple examples in the same file.
2. Yes - the idioms do appear to differ

operator/scripts/install/create_oci_config_secret.sh
```diff
@@ -57,11 +57,11 @@ function usage {
     exit 1
 }
 
-OUTPUT_FILE=$TMP_DIR/oci-config.yaml
+OUTPUT_FILE=$TMP_DIR/oci.yaml
 
 OCI_CONFIG_FILE=~/.oci/config
 SECTION=DEFAULT
-KUBE_KEY_NAME=oci-config
+KUBE_KEY_NAME=oci
```
desagar, line 64: another rename variable request for KUBE_KEY_NAME to something that says it's the dns config secret name - if you're going to rerun tests anyway. If not, later is fine.

desagar (dismissed):
From install scripts side, this looks good to me. Prefer someone who knows operator better to look at that code as well.

operator/controllers/verrazzano_controller.go
```diff
@@ -105,6 +104,13 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		return reconcile.Result{}, err
 	}
 
+	// if an OCI DNS installation, make sure the secret required exists before proceeding
```
bobdonat, line 107: this should only be needed temporarily until we have a webhook to catch this

operator/controllers/verrazzano_controller.go
```diff
@@ -105,6 +104,13 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	if vz.Spec.DNS.OCI != (installv1alpha1.OCI{}) {
+		err := doesOCIDNSConfigSecretExist(r, vz)
```
bobdonat, line 109: nit.  the function could be defined as "func (r *VerrazzanoReconciler) doesOCIDNSConfigSecretExist" and then no need to pass r

operator/internal/installjob/install_config.go
```diff
@@ -120,18 +120,13 @@ type OCIConfigAuth struct {
 
 // DNSAuth provides the authentication structures for external DNS
 type DNSAuth struct {
-	PrivateKeyAuth OCIConfigAuth `yaml:"privateKeyAuth"`
+	PrivateKeyAuth OCIConfigAuth `yaml:"auth"`
```
bobdonat, line 123: should this field be named Auth.  I think typically the field name is the same as the yaml tag.

operator/scripts/install/2-install-system-components.sh
```diff
@@ -169,6 +161,17 @@ function install_cert_manager()
 function install_external_dns()
 {
   if [ "$DNS_TYPE" == "oci" ]; then
+    if ! kubectl get secret $OCI_DNS_CONFIG_SECRET -n cert-manager ; then
+      # secret does not exist, so copy the configured oci config secret from default namespace.
+      # Operator has already checked for existence of secret in default namespace
```
bobdonat, line 166: I think the operator is checking for the secret in the namespace of the vz resource which can be anywhere.

operator/internal/installjob/install_config.go
jmaron99: I left the name as is since Mike had mentioned the desire to explore other auth mechanisms, though I'm not certain the OCI plugin for external DNS supports other auth methods.  But this would clearly indicate we are using private key auth while creating the appropriate json field name for the generated secret.

operator/scripts/install/2-install-system-components.sh
jmaron99: I ran into  installation issues where the secret wasn't found by the external dns and/or cert manager - they were explicitly looking for it to be co-located in the cert-manager namespace.  In addition, the secret needs the "compartment" to be added for external dns component.  I though of adding it on the front end, but that would require some additional user input, and wouldn't negate the need for copying the secret to the cert-manager namespace (which doesn't exist at pre-installation)

bobdonat: this code assumes that this secret is in the default namespace while the operator checks for the existence of this secret in the namespace of the vz resource which could be in a namespace other then default.

jmaron99: Oh - now I understand :). Interestingly my testing shows that the operator check passes, so the VZ resource appears to be in the default namespace but I imagine that is not necessarily going to be the case moving forward?  I could specify the "default" namespace in the operator code.

operator/controllers/verrazzano_controller.go
pfmackin: what is confusing is that secret is defined as an &corev1.Secret vs just a corev1.Secret.  It would be clearer to do that and pass &secret in the call, but a Jon says this pattern is commonly used.

operator/controllers/verrazzano_controller.go
```diff
@@ -105,6 +104,13 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		return reconcile.Result{}, err
 	}
 
+	// if an OCI DNS installation, make sure the secret required exists before proceeding
+	if vz.Spec.DNS.OCI != (installv1alpha1.OCI{}) {
```
pfmackin, line 108: DNS is an optional component and could be nil I think, unless some code is always creating it

operator/controllers/verrazzano_controller.go
pfmackin: agree with Bob this should be changed to a method

operator/scripts/install/2-install-system-components.sh
pfmackin: we should not assume default

jmaron99: @pfmackin @bobdonat I don't understand what the alternative is.  The documentation and the scripts we provide explicitly specify the default namespace for the creation of the secret.  So it's not an assumption - it's a fact or a defined pre-requisite.  I see nothing wrong with checking for the existence of the secret in the default namespace within the operator.  Do you have a different proposal?

jmaron99: Right.  Which appears to be "default" in this case (repeated tests show that the secret check is passing)

operator/controllers/verrazzano_controller.go
jmaron99: Can you think of a situation in which we wouldn't specify DNS?  I think for the time being this is perfectly safe code.

bobdonat: I think this code is fine.  Checking if dns is specified in the spec.

operator/scripts/install/2-install-system-components.sh
bobdonat: but I think things are only going to work if the vz resource is created in the default namespace.  Have you tried creating the vz resource in some other namespace

jmaron99: I guess that's true.  I'll use "default"

jmaron99:
@bobdonat @desagar @pfmackin successful build post merges:  https://build.verrazzano.io/job/verrazzano-oci-dns-acceptance-tests/job/jmaron%252FVZ-1830/15/

Maybe we can get together Monday morning and agree on what is required to approve these PRs (acceptance test and this one)?

operator/controllers/verrazzano_controller.go
pfmackin: The DNS part is ok but if DNS is nil, then DNS.OCI access will panic.  DNS is optional so I don't know that Spec.DNS will always be there.

jmaron99: Please explain what it would mean to deploy without dns specified. Does it mean you’re deploying an app with no exposed endpoints accessible?  I don’t understand how something so fundamental can be optional. 

Also - this PR is languishing for days without closure. What do I need to do to actually bring this to a conclusion?

bobdonat (approved):

PR #436, opened 2020-12-11
VZ-1844.  Restrict upgrade to version change only

michael-cico:
- validate spec for upgrade scenario
- some cleanup left over from VZ-1788 comments that came in at the end


operator/controllers/verrazzano_controller.go
```diff
@@ -97,7 +99,15 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		// If the version is specified and different than the current version of the installation
 		// then proceed with upgrade
 		if len(vz.Spec.Version) > 0 && vz.Spec.Version != vz.Status.Version {
-			return r.reconcileUpgrade(log, req, vz)
+			updated, err := r.installSpecUpdated(ctx, vz)
```
pfmackin, line 102: this logic should go in the reconcileUgrade method

operator/controllers/verrazzano_controller.go
```diff
@@ -511,43 +521,80 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
...
-				r.Log.Errorf("Unable to create installer config map %s: %v", configMapName, err)
-				return err
-			}
+		// Update the configmap if the data has changed
+		currentConfigData := installConfig.Data[configDataKey]
+		if currentConfigData != installSpec {
+			installConfig.Data[configDataKey] = installSpec
+			return r.Update(ctx, installConfig)
+		}
+	} else if errors.IsNotFound(err) {
```
pfmackin, line 536: Normally operators go into the err != nil block first, can you swap this?

operator/controllers/verrazzano_controller.go
```diff
@@ -511,43 +521,80 @@ func (r *VerrazzanoReconciler) setUninstallCondition(log *zap.SugaredLogger, job
...
+			r.Log.Errorf("Unable to create installer config map %s: %v", configMapName, err)
+			return err
 		}
 	}
-	return err
+	return nil
+}
+
+// installSpecUpdated Returns true if the current Spec field of the Verrazzano resource does not match what was saved in the internal ConfigMap
+func (r *VerrazzanoReconciler) installSpecUpdated(ctx context.Context, vz *installv1alpha1.Verrazzano) (bool, error) {
```
pfmackin, line 564: can you rename to isInstallSpecUpdated

operator/controllers/verrazzano_controller.go
```diff
@@ -97,7 +99,15 @@ func (r *VerrazzanoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		// If the version is specified and different than the current version of the installation
 		// then proceed with upgrade
 		if len(vz.Spec.Version) > 0 && vz.Spec.Version != vz.Status.Version {
-			return r.reconcileUpgrade(log, req, vz)
+			updated, err := r.installSpecUpdated(ctx, vz)
+			if err != nil {
+				return reconcile.Result{}, err
+			} else if updated {
```
pfmackin, line 105: The version field is part of the spec, so we need to make sure that filed is the only one that changed to allow upgrade.

michael-cico: OK....since we are doing the version field check before we get to this, or should be, I didn't think we needed to check it again.

michael-cico: OK, based on the comment about reconcileUpgrade above, I can see where we move the that check...let me rework it a bit based on the above.

operator/controllers/verrazzano_controller.go
michael-cico: renamed to isValidUpgradeRequest()

operator/controllers/verrazzano_controller.go
michael-cico: Doing more checks in the upgrade validation, comparing the version fields using the semver package...not sure if that overlaps with Bob's/Bhat's work, but we can sort that out.

operator/controllers/verrazzano_controller.go
michael-cico: Done

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	requestedSemVer, err := semver.NewSemVersion(requestedVersion)
+	if err != nil {
+		return err
+	}
+	chartSemVer, err := getCurrentChartVersion()
+	if err != nil {
+		return err
+	}
+	if requestedSemVer.CompareTo(chartSemVer) != 0 {
```
pfmackin, line 59: can we make this case insensitive

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+		return fmt.Errorf("Requested version %s does not match chart version %s", requestedSemVer.VersionString, chartSemVer.VersionString)
+	}
+	return nil
+}
+
+// CheckValidateUpgradeRequest Ensure that for the upgrade case only the version field has changed
+func CheckValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec) error {
```
pfmackin, line 66: can we call this ValidateUpgradeRequest

pfmackin: should be ... ensures that ...

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+	// Short-circuit if the version strings are the same
+	if currentSpec.Version == newSpec.Version {
+		return nil
+	}
+	if len(newSpec.Version) == 0 {
+		// if we get here, the current version is not empty, but the new version is
```
pfmackin, line 72: i am missing where you are checking if current version is not empty

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// Verify that the new version request is > than the currently version
+	if len(currentSpec.Version) > 0 {
+		currentSemVer, err := semver.NewSemVersion(currentSpec.Version)
+		if err != nil {
+			// Unable to parse the current spec version; this should never happen
+			return err
+		}
+		if requestedSemVer.CompareTo(currentSemVer) < 0 {
```
pfmackin, line 93: will this work going from 0.0.9 to 0.0.10 ?

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: you mean for the "v" part of the version string?

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: The == check above covers the case where both strings are empty, or both are not empty and equal.

If we pass the == check, and newSpec.Version is empty, then the current version can't be.  Otherwise if both were empty the == check would have returned, right?

michael-cico: Does that seem right to you?

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: It should.  I'll add a unit test for that, but basically we break the major/minor/patch into integer values, and compare each integer value.

michael-cico: Just added a test for that, and it works.

I should add better comparison methods for > or <.  Next pass.

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: Made the "v" part case-insensitive.

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: Done

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: made the "v" in the version string regex case-insensitive

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: Done

operator/api/verrazzano/v1alpha1/validate.go
pfmackin: sort of.  If the newspec is empty then we should not upgrade, regardless of what is in the current spec.

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+		return fmt.Errorf("Requested version %s does not match chart version %s", requestedSemVer.VersionString, chartSemVer.VersionString)
+	}
+	return nil
+}
+
+// ValidateUpgradeRequest Ensures that for the upgrade case only the version field has changed
+func ValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec) error {
+	// Short-circuit if the version strings are the same
+	if currentSpec.Version == newSpec.Version {
+		return nil
```
pfmackin, line 69: so if the version did not change then this is not considered an upgrade request, I think you are handling that but just wanted to mention it.

Jenkinsfile
```diff
@@ -186,7 +186,7 @@ pipeline {
                       failNoReports: true,
                       onlyStable: false,
                       fileCoverageTargets: '100, 0, 0',
-                      lineCoverageTargets: '90, 90, 90',
+                      lineCoverageTargets: '85, 85, 85',
```
pfmackin, line 189: We should fix this by moving the code out of this package, but this is ok for now.  Can you create a JIRA for that

operator/api/verrazzano/v1alpha1/validate_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
+	"testing"
+)
+
+const validChartYAML = `
+# Copyright (c) 2020, Oracle and/or its affiliates.
```
pfmackin, line 14: we don't need the copyright here

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+var readFileFunction func(string) ([]byte, error) = ioutil.ReadFile
+
+// getCurrentChartVersion Load the current Chart.yaml into a chartVersion struct
+func getCurrentChartVersion() (*semver.SemVersion, error) {
+	chartDir := env.VzChartDir()
+	chartBytes, err := readFileFunction(chartDir + "/Chart.yaml")
+	if err != nil {
+		return nil, err
+	}
+	chartVersion := &chartVersion{}
```
pfmackin, line 34: need a defer close.

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -0,0 +1,105 @@
...
+	Version     string
+	AppVersion  string
+}
+
+var readFileFunction func(string) ([]byte, error) = ioutil.ReadFile
+
+// getCurrentChartVersion Load the current Chart.yaml into a chartVersion struct
+func getCurrentChartVersion() (*semver.SemVersion, error) {
+	chartDir := env.VzChartDir()
+	chartBytes, err := readFileFunction(chartDir + "/Chart.yaml")
```
pfmackin, line 30: wonder if there is another way other than having to replace this function.  I understand you are doing it for unit tests

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: Yeah, I think the comment was just me thinking through the process.  The newspec can be empty, but it's only an issue if newspec is empty and the current spec is not, at least that's the way I treated it.

operator/api/verrazzano/v1alpha1/validate_test.go
```diff
@@ -0,0 +1,337 @@
...
+		readFileFunction = ioutil.ReadFile
+	}()
+	currentSpec := &VerrazzanoSpec{
+		Profile: "dev",
+	}
+	newSpec := &VerrazzanoSpec{
+		Version: "v0.7.0",
+		Profile: "dev",
+	}
+	assert.Nil(t, ValidateUpgradeRequest(currentSpec, newSpec))
```
pfmackin, line 42: you can use assert.Error and assert.NoError, but this is ok

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: Yes, basically, this method is to detect a) if it it seems like an upgrade request (new version) and b) that we can allow it.

Jenkinsfile
michael-cico: Will do

operator/api/verrazzano/v1alpha1/validate_test.go
michael-cico: True, lol, I copied it out of the actual chart code.

michael-cico: It's what happens when you're in an hurry...

Jenkinsfile
michael-cico: Opened VZ-1897 to track.

operator/api/verrazzano/v1alpha1/validate.go
michael-cico: We might be able to mock it, or do something else, but I did it this way for expediency given the time frame.

operator/api/verrazzano/v1alpha1/validate_test.go
michael-cico: OK

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+type SemVersion struct {
+	Major         int64
+	Minor         int64
+	Patch         int64
+	Prerelease    string
+	Build         string
+	VersionString string
+}
+
+var _regex *regexp.Regexp = nil
```
pfmackin, line 27: why are the vars prefixed with _ ?

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+var _initRegEx sync.Mutex
+
+func getRegex() *regexp.Regexp {
+	if _regex != nil {
+		return _regex
+	}
+	_initRegEx.Lock()
+	defer _initRegEx.Unlock()
```
pfmackin, line 35: Why do we need a lock?

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+	var err error
+	_regex, err = regexp.Compile(semverRegex)
+	if err != nil {
+		panic(fmt.Sprintf("Error compiling semver regex: %v", err))
```
pfmackin, line 39: we should not panic, log and error or just return an error

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return _regex
+}
+
+// NewSemVersion Create an instance of a SemVersion
+func NewSemVersion(version string) (*SemVersion, error) {
+	if len(version) == 0 {
+		return nil, errors.New("SemVersion string can not be empty")
```
pfmackin, line 47: cannot be empty

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+	allMatches := regex.FindAllStringSubmatch(version, -1)
+	zap.S().Debugf("allMatches: %v", allMatches)
+	if len(allMatches) == 0 {
+		return nil, fmt.Errorf("Invalid version string %s", version)
+	}
+
+	versionComponents := allMatches[0]
+	zap.S().Debugf("components: %v", versionComponents)
+	numComponents := len(versionComponents)
+	if numComponents < 3 {
```
pfmackin, line 61: i am assuming none of these errors can ever happen since you used a regex and it succeeded.  Anyway doesn't hurt to keep the checking in

operator/internal/util/semver/semver.go
```diff
@@ -0,0 +1,128 @@
...
+const semverRegex = "^[v|V](0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+
+// SemVersion Implements a basic notion a semantic version (see https://semver.org/, test page: https://regex101.com/r/vkijKf/1/)
+type SemVersion struct {
+	Major         int64
+	Minor         int64
+	Patch         int64
+	Prerelease    string
+	Build         string
+	VersionString string
```
pfmackin, line 24: is VersionString the entire of the semver?

pfmackin (dismissed):
Some comments but nothing blocking.  Can you do a follow up task  to add that special test header to all the tests (GIVEN, etc).

operator/api/verrazzano/v1alpha1/validate_test.go
michael-cico: Fixed

operator/internal/util/semver/semver.go
michael-cico: fixed

operator/internal/util/semver/semver.go
michael-cico: Yes...may not be necessary, but kept it around for now.  Can probably remove it in a subsequent pass.  I'll add it to the cleanup JIRA I opened.

pfmackin: it is fine, just wasn't sure what it was

pfmackin (approved):

pfmackin (approved):

PR #437, opened 2020-12-11
Tvlatas VZ-1848 

agv426:
This adds random selection of the region for acceptance testing to spread the load across regions and reduces the regions to a smaller set. There are similar changes being made to other AT related pipelines as well.


markxnelson (dismissed):

Jenkinsfile
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 def DOCKER_IMAGE_TAG
+def availableRegions = [ "us-ashburn-1", "eu-frankfurt-1", "uk-london-1", "us-phoenix-1" ]
```
agv426, line 5: From discussions we are going to keep the full list of regions rather than the limited list

desagar (approved):
lgtm - we'll have to see in the actual runs whether the randomness is sufficient. Current sample size is too small to judge

PR #438, opened 2020-12-13
KinD support on MAC

guoyongzhang:
On MAC, unlike Linux, container ip is not accessible in docker bridge network.  The workaround is to port forwarding.
Nginx ingresses are now bound to 127.0.0.1.

### DIFF
--- a/operator/scripts/install/config.sh
+++ b/operator/scripts/install/config.sh
@@ -158,8 +158,7 @@ function validate_config_json {
 function get_verrazzano_ingress_ip {
   local ingress_type=$(get_config_value ".ingress.type")
   if [ ${ingress_type} == "NodePort" ]; then
-    # on MAC and Windows, container IP is not accessible.  Port forwarding is required.
-    # ingress_ip=$(kubectl -n ingress-nginx get pods --selector app=nginx-ingress,component=controller -o jsonpath='{.items[0].status.hostIP}')
+    # on MAC and Windows, container IP is not accessible.  Port forwarding from 127.0.0.1 to container IP is needed.
     ingress_ip="127.0.0.1"
   elif [ ${ingress_type} == "LoadBalancer" ]; then
     # Test for IP from status, if that is not present then assume an on premises installation and use the externalIPs hint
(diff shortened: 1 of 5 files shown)


operator/scripts/install/config.sh
```diff
@@ -158,7 +158,8 @@ function validate_config_json {
 function get_verrazzano_ingress_ip {
   local ingress_type=$(get_config_value ".ingress.type")
   if [ ${ingress_type} == "NodePort" ]; then
-    ingress_ip=$(kubectl -n ingress-nginx get pods --selector app=nginx-ingress,component=controller -o jsonpath='{.items[0].status.hostIP}')
+    # on MAC and Windows, container IP is not accessible.  Port forwarding from 127.0.0.1 to container IP is needed.
```
mgianatagh, line 161: Does ingress type of NodePort always imply Mac or Windows?  or always imply the IP must be 127.0.0.1?  Is there an additional profile attribute that could be checked to decide how to derive the ingress_ip for NodePort?

desagar (dismissed):

operator/scripts/install/config.sh
guoyongzhang: For nodeport on linux or mac/windows, this PR is making ingress bound to 127.0.0.1.  In other words, we treat all OS equally.
KinD config is required to have 80/443 porting mapping on 0.0.0.0.

operator/scripts/install/config.sh
```diff
@@ -173,6 +174,7 @@ function get_verrazzano_ingress_ip {
 function get_application_ingress_ip {
   local ingress_type=$(get_config_value ".ingress.type")
   if [ ${ingress_type} == "NodePort" ]; then
+    # on MAC and Windows, container IP is not accessible.  Port forwarding is needed, https://jira.oraclecorp.com/jira/browse/VZ-1902
```
markxnelson, line 177: internal name

markxnelson (changes requested):

markxnelson (approved):

PR #439, opened 2020-12-13
Add generation of client code

mgianatagh:
- Add generation of client code for the verrazzano custom resource
- Remove some Makefile dependencies that were automatically generating code


Makefile
```diff
@@ -97,12 +98,26 @@ go-mod:
 	$(GO) mod tidy
 	$(GO) mod vendor
 
+	# go mod vendor only copies the .go files.  Also need
+	# to populate the k8s.io/code-generator folder with the
+	# scripts for generating informer/lister code
+
+	# Obtain k8s.io/code-generator version
+	$(eval codeGenVer=$(shell grep "code-generator" go.mod | awk '{print $$2}'))
```
markxnelson, line 106: we do not want makefiles to automatically generate code - see decision log

markxnelson (changes requested):

Makefile
pfmackin: this client gen should be under the generate target

mgianatagh: this it not generating code, it is populating the vendor folder

However, while reviewing your comment I noticed the target `docker-build` has an pre-existing dependency on `generate` and `manifests`.  I will remove those.

mgianatagh: I pushed a change that removes the pre-existing dependencies on targets that generate code

markxnelson (approved):

pfmackin (approved):

PR #442, opened 2020-12-14
Add validation webhook configuration

bobdonat:
This pull request adds the needed configuration and framework for implementing a validation admissions webhook in the platform-operator.

In addition, a few other fixes were made.
1. Arguments are now passed to the operator in run.sh
2. The create-kubeconfig call was failing in run.sh when launching the operator.  Removed the call when launching the operator.
3. Changes to logging to make logging consistent.
4. Set the code coverage threshold to 90% instead of having a moving target.


markxnelson (approved):

operator/deploy/operator.yaml
```diff
@@ -46,17 +46,68 @@ spec:
         - name: verrazzano-platform-operator
```
pfmackin, line 46: why do we have operator.yaml and verrazzano-platform-operator.yaml?

operator/internal/certificates/certicates.go
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2020, Oracle and/or its affiliates.
```
pfmackin, line 1: package name (and directory name) should be singular

operator/deploy/operator.yaml
bobdonat: verrazzano-platform-operator.yaml is a template
operator.yaml is generated from the template and additionally includes the CRD

operator/internal/certificates/certicates.go
```diff
@@ -0,0 +1,173 @@
...
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(serverPrivKey),
+	})
+
+	err = os.MkdirAll(certDir, 0666)
+	if err != nil {
+		return nil, err
+	}
+
+	err = writeFile(fmt.Sprintf("%s/tls.crt", certDir), serverCertPEM)
```
pfmackin, line 123: Could this be an emphermal platform operator pod dir?   If so we will need to create the cert every time the pod starts.

operator/internal/certificates/certicates.go
bobdonat: Is that a GO thing?  We have other packages such as controllers.  I will change it.

operator/internal/certificates/certicates.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = writeFile(fmt.Sprintf("%s/tls.key", certDir), serverPrivKeyPEM)
```
pfmackin, line 128: Do we need any special file permissions on the private key?

operator/internal/certificates/certicates.go
bobdonat: I don't see this as a problem that the certs are created when the pod is started.
Do yo know of any specific problem this will cause?

operator/deploy/operator.yaml
pfmackin: ok it should be in the config dir or someplace else.  Not an issue for this PR

operator/internal/certificates/certicates.go
pfmackin: yeah it is a go best practice, file name too.  Sorry for not mentioning it earlier.  Should not block the merge.

bobdonat: I will make this change in my next branch for the state stuff

operator/internal/certificates/certicates.go
```diff
@@ -0,0 +1,173 @@
...
+		Bytes: serverCertBytes,
+	})
+
+	serverPrivKeyPEM = new(bytes.Buffer)
+	_ = pem.Encode(serverPrivKeyPEM, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(serverPrivKey),
+	})
+
+	err = os.MkdirAll(certDir, 0666)
```
pfmackin, line 118: Who is responsible for getting rid of an existing cert/key if the container restarts (but not the pod)

operator/internal/certificates/certicates.go
bobdonat: I am not sure if any special permissions are needed.

operator/internal/certificates/certicates.go
bobdonat: The cert/key will be overwritten if the container restarts

operator/internal/certificates/certicates.go
pfmackin: no what i am saying is that the cert needs to be created/re-created when the pod starts.

pfmackin (approved):

PR #453, opened 2020-12-17
Create release v0.7.0

pfmackin:
Create release v0.7.0


operator/scripts/install/4-install-keycloak.sh
```diff
@@ -37,7 +37,8 @@ function install_mysql {
   fi
 
   log "Update MySQL configuration template"
-  sed -e "s|MYSQL_IMAGE_TAG|${MYSQL_IMAGE_TAG}|g" \
+  sed -e "s|MYSQL_IMAGE|${MYSQL_IMAGE}|g" \
```
mgianatagh, line 40: Does this have the right MYSQL changes?  The original transaction was reverted recently and then an updated version of the PR was committed today.

markxnelson (dismissed):

operator/scripts/install/4-install-keycloak.sh
pfmackin: fixed

PR #454, opened 2020-12-17
Move updating of webhook configuration to init container

bobdonat:
This pull request moves the creation of certificates and update of the validatingwebhookconfiguration resource to an init container.  With this change the webhook configuration will be up-to-date when the operator pod goes to a ready state.

### DIFF
--- a/operator/main.go
+++ b/operator/main.go
@@ -64,7 +64,7 @@ func main() {
 	// validatingWebhookConfiguration resource before the operator container runs.
 	if initWebhooks {
 		setupLog.Info("Setting up certificates for webhook")
-		caCert, err := certificates.SetupCertificates(certDir)
+		caCert, err := certificates.CreateCertificates(certDir)
 		if err != nil {
 			setupLog.Error(err, "unable to setup certificates for webhook")
 			os.Exit(1)
--- a/operator/main.go
+++ b/operator/main.go
@@ -64,7 +64,7 @@ func main() {
 	// validatingWebhookConfiguration resource before the operator container runs.
 	if initWebhooks {
 		setupLog.Info("Setting up certificates for webhook")
-		caCert, err := certificates.CreateCertificates(certDir)
+		caCert, err := certificates.CreateWebhookCertificates(certDir)
 		if err != nil {
 			setupLog.Error(err, "unable to setup certificates for webhook")
 			os.Exit(1)
--- a/operator/main.go
+++ b/operator/main.go
@@ -38,6 +38,7 @@ func main() {
 	var enableLeaderElection bool
 	var certDir string
 	var enableWebhooks bool
+	var initWebhooks bool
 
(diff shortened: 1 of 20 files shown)


jmaron99:
Just so I'm clear on the approach:

The theory here is that performing the cert initialization in the init container, prior to when the pod comes up as ready, will ensure that the pod startup time is shorter.  In addition, the API server, even though it is configured to know of the webhook, will not actually attempt forwarding a request to the webhook/operator till the pod state is ready.  Correct?

operator/main.go
```diff
@@ -46,6 +47,8 @@ func main() {
 	flag.StringVar(&certDir, "cert-dir", "/etc/webhook/certs", "The directory containing tls.crt and tls.key.")
 	flag.BoolVar(&enableWebhooks, "enable-webhooks", true,
 		"Enable webhooks for the operator")
+	flag.BoolVar(&initWebhooks, "init-webhooks", false,
```
pfmackin, line 50: Can you call this createWebhookCert since that is all it does (i think)

bobdonat: Yes, I will do that.

bobdonat:
> Just so I'm clear on the approach:
> 
> The theory here is that performing the cert initialization in the init container, prior to when the pod comes up as ready, will ensure that the pod startup time is shorter. In addition, the API server, even though it is configured to know of the webhook, will not actually attempt forwarding a request to the webhook/operator till the pod state is ready. Correct?

This will not shorten the pod startup time.  This will delay the  pod going to a ready state (the init code takes 6-10 seconds to create certs).  We are documenting that after applying operator.yaml you should execute the command "kubectl -n verrazzano-install rollout status deployment/verrazzano-platform-operator".   The webhook server is more likely to be available after this command completes since the init code has already run.

jmaron99:
> > Just so I'm clear on the approach:
> > The theory here is that performing the cert initialization in the init container, prior to when the pod comes up as ready, will ensure that the pod startup time is shorter. In addition, the API server, even though it is configured to know of the webhook, will not actually attempt forwarding a request to the webhook/operator till the pod state is ready. Correct?
> 
> This will not shorten the pod startup time. This will delay the pod going to a ready state (the init code takes 6-10 seconds to create certs). We are documenting that after applying operator.yaml you should execute the command "kubectl -n verrazzano-install rollout status deployment/verrazzano-platform-operator". The webhook server is more likely to be available after this command completes since the init code has already run.

Ok. So the idea is: The application of the operator yaml will trigger the init container, it will create the certs and complete, then the operator container will start. When the rollout status command is executed and it returns ready it is more likely to represent an operator with a ready webhook listener?  I imagine we should mimic this sequence in the acceptance tests?

operator/main.go
bobdonat: Fixed

pfmackin (approved):

PR #457, opened 2020-12-18
Update remaining OCR image references to use ghcr.io

mgianatagh:
Update remaining OCR image references to use ghcr.io

### DIFF
--- a/operator/scripts/install/1-install-istio.sh
+++ b/operator/scripts/install/1-install-istio.sh
@@ -104,7 +104,7 @@ function install_istio()
         --set grafana.image.repository=$GRAFANA_REPO \
         --set grafana.image.tag=$GRAFANA_TAG \
         --set prometheus.hub=$GLOBAL_HUB_REPO \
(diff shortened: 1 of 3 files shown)


markxnelson (approved):

operator/scripts/install/1-install-istio.sh
```diff
@@ -104,7 +104,7 @@ function install_istio()
         --set grafana.image.repository=$GRAFANA_REPO \
         --set grafana.image.tag=$GRAFANA_TAG \
         --set prometheus.hub=$GLOBAL_HUB_REPO \
-        --set prometheus.tag=v2.13.1 \
+        --set prometheus.tag=$PROMETHEUS_TAG \
```
pfmackin, line 107: where is PROMETHEUS_TAG getting set?

mgianatagh: in common.sh.  I don't know why we had a version specified in this file.  I made it consistent with other components, like GRAFANA_TAG above.

PR #460, opened 2020-12-18
Add additional webhook validations

bobdonat:
This pull request adds a state field to the status section of the verrazzano resource and adds a couple of validations.
Validations added are:
1. Don't allow more than one install - meaning only one verrazzano resource can be applied at any time.
2. Don't allow updates to a verrazzano resource while an upgrade or install is in progress.


operator/api/verrazzano/v1alpha1/verrazzano_webhook_test.go
```diff
@@ -61,6 +76,12 @@ func TestCreateCallbackFailsWithInvalidVersion(t *testing.T) {
 	defer func() {
 		readFileFunction = ioutil.ReadFile
 	}()
+
+	getControllerRuntimeClient = func() (client.Client, error) {
```
michael-cico, line 80: It's too bad we have to replicate a lot of this for all the tests...is there something similar to an @BeforeClass method we could tag like in JUnit?

michael-cico: Probably don't need to do it now, but check out the Main section in the overview on golang testing:

https://golang.org/pkg/testing/

Not quite sure how exactly it works, but might be useful.  This is assuming we always want to override the getClient() function.

operator/api/verrazzano/v1alpha1/verrazzano_webhook.go
```diff
@@ -40,6 +54,11 @@ func (newResource *Verrazzano) ValidateUpdate(old runtime.Object) error {
 	log.Infof("oldResource Annotations: %v, Finalizers: %v, Spec: %v", oldResource.Annotations, oldResource.Finalizers, oldResource.Spec)
 	log.Infof("to Annotations: %v, Finalizers: %v, Spec: %v", newResource.Annotations, newResource.Finalizers, newResource.Spec)
 
+	// Updates are not allowed when an install or an upgrade is in in progress
```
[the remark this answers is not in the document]
bobdonat: Not sure I understand.  This the first check on an update request for a verrazzano resource

operator/api/verrazzano/v1alpha1/verrazzano_webhook_test.go
bobdonat: Looks like a TestMain may help.
https://tip.golang.org/pkg/testing

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -108,3 +111,28 @@ func ValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec
...
+	if len(vzList.Items) != 0 {
+		return fmt.Errorf("Only one install of Verrazzano is allowed")
+	}
+
+	return nil
+}
+
+// ValidateInProgress makes sure there is not an install or upgrade in progress
+func ValidateInProgress(state StateType) error {
+	if state == Installing || state == Upgrading {
```
pfmackin, line 133: should uninstalling be done here.  Also if verrazzano is installing, we don't want to start another install job, but do we want to allow the user to cancel an installation?  Maybe not until we have a more robust component based install.

operator/api/verrazzano/v1alpha1/validate.go
```diff
@@ -108,3 +111,28 @@ func ValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec
...
+
+	if len(vzList.Items) != 0 {
+		return fmt.Errorf("Only one install of Verrazzano is allowed")
+	}
+
+	return nil
+}
+
+// ValidateInProgress makes sure there is not an install or upgrade in progress
+func ValidateInProgress(state StateType) error {
```
pfmackin, line 132: I assume that the webhook is not called when our reconciler updates the status, otherwise we would have a problem.

operator/api/verrazzano/v1alpha1/validate.go
bobdonat: I will add uninstalling here.  
There is another validation called in ValidateCreate that does not allow another vz resource to  be created when one already exist.  This covers the case of starting another install job.
Cancelling an install was not part of this task.  I think we need a Jira to capture that.

operator/api/verrazzano/v1alpha1/validate.go
bobdonat: webhook is not being called when reconciler updates the status

operator/api/verrazzano/v1alpha1/validate.go
bobdonat: I have pushed change to handle uninstalling.

pfmackin: i saw that validation ensuring a single verrazzano

pfmackin (approved):

PR #462, opened 2020-12-19
Tvlatas/reduce regions image pull

agv426:
Reduce list of regions to larger ones only


agv426:
Needs to be from develop will recreate

PR #463, opened 2020-12-19
Cut region list down to larger ones only

agv426:
Cut list down to larger regions


agv426:
We haven't had to reduce the region list, closing this out

PR #464, opened 2020-12-19
Fix make manifests

bobdonat:
The pull requests fixes the Makefile manifests target.  It was generating a bad operator.yaml file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ manifests: controller-gen
 	./operator/hack/update-codegen-verrazzano.sh
 
 	# Re-generate operator.yaml using template yaml file
-	cat operator/config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|$(shell grep "image:" operator/deploy/operator.yaml | awk '{ print $$2 }')|g" > operator/deploy/operator.yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -120,7 +120,7 @@ manifests: controller-gen
 	./operator/hack/update-codegen-verrazzano.sh
 
 	# Re-generate operator.yaml using template yaml file
-	cat operator/config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|$(shell grep "image:" operator/deploy/operator.yaml | awk '{ print $$2 }')|g" > operator/deploy/operator.yaml
+	cat operator/config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|$(shell grep --max-count=1 "image:" operator/deploy/operator.yaml | awk '{ print $$2 }')|g" > operator/deploy/operator.yaml
```
pfmackin, line 123: I don't understand why we are changing operator.yaml on controller-gen target.  Shouldn't that be done after the docker-build?

bobdonat: The make target is only changing operator.yaml if there are changes to the crd or any of the deployment related yaml. The image name in operator.yaml only gets changed in the junkins pipeline

pfmackin (approved):

PR #466, opened 2020-12-21
JIRA:VZ-1911 create startup probe for verrazzano operator

jmaron99:
startup probe added to operator which monitors the webhook listening port to discern when that listener is up and ready to process incoming requests.

### DIFF
--- a/operator/deploy/operator.yaml
+++ b/operator/deploy/operator.yaml
@@ -50,6 +50,10 @@ spec:
             - containerPort: 9443
               name: webhook
               protocol: TCP
+          readinessProbe:
+            httpGet:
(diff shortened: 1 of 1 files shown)


operator/deploy/operator.yaml
```diff
@@ -63,6 +63,14 @@ spec:
             - containerPort: 9443
               name: webhook
               protocol: TCP
+          startupProbe:
```
bobdonat, line 66: If I understand correctly, the pod will restart if the startup probe is not successful after 30 seconds (default).  Should we allow longer for the startup probe to complete

jmaron99: OK

mgianatagh (approved):

bobdonat (approved):

PR #467, opened 2020-12-21
Dockerfile Updates

gilbode:
Dockerfile updates

    - Add .dockerignore

    - Remove extraneous yum install calls

    - No longer install 'release' yum packages as they reset your yum config
      to use the OCI regional mirror when on OCI but when running inside
      Docker on OCI you can get interleaved mirrors (some from yum.oracle.com
      and some from the OCI regional mirror) and when mirrors are being synced
      you can get build failures.

    - Remove extraneous package installs in the release image & avoid some
      other unnecessary layers to make image smaller.


mgianatagh (approved):
I tested an install and uninstall and both worked fine.

PR #469, opened 2020-12-21
Remove creating kubeconfig for platform operator

pfmackin:
Remove creating kubeconfig for platform operator in run.sh.  Also remove setting kubeconfig ENV var from helmcli.go, since it is not needed.

### DIFF
--- a/operator/config/scripts/run.sh
+++ b/operator/config/scripts/run.sh
@@ -65,6 +65,5 @@ elif [ ${MODE} == "UNINSTALL" ]; then
   dump-uninstall-logs 0
 else
   # Run the operator
-    create-kubeconfig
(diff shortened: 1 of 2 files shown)


mgianatagh (approved):

gilbode:
I think there are 2 other spots that set VERRAZZANO_KUBECONFIG - install job and uninstall job.  We should probably remove traces of them.

mgianatagh:
> I think there are 2 other spots that set VERRAZZANO_KUBECONFIG - install job and uninstall job. We should probably remove traces of them.

The creation of that environment variable is required for the install and uninstall scripts to run.  You could remove the environment variable from the job, but instead you would need to create the environment variable in the run.sh script.

bobdonat (approved):

PR #477, opened 2020-12-23
Add keycloak/mysql component during install

bobdonat:
This pull request adds the Keycloak component to the verrazzano CRD and updates the install scripts to allow helm values to be set for Keycloak.  With this change, MySQL which is used by Keycloak can be configured to not use persistent volumes but instead use an emptyDir volume.


operator/api/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -197,6 +200,23 @@ type IstioComponent struct {
...
+	// MySql contains the MySQL component configuration needed for keycloak
+	// +optional
+	MySQL MySQLComponent `json:"mysql,omitempty"`
+}
+
+// MySQLComponent specifies the MySQL configuration
+type MySQLComponent struct {
+	// Arguments for installing MySQL
+	// +optional
+	MySQLInstallArgs []InstallArgs `json:"mysqlInstallArgs,omitempty"`
```
pfmackin, line 217: I don't see where we would specify to use EmptyDir.  Is the user supposed to know the install args to use?

operator/internal/installjob/install_config.go
```diff
@@ -83,21 +83,21 @@ type IngressPort struct {
 	TargetPort int32  `json:"targetPort,omitempty"`
 }
 
-// IngressArg configuration
-type IngressArg struct {
+// InstallArg configuration
+type InstallArg struct {
```
pfmackin, line 87: Not sure why InstallArgs from the API cannot be used here

operator/api/verrazzano/v1alpha1/verrazzano_types.go
bobdonat: the verrazzano CRD would need to have this:
component:
   keycloak:
     mysql:
        mysqlInstallArgs:
         - name: persistence.enabled
           value: "false"
 
The MySQL helm chart has a value persistence.enabled which when set to false uses an emptyDir volume instead of a PV

operator/internal/installjob/install_config.go
```diff
@@ -154,30 +154,42 @@ const (
 	InstallProfileDev InstallProfile = "dev"
 )
 
+// Keycloak configuration
```
pfmackin, line 157: Same comment as before, not sure why we are duplicating structs defined in the API

operator/internal/installjob/install_config.go
```diff
@@ -278,42 +293,52 @@ func getIngressPorts(ports []corev1.ServicePort) []IngressPort {
...
 	}
 
-	return ingressArgs
+	return installArgs
+}
+
+// getKeycloak returns the json representation for the keycloak configuration
+func getKeycloak(keycloak installv1alpha1.KeycloakComponent) Keycloak {
+	return Keycloak{
+		KeycloakInstallArgs: getInstallArgs(keycloak.KeycloakInstallArgs),
```
pfmackin, line 325: We are not consistent using InstallArgs and xyzInstallArgs.   Maybe we should create a JIRA to use use Args across the board, since they may be needed for upgrade also.

operator/internal/installjob/install_config.go
bobdonat: We could look into that but today they are not the same.

pfmackin (approved):

operator/internal/installjob/install_config.go
bobdonat: Where are we not consistent?  Args may be better

bobdonat: I think you are saying no need to say xyzInstallArgs but instead always say installArgs.

bobdonat: they are realy helm args

pfmackin: Yes, at least we just use InstallArgs across the board.  Not a big deal

PR #485, opened 2021-01-07
ineffassign behaviour changed, update to address it

agv426:
ineffassign behaviour changed, update to address it


markxnelson (commented):
not a big deal, but it would be better if we did not commit the go.mod and go.sum changes in this commit, since they are just derived stuff anyway

markxnelson (approved):

agv426:
go.sum/mod were inadvertently changed, deleting this PR and creating a new one

PR #486, opened 2021-01-07
Remove use of vendor directory

gilbode:


Makefile
```diff
@@ -41,7 +41,7 @@ CRD_PATH=operator/config/crd/bases
 CODEGEN_PATH = k8s.io/code-generator
 
 .PHONY: build
-build: go-mod
+build:
```
markxnelson, line 44: did you intentionally leave the `go-mod` target in the makefile?

gilbode: Sort of - mainly b/c i didn't understand the implications of removing it given all the code generator related stuff

gilbode: I was hoping some comment might explain why it could just be tossed :)

markxnelson: there's one way to find out ;)

markxnelson (approved):

PR #489, opened 2021-01-07
Fix the chart location used by upgrade

pfmackin:
A recent merge moved the charts on the image which broken upgrade.
Another merge changed rancher helm schema for values that do not have defaults, which also broke upgrade
Fix both issues.

### DIFF
--- a/operator/internal/config/config.go
+++ b/operator/internal/config/config.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, 2021 Oracle and/or its affiliates.
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package config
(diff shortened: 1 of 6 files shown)


operator/internal/config/config.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, Oracle and/or its affiliates.
+// Copyright (c) 2020, 2021 Oracle and/or its affiliates.
```
markxnelson, line 1: missing comma

markxnelson (changes requested):

michael-cico (dismissed):

michael-cico (approved):

operator/internal/config/config.go
pfmackin: fixed in all the places where i added 2021

markxnelson (approved):
i did not checkout and manually test this, but just reading it, looks ok

PR #495, opened 2021-01-08
Upgrade external-dns

pfmackin:
WIP: Needs more testing

Add upgrade support for external DNS.  This includes the following:

1) Add external-dns values override file
2) Change installer to use the override file, remove fields from common.sh
3) Add generic helm upgrade component
4) Add and minor refactor units tests related to component upgrade
5) Remove upgrade version validation in reconciler since it is done in webhook and preventing disabling the validation.


operator/internal/component/conponent.go
```diff
@@ -1,8 +1,13 @@
-// Copyright (c) 2020, Oracle and/or its affiliates.
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package component
 
+import (
+	config2 "github.com/verrazzano/verrazzano/operator/internal/config"
```
michael-cico, line 7: Just a nit, but why is this import alias called "config2"?

operator/internal/component/helm_component.go
```diff
@@ -0,0 +1,61 @@
...
+	// chartDir is the helm chart directory
+	chartDir string
+
+	// chartNamespace is the namespace passed to the helm command
+	chartNamespace string
+
+	// namespaceHardcoded bool indicates that a component has a hardcoded namespace
+	// and ignores the namespace param passed to Upgrade
+	namespaceHardcoded bool
+
```
michael-cico, line 24: Another nit...maybe call this allowsNamespaceOverride, which indicates you can use chartNamespace?  You can default the behavior to true if that's the majority of cases.  I just think it sounds a little better than "hardcoded".

operator/internal/component/conponent.go
pfmackin: fixed

operator/internal/component/helm_component.go
pfmackin: fixed

operator/internal/component/component.go
```diff
@@ -15,5 +20,17 @@ type Component interface {
 // GetComponents returns the list of components that are installable and upgradeable.
 // The components will be processed in the order items in the array
 func GetComponents() []Component {
-	return []Component{Verrazzano{}, Nginx{}}
+	componentDir := filepath.Join(config.Get().VerrazzanoInstallDir, "components")
+
+	return []Component{
+		Verrazzano{},
+		Nginx{},
```
bobdonat, line 27: Does NGINX need to be a helm component now

operator/internal/component/component_test.go
```diff
@@ -9,10 +9,15 @@ import (
...
 )
 
+// TestGetComponents tests getting the components
+// GIVEN a component
+//  WHEN I call GetComponents
+//  THEN the Get returns the correct components
 func TestGetComponents(t *testing.T) {
 	assert := assert.New(t)
 	comps := GetComponents()
 	assert.Len(comps, 2, "Wrong number of components")
```
bobdonat, line 19: should this be 3

operator/internal/component/component_test.go
```diff
@@ -9,10 +9,15 @@ import (
...
 	assert.Equal(comps[0].Name(), "verrazzano")
-	assert.Equal(comps[1].Name(), "ingress-nginx")
+	assert.Equal(comps[2].Name(), "ingress-nginx")
```
bobdonat, line 21: should these be in order

operator/internal/component/helm_component.go
bobdonat: I don't understand what this doed

operator/internal/component/component_test.go
pfmackin: yes fixed

operator/internal/component/component_test.go
pfmackin: fixed

operator/internal/component/component.go
pfmackin: should be but doesn't have to.  We can cycle back and fix this after

operator/scripts/install/2-install-system-components.sh
```diff
@@ -165,21 +165,11 @@ function install_external_dns()
...
-        --set sources[1]=service \
+        -f $SCRIPT_DIR/components/external-dns-values.yaml \
         --set domainFilters[0]=${DNS_SUFFIX} \
         --set zoneIdFilters[0]=$(get_config_value ".dns.oci.dnsZoneOcid") \
         --set txtOwnerId=v8o-local-${NAME} \
         --set txtPrefix=_v8o-local-${NAME}_ \
-        --set policy=sync \
-        --set interval=24h \
-        --set triggerLoopOnEvent=true \
         --set extraVolumes[0].name=config \
```
bobdonat, line 173: Can't this be in the overrides files
--set extraVolumes[0].name=config
same with
--set extraVolumeMounts[0].name=config

operator/internal/component/helm_component.go
pfmackin: The upgrade-reconcile passes a namespace (from the vz cr).  Most components need to ignore this

bobdonat: Do any components use it

operator/scripts/install/2-install-system-components.sh
pfmackin: I don't want to split the extraVolumes and extraVolumeMount config.   Since the following 2 cannot be in the yaml, i kept them all here
        --set extraVolumes[0].secret.secretName=$OCI_DNS_CONFIG_SECRET \
        --set extraVolumeMounts[0].mountPath=/etc/kubernetes/ \

michael-cico (approved):
Approving, looks like comments I had were addressed.

bobdonat (approved):

PR #503, opened 2021-01-13
Check if component installed before upgrading it

pfmackin:
Certain components may not be installed, like external-dns.  Before doing the helm upgrade of a component, check if it is installed, otherwise upgrade will fail.

### DIFF
--- a/operator/internal/component/helm_component.go
+++ b/operator/internal/component/helm_component.go
@@ -4,7 +4,9 @@
 package component
 
 import (
+	"fmt"
 	"github.com/verrazzano/verrazzano/operator/internal/util/helm"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 // helmComponent struct needed to implement a component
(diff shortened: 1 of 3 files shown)


operator/internal/util/helm/helmcli_test.go
```diff
@@ -68,7 +116,24 @@ func (r goodRunner) Run(cmd *exec.Cmd) (stdout []byte, stderr []byte, err error)
...
+// Run should return an error with stderr contents
 func (r badRunner) Run(cmd *exec.Cmd) (stdout []byte, stderr []byte, err error) {
 	return []byte(""), []byte("error"), errors.New("error")
 }
+
+// Run should assert the command parameters are correct then return a success or error depending on release name
+func (r foundRunner) Run(cmd *exec.Cmd) (stdout []byte, stderr []byte, err error) {
+	assert := assert.New(r.t)
+	assert.Contains(cmd.Path, "helm", "exec command should contain helm")
+	assert.Contains(cmd.Args[0], "helm", "exec args should contain helm")
```
michael-cico, line 128: I think the error message on this check is off, "args should contain helm"?

michael-cico (approved):

operator/internal/component/helm_component.go
```diff
@@ -44,11 +46,24 @@ func (h helmComponent) Name() string {
...
 		namespace = h.chartNamespace
 	}
-	_, _, err := upgradeFunc(h.releaseName, namespace, h.chartDir, h.valuesFile)
+	// Check if the component is installed before trying to upgrade
+	found, err := helm.IsReleaseInstalled(h.releaseName, namespace)
+	if err != nil {
+		return err
+	}
+	if !found {
+		log.Info(fmt.Sprintf("Skipping upgrade of component %s since it is not installed", h.releaseName))
```
bobdonat, line 61: FYI. there is a log.Infof function that can called here instead log.Info with fmt

operator/internal/component/helm_component.go
```diff
@@ -44,11 +46,24 @@ func (h helmComponent) Name() string {
...
 func (h helmComponent) Upgrade(ns string) error {
+	var log = ctrl.Log.WithName("upgrade")
+
 	namespace := ns
 	if h.ignoreNamespaceOverride {
 		namespace = h.chartNamespace
 	}
-	_, _, err := upgradeFunc(h.releaseName, namespace, h.chartDir, h.valuesFile)
+	// Check if the component is installed before trying to upgrade
+	found, err := helm.IsReleaseInstalled(h.releaseName, namespace)
```
bobdonat, line 56: Wondering if the helm_component should have an optional field so that we only call IsReleaseInstalled if optional is true. Most installs are not optional and should fail.

PR #506, opened 2021-01-13
Add patch support for Keycloak

bobdonat:
Add patch support for Keycloak. This includes the following:

- add Keycloak values override file
- remove Keycloak template file
- change installer to use the override file, remove fields from common.sh
- add a helm component for Keycloak


michael-cico (approved):

pfmackin (approved):

operator/scripts/install/components/keycloak-values.yaml
```diff
@@ -1,9 +1,9 @@
-# Copyright (c) 2020, Oracle and/or its affiliates.
+# Copyright (c) 2021, Oracle and/or its affiliates.
```
markxnelson, line 1: should be `2020, 2021`

bobdonat: That was a new file so I only added 2021.  The diffs shows the file moved because I deleted a file with similar content and added a new file.

PR #509, opened 2021-01-13
Jenkins Pipeline Git Commimt

gilbode:
Jenkins pipeline makes a git commit at the end of the build & if there were subsequent commits since the build started git push fails.  To make this failure less likely move the git commit to beginning of Jenkinsfile.  In the event the commit fails, uses the groovy postbuild plugin to update build description to describe why it failed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,6 +102,46 @@ pipeline {
             }
         }
 
+        stage('Generate operator.yaml') {
+            when { not { buildingTag() } }
+            steps {
+                sh """
+                    cd ${GO_REPO_PATH}/verrazzano/operator
+                    cat config/deploy/verrazzano-platform-operator.yaml | sed -e "s|IMAGE_NAME|${env.DOCKER_REPO}/${env.DOCKER_NAMESPACE}/${DOCKER_PLATFORM_IMAGE_NAME}:${DOCKER_IMAGE_TAG}|g" > deploy/operator.yaml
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -99,6 +99,46 @@ pipeline {
...
+                    git checkout -b ${env.BRANCH_NAME}
+                    git add deploy/operator.yaml
+                    git commit -m "[verrazzano] Update verrazzano-platform-operator image version to ${DOCKER_IMAGE_TAG} in operator.yaml"
+                    git push origin ${env.BRANCH_NAME}
+                   """
+            }
+            post {
+                unsuccessful {
+                    script {
+                        currentBuild.description = "Git commit of operator.yaml failed"
```
markxnelson, line 136: could we add an instruction to retry/rerun?

markxnelson (dismissed):
one small request, but approving anyway

Jenkinsfile
gilbode: i thought about that but worried it would be confusing if you didn't read the message in real time.

Jenkinsfile
```diff
@@ -99,6 +99,46 @@ pipeline {
             }
         }
 
+        stage('Generate operator.yaml') {
```
mgianatagh, line 102: should we combine these into a single stage?

gilbode: i think generate operator.yaml might be needed on all branches

PR #520, opened 2021-01-14
The makefile target "manifests" no longer works for the verrazzano-pl…

ddrechse:
…atform-operator


mgianatagh (approved):
I tested and it worked for me.

PR #524, opened 2021-01-20
Add startup probe to the oam-application-operator container

tbeerbower:
Add startup probe to the oam-application-operator container to ensure webhooks are ready.

### DIFF
--- a/oam-application-operator/deploy/verrazzano.yaml_template
+++ b/oam-application-operator/deploy/verrazzano.yaml_template
@@ -58,6 +58,16 @@ spec:
         - name: verrazzano-application-operator
           imagePullPolicy: IfNotPresent
           image: IMAGE_NAME
+          startupProbe:
+            exec:
+              command:
+              - sh
+              - -c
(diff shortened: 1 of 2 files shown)


oam-application-operator/deploy/verrazzano.yaml_template
```diff
@@ -58,6 +58,16 @@ spec:
         - name: verrazzano-application-operator
           imagePullPolicy: IfNotPresent
           image: IMAGE_NAME
+          startupProbe:
```
kminder, line 61: I think this change now needs to go here.
operator/thirdparty/charts/oam-application-operator/templates/verrazzano-application-operator.yaml
We may need to discuss this.

kminder (changes requested):

oam-application-operator/deploy/verrazzano.yaml_template
tbeerbower: Updated to include this.  Acceptance tests https://build.verrazzano.io/blue/organizations/jenkins/verrazzano/detail/tbeerbower%2FVZ-1982/3/pipeline

kminder (approved):

PR #527, opened 2021-01-20
VZ-2011: use a more generic namespace label selector for oam based weblogic workloads

kminder:
Changes the namespace label the WebLogic operator watches to detect Domain CRs from `verrazzano-domain` to `verrazzano-managed`.  This is more future proof as this namespace label can then be used for other Verrazzano features that might apply for a given namespace.


tbeerbower (approved):

wmhopkins (commented):
Lgtm.

PR #530, opened 2021-01-20
Patch MySQL

bobdonat:
Patch MySQL

    Move MySQL values into a new component values YAML file
    Change install to use that new mysql-values.yaml
    Register MySQL component
 
Also, fixed bug where keycloakadmin password was being regenerated each time a helm upgrade was done. 

Tested with a new mysql image


operator/internal/component/registry.go
```diff
@@ -4,60 +4,69 @@
...
 		},
-		helmComponent{
-			releaseName:             "istio",
-			chartDir:                filepath.Join(config.Get().ThirdpartyChartsDir, "istio"),
-			chartNamespace:          "istio-system",
-			ignoreNamespaceOverride: true,
-			valuesFile:              filepath.Join(componentDir, "istio-values.yaml"),
-			preUpgradeFunc:          PreUpgrade,
-		},
+		Verrazzano{},
```
michael-cico, line 49: Just a general question, is there a reason behind the component ordering?  i.e., do we want things to be upgraded in this specific sequence?

No real stake/preference on my part, just wondering if we needed to keep a certain ordering.

operator/internal/component/registry.go
```diff
@@ -4,60 +4,69 @@
...
 		helmComponent{
 			releaseName:             "coherence-operator",
 			chartDir:                filepath.Join(config.Get().ThirdpartyChartsDir, "coherence-operator"),
```
michael-cico, line 52: Use the new thirdpartyDir variable here?

michael-cico (approved):

operator/internal/component/registry.go
bobdonat: I changed the order to the same order as the install order. Doesn't seem to make a difference now but maybe it will in the future

operator/internal/component/registry.go
bobdonat: I noticed that after I rebased and pushed. I plan to fix that in my next branch.

operator/scripts/install/4-install-keycloak.sh
```diff
@@ -75,6 +77,18 @@ function install_keycloak {
...
+  # Create a random secret for the keycloakadmin user
+  kubectl apply -f <(echo "
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ${KCADMIN_SECRET}
+  namespace: ${KEYCLOAK_NS}
+type: Opaque
+data:
+  password: $(cat /dev/urandom | tr -dc "a-zA-Z0-9" | fold -w 10 | head -n 1 | base64)
```
pfmackin, line 89: does this password get displayed in the log file?  If so we need to prevent that.

operator/scripts/install/4-install-keycloak.sh
```diff
@@ -38,23 +39,24 @@ function install_mysql {
...
-      -e "s|MYSQL_IMAGE|${MYSQL_IMAGE}|g" \
-      -e "s|MYSQL_USERNAME|${MYSQL_USERNAME}|g" \
-      $SCRIPT_DIR/config/mysql-values-template.yaml > ${TMP_DIR}/mysql-values-sed.yaml
-
-  # Handle any additional MySQL install args
+  # Handle any additional MySQL install args that cannot be in mysql-values.yaml
   local EXTRA_MYSQL_ARGUMENTS=$(get_mysql_helm_args_from_config)
+  EXTRA_MYSQL_ARGUMENTS="$EXTRA_MYSQL_ARGUMENTS --set mysqlUser=${MYSQL_USERNAME}"
+
+  echo "CREATE DATABASE IF NOT EXISTS keycloak DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;" > ${TMP_DIR}/create-db.sql
```
pfmackin, line 46: What is all this new create-db.sql code for?

bobdonat: This is the database initialization SQL that initially runs. It used to be in a mysql yaml template file and sed was used to replace the mysql username.  Since the template was removed this stuff was moved into the script.

pfmackin (approved):

PR #531, opened 2021-01-21
Upgrade WebLogic Kubernetes Operator

bhatpmk:
Upgrade WebLogic Kubernetes Operator

- Move WebLogic values into a new component values YAML file
- Change install script to use the new weblogic-values.yaml
- Register the component weblogic-operator

The change is tested with a new image and helm version on upgrade is validated. Before upgrading the operator, Bob's Books example was deployed, confirmed that the application endpoints are accessible after the upgrade


markxnelson (approved):

michael-cico (approved):

operator/scripts/install/components/weblogic-values.yaml
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+image: oracle/weblogic-kubernetes-operator:3.1.0
```
markxnelson, line 4: image is available in ghcr now, please change this to use the new address
`ghcr.io/oracle/weblogic-kubernetes-operator:3.1.0`

markxnelson (changes requested):

operator/internal/component/registry.go
```diff
@@ -54,6 +54,13 @@ func GetComponents() []Component {
 			ignoreNamespaceOverride: true,
 			valuesFile:              filepath.Join(componentDir, "coherence-values.yaml"),
 		},
+		helmComponent{
+			releaseName:             "weblogic-operator",
+			chartDir:                filepath.Join(thirdpartyDir, "weblogic-operator"),
+			chartNamespace:          "verrazzano-system",
+			ignoreNamespaceOverride: true,
+			valuesFile:              filepath.Join(componentDir, "weblogic-values.yaml"),
```
bobdonat, line 62: should the name of the overrides file be weblogic-operator-values.yaml to be consistent with the naming of the chart name?  @pfmackin

bobdonat: seems like coherence should also use consistent naming conventions too @pfmackin

pfmackin: i think it is ok for now, we can cleanup during our cleanup pass.

pfmackin (approved):

operator/scripts/install/components/weblogic-values.yaml
markxnelson: resolving on the understanding that this will be fixed in another pr today @pfmackin @gilbode

markxnelson (approved):

PR #534, opened 2021-01-21
Upgrade application operator, move chart and all override files

pfmackin:
Upgrade the application operator and do some chart and overrides cleanup

1. Move application operator chart to operator/helm_config/charts
2. Move all component override values YAML files to operator/helm_config/charts
3. Get rid of the install usage of VERRAZZANO_APPLICATION_OPERATOR_IMAGE env var (still used in docker build)
4. Add application operator override file and register it in the component registry used by upgrade
5. Change application operator to use new chart location and override file


operator/helm_config/charts/verrazzano-application-operator/values.yaml
```diff
@@ -6,7 +6,8 @@ namespace: verrazzano-system
 global:
   imagePullSecrets: []
 
-image: ghcr.io/verrazzano/verrazzano-application-operator:0.5.0-20201218192145-6954bf6
+# NOTE: The image value gets set in the override file verrazzano-application-operator-values.yaml
```
bobdonat, line 9: Should the image tag be included in the values.yaml but empty.  I would expect to see all possible values in this file.

kminder, line 9: Doesn't `image` need to be in here to override it later?

pfmackin: no, but i can put it there as Bob suggested

pfmackin: ok, i can do that

operator/scripts/install/common.sh
```diff
@@ -10,6 +10,8 @@ SCRIPT_DIR=${SCRIPT_DIR:-$(cd $(dirname ${BASH_SOURCE[${#BASH_SOURCE[@]} - 1]});
 # The directory where any generated artifacts should be stored.
 BUILD_DIR="${SCRIPT_DIR}/build"
 CHARTS_DIR=$(cd $SOURCE_DIR/../../../thirdparty/charts; pwd -P)
+HELM_CONFIG_DIR=$(cd $SOURCE_DIR/../../../helm_config; pwd -P)
```
bobdonat, line 13: could this be HELM_CONFIG_OVERRIDES_DIR and then there is no need to do $HELM_CONFIG_DIR/overrides everywhere.

pfmackin: ok, but we need a HELM charts dir also so i can add that

pfmackin: fixed

operator/helm_config/charts/verrazzano-application-operator/values.yaml
pfmackin: fixed

markxnelson (approved):
i don't like the directory name at all fwiw, but approving..

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -62,16 +62,6 @@ spec:
         - name: {{ .Values.name }}
           image: {{ .Values.image }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
-          startupProbe:
```
bobdonat, line 65: why is this being removed?

bobdonat (approved):

PR #545, opened 2021-01-22
oam-kubernetes-runtime patching

bobdonat:
1. Add oam-kubernetes-runtime override file and register it in the component registry used by upgrade
2. Change install of oam-kubernetes-runtime to use new chart location and override file
3. Remove unneeded variables from commom.sh
4. Cleanup file added back from previous merge.

Validated upgrade by changing the oam-kubernetes-runtime image and verifying the hello-helidon app is still working after upgrade.


markxnelson (approved):

michael-cico (approved):

platform-operator/scripts/install/common.sh
```diff
@@ -220,13 +220,4 @@ command -v curl >/dev/null 2>&1 || {
 ##################################################
 ####Constants for Docker images, versions, tags
 ##################################################
-GLOBAL_HUB_REPO=ghcr.io/verrazzano
 GLOBAL_IMAGE_PULL_SECRET=verrazzano-container-registry
-
-GRAFANA_REPO=ghcr.io/verrazzano/grafana
```
pfmackin, line 226: Thanks for cleaning these up, i confirmed they are not used

pfmackin (approved):

PR #554, opened 2021-01-25
Upgrade minor restructuring

pfmackin:
Move some code to the controllers package, since it is closely related to the install/upgrade logic


michael-cico:
Any reason why the common Job structure at internal/k8s/job.go wasn't moved also?

michael-cico (approved):
Had a minor question about moving the internal/k8s/job.go code, if it should go to controllers also or not.

PR #556, opened 2021-01-25
Add tools/scripts directory for general tools, add k8s-dump-cluster.s…

agv426:
Add tools/scripts directory for general tools, add k8s-dump-cluster.sh (initial full dump, will be evolving and more tools will be getting added).

### DIFF
--- a/tools/scripts/k8s-dump-cluster.sh
+++ b/tools/scripts/k8s-dump-cluster.sh
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2021, Oracle and/or its affiliates.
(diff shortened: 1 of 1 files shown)


tools/scripts/k8s-dump-cluster.sh
```diff
@@ -0,0 +1,131 @@
...
+#		pods.json
+#		replicasets.json
+#		replication-controllers.json
+#		services.json
+#		directory per pod
+#			logs.txt
+#	application-configurations.json
+#	crd.json
+#	gateways.json
+#	helm-ls.json
```
agv426, line 66: will remove the helm file comments in the next set of updates. We removed the dependency on helm

michael-cico (approved):

PR #557, opened 2021-01-25
Fix for "operation cannot be fulfilled" messages and upgrade/patching running twice

bobdonat:
This pull request contains two fixes.

1. Handle benign "operation cannot be fulfilled" messages when updating the status.
2. Fix bug where upgrade/patching would be run twice. This was resolved by using the GenerationChangedPredicate event filter which allows the skipping of update events (reconciler not called) unless there are changes to the vz spec.  Updates to the status and metadata fields will no longer result in the reconciler being called.


platform-operator/controllers/controller.go
```diff
@@ -344,14 +350,6 @@ func (r *VerrazzanoReconciler) createUninstallJob(log *zap.SugaredLogger, vz *in
...
-		err = r.setUninstallCondition(log, jobFound, vz)
-		if err != nil {
-			return err
-		}
-
-		// Job created successfully and status set successfully
-		return nil
 	} else if err != nil {
 		return err
 	}
```
pfmackin, line 355: looks like we create the uninstall condition (line 357) in the case where the job was already started.  Can you double check this?

platform-operator/controllers/controller.go
```diff
@@ -424,8 +422,8 @@ func (r *VerrazzanoReconciler) updateStatus(log *zap.SugaredLogger, cr *installv
 
 	// Update the status
 	err := r.Status().Update(context.TODO(), cr)
-	if err != nil {
-		log.Error(err, "Failed to update verrazzano resource status")
+	if err != nil && !errors.IsConflict(err) {
```
pfmackin, line 425: what is this isConflict check?

platform-operator/controllers/upgrade.go
```diff
@@ -5,33 +5,26 @@ package controllers
...
-// The max upgrade failures for a given upgrade attempt is 2
-const failedUpgradeLimit = 2
-
 // Reconcile upgrade will upgrade the components as required
 func (r *VerrazzanoReconciler) reconcileUpgrade(log *zap.SugaredLogger, req ctrl.Request, cr *installv1alpha1.Verrazzano) (ctrl.Result, error) {
 	// Upgrade version was validated in webhook, see ValidateVersion
 	targetVersion := cr.Spec.Version
 
-	// Only allow upgrade to retry a certain amount of times during any upgrade attempt.
-	if upgradeFailureCount(cr.Status, cr.Generation) > failedUpgradeLimit {
```
pfmackin, line 25: We have to be absolutely sure the reconcile loop doesn't keep running (meaning we return an error from upgrade every time we are called) or else the CR will fill up with status conditions (which used to happen).  Taking this out might be risky in that sense.

platform-operator/controllers/controller.go
bobdonat: yes, that is the case that is why I remove these lines of code.

platform-operator/controllers/controller.go
bobdonat: Checking for "operation cannot be fulfilled" error message

platform-operator/controllers/upgrade.go
bobdonat: Agreed.  I will add it back.

bobdonat: Fixed.

pfmackin (approved):

PR #559, opened 2021-01-26
Platform operator changes

bobdonat:
This pull request contains two changes.

1. Added a printcolumn for showing the installed version of Verrazzano as shown below.
```
kubectl get verrazzano
NAME            STATUS           VERSION
my-verrazzano   InstallStarted   0.9.0
```
2. Fixed bug where the clientset was not being generated.  This broke when the directory name changed from operator to platform-operator.


platform-operator/controllers/controller.go
```diff
@@ -283,6 +283,13 @@ func (r *VerrazzanoReconciler) createInstallJob(ctx context.Context, log *zap.Su
 		return err
 	}
 
+	// Set the version in the status.  This will be updated when the starting install condition is updated.
+	chartSemVer, err := installv1alpha1.GetCurrentChartVersion()
```
pfmackin, line 287: This is ok, but better if we had that GetCurrentChartVersion method in the util/helm package.

platform-operator/deploy/operator.yaml
```diff
@@ -162,6 +162,10 @@ spec:
       jsonPath: .status.conditions[-1:].type
       name: Status
       type: string
+    - description: The current version of the install
```
pfmackin, line 165: I think the term installation is better "of the Verrazzano installation"

pfmackin (dismissed):

platform-operator/hack/update-codegen-verrazzano.sh
```diff
@@ -27,7 +27,7 @@ rm -rf $GENERATED_CLIENT_DIR
 #                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
```
pfmackin, line 27: needs 2021 copyright

platform-operator/deploy/operator.yaml
```diff
@@ -162,6 +162,10 @@ spec:
       jsonPath: .status.conditions[-1:].type
```
pfmackin, line 162: needs 2021 copyright

platform-operator/deploy/operator.yaml
bobdonat: fixed

platform-operator/hack/update-codegen-verrazzano.sh
bobdonat: fixed

platform-operator/deploy/operator.yaml
bobdonat: fixed

platform-operator/controllers/controller.go
bobdonat: I will fix this on a followup PR the first thing in the morning

pfmackin (approved):

PR #560, opened 2021-01-28
JIRA:VZ-2007 An OAM Sock Shop application and an associated validation test

jmaron99:
Tests results are available here:  https://build.verrazzano.io/job/verrazzano/job/jmaron%252FVZ-2007/


examples/sock-shop/README_OAM.md
```diff
@@ -0,0 +1,128 @@
...
+  "password":"***",
+  "email":"foo@example.com",
+  "firstName":"foo",
+  "lastName":"coo"
+}` to create a user. This
+endpoint accepts the `POST` HTTP request method.
+
+Follow these steps to test the endpoints:
+
+1. Get the `EXTERNAL_IP` address of the istio-ingressgateway service.
```
kminder, line 49: Does this work for both OKE and KIND?  I'm trying to figure out what to do in my README.

examples/sock-shop/README_OAM.md
```diff
@@ -0,0 +1,128 @@
...
+1. Prerequisites: Install Verrazzano following the [installation instructions](../../README.md).
+   The Sock Shop application deployment artifacts are contained in the Verrazzano project located at 
+   `<VERRAZZANO_HOME>/examples/sockshop`, where `<VERRAZZANO_HOME>` is the root of the Verrazzano project.
+
+   **NOTE:** All files and paths in this document are relative to 
+   `<VERRAZZANO_HOME>/examples/sockshop`.
+
+1. Create a namespace for the Sock Shop application and add a label identifying the namespace as managed by Verrazzano.
+   ```
+   kubectl create namespace oam-sockshop
```
kminder, line 17: I don't think we should be using and oam prefix/suffix at this point if possible.

examples/sock-shop/README_OAM.md
```diff
@@ -0,0 +1,128 @@
...
+
+1. Get the `EXTERNAL_IP` address of the istio-ingressgateway service.
+   ```
+   kubectl get service istio-ingressgateway -n istio-system
+
+   NAME                   TYPE           CLUSTER-IP    EXTERNAL-IP   PORT(S)                      AGE
+   istio-ingressgateway   LoadBalancer   10.96.97.98   11.22.33.44   80:31380/TCP,443:31390/TCP   13d
+   ```   
+
+1. The application is deployed by default with a host value of `oam-sockshop.example.com`.
```
kminder, line 57: Remove oam- prefix

examples/sock-shop/README_OAM.md
```diff
@@ -0,0 +1,128 @@
...
+
+   * **Using your own DNS Name:**
+   
+     * Point your own DNS name to the ingress gateway's `EXTERNAL-IP` address
+     * In this case, you would need to edit the `sock-shop-app.yaml` file 
+       to use the appropriate value under the `hosts` section (such as `yourhost.your.domain`), 
+       before deploying the Sock Shop application.
+     * Then, you can use a browser to access the application at `http://<yourhost.your.domain>/catalogue`
+   
+1. ## Troubleshooting
```
kminder, line 97: Should we add anything about accessing metrics?

examples/sock-shop/README_OAM.md
```diff
@@ -0,0 +1,128 @@
...
+   kubectl get Domain -n oam-sockshop
+   kubectl get IngressTrait -n oam-sockshop
+   ```   
+
+1. Verify that the Sock Shop service pods are successfully created and transition to the ready state.
+   Note that this may take a few minutes and that you may see some of the services terminate and restart.
+   ```
+    kubectl get pods -n oam-sockshop -w
+   
+    NAME             READY   STATUS        RESTARTS   AGE
```
kminder, line 111: I think I'd find this output confusing.  For example what pods are required?  I also think the -w is more confusing that a non-watch get pods where all the required pods show the expected state.

examples/sock-shop/README_OAM.md
jmaron99: Are we expecting KIND as a deployment env for end users?  I guess we could add a "this assumes the USE of OKE or similar Kubernetes environment" at the top?

examples/sock-shop/README_OAM.md
jmaron99: This is iffy for me.  I assume there is a certain amount of time during which the model binding and hockshop versions may be tested side by side?  Is that assumption incorrect?

examples/sock-shop/README_OAM.md
jmaron99: I could try to add something.  Has it been added to readme's previously?

examples/sock-shop/README_OAM.md
jmaron99: OK

examples/sock-shop/carts.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: coherence.oracle.com/v1
+kind: Coherence
+metadata:
+  name: carts-coh
+  namespace: temp
```
kminder, line 5: Why namespace temp and how does that work?
Is this name hard coded somewhere else or could it be changed to something easier to understand?

jmaron99: Oops - part of a debugging session.  needs to be removed from commit.

examples/sock-shop/README_OAM.md
kminder: The plan is to remove model/binding in the next sprint.

examples/sock-shop/README_OAM.md
kminder: There is at least a reference to telemetry in the Bob's Books example.  That being said might be best as a second pass.

examples/sock-shop/sock-shop-app.yaml
```diff
@@ -0,0 +1,201 @@
...
+  annotations:
+    version: v1.0.0
+    description: "OAM Sock Shop Carts Application"
+spec:
+  components:
+    - componentName: carts
+      traits:
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: IngressTrait
```
gilbode, line 17: An ingress trait on every component seems a whole lot more verbose/redundnant than just creating a single ingress...
https://github.com/helidon-sockshop/sockshop/blob/master/k8s/optional/ingress.yaml#L12

I think we should re-visit if trait is really the best way to express ingress, not suggesting you do anything now but raising the point for discussion

tests/e2e/examples/sock-shop/sock_shop_example_test.go
```diff
@@ -0,0 +1,269 @@
...
+	It("SockShop can log in with default user", func() {
+		sockShop.Cookies = login(username, password)
+	})
+
+	It("SockShop can access Calatogue and choose item", func() {
+		webpage := sockShop.ConnectToCatalog()
+		sockShop.VerifyCatalogItems(webpage)
+	})
+
+	It("SockShop can add item to cart", func() {
```
gilbode, line 105: Do we really need to test the application?  It's probably safe to assume it works.  However, it's not clear this actually proves all the microservices are actually accessible through the ingress gateway which seems to be an important thing to test.

examples/sock-shop/sock-shop-app.yaml
```diff
@@ -0,0 +1,201 @@
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: carts-appconf
+  namespace: sockshop
+  annotations:
+    version: v1.0.0
+    description: "OAM Sock Shop Carts Application"
```
kminder, line 10: Should not block merge but eventually all of the OAM references probably need to be removed.

tests/e2e/examples/sock-shop/sock_shop_example_test.go
jmaron99: The test is focused on proving that the common tasks performed by a user are functioning.  I think it's also relatively safe to assume that the ability to perform all the common functions speaks to the availability and functioning of the micro services.

examples/sock-shop/sock-shop-app.yaml
```diff
@@ -0,0 +1,201 @@
...
+                  paths:
+                    - path: "/carts"
+                      pathType: Prefix
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: MetricsTrait
+            metadata:
+              name: carts-metrics
+            spec:
+              port: 7001
```
kminder, line 33: This is to work around the coherence/helidon metrics issue?

jmaron99: yes

kminder (approved):

PR #564, opened 2021-01-29
VZ-1547.  Storage setup & config

michael-cico:
### Storage Configuration

Allow users to define storage templates and configurations for components that require it.  Configuration is done through `VolumeSources` and `VolumeClaimSpecTemplate`, which do not configure actual Volumes or PVCs, but are used to provide configuration to underlying Verrazzano components that support it.

The Verrazzano components supporting storage configuration in this manner at present are

- Keycloak/MySQL
- VMI (Elasticsearch, Prometheus, Grafana)

Users can configure a `DefaultVolumeSource` for all components, which components that support it can override by declaring their own `VolumeSource`.  

At present, the supported VolumeSources are

- `EmptyDir`
- `PersistentVolumeClaimVolumeSource`

If a `PersistentVolumeClaimVolumeSource` is configured, the `claimSource` field must reference a `VolumeClaimSpecTemplate` declared in the `VolumeClaimSpecTemplates` section of the Verrazzano CR.

#### Examples:

[prod-storage-example.txt](https://github.com/verrazzano/verrazzano/files/5896114/prod-storage-example.txt)
[dev-with-persistence-example.txt](https://github.com/verrazzano/verrazzano/files/5896115/dev-with-persistence-example.txt)
[dev-storage-example.txt](https://github.com/verrazzano/verrazzano/files/5896116/dev-storage-example.txt)
[dev-mysql-storage-example.txt](https://github.com/verrazzano/verrazzano/files/5896117/dev-mysql-storage-example.txt)

### Other changes:

- Declare an explicit INIT mode on the platform operator to fix an issue where `init-webhooks` is when run in the init-container, which causes the operator code to try to start


platform-operator/api/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -64,6 +64,17 @@ type VerrazzanoSpec struct {
...
+	// DefaultVolumeSource Defines the type of volume to be used for persistence, if not explicitly declared by a component;
+	// at present only EmptyDirVolumeSource or PersistentVolumeClaimVolumeSource are supported. If PersistentVolumeClaimVolumeSource
+	// is used, it must reference a PersistentVolumeClaim in the VolumeClaimTemplates section.
+	// +optional
+	DefaultVolumeSource corev1.VolumeSource `json:"defaultVolumeSource,omitempty"`
+
+	// VolumeClaimTemplates Defines a named set of PVC options that can be referenced as templates by other
+	// components in the resource;
+	// +optional
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty" protobuf:"bytes,4,rep,name=volumeClaimTemplates"`
```
pfmackin, line 77: shouldn't this be the PersistentVolumeClaimSpec, since PersistentVolumeClaim has metadata and status.  Also, the 
PersistentVolumeClaimSpec has a VolumeName in it so can it be shared across components?

michael-cico: I thought what we discussed yesterday was that it would just be a straightforward VolumeClaimTemplates list?  To be consistent with k8s things.

I though this was what @gilbode was looking for.  I can easily switch it back to what I had before.

pfmackin: ok, i thought it was just the spec part.

pfmackin:
The curly braces for emptyDir seem strange, do you see that. being used in other YAML in general?

  defaultVolumeSource:
    emptyDir: {}

platform-operator/api/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -64,6 +64,28 @@ type VerrazzanoSpec struct {
...
+// VolumeClaimSpecTemplate Contains common PVC configuration that can be referenced from Components; these
+// do not actually result in generated PVCs, but can used to provide common configuration to components that
+// declare a PersistentVolumeClaimVolumeSource
+type VolumeClaimSpecTemplate struct {
+	// Name The name of the configuration template
+	Name string `json:"name"`
+
+	// Spec The configuration specs for the template
+	// +optional
+	Spec corev1.PersistentVolumeClaimSpec `json:"spec,omitempty"`
```
pfmackin, line 88: Is the spec really optional?

michael-cico: No, probably not...just habit I think.

platform-operator/api/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -64,6 +64,28 @@ type VerrazzanoSpec struct {
 	// Core specifies core Verrazzano configuration
 	// +optional
 	Components ComponentSpec `json:"components,omitempty"`
+
+	// DefaultVolumeSource Defines the type of volume to be used for persistence, if not explicitly declared by a component;
+	// at present only EmptyDirVolumeSource or PersistentVolumeClaimVolumeSource are supported. If PersistentVolumeClaimVolumeSource
+	// is used, it must reference a VolumeClaimSpecTemplate in the VolumeClaimSpecTemplates section.
+	// +optional
+	DefaultVolumeSource *corev1.VolumeSource `json:"defaultVolumeSource,omitempty"`
```
pfmackin, line 72: Since this is optional, i assume you default to emptydir

michael-cico: Right now, we don't.  If I force things to default to emptyDir now, it will break the prod profile default.  So if this isn't specified, the defaults in the profile values files (in the helm charts) will be the default.

The intent is to make EmptyDir the default behavior, and then prod will need to specify persistence.

platform-operator/main.go
```diff
@@ -73,7 +73,12 @@ func main() {
 
 	// initWebhooks flag is set when called from an initContainer.  This allows the certs to be setup for the
 	// validatingWebhookConfiguration resource before the operator container runs.
-	if config.InitWebhooks {
+	mode := os.Getenv("MODE")
```
pfmackin, line 76: why did you change this to use ENV VAR?

platform-operator/config/deploy/verrazzano-platform-operator.yaml
```diff
@@ -51,7 +51,7 @@ spec:
             - --init-webhooks=true
           env:
             - name: MODE
-              value: RUN_OPERATOR
```
pfmackin, line 54: why did you change this?

platform-operator/controllers/installjob/install_config.go
```diff
@@ -329,11 +334,63 @@ func getInstallArgs(args []installv1alpha1.InstallArgs) []InstallArg {
...
+	mySQLArgs := getInstallArgs(keycloak.MySQL.MySQLInstallArgs)
+
+	// Use a volume source specified in the Keycloak config, otherwise use the default spec
+	mysqlVolumeSource := keycloak.MySQL.VolumeSource
+	if mysqlVolumeSource == nil {
+		mysqlVolumeSource = defaultVolumeSpec
+	}
+
+	if mysqlVolumeSource.EmptyDir != nil {
+		// TODO: Eventually, we may wan to pass down EmptyDir settings, right now we just use helm settings for the
```
pfmackin, line 348: need to remove TODO

platform-operator/controllers/installjob/install_config.go
```diff
@@ -329,11 +334,63 @@ func getInstallArgs(args []installv1alpha1.InstallArgs) []InstallArg {
...
+		mySQLArgs = append(mySQLArgs, InstallArg{
+			Name:      "persistence.enabled",
+			Value:     "false",
+			SetString: true,
+		})
+	} else if mysqlVolumeSource.PersistentVolumeClaim != nil {
+		pvcs := mysqlVolumeSource.PersistentVolumeClaim
+		storageSpec, found := findVolumeTemplate(pvcs.ClaimName, templates)
+		if !found {
+			log.Errorf("No VolumeClaimTemplate found for %s", pvcs.ClaimName)
```
pfmackin, line 361: shouldn't this return an error?

platform-operator/main.go
michael-cico: It's the same behavior, the MODE indicates that we execute the code for init only in this case.  The reason is, if you just try to disable --init-webhooks, the code drops out of the conditonal and then tries to start the operator.

We still use init-webhooks to check if we are actually enabling the webhook or not.  

With this, if you want to run the operator without valiation you can just apply the CRD and run the operator, and don't have to create the validatingwebhook configuration and delete it after.

platform-operator/config/deploy/verrazzano-platform-operator.yaml
michael-cico: For the reason in the last comment.

platform-operator/controllers/installjob/install_config.go
michael-cico: Should we open a backlog item to change the defaults later?

platform-operator/controllers/installjob/install_config.go
```diff
@@ -402,6 +459,68 @@ func getProfile(profileType installv1alpha1.ProfileType) InstallProfile {
 	return InstallProfileDev
 }
 
+// getVerrazzanoInstallArgs Set custom helm args for the Verrazzano internal component as needed
+func getVerrazzanoInstallArgs(vzSpec *installv1alpha1.VerrazzanoSpec, log *zap.SugaredLogger) []InstallArg {
+	if vzSpec.DefaultVolumeSource.EmptyDir != nil {
+		// TODO: Eventually, we may wan to pass down EmptyDir settings, right now we just use helm settings for the
```
pfmackin, line 465: need to remove TODO

platform-operator/controllers/installjob/install_config.go
```diff
@@ -402,6 +459,68 @@ func getProfile(profileType installv1alpha1.ProfileType) InstallProfile {
...
+				Value:     "",
+				SetString: true,
+			},
+			{
+				Name:      "verrazzanoOperator.prometheusDataStorageSize",
+				Value:     "",
+				SetString: true,
+			},
+		}
+	} else if vzSpec.DefaultVolumeSource.PersistentVolumeClaim != nil {
```
pfmackin, line 486: no need for else

platform-operator/controllers/installjob/install_config.go
michael-cico: Probably should, I'll have to add hooks to propagate an error up to the controller.

platform-operator/controllers/installjob/install_config.go
```diff
@@ -402,6 +459,68 @@ func getProfile(profileType installv1alpha1.ProfileType) InstallProfile {
...
+		}
+	}
+	return []InstallArg{}
+}
+
+// findVolumeTemplate Find a named VolumeClaimTemplate in the list
+func findVolumeTemplate(templateName string, templates []installv1alpha1.VolumeClaimSpecTemplate) (*corev1.PersistentVolumeClaimSpec, bool) {
+	for _, template := range templates {
+		if templateName == template.Name {
+			return &template.Spec, true
```
pfmackin, line 518: i think you need to use and index and return templates[i].Spec

platform-operator/controllers/installjob/install_config.go
```diff
@@ -402,6 +459,68 @@ func getProfile(profileType installv1alpha1.ProfileType) InstallProfile {
...
+	} else if vzSpec.DefaultVolumeSource.PersistentVolumeClaim != nil {
+		pvcs := vzSpec.DefaultVolumeSource.PersistentVolumeClaim
+		storageSpec, found := findVolumeTemplate(pvcs.ClaimName, vzSpec.VolumeClaimSpecTemplates)
+		if !found {
+			log.Errorf("No VolumeClaimTemplate found for %s", pvcs.ClaimName)
+			return []InstallArg{}
+		}
+		return []InstallArg{
+			{
+				Name:      "verrazzanoOperator.esDataStorageSize",
```
pfmackin, line 495: I don't understand this part.  Does the ES, Grafana, etc components always use the default volume source?

michael-cico:
> The curly braces for emptyDir seem strange, do you see that. being used in other YAML in general?
> 
> defaultVolumeSource:
> emptyDir: {}

I think this is the way you specify EmptyDir in yaml, AFAICT.

platform-operator/controllers/installjob/install_config.go
michael-cico: Right now, yes.  The problem is we don't surface the monitoring components yet, and I wasn't sure if we wanted to do that as part of this change.

platform-operator/controllers/installjob/install_config.go
michael-cico: Is that just a preferred go style?

platform-operator/controllers/installjob/install_config.go
michael-cico: Spoke with Paul, this is OK for now, and we will address surfacing the monitoring components in the CR later.

platform-operator/api/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: Spoke with Paul, we'll refine how the defaults are handled later as the profiles concept/impl evolves.

platform-operator/api/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: fixed

platform-operator/main.go
michael-cico: Backing out this change.

platform-operator/config/deploy/verrazzano-platform-operator.yaml
michael-cico: Backing out this change.

platform-operator/controllers/installjob/install_config.go
michael-cico: Fixed.

platform-operator/controllers/installjob/install_config.go
michael-cico: Fixed.

platform-operator/controllers/installjob/install_config.go
michael-cico: Fixed.

pfmackin (approved):

PR #567, opened 2021-01-29
Tvlatas/update runsh loop test

agv426:
This updates the run.sh and the config spots where the move from operator to platform-operator were missed


agv426:
Oops, it defaulted to master, needs to base on develop, deleting this one

PR #568, opened 2021-01-29
update some spots missed during move of operators

agv426:


jmaron99 (approved):

bhatpmk (approved):

agv426:
Got back to looking at this. The tests failed, it looks like the paths I tweaked here were for inside the docker image which is still using "operator" and not "platform-operator". Closing this out

PR #571, opened 2021-01-29
WIP:  ES rollover index name with date

guoyongzhang:


fredtibbitts:
This is going to require updates to `controllers/cohworkload/coherenceworkload_controller.go` as well, since that's where logging gets configured for Coherence. Not sure if it should be included in this PR or if it should be done separately.

PR #577, opened 2021-02-01
Create verrazzano-mc namespace as part of install

mgianatagh:
The the verrazzano-mc namespace as part of an install, it will be used to contain the verrazzano multi-cluster objects.

Add a test to verify the namespace is created.


markxnelson (approved):
I don't love the name but...

pfmackin (approved):

markxnelson (approved):

PR #582, opened 2021-02-03
VZ-2044: Additional support for VerrazzanoCoherenceWorkload

fredtibbitts:
I'm breaking this ticket into multiple PRs to try to make them easier to review rather than submitting a single, bigger PR. This PR continues support for VerrazzanoCoherenceWorkload.

Specifically, this PR updates the ingress and metrics traits so that they are Verrazzano workload aware. When fetching the workload for the trait, if the workload is a Verrazzano workload type, we unwrap the contained workload before fetching the workload definition and children. Most of the code here is in additional unit tests.

**Note**: Based on the late-breaking decision to remove the kind and API version from the wrapped CRs, I will have to make corresponding CRD and implementation changes. I'm not going to do that yet, I'll submit a separate PR for those changes.

### DIFF
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
@@ -184,6 +184,144 @@ func TestSuccessfullyCreateNewIngress(t *testing.T) {
 	assert.Equal(time.Duration(0), result.RequeueAfter)
 }
 
+// TestSuccessfullyCreateNewIngressForVerrazzanoWorkload tests the Reconcile method for the following use case.
+// GIVEN a request to reconcile an ingress trait resource that applies to a Verrazzano workload type
+// WHEN the trait exists but the ingress does not
+// THEN ensure that the workload is unwrapped and the trait is created.
+func TestSuccessfullyCreateNewIngressForVerrazzanoWorkload(t *testing.T) {
+	assert := asserts.New(t)
+	mocker := gomock.NewController(t)
+	mock := mocks.NewMockClient(mocker)
+	mockStatus := mocks.NewMockStatusWriter(mocker)
+	// Expect a call to get the ingress trait resource.
+	mock.EXPECT().
+		Get(gomock.Any(), types.NamespacedName{Namespace: "test-space", Name: "test-trait-name"}, gomock.Not(gomock.Nil())).
+		DoAndReturn(func(ctx context.Context, name types.NamespacedName, trait *vzapi.IngressTrait) error {
+			trait.TypeMeta = metav1.TypeMeta{
+				APIVersion: "oam.verrazzano.io/v1alpha1",
+				Kind:       "IngressTrait"}
+			trait.ObjectMeta = metav1.ObjectMeta{
+				Namespace: name.Namespace,
+				Name:      name.Name}
+			trait.Spec.Rules = []vzapi.IngressRule{{
+				Hosts: []string{"test-host"},
+				Paths: []vzapi.IngressPath{{Path: "test-path"}}}}
+			trait.Spec.WorkloadReference = oamrt.TypedReference{
+				APIVersion: "oam.verrazzano.io/v1alpha1",
+				Kind:       "VerrazzanoCoherenceWorkload",
+				Name:       "test-workload-name"}
+			return nil
+		})
+
+	containedName := "test-contained-workload-name"
+	containedResource := map[string]interface{}{
+		"apiVersion": "coherence.oracle.com/v1",
+		"kind":       "Coherence",
+		"metadata": map[string]interface{}{
+			"name": containedName,
+		},
+	}
+
+	// Expect a call to get the Verrazzano Coherence workload resource
+	mock.EXPECT().
+		Get(gomock.Any(), types.NamespacedName{Namespace: "test-space", Name: "test-workload-name"}, gomock.Not(gomock.Nil())).
+		DoAndReturn(func(ctx context.Context, name types.NamespacedName, workload *unstructured.Unstructured) error {
+			workload.SetAPIVersion("oam.verrazzano.io/v1alpha1")
+			workload.SetKind("VerrazzanoCoherenceWorkload")
+			workload.SetNamespace(name.Namespace)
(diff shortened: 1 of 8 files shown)


application-operator/controllers/loggingscope/loggingscope_controller.go
```diff
@@ -81,7 +82,13 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 		handler := r.Handlers[handlerKey(resource)]
 		if handler == nil {
-			log.Error(nil, "Unknown Resource Kind encountered in Logging Scope Controller", "resource", resource)
+			// if this is one of our wrapper resources, we expect to not find a handler since logging is
+			// added at resource creation time
+			if vznav.IsVerrazzanoWorkloadKind(workload) {
+				log.Info("Skipping logging scope processing for Verrazzano workload kind", "resource", resource)
+			} else {
+				log.Error(nil, "Unknown Resource Kind encountered in Logging Scope Controller", "resource", resource)
```
pfmackin, line 90: should this be a fatal error?

fredtibbitts: No, this is an expected condition. The controller for the v8o workload (e.g. VerrazzanoCoherenceWorkload) adds logging up front, so we need to ignore it here.

application-operator/controllers/navigation/workload.go
```diff
@@ -119,3 +125,57 @@ func LoggingScopeFromWorkloadLabels(ctx context.Context, cli client.Reader, name
...
+// IsVerrazzanoWorkloadKind returns true if the workload is a Verrazzano workload kind
+// (e.g. VerrazzanoWebLogicWorkload), false otherwise.
+func IsVerrazzanoWorkloadKind(workload *unstructured.Unstructured) bool {
+	kind := workload.GetKind()
+	return strings.HasPrefix(kind, "Verrazzano") && strings.HasSuffix(kind, "Workload")
+}
+
+// GetContainedWorkloadVersionKindName returns the api version, kind, and name of the contained workload
+// inside a Verrazzano*Workload.
+func GetContainedWorkloadVersionKindName(workload *unstructured.Unstructured) (string, string, string, error) {
```
kminder, line 138: Won't this need to change based on recent decisions to not include apiVersion/Kind in the wrapper?

application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
```diff
@@ -179,6 +179,144 @@ func TestSuccessfullyCreateNewIngress(t *testing.T) {
...
+		DoAndReturn(func(ctx context.Context, name types.NamespacedName, workloadDef *v1alpha2.WorkloadDefinition) error {
+			workloadDef.Namespace = name.Namespace
+			workloadDef.Name = name.Name
+			workloadDef.Spec.ChildResourceKinds = []v1alpha2.ChildResourceKind{
+				{APIVersion: "apps/v1", Kind: "Deployment", Selector: nil},
+				{APIVersion: "v1", Kind: "Service", Selector: nil},
+			}
+			return nil
+		})
+	// Expect a call to list the child Deployment resources of the coherence workload definition
```
kminder, line 252: Capitalization

application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
```diff
@@ -179,6 +179,144 @@ func TestSuccessfullyCreateNewIngress(t *testing.T) {
...
+	mock.EXPECT().
+		List(gomock.Any(), gomock.Not(gomock.Nil()), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, list *unstructured.UnstructuredList, opts ...client.ListOption) error {
+			assert.Equal("Deployment", list.GetKind())
+			return nil
+		})
+	// Expect a call to list the child Service resources of the coherence workload definition
```
kminder, line 259: Capitalization

application-operator/controllers/navigation/workload.go
fredtibbitts: Yes it will. I'll make those updates in a subsequent PR.

application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
fredtibbitts: Fixed

application-operator/controllers/ingresstrait/ingresstrait_controller_test.go
fredtibbitts: Fixed

kminder (approved):

application-operator/controllers/navigation/workload.go
```diff
@@ -119,3 +125,57 @@ func LoggingScopeFromWorkloadLabels(ctx context.Context, cli client.Reader, name
...
+	// container workloads use different properties to hold the resource, so use a map
+	// to get the property name (e.g. VerrazzanoWebLogicWorkload has a "spec" containing a "domain" property)
+	key := fmt.Sprintf("%s.%s", workload.GetAPIVersion(), workload.GetKind())
+	if specProperty, ok := workloadPropertyNameMap[key]; ok {
```
pfmackin, line 142: It is a better go style to return the error immediately if !ok, then get rid of the subsequent nested block.  This keeps code closer to the left margin happy path.

markxnelson: i prefer to see code that i can read without needing to stop and think about it - i worry this particular construction would make it more likely that a future developer might miss something and make a bad change.  i do know it is common though

fredtibbitts: @pfmackin done. @markxnelson I don't like this either. If we get rid of the specific property names and replace them with `template`, then this function goes away (since we're removing api version and kind from the contained resource as well). If we leave these properties, do you have suggestions on how to make this code better?

fredtibbitts: @markxnelson unless you object, I'm going to move forward with this and will revisit once we have the developer drop-in meeting to discuss.

kminder (approved):

PR #588, opened 2021-02-03
Changes for new api server

lake-of-dreams:
- Changes to Jenkinsfile to enable dumping the verrazzano-api logs
- Change verrazzano api ingress to point to new api server
- New helm template for api server
- Update console image version to latest (which uses new api)
- One unrelated change to 4-install-keycloak.sh to enable LC_ALL=C so that tr works on macos (in case user wants to install/uninstall keycloak using scripts for debugging)
- Changes to keyclaok config json to add mapper for groups and k8s_user attribute
- /instance api endpoint will no longer exist in new api as it is just a proxy for real k8s resource and nor will exist customized resources like Secret/Instance etc. - made relevant changes to all tests using it and divided test cases in instance_api.test to separate tests (e.g. tests for fetching mgmt cluster api, fetch kecycloak/api urls etc.)
- Changes to all rest api tests according to new api


platform-operator/helm_config/charts/verrazzano/values.yaml
```diff
@@ -95,7 +95,17 @@ verrazzanoAdmissionController:
 console:
   name: verrazzano-console
   imageName: ghcr.io/verrazzano/console
-  imageVersion: 0.9.0-20210122131824-c7cfc9b
+  imageVersion: 0.9.0-20210202153634-5d96a30
+
+api:
```
markxnelson, line 100: are you adding another nginx ingress controller? or just using the one that was already there - for system vmi, etc.  we want to share, and only have the auth on the `/api/xxx` endpoints

tests/e2e/pkg/common.go
```diff
@@ -17,24 +17,6 @@ import (
...
-	ID               string `json:"id"`
-	Name             string `json:"name"`
-	MgmtCluster      string `json:"mgmtCluster"`
-	MgmtPlatform     string `json:"mgmtPlatform"`
-	Status           string `json:"status"`
-	Version          string `json:"version"`
-	KeyCloakURL      string `json:"keyCloakUrl"`
-	RancherURL       string `json:"rancherUrl"`
-	VzAPIURI         string `json:"vzApiUri"`
-	ElasticURL       string `json:"elasticUrl"`
```
markxnelson, line 31: these are used elsewhere - have not read the whole PR yet, you might have already addressed that

markxnelson: ok looks like you have thanks

platform-operator/helm_config/charts/verrazzano/values.yaml
lake-of-dreams: It not adding another nginx ingress controller, I am just using the image being used by the ingress controller, it will have separate deployment and service and will act as api server..

tests/e2e/pkg/common.go
lake-of-dreams: yes

markxnelson (approved):

PR #589, opened 2021-02-03
Fix boilerplate copyright headers

desagar:
- Update copyright boilerplate
- Fix copyright header scripts to distinguish existing files from new files and add the correct headers

### DIFF
--- a/application-operator/Makefile
+++ b/application-operator/Makefile
@@ -70,8 +70,9 @@ manifests: go-mod controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	# Add copyright headers to the kubebuildr generated CRDs
 	./hack/add-crd-header.sh
-	# Add copyright headers to the kubebuildr generated manifests
+	# Add copyright headers to the kubebuildr generated manifests and to the edited PROJECT file
 	./hack/add-yml-header.sh config/webhook/manifests.yaml
+	./hack/add-yml-header.sh PROJECT
 
 # Generate code
 .PHONY: generate
--- a/application-operator/hack/add-crd-header.sh
+++ b/application-operator/hack/add-crd-header.sh
@@ -14,8 +14,19 @@ CRD_HEADER=$(dirname $0)/boilerplate.yaml.txt
 GENERATED_CRDS_DIR=$SCRIPT_ROOT/config/crd/bases
 
 for CRD_FILENAME in $(ls $GENERATED_CRDS_DIR/*.y*ml) ; do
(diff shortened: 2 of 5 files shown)


application-operator/Makefile
```diff
@@ -70,8 +70,9 @@ manifests: go-mod controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
mgianatagh, line 70: can you also make these changes for platform-operator?

desagar: So I made an effort to make similar changes for platform-operator, but it turns out to be more complicated. Platform operator passes the header file to a client generator script from Kubernetes, and this would always result in it being half-wrong no matter if we updated the boilerplate header or not. Discussed with @mgianatagh...he is working around this in his branch and might add that workaround in a future PR.

For application-operator, these changes work since our scripts control all the copyright headers, so I'd like to keep these changes.

application-operator/hack/add-crd-header.sh
```diff
@@ -14,8 +14,19 @@ CRD_HEADER=$(dirname $0)/boilerplate.yaml.txt
...
-  mv $TMP_CRD $CRD_FILENAME
+  GIT_HISTORY_LENGTH=$(git log $CRD_FILENAME | wc -l)
+  if [ "$GIT_HISTORY_LENGTH" -eq 0 ] ; then
+    echo "Adding header from $CRD_HEADER to generated NEW CRD file $CRD_FILENAME"
+    TMP_CRD=${CRD_FILENAME}.tmp
+    cat $CRD_HEADER $CRD_FILENAME > $TMP_CRD
+    mv $TMP_CRD $CRD_FILENAME
+  else
+    echo "Adding back previous header to re-generated existing file $CRD_FILENAME"
+    # get and use existing copyright header by getting first 2 lines of file at
```
markxnelson, line 25: seems pretty unreliable?

desagar: these are generated CRD files, and the way we add copyright headers to them in the first place is by prepending the first 2 lines of the boilerplate before the rest of the file. Not bulletproof of course but good enough nearly all of the time  - how do you think it should be made "smarter"?

desagar: another option is to use `git show origin/develop:$CRD_FILENAME` instead of relying on whatever branch we're on. Similarly we could use origin/develop to determine if it's an existing file. Do you think that's a little better?

markxnelson: oh ok - makes sense then.

markxnelson (approved):

PR #591, opened 2021-02-03
VZ-2044: Update Coherence ES index

fredtibbitts:
This PR updates the Coherence ES index to match the work that GZ recently did for WLS. The index uses the pattern `oam-#{ENV['NAMESPACE']}-#{ENV['APP_CONF_NAME']}-#{ENV['COMPONENT_NAME']}` so that users can more easily search logs for their applications.

The work in the VerrazzanoCoherenceWorkload controller was simply to update the fluentd rules and also to copy the OAM app and component name labels into the Coherence CR that the controller writes out.

I also moved 3 fluentd container environment variables out of `wls.go` into `fluentd.go` as they will be needed by all workload types.

### DIFF
--- a/application-operator/controllers/cohworkload/coherenceworkload_controller.go
+++ b/application-operator/controllers/cohworkload/coherenceworkload_controller.go
@@ -62,7 +62,7 @@ multiline_flush_interval 20s
   port "#{ENV['ELASTICSEARCH_PORT']}"
   user "#{ENV['ELASTICSEARCH_USER']}"
   password "#{ENV['ELASTICSEARCH_PASSWORD']}"
-  index_name "oam-#{ENV['NAMESPACE']}-#{ENV['APP_CONF_NAME']}-#{ENV['COMPONENT_NAME']}"
+  index_name "` + loggingscope.ElasticSearchIndex + `"
   scheme http
   key_name timestamp 
   types timestamp:time
--- a/application-operator/controllers/cohworkload/coherenceworkload_controller.go
+++ b/application-operator/controllers/cohworkload/coherenceworkload_controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
 	"github.com/go-logr/logr"
 	vzapi "github.com/verrazzano/verrazzano/application-operator/apis/oam/v1alpha1"
 	"github.com/verrazzano/verrazzano/application-operator/controllers"
@@ -48,6 +49,10 @@ multiline_flush_interval 20s
  role "#{ENV['COH_ROLE']}"
  host "#{ENV['HOSTNAME']}"
  pod-uid "#{ENV['COH_POD_UID']}"
+ oam.applicationconfiguration.namespace "#{ENV['NAMESPACE']}"
+ oam.applicationconfiguration.name "#{ENV['APP_CONF_NAME']}"
+ oam.component.namespace "#{ENV['NAMESPACE']}"
+ oam.component.name  "#{ENV['COMPONENT_NAME']}"
 </record>
 </filter>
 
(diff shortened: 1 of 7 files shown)


guoyongzhang (approved):
LGTM

application-operator/controllers/cohworkload/coherenceworkload_controller.go
```diff
@@ -57,7 +62,7 @@ multiline_flush_interval 20s
   port "#{ENV['ELASTICSEARCH_PORT']}"
   user "#{ENV['ELASTICSEARCH_USER']}"
   password "#{ENV['ELASTICSEARCH_PASSWORD']}"
-  index_name "#{ENV['COH_CLUSTER_NAME']}"
+  index_name "oam-#{ENV['NAMESPACE']}-#{ENV['APP_CONF_NAME']}-#{ENV['COMPONENT_NAME']}"
```
kminder, line 65: I'm not a fan of exposing oam any more than necessary.  Could this be vz- or verrazzano- instead.  I'm aware that this probably means more work elsewhere too.

fredtibbitts: I was following what GZ did. @guoyongzhang can you comment on having "oam" in the index?

fredtibbitts: Also we should probably make the index a constant in fluentd.go so we all use the same pattern. I can make that update.

guoyongzhang: how aout app-....?

kminder: Someone should check with @gilbode for an official opinion.

fredtibbitts: I'm not sure the "oam-" adds value, maybe just remove it altogether? I agree with @kminder we should get sign-off on the index pattern.

markxnelson: i see no reason to include `oam` in this name - the fact that we use oam to deploy something is an implementation detail

application-operator/controllers/cohworkload/coherenceworkload_controller_test.go
```diff
@@ -106,6 +108,9 @@ func TestReconcileCreateCoherence(t *testing.T) {
 		DoAndReturn(func(ctx context.Context, u *unstructured.Unstructured, opts ...client.CreateOption) error {
 			assert.Equal("coherence.oracle.com/v1", u.GetAPIVersion())
 			assert.Equal("Coherence", u.GetKind())
+
+			// make sure the oam component and app name labels were copied
```
markxnelson, line 112: OAM is captitalized

markxnelson (changes requested):

application-operator/controllers/cohworkload/coherenceworkload_controller_test.go
fredtibbitts: Fixed.

application-operator/controllers/cohworkload/coherenceworkload_controller.go
fredtibbitts: I removed `oam-` and moved the index pattern to a shared constant.

application-operator/controllers/loggingscope/fluentd.go
```diff
@@ -30,6 +31,9 @@ const (
 	elasticSearchPwdField  = "ELASTICSEARCH_PASSWORD"
 )
 
+// ElasticSearchIndex defines the common index pattern
+const ElasticSearchIndex = "#{ENV['NAMESPACE']}-#{ENV['APP_CONF_NAME']}-#{ENV['COMPONENT_NAME']}"
```
gilbode, line 35: Is this the same pattern used for other component types?  I'd like this to be consistent.

fredtibbitts: Yes, I made them all use this constant.

markxnelson (approved):

PR #597, opened 2021-02-04
Reorganize generated client code in verrazzano-platform-operator

mgianatagh:
The generated client code is now under the folders listed below to be consistent with how the platform-operator/apis folder is structured.  Both apis and clients now have the same sub-folders.

platform-operator/clients/verrazzano
platform-operator/clients/clusters


desagar (approved):
looks good.

PR #600, opened 2021-02-04
Add remaining MultiClusterXXX CRDs

desagar:
Add the following MultiCluster CRDs

- MultiClusterConfigMap
- MultiClusterComponent
- MultiClusterApplicationConfiguration
- MultiClusterLoggingScope


application-operator/PROJECT
```diff
@@ -1,5 +1,3 @@
-# Copyright (C) 2020, 2021, Oracle and/or its affiliates.
```
mgianatagh, line 1: Why was the license removed?

desagar: Good catch. Fixed. I don't know why it didn't get added back, and the copyright check didn't catch it during the build. Will look into that a bit.

mgianatagh (dismissed):

mgianatagh (approved):

PR #604, opened 2021-02-04
VZ-2044: Updates to VerrazzanoCoherenceWorkload CRD and related files

fredtibbitts:
This PR updates the VerrazzanoCoherenceWorkload CRD removing the API version and kind from the embedded Coherence resource definition. Also, the spec field `coherence` has been renamed to `template`.

Most of the changes are in a few utility functions and in unit tests. The controller for VerrazzanoCoherenceWorkload sets the appropriate API version and kind in the Coherence CR before writing it out.

Note that there is a utility function `GetContainedWorkloadVersionKindName` in `navigation/workload.go` that I was hoping would go away with these changes, however because the embedded resource no longer has a version and kind, we need a map that maps the Verrazzano workload wrappers to the version and kind of thing that they contain. I tried to refactor that code a bit to make it a little easier to understand. I still don't like having a map, developers will need to know to update it when adding a new workload type, but I can't think of a better solution (at least one that doesn't require lots of work).


markxnelson (dismissed):
LGTM

application-operator/controllers/navigation/workload.go
```diff
@@ -133,28 +137,36 @@ func IsVerrazzanoWorkloadKind(workload *unstructured.Unstructured) bool {
...
 
-	kind, found, err := unstructured.NestedString(workload.Object, "spec", specProperty, "kind")
-	if !found || err != nil {
-		return "", "", "", errors.New("Unable to find kind in contained workload")
+// GetContainedWorkloadVersionKindName returns the API version, kind, and name of the contained workload
+// inside a Verrazzano*Workload.
+func GetContainedWorkloadVersionKindName(workload *unstructured.Unstructured) (string, string, string, error) {
+	gvk := WorkloadToContainedGVK(workload)
+	if gvk == nil {
+		return "", "", "", fmt.Errorf("Unable to find contained GroupVersionKind for workload: %v", workload)
```
kminder, line 162: Not worth fixing but I believe Go errors are supposed to start with a lower case character.  Did make check pass?

kminder (approved):

application-operator/controllers/navigation/workload.go
fredtibbitts: Make check did pass, odd that the linter didn't catch it. I'll fix this when I do the next ticket to add VerrazzanoWebLogicWorkload. Thanks for catching it.

PR #605, opened 2021-02-05
Move to istio 1.7.3

bobdonat:
This pull request includes changes to install, uninstall, and patching for moving to istio 1.7.3.
Changes to Kind config were also required for setting api server parameters that were not set by default.


platform-operator/controllers/component/registry.go
```diff
@@ -18,12 +18,52 @@ func GetComponents() []Component {
...
 	return []Component{
 		helmComponent{
-			releaseName:             "istio",
-			chartDir:                filepath.Join(thirdPartyChartsDir, "istio"),
+			releaseName:             "istio-base",
+			chartDir:                filepath.Join(thirdPartyChartsDir, "istio/base"),
+			chartNamespace:          "istio-system",
+			ignoreNamespaceOverride: true,
+		},
+		helmComponent{
```
pfmackin, line 26: I was thinking we would have 1 component with 7 charts.  This is ok for now, but we just need to make sure that the rollback work gets done or we could have a partial istio upgrade if a failure happens for one of the istio components.  We also don't want to expose Istio as 7 components in the verrazzano CR

platform-operator/helm_config/overrides/istio-values.yaml
```diff
@@ -3,28 +3,53 @@
 
 global:
   hub: ghcr.io/verrazzano
-  tag: 1.4.6
+  tag: 1.7.3
+  arch:
```
pfmackin, line 7: are we sure about these arch values?

platform-operator/scripts/install/1-install-istio.sh
```diff
@@ -65,45 +65,74 @@ function create_secret {
...
+    log "Installing istiocoredns"
+    helm upgrade istiocoredns ${ISTIO_CHART_DIR}/istiocoredns \
+      --install \
+      --namespace istio-system \
+      -f $VZ_OVERRIDES_DIR/istio-values.yaml \
+      ${IMAGE_PULL_SECRETS_ARGUMENT} \
+      || return $?
+
+    log "Installing Istio Grafana"
+    helm upgrade grafana ${ISTIO_CHART_DIR}/istio-telemetry/grafana \
```
pfmackin, line 122: Is Istio Grafana needed?

platform-operator/scripts/install/1-install-istio.sh
```diff
@@ -65,45 +65,74 @@ function create_secret {
...
+      --install \
+      --namespace istio-system \
+      -f $VZ_OVERRIDES_DIR/istio-values.yaml \
+      ${IMAGE_PULL_SECRETS_ARGUMENT} \
+      || return $?
+
+    log "Installing Istio Prometheus"
+    helm upgrade prometheus ${ISTIO_CHART_DIR}/istio-telemetry/prometheus \
```
pfmackin, line 130: Is Istio Grafana needed?

platform-operator/controllers/component/istio.go
```diff
@@ -1,28 +0,0 @@
-// Copyright (c) 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: We had this code because I saw that istio had a job hanging around in some use cases.   We have to test and make sure we don't see this.

bobdonat: There are no jobs with istio 1.7.3. You will notice fewer pods too.

platform-operator/helm_config/overrides/istio-values.yaml
bobdonat: These are the defaults.  Without explicitly setting the arch setting some of the installs would fail.

platform-operator/controllers/component/registry.go
bobdonat: I was hoping for one chart too but istio is not setup that way. We should be able  to only have 1 istio component in the crd.

pfmackin (approved):

PR #609, opened 2021-02-05
VZ-2105: remove ability to disable oam component installation

kminder:
Removes the configurability that made installing OAM components optional.  With this change the OAM components are now always installed.


bobdonat (approved):
Looks good to me

michael-cico (approved):

PR #614, opened 2021-02-05
Update controller-gen to 0.4.1

michael-cico:
Update the Makefiles for the platform and application operators to use controller-gen 0.4.1.

This also required updating the VAO webhook configurations to declare values for 2 attributes:
- sideEffects
- admissionReviewVersions


application-operator/PROJECT
```diff
@@ -1,5 +1,11 @@
 # Copyright (C) 2020, 2021, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+# Copyright (C) 2020, 2021, Oracle and/or its affiliates.
```
pfmackin, line 3: what are all these extra copyrights?

michael-cico: not sure, it got generated when I ran make manifests.  Let me try to revert that.

markxnelson (approved):

application-operator/PROJECT
michael-cico: Fixed, and also fixed an issue with the header generation on application-operator/config/rbac/role.yaml.

application-operator/config/webhook/manifests.yaml
```diff
@@ -26,16 +28,19 @@ webhooks:
     - UPDATE
     resources:
```
tbeerbower, line 29: I thought we had DELETE here also.  If an appconfig is deleted there may be a default logging scope to clean up.

michael-cico: I did a diff against develop, it looks like it's not changed from that perspective, but I'll add it back in.

michael-cico: Fixed.... I added it to both webhook marker statements.

application-operator/PROJECT
mgianatagh: @desagar could the extra header generation be related to some recent changes to you made to have the copyright years set based on their previous value?

michael-cico: Is the PROJECT file generated?  I didn't think it was, I might need to add that back.

mgianatagh: The PROJECT file was originally generated by kubebuilder and I think it gets updated when you add a new API.  The file does not get re-generated.

@desagar recently made a change to how copyrights get updated and I think the extra line resulted from a bug in that logic.

application-operator/Makefile
```diff
@@ -70,9 +70,9 @@ manifests: go-mod controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	# Add copyright headers to the kubebuildr generated CRDs
 	./hack/add-crd-header.sh
-	# Add copyright headers to the kubebuildr generated manifests and to the edited PROJECT file
+	# Add copyright headers to the kubebuildr generated manifests
 	./hack/add-yml-header.sh config/webhook/manifests.yaml
-	./hack/add-yml-header.sh PROJECT
```
mgianatagh, line 75: I do not feel this should be removed.

application-operator/apis/oam/v1alpha1/ingresstrait_webhook.go
```diff
@@ -30,7 +30,7 @@ func (r *IngressTrait) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// +kubebuilder:webhook:verbs=create;update,path=/validate-oam-verrazzano-io-v1alpha1-ingresstrait,mutating=false,failurePolicy=fail,groups=oam.verrazzano.io,resources=ingresstraits,versions=v1alpha1,name=vingresstrait.kb.io
+// +kubebuilder:webhook:admissionReviewVersions={"v1","v1alpha1"},verbs=create;update,path=/validate-oam-verrazzano-io-v1alpha1-ingresstrait,mutating=false,failurePolicy=fail,groups=oam.verrazzano.io,resources=ingresstraits,sideEffects=None,versions=v1alpha1,name=vingresstrait.kb.io
```
mgianatagh, line 33: Why is the api version being changed as part of updating to controller-get 0.4.1?

application-operator/config/webhook/manifests.yaml
```diff
@@ -24,18 +26,22 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
+    - DELETE
     resources:
     - ApplicationConfigurations
+  sideEffects: NoneOnDryRun
```
bobdonat, line 32: why is this NoneOnDryRun while other webhook is Non

application-operator/controllers/webhooks/appconfig_defaulter.go
```diff
@@ -19,7 +19,7 @@ var log = ctrl.Log.WithName("webhooks.appconfig-defaulter")
 // AppConfigDefaulterPath specifies the path of AppConfigDefaulter
 const AppConfigDefaulterPath = "/appconfig-defaulter"
 
-// +kubebuilder:webhook:verbs=create;update,path=/appconfig-defaulter,mutating=true,failurePolicy=fail,groups=core.oam.dev,resources=ApplicationConfigurations,versions=v1alpha2,name=appconfig-defaulter.kb.io
+// +kubebuilder:webhook:admissionReviewVersions={"v1","v1alpha1"},verbs=create;update;delete,path=/appconfig-defaulter,mutating=true,failurePolicy=fail,groups=core.oam.dev,resources=ApplicationConfigurations,sideEffects=NoneOnDryRun,versions=v1alpha2,name=appconfig-defaulter.kb.io
```
mgianatagh, line 22: Similar question as earlier - if the api version is changing for a reason other than updating controller-gen I feel that should be handled in a separate PR.

application-operator/Makefile
michael-cico: yeah, adding it back.

application-operator/apis/oam/v1alpha1/ingresstrait_webhook.go
michael-cico: The generation was failing, and required admissionReviewVersions.  I wasn't sure what should go in here for versions, this is one of the things I needed help on.

application-operator/config/webhook/manifests.yaml
michael-cico: None isn't allowed on mutating webhooks.

application-operator/controllers/webhooks/appconfig_defaulter.go
michael-cico: Similar answer here...wasn't sure.  Do we only want v1alpha1?

application-operator/Makefile
michael-cico: This should be fixed, as with the header duplication in the PROJECT file

application-operator/apis/oam/v1alpha1/ingresstrait_webhook.go
bobdonat: I have the same question as Mike G.   Also, we are not changing the platform operator api version but we are changing the application operator api version.

i think this annotation can go if it is problematic

michael-cico: I only added it because it was failing without it.

bobdonat: platform-operator does not have this annotation.

application-operator/config/webhook/manifests.yaml
michael-cico: Actually, I was mistaken here, seems to be OK for None here.

application-operator/PROJECT
michael-cico: I re-added this and fixed the hack script for the header update for existing files.

desagar: The hack script for copyright header generation is pretty stupid. It gets the most recent version out of the current branch to tack on the copyrights. If you run make manifests without there being any changes to the PROJECT file, it's possible this could have happened.

michael-cico: OK, backed out this change.

application-operator/apis/oam/v1alpha1/ingresstrait_webhook.go
michael-cico: admissionReviewVersions indicates what version of the AdmissionReview the webhook will accept, it's not our object version.  I've set this to v1, v1beta1 per the examples on this page below.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

application-operator/config/webhook/manifests.yaml
michael-cico: Updated to None.

application-operator/controllers/webhooks/appconfig_defaulter.go
michael-cico: See above comment.

application-operator/apis/oam/v1alpha1/ingresstrait_webhook.go
michael-cico: Reverted this change, and changed the build to not generate the webhook manifests; this matches how we manage the platform operator webhook configurations, and will let us uptake v1 webhook API when we desire to.

mgianatagh (approved):

bobdonat (approved):

PR #615, opened 2021-02-05
Initial setup of controller and reconciler for VerrazzanoManagedCluster objects

mgianatagh:
The initial set of a controller and reconciler for VerrazzanoManagedCluster resources.  The reconciler does not do any real work yet.

Added integration tests for testing the creating and deletion of VerrazzanoManagedCluster resources.

Added unit tests for the new controller.

### DIFF
--- a/platform-operator/controllers/verrazzanomanagedcluster/verrazzanomanagedcluster_controller.go
+++ b/platform-operator/controllers/verrazzanomanagedcluster/verrazzanomanagedcluster_controller.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	clustersv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/clusters/v1alpha1"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// VerrazzanoManagedClusterReconciler reconciles a VerrazzanoManagedCluster object
+type VerrazzanoManagedClusterReconciler struct {
+	client.Client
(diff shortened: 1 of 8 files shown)


markxnelson (approved):

platform-operator/controllers/verrazzanomanagedcluster/verrazzanomanagedcluster_controller.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
```
pfmackin, line 2: You should only have one controller per controllers/xyz directory, so you don’t need such a long name for the go file.

platform-operator/go.mod
```diff
@@ -1,11 +1,12 @@
-// Copyright (c) 2020, Oracle and/or its affiliates.
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 module github.com/verrazzano/verrazzano/platform-operator
 
 go 1.13
 
 require (
+	github.com/go-logr/logr v0.1.0
```
bobdonat, line 9: this is an extra dependency which is not required.  the zap logger is the only logger needed.

platform-operator/main.go
```diff
@@ -146,6 +147,15 @@ func main() {
 		mgr.GetWebhookServer().CertDir = config.CertDir
 	}
 
+	// Setup the reconciler for VerrazzanoManagedCluster objects
+	if err = (&clusterscontroller.VerrazzanoManagedClusterReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("VerrazzanoManagedCluster"),
```
bobdonat, line 153: the "Log" field is not needed and if used you will not get consistent logging with the caller field.

platform-operator/go.mod
mgianatagh: fixed in next PR

platform-operator/main.go
mgianatagh: removed in next PR

platform-operator/controllers/verrazzanomanagedcluster/verrazzanomanagedcluster_controller.go
mgianatagh: I was following the pattern I saw in the application-operator folders containing controllers.  Are you suggesting to simplify the name to "controller.go"?

PR #620, opened 2021-02-08
VZ-2047: VerrazzanoWebLogicWorkload CRD and supporting files and updates

fredtibbitts:
This PR adds the CRD for VerrazzanoWebLogicWorkload and supporting generated files, types, and skeleton controller.


kminder (approved):

application-operator/Makefile
```diff
@@ -116,14 +116,14 @@ go-build:
 		main.go
 
 .PHONY: go-build-linux
-go-build-linux:
+go-build-linux: check
```
mgianatagh, line 119: Adding this dependency will cause the format checks to run twice in the build pipeline.  They are current run as a separate stage.

fredtibbitts: The problem is that, when doing local development, you have to remember to run `make check`. Developers shouldn't have to remember to do that, it should just be part of `make go-install`, `make unit-test`, etc. I think it would be better to remove the explicit pipeline step and have it done as part of another target, but I don't want to take that on now so I guess I'll back this out.

mgianatagh: Yes, I agree it is better to back this out.  If you want to make a change to the behavior of the build pipeline, that would be preferable to address in a separate PR.

fredtibbitts: Done

kminder (approved):

PR #621, opened 2021-02-08
fix appconfig UPDATE operation

shihchang:
fix appconfig UPDATE operation by requeue the loggingscope.reconcile event

### DIFF
--- a/application-operator/controllers/loggingscope/helidon.go
+++ b/application-operator/controllers/loggingscope/helidon.go
@@ -125,27 +125,25 @@ func (h *HelidonHandler) Apply(ctx context.Context, workload vzapi.QualifiedReso
 		duration := time.Duration(rand.IntnRange(5, 10)) * time.Second
 		return &ctrl.Result{Requeue: true, RequeueAfter: duration}, nil
 	}
-	if !fluentdFound {
-		err = h.ensureFluentdConfigMap(ctx, scope.GetNamespace(), workload.Name)
-		if err != nil {
-			return nil, err
-		}
-		err = h.ensureEsSecret(ctx, scope.GetNamespace(), scope.Spec.SecretName)
-		if err != nil {
-			return nil, err
-		}
-		volumes := CreateFluentdHostPathVolumes()
-		for _, volume := range volumes {
-			deploy.Spec.Template.Spec.Volumes = append(deploy.Spec.Template.Spec.Volumes, volume)
-		}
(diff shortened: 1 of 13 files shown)


application-operator/controllers/loggingscope/helidon.go
```diff
@@ -105,21 +110,29 @@ type HelidonHandler struct {
...
 	appContainer, fluentdFound := searchContainers(deploy.Spec.Template.Spec.Containers)
+	h.Log.V(1).Info("Update Deployment", "Deployment", deploy.Name, "fluentdFound", fluentdFound)
+	if fluentdFound {
+		// If the Deployment does contain a FLUENTD container
+		// requeue with a jittered delay to account for situation where the Deployment should be
+		// updated by the oam-kubernetes-runtime
+		duration := time.Duration(rand.IntnRange(5, 10)) * time.Second
+		return &ctrl.Result{Requeue: true, RequeueAfter: duration}, nil
+	}
 	if !fluentdFound {
```
tbeerbower, line 128: Is this check still required since the above `if fluentdFound` always returns?

shihchang: removed !fluentdFound check which is no longer needed

application-operator/controllers/loggingscope/loggingscope_controller.go
```diff
@@ -91,7 +91,10 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			}
 			continue
 		}
-		err = handler.Apply(ctx, resource, scope)
+		result, err := handler.Apply(ctx, resource, scope)
+		if result != nil {
+			return *result, nil
```
tbeerbower, line 96: This is in a loop on the workloads.  If there are multiple workloads could returning from within the loop be bad (some workloads not yet processed)?  Or is it all just assumed that the result is a re-queue and it will all get handled eventually?

shihchang: I do assume it will all get handled eventually. Ideally, we should requeue only the hanlder.Apply of the workload which need to wait for expected deployment/state. But the reconcile API does not seem allow such result. I did try an alternative of returning an error to cause the reconcile to re-run, but that has similar issue of the returned error cannot be made specific to a hanlder/workload.

tbeerbower (approved):

PR #625, opened 2021-02-09
Do not page on PR issues

agv426:
Looking at how to best verify this, the change is to add a conditional to the job which will not execute on PR's

From the github examples:
     https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issue_comment


jmaron99:
LGTM.  Let me know when you've managed to conduct a test.

markxnelson (approved):

PR #626, opened 2021-02-09
Add  sidecar.istio.io/inject annotation to Coherence workload

bobdonat:
This pull request adds the sidecar.istio.io/inject annotation as part of the Coherence workload and sets the value to false.
Setting the value to false insures that the sidecar istio-provy is not injected into a Coherence pod.  Coherence does not work with istio so this is required.


ddrechse (dismissed):

fredtibbitts (commented):
Looks like you accidentally committed `go.sum`? Aside from that, this LGTM.

bobdonat:
> Looks like you accidentally committed `go.sum`? Aside from that, this LGTM.

Thanks.  I have reverted the commit to go.sum.

fredtibbitts (approved):

PR #661, opened 2021-02-11
VZ-2066 - add user rbac cluster roles and role bindings at install time

wmhopkins:
# Description

VZ-2066

Changes to install cluster roles and cluster role bindings for user RBAC:
- helm template changes to add new CRs and CRBs
- values.yaml updates to add default binding values
- api changes and controller logic changes to support user-specified binding subjects.

API changes: `platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go`

# Checklist 

- [x] Addressed the requirement and meets the acceptance criteria
- [x] Does not introduce any unapproved dependency
- [x] Makes sense and it easy to understand, and/or difficult areas of code are clearly documented so that they can be understood


platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -482,53 +482,93 @@ func getProfile(profileType installv1alpha1.ProfileType) InstallProfile {
 
 // getVerrazzanoInstallArgs Set custom helm args for the Verrazzano internal component as needed
 func getVerrazzanoInstallArgs(vzSpec *installv1alpha1.VerrazzanoSpec, log *zap.SugaredLogger) ([]InstallArg, error) {
-	if vzSpec.DefaultVolumeSource == nil {
-		return []InstallArg{}, nil
+	args := make([]InstallArg, 0, 10) // need to ensure the size grows as more args are added here
```
markxnelson, line 485: feels like unnecessary debt and brittleness

markxnelson: what's all this volume stuff? is that your code, or do you need a rebase maybe?

markxnelson:
i note that AT's are coming in a separate PR, from me :)

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -76,6 +80,24 @@ type VerrazzanoSpec struct {
 	VolumeClaimSpecTemplates []VolumeClaimSpecTemplate `json:"volumeClaimSpecTemplates,omitempty"`
 }
 
+// RoleBindingSubject specifes the kind and name of a subject to bind to
+type RoleBindingSubject struct {
+	// Kind specifies the kind value for an rbac subject for a RoleBinding
```
pfmackin, line 85: Does it make sense to use k8s rbac subject?  In general we want to use Kubernetes types directly if possibe https://github.com/kubernetes/api/blob/master/rbac/v1/types.go

platform-operator/controllers/verrazzano/installjob/install_config.go
wmhopkins: Yeah, this is the first go code I've written that wasn't copy/paste, so wasn't sure what the efficient/idiomatic thing to do was w.r.t. slices. Suggestions?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
wmhopkins: I originally started out that way, but ran in to some issues and realized we didn't really need the object, we just needed the spec --we can't use the CR directly. Had trouble trying to use just rbacv1.Subject.spec.

pfmackin: not sure what you mean, the rbac subject contains all the reference info for the subject.  There is no subject.spec, it is not a top level CR.

platform-operator/controllers/verrazzano/installjob/install_config.go
wmhopkins: re: the volume stuff, that was there before, I just added InstallArgs processing for the Security element, and had to revise the logic for the existing code. Probably could have left the existing logic alone and simply written a separate method for mine, but at the time my understanding of appending slices wasn't good.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
wmhopkins: You're right -- that explains why trying to use rbac.Subject.spec didn't work. ;)

It's still got apiGroup and namespace fields, though, which we don't need and likely wouldn't honor, at least for the namespace field. We can do it that way, but I didn't want to force people to specify uneccessary info and have to process fields we won't use.

markxnelson (approved):

PR #673, opened 2021-02-13
use nodeport instead of lb address to access rancher during install

guoyongzhang:
# Description

Use nodeport instead of lb address to access rancher during install.  This way, installer will not access public endpoint, hence no DNS resolution needed.

Manually tested with OKE magicdns and OKE OCIDNS as well.

# Checklist 


gilbode:
Why do we need this?  Also why not just use the cluster address vs the ingress address?

markxnelson (approved):

PR #676, opened 2021-02-16
Initial setup of cluster agent to sync multi-cluster objects to a managed cluster

mgianatagh:
# Description

Initial setup of a cluster agent to sync multi-cluster objects to a managed cluster.  These changes are only syncing the MultiClusterSecret and MultiClusterComponent objects.  Subsequent PR's will incrementally complete all multi-cluster objects.

Fixes VZ-2065

# Checklist 


application-operator/mcagent/multicluster_agent_comonent.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: filename spelling (comonent).  I like how you use mc abbrev, wonder if we can use for filenames also (mcagent_component.go or mc_agent_component.go)

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/verrazzano/verrazzano/application-operator/constants"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Syncer contains context for synchronize operations
+type Syncer struct {
```
pfmackin, line 24: Is this a generic type used by other syncers?  If so it should be in a common file

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	// Start the thread for syncing multi-cluster objects
+	s := &Syncer{
+		AdminClient: adminClient,
+		MCClient:    client,
+		Log:         log,
+		ClusterName: secret.ClusterName,
+		Context:     context.TODO(),
+	}
+
+	go s.StartSync()
```
pfmackin, line 73: We already have a thread started by main, do we really need that if we are going to start another thread?

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		err = s.syncMCComponentObjects()
+		if err != nil {
+			s.Log.Error(err, "Error syncing MultiClusterComponent objects")
+		}
+		err = s.syncMCApplicationConfigurationObjects()
+		if err != nil {
+			s.Log.Error(err, "Error syncing MultiClusterApplicationConfiguration objects")
+		}
+		time.Sleep(5 * time.Minute)
```
pfmackin, line 102: 5 minutes seems like a very long time to wait for an app to get propagated to a remote cluster.  We should have some way of doing a lightweight poll more frequently.  Have you considered using watchers?

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+// Validate the cluster secret
+func validateClusterSecret(secret *corev1.Secret) error {
+	// The secret must contain a cluster-name
```
pfmackin, line 108: Do we do this validation on the admin cluster with a WebHook?

application-operator/mcagent/multicluster_agent.go
mgianatagh: Ok, I will do that in the next PR for this Jira.

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Get the clientset for accessing the admin cluster
+func getAdminClient(secret *corev1.Secret) (client.Client, error) {
+	// Create a temp file that contains the kubeconfig
+	tmpFile, err := ioutil.TempFile("/tmp", "kubeconfig")
+	if err != nil {
+		return nil, err
+	}
+
+	err = ioutil.WriteFile(tmpFile.Name(), secret.Data["kubeconfig"], 0777)
```
pfmackin, line 131: i think we only need rw permissions

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	defer os.Remove(tmpFile.Name())
```
pfmackin, line 132: maybe defer should be after the error check

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	clientset, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, err
+	}
+
+	return clientset, nil
+}
+
+// Check if the placement is for this cluster
+func isThisCluster(clusterName string, placement clustersv1alpha1.Placement) bool {
```
pfmackin, line 153: The name This implies that there is some "this" context being used.  Really, it just checks if the given cluster name is in a placement cluster list.

application-operator/mcagent/multicluster_agent_secret.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	clustersv1alpha1 "github.com/verrazzano/verrazzano/application-operator/apis/clusters/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// Synchronize MultiClusterSecret objects to the local cluster
+func (s *Syncer) syncMCSecretObjects() error {
+	// Get all the MultiClusterSecret objects from the admin cluster
+	allMCSecrets := &clustersv1alpha1.MultiClusterSecretList{}
```
pfmackin, line 15: this is not consistent with line 42 where we do no use the address.  We should use a consistent style, I prefer the line 42 style where the & is used when needed, getting rid of & here.

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+func StartAgent(client client.Client, log logr.Logger) {
+	// Wait for the existence of the verrazzano-cluster secret.  It contains the credentials
+	// for connecting to a managed cluster.
+	log.Info("Starting multi-cluster agent")
+	secret := corev1.Secret{}
+
+	for {
+		err := client.Get(context.TODO(), types.NamespacedName{Name: constants.MCRegistrationSecret, Namespace: constants.VerrazzanoSystemNamespace}, &secret)
+		if err != nil {
+			time.Sleep(60 * time.Second)
```
pfmackin, line 42: 60 secs seems way to long.   Shouldn't the secret exist (Is this the secret the VO creates?

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	go s.StartSync()
+}
+
+// StartSync - start the thread for syncing multi-cluster objects
+func (s *Syncer) StartSync() {
+	s.Log.Info("Starting sync of multi-cluster objects")
+
+	// Periodically loop looking for multi-cluster objects
+	for {
+		err := s.syncMCSecretObjects()
```
pfmackin, line 82: These syncs don't handle delete.  Can you make sure we have a JIRA to track delete processing.

application-operator/mcagent/multicluster_agent.go
mgianatagh: fixed in next push - will call inline

application-operator/mcagent/multicluster_agent.go
mgianatagh: I feel we need to discuss this frequency, but for initial push I picked 5 minutes.  Reducing to 1 minute for now.

Mike may not be keen on cross cluster watchers.  I feel we need to discuss that.

application-operator/mcagent/multicluster_agent_comonent.go
mgianatagh: I have renamed the files to use "mcagent" as the prefix

application-operator/mcagent/multicluster_agent.go
mgianatagh: no, this secret only would exist on the managed cluster.  Currently we create it manually.  Part of the registration process will be to generate the yaml to create.  But this is a generic Kubernetes secret so I don't think you can an additional validating web hook for that.

application-operator/mcagent/multicluster_agent.go
```diff
@@ -0,0 +1,161 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Syncer contains context for synchronize operations
+type Syncer struct {
+	AdminClient client.Client
+	MCClient    client.Client
```
pfmackin: might be better to call this localClient, MCClient sort of implies it is used for multiple cluster, where really it is just the local cluster if i understand it.

application-operator/mcagent/multicluster_agent.go
mgianatagh: ok, will change to 666 on next PR

application-operator/mcagent/multicluster_agent.go
mgianatagh: I don't know but just in case it seems safer to have it before the error check.  Worst case is the remove fails silently.

pfmackin (approved):

application-operator/mcagent/multicluster_agent.go
mgianatagh: I just pushed a change to make this function a method on the syncer.  So now there is a "this" context.

application-operator/mcagent/multicluster_agent_secret.go
```diff
@@ -0,0 +1,52 @@
...
+	listOptions := &client.ListOptions{}
+	err := s.AdminClient.List(s.Context, allMCSecrets, listOptions)
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	// Write each of the records that are targeted to this cluster
+	for _, mcSecert := range allMCSecrets.Items {
```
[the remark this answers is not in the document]
mgianatagh: good catch, will fix in the next PR

application-operator/mcagent/multicluster_agent_secret.go
mgianatagh: will look at changing that style in the next PR

desagar (approved):
We did a detailed interactive review and Michael also addressed some of Paul's comments. This looks good for an initial cluster agent create.

application-operator/mcagent/multicluster_agent.go
mgianatagh: yes, we have this on our list of Jiras to write

application-operator/mcagent/multicluster_agent.go
mgianatagh: will change in next PR

PR #677, opened 2021-02-16
remove k8s_user related settings for impersonation

lake-of-dreams:
# Description

The existing implementation of verrazzano-api requires a user for impersonation and that user was being read from `k8s_user` attribute of keycloak user. As per review from @wmhopkins it was decided that we might not require this attribute and the `sub` field from jwt token may directly be used for impersonation and there is no need to assign the concerned clusterrole to users individually as long as the group being impersonated is bound to that role. This was tested and found to be working fine. Hence this PR to remove the keycloak settings and role assignments/impersonation settings which are no longer required.

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
@@ -122,9 +122,6 @@ subjects:
 - apiGroup: rbac.authorization.k8s.io
   kind: Group
   name: {{ .Values.api.consoleUserGroup }}
-- apiGroup: rbac.authorization.k8s.io
-  kind: User
-  name: {{ .Values.api.consoleUser }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -233,10 +230,10 @@ data:
       ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
 
-    logJsonMessage(ngx.INFO, "Check for k8s_user in jwt token.")
-    if ( not (jwt_obj.payload) or not (jwt_obj.payload.k8s_user)) then
+    logJsonMessage(ngx.INFO, "Check for sub in jwt token.")
+    if ( not (jwt_obj.payload) or not (jwt_obj.payload.sub)) then
       ngx.status = 401
-      logJsonMessage(ngx.ERR, "No k8s_user asscoiated with user.")
+      logJsonMessage(ngx.ERR, "No sub asscoiated with user.")
       ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
 
@@ -261,7 +258,7 @@ data:
     for i,group in pairs(jwt_obj.payload.groups) do
       ngx.req.set_header("Impersonate-Group", group)
     end
-    ngx.req.set_header("Impersonate-User", jwt_obj.payload.k8s_user)
+    ngx.req.set_header("Impersonate-User", jwt_obj.payload.sub)
   nginx.conf: |
     #user  nobody;
     worker_processes  1;
--- a/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
@@ -226,14 +226,14 @@ data:
     logJsonMessage(ngx.INFO, "Check for groups in jwt token.")
     if ( not (jwt_obj.payload) or not (jwt_obj.payload.groups)) then
       ngx.status = 401
-      logJsonMessage(ngx.ERR, "No groups asscoiated with user.")
+      logJsonMessage(ngx.ERR, "No groups associated with user.")
       ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
 
     logJsonMessage(ngx.INFO, "Check for sub in jwt token.")
     if ( not (jwt_obj.payload) or not (jwt_obj.payload.sub)) then
       ngx.status = 401
-      logJsonMessage(ngx.ERR, "No sub asscoiated with user.")
+      logJsonMessage(ngx.ERR, "No sub associated with user.")
       ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
 
--- a/platform-operator/helm_config/charts/verrazzano/values.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/values.yaml
@@ -90,7 +90,6 @@ api:
   port: 8775
   clusterRole: console
   consoleUserGroup: console-users
-  consoleUser: console-user
 
 # OCI-related values
 oci:
--- a/platform-operator/scripts/install/config/keycloak.json
+++ b/platform-operator/scripts/install/config/keycloak.json
@@ -401,12 +401,7 @@
       "realm-management" : [ "manage-users" ]
     },
     "notBefore" : 0,
-    "groups" : [ "console-users" ],
-    "attributes": {
-      "k8s_user": [
-        "console-user"
-      ]
-    }
+    "groups" : [ "console-users" ]
   }, {
     "id" : "ca205b69-e8a8-45c9-88b1-ea9c1b072152",
     "createdTimestamp" : 1564417839778,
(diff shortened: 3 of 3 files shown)


platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -233,10 +230,10 @@ data:
       ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
 
-    logJsonMessage(ngx.INFO, "Check for k8s_user in jwt token.")
-    if ( not (jwt_obj.payload) or not (jwt_obj.payload.k8s_user)) then
+    logJsonMessage(ngx.INFO, "Check for sub in jwt token.")
+    if ( not (jwt_obj.payload) or not (jwt_obj.payload.sub)) then
       ngx.status = 401
-      logJsonMessage(ngx.ERR, "No k8s_user asscoiated with user.")
+      logJsonMessage(ngx.ERR, "No sub asscoiated with user.")
```
mgianatagh, line 236: "associated"

lake-of-dreams: fixed

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -229,14 +226,14 @@ data:
     logJsonMessage(ngx.INFO, "Check for groups in jwt token.")
     if ( not (jwt_obj.payload) or not (jwt_obj.payload.groups)) then
       ngx.status = 401
-      logJsonMessage(ngx.ERR, "No groups asscoiated with user.")
+      logJsonMessage(ngx.ERR, "No groups associated with user.")
```
wmhopkins, line 229: We should not exit here just because there are no groups - we should let k8s RBAC decide if a user with no groups should be allowed access.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -122,9 +122,6 @@ subjects:
 - apiGroup: rbac.authorization.k8s.io
```
wmhopkins, line 122: This should also be removed - we don't want/need to create a group.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -229,14 +226,14 @@ data:
     logJsonMessage(ngx.INFO, "Check for groups in jwt token.")
```
wmhopkins, line 226: What specifically does logJsonMessage() do? I did a quick google but didn't find any documentation. Want to make sure this writes only to nginx logs and is not returned to the client.

platform-operator/helm_config/charts/verrazzano/values.yaml
```diff
@@ -90,7 +90,6 @@ api:
   port: 8775
   clusterRole: console
   consoleUserGroup: console-users
```
wmhopkins, line 92: Should delete this, too.

platform-operator/scripts/install/config/keycloak.json
```diff
@@ -401,12 +401,7 @@
       "realm-management" : [ "manage-users" ]
     },
     "notBefore" : 0,
-    "groups" : [ "console-users" ],
-    "attributes": {
-      "k8s_user": [
-        "console-user"
-      ]
-    }
+    "groups" : [ "console-users" ]
```
wmhopkins, line 404: Should not create this group.

wmhopkins (commented):
Not directly related to this change, but I noticed you are creating a ClusterRole and ClusterRoleBinding for "console-user". Instead of having a console-specific role, we should use existing verrazzano-admin and verrazzano-monitor roles to control access. Those roles are now being installed, and are bound to subjects (user/group) based on the userrolebindings section of the verrazzano CR used for installation.

I noticed your roles include what's needed to access, e.g., weblogic operator as well as VZ resources. We should capture the details of the grants you make here, and ensure that they are incorporated into the verrazzano-admin role somehow, either through aggregation or through binding multiple roles to the same subject.

lake-of-dreams:
> Not directly related to this change, but I noticed you are creating a ClusterRole and ClusterRoleBinding for "console-user". Instead of having a console-specific role, we should use existing verrazzano-admin and verrazzano-monitor roles to control access. Those roles are now being installed, and are bound to subjects (user/group) based on the userrolebindings section of the verrazzano CR used for installation.
> 
> I noticed your roles include what's needed to access, e.g., weblogic operator as well as VZ resources. We should capture the details of the grants you make here, and ensure that they are incorporated into the verrazzano-admin role somehow, either through aggregation or through binding multiple roles to the same subject.

@wmhopkins can you please let me know what would be the best practice? Should I keep this role as is and tie this to `verrazznao-project-admin` using aggregation? Or should I create separate roles based on api groups for example console will need access to **any**  resource type that can be used in a component, trait or scope e.g. `Domain` which can be a workload type and is in `weblogic.oracle` apiGroup and as well as `IngressTrait` which is in `oam.verrazzano.io` group?

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: it is not a public api function but defined in this template itself - see [here](https://github.com/verrazzano/verrazzano/blob/d75e4fee75ec6225f28a673cef14f500d432e9dc/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml#L156). this just writes to error.log

wmhopkins: Thanks, seems fine.

wmhopkins:
> > Not directly related to this change, but I noticed you are creating a ClusterRole and ClusterRoleBinding for "console-user". Instead of having a console-specific role, we should use existing verrazzano-admin and verrazzano-monitor roles to control access. Those roles are now being installed, and are bound to subjects (user/group) based on the userrolebindings section of the verrazzano CR used for installation.
> > I noticed your roles include what's needed to access, e.g., weblogic operator as well as VZ resources. We should capture the details of the grants you make here, and ensure that they are incorporated into the verrazzano-admin role somehow, either through aggregation or through binding multiple roles to the same subject.
> 
> @wmhopkins can you please let me know what would be the best practice? Should I keep this role as is and tie this to `verrazznao-project-admin` using aggregation? Or should I create separate roles based on api groups for example console will need access to **any** resource type that can be used in a component, trait or scope e.g. `Domain` which can be a workload type and is in `weblogic.oracle` apiGroup and as well as `IngressTrait` which is in `oam.verrazzano.io` group?

I guess I'd argue that you shouldn't do **anything** w.r.t. roles. This code is meant to do authentication, only. As a completely separate effort, we are creating user-facing roles for the k8s API objects, which would also protect the UI.

It's a good idea to preserve what you have here -- the work of figuring out which permissions are needed is valuable -- but only the standard VZ user roles will should be needed. Rather than create new/different roles here, we should make sure the standard roles include the necessary access.

wmhopkins:
Doesn't look like I can comment on lines you didn't change, but I'd also suggest:
- This is just a nit-pick, but not being able to get an authorization header from the request isn't really an error; I'd argue that's normal processing. We may fail login, but there was no processing error, just a lack of a token supplied by a user (assuming no systemic problems with the HTTP server, etc.). Might want to log those (and similar) conditions as INFO, not ERR. From our point of view, a missing or invalid token isn't an error, it's normal processing that did not result in a successful authentication.
- More importantly, we should probably do a better job of validating the token. I'm not sure if the verify_jwt_obj() checks that the key matches the keyid from the token, but if not we should verify it, and make sure the keyid matches the keycloak verrazzano-system key that we're using to verify. We should verify that the iss (issuer) field matches keycloak. And we should verify that the token is valid and unexpired, time-wise. Looks like that's all pretty straightforward using claim_specs.

lake-of-dreams:
closed as implemented as part of https://github.com/verrazzano/verrazzano/pull/801

PR #698, opened 2021-02-18
Move the load of the example images

tbeerbower:
# Description

Updates the Jenkinsfile to move the loading of the todo images into the todo stage and the bob's books images into the bob's books stage.  This ensures that the image loads only happen if required.

# Checklist 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -383,27 +383,6 @@ pipeline {
 
                             # wait for Verrazzano install to complete
                             ./tests/e2e/config/scripts/wait-for-verrazzano-install.sh
-
-                            # Hack
-                            # OCR images don't work with KIND.
-                            # The ToDoList example image currently cannot be pulled in KIND.
-                            docker pull container-registry.oracle.com/verrazzano/example-todo:0.8.0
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -452,6 +452,13 @@ pipeline {
                                 expression {params.RUN_EXAMPLE_TESTS == true}
                             }
                             steps {
+                                sh """
+                                      # Hack
```
kminder, line 456: Would it be too much to replace "Hack" with a brief statement about what is required to remove the workaround?  If this persists forever don't really want "Hack" in a public repo.

Jenkinsfile
```diff
@@ -452,6 +452,13 @@ pipeline {
...
+                                      # OCR images don't work with KIND.
```
kminder, line 457: Are we allowed to use the OCR abbreviation?

Jenkinsfile
tbeerbower: I removed "hack".  I'm not exactly sure what is required to remove the workaround.

Jenkinsfile
tbeerbower: I removed OCR.

PR #701, opened 2021-02-19
Process only locally placed multi cluster resources

desagar:
# Description

Multi cluster controllers process only locally placed multi cluster resources.

Fixes VZ-2115

Cleanup changes:
- MC controllers should also set labels and annotations when mutating resources
- Added deletion test case for MCxxx resources
- Don’t log errors on notfound MCxxx resources - only if some other problem occurred
- separated test utility and non-test utility methods in the clusters package

# Checklist 

### DIFF
--- a/application-operator/controllers/clusters/cluster_utils.go
+++ b/application-operator/controllers/clusters/cluster_utils.go
@@ -4,20 +4,25 @@
 package clusters
 
 import (
-	"encoding/json"
+	"context"
 	"fmt"
-	"io/ioutil"
 	"time"
 
-	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	"github.com/go-logr/logr"
 	clustersv1alpha1 "github.com/verrazzano/verrazzano/application-operator/apis/clusters/v1alpha1"
+	"github.com/verrazzano/verrazzano/application-operator/constants"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/yaml"
 )
+
+// MCRegistrationSecretFullName is the full NamespacedName of the cluster registration secret
+var MCRegistrationSecretFullName = types.NamespacedName{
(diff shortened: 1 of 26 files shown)


application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -76,40 +81,31 @@ func NewScheme() *runtime.Scheme {
...
-	return ctrl.Request{
-		NamespacedName: types.NamespacedName{
-			Namespace: namespace,
-			Name:      name,
-		},
-	}
-}
+// IsPlacedInThisCluster determines whether the given Placement represents placement in the current
+// cluster. Current cluster's identity is determined from the verrazzano-cluster secret
+func IsPlacedInThisCluster(ctx context.Context, rdr client.Reader, placement clustersv1alpha1.Placement) bool {
```
pfmackin, line 86: I think GZ just did an equiv function, we should share the code

application-operator/controllers/clusters/multiclusterapplicationconfiguration/controller.go
```diff
@@ -80,6 +83,8 @@ func (r *Reconciler) createOrUpdateAppConfig(ctx context.Context, mcAppConfig cl
 // MultiClusterApplicationConfiguration
 func (r *Reconciler) mutateAppConfig(mcAppConfig clustersv1alpha1.MultiClusterApplicationConfiguration, oamAppConfig *v1alpha2.ApplicationConfiguration) {
 	oamAppConfig.Spec = mcAppConfig.Spec.Template.Spec
+	oamAppConfig.Labels = mcAppConfig.Spec.Template.Metadata.Labels
```
pfmackin, line 86: so are MC resources supposed to have the labels on the MC part or the YAML, I would have expected them to be on the wrapped YAML (i.e. OAM component metadata)

mgianatagh: I do not know if they are supposed to have labels, but I feel the intent of this code is to make sure that if an Mcxxxx resource has labels or annotations applied, they should not get lost and be applied.

application-operator/controllers/clusters/cluster_utils.go
desagar: I agree we should share code, but currently those are in 2 different packages and the method Michael wrote is on a type that is specific to the mc agent.
I'd rather try to factor out this functionality into a common place in a separate PR.

application-operator/controllers/clusters/multiclusterapplicationconfiguration/controller.go
desagar: yes - we have an action item to determine what labels or annotations we are putting on these embedded resources. This is just to make sure we don't lose any that have been explicitly provided by the users.

mgianatagh (approved):

PR #703, opened 2021-02-19
Create secret with kubeconfig that uses serviceaccount token

pfmackin:
# Description

When a VMC is created, the new code will create a secret that contains a kubeconfig YAML which restricts access as specified in verrazzano-managed-cluster ClusterRole.  The secret name is then written to the VMC spec.  This secret will then be used by the MC agent (not included of this JIRA) to access the admin cluster thereby restricting access to MCxyz read an MCxyz/status write.  

Fixes VZ-2058

# Checklist 


platform-operator/controllers/clusters/kubeconfig_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if err := r.Get(context.TODO(), saNsn, &sa); err != nil {
+		return fmt.Errorf("Failed to fetch the service account for VMC %s/%s, %v", managedNamespace, saName, err)
+	}
+	if len(sa.Secrets) == 0 {
+		return fmt.Errorf("Service account %s/%s is missing a secret name", managedNamespace, saName)
+	}
+
+	// Get the service account token from the secret
+	tokenName := sa.Secrets[0].Name
```
desagar, line 94: do we know for sure this _always_ guaranteed to be the zeroth item in the sa secrets array?

desagar: I do see that in the token secret, there is an annotation that points to the service account name.  
I was able to edit the service account and add any secret names I wanted before or after the token secret.

platform-operator/controllers/clusters/kubeconfig_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = r.createOrUpdateSecret(vmc, string(kcBytes), secretName, managedNamespace)
+	if err != nil {
+		return err
+	}
+
+	// Save the KubeconfigSecret in the VMC
+	vmc.Spec.KubeconfigSecret = secretName
```
desagar, line 148: in my manual tests, I don't see the KubeconfigSecret field of the VMC being updated for some reason

platform-operator/controllers/clusters/kubeconfig_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+//   3. get the in-memory client configuration used to access the admin cluster
+//   4. build a kubeconfig struct using data from the client config and the service account token
+//   5. save the kubeconfig as a secret
+//   6. update VMC with the kubeconfig secret name
+func (r *VerrazzanoManagedClusterReconciler) reconcileKubeConfig(vmc *clusterapi.VerrazzanoManagedCluster) error {
+
+	// The same managed name and  vmc namespace is used for the service account and the kubeconfig secret,
+	// for clarity use different vars
+	saName := generateManagedResourceName(vmc.Name)
+	secretName := generateManagedResourceName(vmc.Name)
```
desagar, line 77: The managedResourceName is not a good name for the secret. It is simply something like `cluster1-managed-cluster` which doesn't really tell you that it is the kubeconfig secret.
I'd prefer this to be changed.

platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -92,6 +98,46 @@ func TestCreateVMC(t *testing.T) {
...
+		})
+
+	// Expect a call to get the kubeconfig secret - return that it does not exist
+	mock.EXPECT().
+		Get(gomock.Any(), types.NamespacedName{Namespace: namespace, Name: generateManagedResourceName(name)}, gomock.Not(gomock.Nil())).
+		Return(errors.NewNotFound(schema.GroupResource{Group: namespace, Resource: "Secret"}, generateManagedResourceName(name)))
+
+	// Expect a call to create the kubeconfig secret
+	mock.EXPECT().
+		Create(gomock.Any(), gomock.Any()).
```
desagar, line 128: can this create call be parameterized with the expected kubeconfig secret name that would be passed to the create?

platform-operator/controllers/clusters/kubeconfig_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+		// will automatically get deleted when the VerrazzanoManagedCluster is deleted
+		return controllerutil.SetControllerReference(vmc, &secret, r.Scheme)
+	})
+}
+
+// Mutate the secret, setting the kubeconfig data
+func (r *VerrazzanoManagedClusterReconciler) mutateSecret(secret *corev1.Secret, b64KubeConfig string) error {
+	secret.Type = corev1.SecretTypeOpaque
+	secret.Data = map[string][]byte{
+		"kubeconfig": []byte(b64KubeConfig),
```
desagar, line 175: I would suggest a constant for this - for other secret data fields in application operator, we used constants. This is not a dealbreaker and can be changed in a future PR.

platform-operator/controllers/clusters/kubeconfig_reconciler.go
desagar: never mind this worked.

platform-operator/controllers/clusters/kubeconfig_reconciler.go
pfmackin: I can append -kubeconfig to the name.  It is a common pattern to use the same name for related resources, you see this with certifcates and the secret that contains the actual cert.

mgianatagh: It may be helpful if the name of the secret was more indicative of what it is.  The secret contains a kubeconfig that is just for one managed cluster to connect to the admin cluster.  Maybe something like "cluster1-admin-kubeconfig"?

desagar (approved):

platform-operator/controllers/clusters/kubeconfig_reconciler.go
pfmackin: I am using the approach described here.  I think it is ok https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengaddingserviceaccttoken.htm

platform-operator/controllers/clusters/kubeconfig_reconciler.go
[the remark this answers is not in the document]
pfmackin: will do in follow up

PR #705, opened 2021-02-19
Ingress support for Helidon Workload

arjav-desai:
# Description

This PR contains changes to support ingress trait for VerrazzanoHelidonWorkload.

Fixes VZ-2120

# Checklist 

### DIFF
--- a/application-operator/controllers/navigation/trait.go
+++ b/application-operator/controllers/navigation/trait.go
@@ -5,6 +5,7 @@ package navigation
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


application-operator/controllers/navigation/trait.go
```diff
@@ -47,7 +48,11 @@ func FetchWorkloadFromTrait(ctx context.Context, cli client.Reader, log logr.Log
 		return nil, err
 	}
 
-	if IsVerrazzanoWorkloadKind(workload) {
+	// Getting kind of helidon workload i.e. "VerrazzanoHelidonWorkload"
+	helidonWorkloadKind := reflect.TypeOf(vzapi.VerrazzanoHelidonWorkload{}).Name()
+
+	// This is required only if the workload wraps unstructured data
+	if IsVerrazzanoWorkloadKind(workload) && (helidonWorkloadKind != workload.GetKind()) {
```
fredtibbitts, line 55: Since `IsVerrazzanoWorkloadKind` is already doing string comparisons on the kind, you could just do `workload.GetKind() != "VerrazzanoHelidonWorkload"` and not use `reflect`, but I suppose your way is a little safer in the event that the kind changes in the future.

fredtibbitts (approved):

application-operator/controllers/navigation/trait.go
fredtibbitts: After thinking about this a little more, I'm wondering if it would be better to change `IsVerrazzanoWorkload` to something like `WorkloadNeedsUnwrapping` and then move the Helidon check into that function. That seems cleaner to me, what do you think?

arjav-desai: Based on usage search, I see this being used in Logging Scope as well and not sure if renaming it to "WorkloadNeedsUnwrapping" would help there. I will take a note and revisit it when I add Logging Scope for Helidon workload.

PR #712, opened 2021-02-22
Incremental improvements to user-facing roles

wmhopkins:
# Description

This PR tweaks roles/bindings as follows:
- creates a separate role for install (aggregated to verrazzano admin)
- use correct verbs for crud operations
- add core.oam.dev apigroup
- add access to weblogic and coherence operator apis in application read/write roles
- add default bindings for k8s admin (verrazzano-admin) and view (verrazzano-monitor). Namespaced bindings will need to be created as part of the projects workflow that creates the project-based role bindings.

# Checklist 


platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
```diff
@@ -276,6 +276,27 @@ subjects:
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
+metadata:
+  name: verrazzano-installer
```
pfmackin, line 280: Why do we need this role?  What about upgrade and other operations that the platform operator might do, like register cluster

platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
```diff
@@ -285,9 +306,10 @@ rules:
 - apiGroups:
     - verrazzano.io
     - oam.verrazzano.io
-    - install.verrazzano.io
+    - core.oam.dev
```
pfmackin, line 309: There is a stack of products that verrazzano installs, not just oam, such as cert manager, istio , etc.   Are we going to enumerate all of these api groups or is this just focused on OAM related resources.

platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
```diff
@@ -303,12 +325,15 @@ rules:
...
   resources:
     - "*"
   verbs:
-    - put
-    - post
+    - create
```
pfmackin, line 332: what about list

platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
wmhopkins: We don't technically need it, but I figured it could be helpful to break it out separately, so that permission can be granted separately, and/or aggregated into multiple roles. It's aggregated into verrazzano-admin now, and won't be separately bound. It's more like an implementation detail than a role we expect people to bind directly, although they could.

verrazzano-admin will be able to do upgrade since it incudes verrazzano-installer (maybe verrazzano-install is a better name?).

I thought cluster registration would be part of the MC operator, not the platform operator -- are those functions implemented by platform operator?

platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
wmhopkins: We're including wls and coherence operator permissions in the VZ roles. We're going to separately bind k8s roles -- k8s admin, edit, or view when we bind VZ roles. This is a first approximation (or second approximation, really). We can add those in, but as much as possible I'd like to leverage existing roles created by the other products, rather than hard-wire specific grants into our roles.

platform-operator/helm_config/charts/verrazzano/templates/01-verrazzano-operator.yaml
wmhopkins: This is a "write" role -- the read verbs are in the corresponding "read" roles. Those who need both get them via aggregation (e.g., vz-admin includes both vz-system-read and vz-system-write).

desagar (approved):
interactively reviewed with Will

mgianatagh (approved):

PR #713, opened 2021-02-22
Bob's Book example e2e test fix

fredtibbitts:
# Description

Two PRs were merged that resulted in the Bob's Books example end-to-end test failing. When there were three separate app configs, we had to find the correct app ingress gateway for the app to get the right hostname. When the Bob's Books example app changed to use a single config, the test was looking for ingress gateways that no longer existed, and the tests could not access the 3 separate UI endpoints.

Since there is now a single app config, there is a single gateway and single ingress host name. The fix is to simply not look for specific gateways and just use the first (and only) gateway.

I also updated the Bob's Books README to reflect the fact that there is now a single app.

# Checklist 

### DIFF
--- a/examples/bobs-books/README.md
+++ b/examples/bobs-books/README.md
@@ -1,6 +1,6 @@
 # Bob's Books
 
-The Bob's Books example is a set of three applications based on WebLogic, Helidon, and Coherence. For more information and the source code of this application, see the [Verrazzano examples](https://github.com/verrazzano/examples).
+The Bob's Books example is an application based on WebLogic, Helidon, and Coherence. For more information and the source code of this application, see the [Verrazzano examples](https://github.com/verrazzano/examples).
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


examples/bobs-books/README.md
```diff
@@ -1,6 +1,6 @@
 # Bob's Books
 
-The Bob's Books example is a set of three applications based on WebLogic, Helidon, and Coherence. For more information and the source code of this application, see the [Verrazzano examples](https://github.com/verrazzano/examples).
+The Bob's Books example is an application based on WebLogic, Helidon, and Coherence. For more information and the source code of this application, see the [Verrazzano examples](https://github.com/verrazzano/examples).
```
rosemarymarano, line 3: I think that for the link to the application code, perhaps, you should use https://github.com/verrazzano/examples/blob/master/bobs-books/README.md, because it gives the user the concept of Bob's Books having three main parts (which correlates to the previous sentence about how the application is based on WebLogic, Helidon, and Coherence.) And, also helps explain the three UI URLs and such that follow.

rosemarymarano (approved):
Other than my suggestion, the revised doc is fine.

examples/bobs-books/README.md
fredtibbitts: That README you linked to is this README. :) The `github.com/verrazzano/examples` link has the source code for the stuff that this README talks about.

PR #719, opened 2021-02-23
Add Authorization Test to Acceptance Test Suite VZ2175

ddrechse:
# Description

Fixes VZ-9999

# Checklist 

- [ x] Checked that I included or updated copyright and license notices in all files that I altered
- [ x] Added or updated acceptance tests if appropriate

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -464,6 +464,14 @@ pipeline {
                                 runGinkgo('examples/hello-helidon')
                             }
                         }
+                        stage('examples authorization policy') {
+                            when {
+                                expression {params.RUN_EXAMPLE_TESTS == true}
+                            }
+                            steps {
+                                runGinkgo('examples/authorization-policy')
+                            }
+                        }
                         stage('examples bobs') {
                             when {
                                 expression {params.RUN_EXAMPLE_TESTS == true}
--- a/examples/authorization-policy/bar/istio-securitytest-app.yaml
+++ b/examples/authorization-policy/bar/istio-securitytest-app.yaml
@@ -1,0 +1,45 @@
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: authpolicy-appconf
+  namespace: bar
+  annotations:
+    version: v1.0.0
+    description: "Authorization Policy Test App"
+spec:
+  components:
+    - componentName: springboot-frontend
+      traits:
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: MetricsTrait
+            spec:
+              scraper: verrazzano-system/vmi-system-prometheus-0
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: IngressTrait
+            metadata:
+              name: springboot-ingress
+            spec:
+              rules:
+                - hosts:
+                    - "authpolicy.bar.com"
+                  paths:
+                    - path: "/"
+                      pathType: Prefix
+                    - path: "/verrazzanoFact"
+                      pathType: Prefix
+                    - path: "/externalCall"
+                      pathType: Prefix
+    - componentName: springboot-backend
+      traits:
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: MetricsTrait
+            spec:
+                scraper: verrazzano-system/vmi-system-prometheus-0
+    - componentName: sleep
+
+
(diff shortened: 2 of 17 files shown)


examples/authorization-policy/bar/istio-securitytest-app.yaml
```diff
@@ -0,0 +1,45 @@
...
+    description: "Authorization Policy Test App"
+spec:
+  components:
+    - componentName: springboot-frontend
+      traits:
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: MetricsTrait
+            spec:
+              scraper: verrazzano-system/vmi-system-prometheus-0
```
kminder, line 19: I think this is the default.  Was it required to have this explicitly?

examples/authorization-policy/foo/istio-securitytest-app.yaml
```diff
@@ -0,0 +1,45 @@
...
+              scraper: verrazzano-system/vmi-system-prometheus-0
+        - trait:
+            apiVersion: oam.verrazzano.io/v1alpha1
+            kind: IngressTrait
+            metadata:
+              name: springboot-ingress
+            spec:
+              rules:
+                - hosts:
+                    - "authpolicy.foo.com"
```
kminder, line 28: Is this explicit host name required or could the default have been used?

examples/authorization-policy/bar/istio-securitytest-app.yaml
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
kminder, line 1: Should all of these new files just be copyright 2021?

tests/e2e/pkg/common.go
```diff
@@ -74,6 +75,25 @@ func PodsRunning(namespace string, namePrefixes []string) bool {
...
+		Log(Info, fmt.Sprintf("Pods %v were TERMINATED in %v", missing, namespace))
+		time.Sleep(15 * time.Second)
+		pods := ListPods(namespace)
+		missing = notRunning(pods.Items, namePrefixes...)
+		i++
+		if i > 10 {
+			break
+		}
+	}
+	Log(Info, fmt.Sprintf("ALL Pods %v were TERMINATED in %v", missing, namespace))
```
kminder, line 93: For debugging shouldn't this log the pods that were not terminated at some point?

Jenkinsfile
```diff
@@ -464,6 +464,14 @@ pipeline {
                                 runGinkgo('examples/hello-helidon')
                             }
                         }
+                        stage('examples authorization policy') {
+                            when {
+                                expression {params.RUN_EXAMPLE_TESTS == true}
```
kminder, line 469: I realize there isn't a better option at the moment but this test doesn't really belong with the examples.  Should this be a merge test (i.e. always run) or a nightly test or some other grouping of test?

examples/authorization-policy/bar/istio-securitytest-app.yaml
ddrechse: idk, I copied the springboot example app

examples/authorization-policy/foo/istio-securitytest-app.yaml
ddrechse: I want to differentiate between the frontends so the test expects to use that value unless there is some way to programmatically access the default value

examples/authorization-policy/bar/istio-securitytest-app.yaml
ddrechse: idk, you tell me. Not in charge of copyright comment.  Again, just copied from the current publicly available springboot example app

tests/e2e/pkg/common.go
ddrechse: Yeah, I could do that one

Jenkinsfile
ddrechse: It was approved by others in the team.  If you want them somewhere else, lemme know

examples/authorization-policy/foo/istio-securitytest-app.yaml
kminder: ok

Jenkinsfile
kminder: I really think these need to be located somewhere else.
In particular the data files should go in something like tests/testdata/istio/authz
and the test code in something like tests/e2e/istio/authz
For now the "example" switch can be used in the Jenkinsfile since it pretty much doesn't mean anything anyway as it is always defaulted to true.

kminder:
I really think these files need to be located somewhere else.
In particular the data files should go in something like tests/testdata/istio/authz
and the test code in something like tests/e2e/istio/authz
For now the "example" switch can be used in the Jenkinsfile since it pretty much doesn't mean anything anyway as it is always defaulted to true.

kminder (changes requested):
Relocate files under the tests directory.

PR #724, opened 2021-02-23
VZ-2189: Fix ES logging when app has multiple workload types

fredtibbitts:
# Description

This is a short/medium-term fix for ES logging when an application is compromised of multiple workload types. Previously we created a single shared fluentd configmap for all pods regardless of workload type. This change results in a separate configmap per workload type.

I tested by running multiple examples and looking in Kibana to make sure I saw logs for all pods. For Bob's Books, I confirmed that both WebLogic and Coherence logs were sent. I also added another test to the Bob's Books test suite to look for Coherence log records in addition to WebLogic log records.

# Checklist 

### DIFF
--- a/application-operator/controllers/loggingscope/fluentd.go
+++ b/application-operator/controllers/loggingscope/fluentd.go
@@ -48,6 +48,7 @@ type Fluentd struct {
 	ParseRules             string
 	StorageVolumeName      string
 	StorageVolumeMountPath string
+	WorkloadType           string
 }
 
 // FluentdPod contains pod information for pods which require FLUENTD integration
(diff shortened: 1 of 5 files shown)


application-operator/controllers/loggingscope/fluentd.go
```diff
@@ -161,7 +162,7 @@ func (f *Fluentd) ensureFluentdVolumeMountExists(fluentdPod *FluentdPod) {
 // to do. If it doesn't exist, create it.
 func (f *Fluentd) ensureFluentdConfigMapExists(namespace string) error {
 	// check if configmap exists
-	configMapExists, err := resourceExists(f.Context, f, configMapAPIVersion, configMapKind, configMapName, namespace)
+	configMapExists, err := resourceExists(f.Context, f, configMapAPIVersion, configMapKind, configMapName+"-"+f.WorkloadType, namespace)
```
kminder, line 165: I'm surprised go fmt doesn't have an opinion about space vs no-space around string operators.  You have two different styles in this change.

kminder: @fredtibbitts confirmed that this spacing is what go fmt requires.  Resolved.

kminder (approved):
Approved.

application-operator/controllers/loggingscope/fluentd.go
fredtibbitts: That is `gofmt`ed. It does use inconsistent rules for string concatenation (function call vs struct field value), but apparently that's on purpose. 🤷

PR #727, opened 2021-02-23
Delete VerrazzanoProject resources on managed cluster when they no longer exist on the admin cluster

mgianatagh:
# Description

Handle the use case where a VerrazzanoProject object is deleted on the admin cluster - the copies of the object on the managed clusters will also get deleted.

Fixes VZ-2210

# Checklist 

### DIFF
--- a/application-operator/mcagent/mcagent_project.go
+++ b/application-operator/mcagent/mcagent_project.go
@@ -18,8 +18,9 @@ import (
 // and create namespaces specified in the Project resources in the local cluster
 func (s *Syncer) syncVerrazzanoProjects() error {
 	// Get all the Project objects from the admin cluster
-	allProjects := &clustersv1alpha1.VerrazzanoProjectList{}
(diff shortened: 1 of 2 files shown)


desagar (approved):

application-operator/mcagent/mcagent_project.go
```diff
@@ -49,6 +50,26 @@ func (s *Syncer) syncVerrazzanoProjects() error {
 		}
 	}
 
+	// Delete orphaned VerrazzanoProject resources.
```
pfmackin, line 53: If we are deleting this then how can we clean up the namespaces?  I know that we have deferred that functionality but we will lose track of the namespaces if we delete the projects.

mgianatagh: I expect this code will change when decisions are made regarding what to do about namespaces when projects are deleted.

PR #736, opened 2021-02-25
Refactoring cleanup and add more info to registration secret

pfmackin:
# Description

1. Build the API Server external URL using the kubeadmin configmap contents
2. Minor refactoring and unit tests for K8s code
3. Minor cluster API field name change
4. Add managed cluster name to the registration secret
5. Add minor integration test check for the secret contents
6. Minor name function changes from reconcileXyz to syncXyz as discussed with team

Contributes to VZ-2124

# Checklist 


platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -112,9 +123,6 @@ func TestCreateVMC(t *testing.T) {
 	mock.EXPECT().
```
desagar, line 123: comment above this line says "Expect a call to get the Secret with the service account token, return one with the token set" but this code has been changed to return nil.

platform-operator/internal/config/config_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // TestConfigDefaults tests the config default values
 // GIVEN a new OperatorConfig object
-//  WHEN I call New
+//  WHEN I call Build
```
desagar, line 14: nothing has changed in this code except for the comment - and the method is calling Get, not New or Build

platform-operator/internal/k8s/kubeadminconfig.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// ApiEndpoint contains the kubeadmin config information needed to access the API server
+type ApiEndpoint struct {
+	AdvertiseAddress string `json:"advertiseAddress"`
+	BindPort         string `json:"bindPort"`
+}
+
+const (
+	kubeSystem       = "kube-system"
```
desagar, line 27: these same constants are declared in multiple files - I think they should be publicly declared in one location (such as here in kubeadminconfig.go) and referenced in other places.

platform-operator/controllers/clusters/vmc_controller_test.go
pfmackin: fixed

platform-operator/internal/config/config_test.go
pfmackin: fixed

mgianatagh (dismissed):

mgianatagh (approved):

desagar (approved):

PR #737, opened 2021-02-25
Configuration for KIND multi-cluster environment for AT

bhatpmk:
# Description

This PR provides the initial environment to run acceptance tests in KIND multi-cluster setup. A new Jenkins job 'verrazzano-multi-cluster-acceptance-tests' is added to accept the following parameters -
    1. Topology - how many clusters? minimum 1, maximum 3 (for now)
    2. Version of Kubernetes - 1.17 and 1.18
    3. Distribution of Kubernetes - OKE or KIND 
    4. Type of DNS to use - magic xip.io, or OCI DNS

The job creates 1..n KIND clusters of the requested version based on the parameter. A basic test is added to validate the following -
 1. The clusters are accessible with appropriate kubeconfig
 2. The namespaces "default", "kube-public", "kube-system" and "kube-node-lease" are present in each of the cluster
 3. The pods "coredns" and "kube-proxy" are up and running in each of the cluster

The Jenkins job takes care of deleting the clusters at the end of the run, archiving the cluster dump all of the clusters on failure, along with an option to dump on success.

The change to configure OKE multi-cluster setup will be done separately.

# Checklist 


tests/e2e/config/scripts/create_kind_cluster.sh
```diff
@@ -25,17 +26,33 @@ create_kind_cluster() {
     exit 1
   fi
 
-  cd ${PLATFORM_OPERATOR_DIR}/build/scripts
-  ./cleanup.sh ${CLUSTER_NAME}
+  if [ ${IS_FIRST_CLUSTER} == true ]; then
```
gilbode, line 29: What does IS_FIRST_CLUSTER mean and what does cleanup.sh do?  This is confusing to me.  Why not just call cleanup.sh whenever you want it and remove it from the create cluster script?  Minimally the flag should be called something else (like CLEANUP?).

bhatpmk: platform-operator/build/scripts/cleanup.sh stops the running containers and deletes them, after deleting the cluster (if any). This script is also used by the Makefile target in platform-operator, ensures that there are no running KIND containers. In a multi-cluster setup, stopping and deleting the containers is avoided while creating the second and third (actually any number) clusters.

ci/multicluster/Jenkinsfile
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
kminder, line 1: Only 2021 for new files.

tests/e2e/multicluster/verify-install/defaultresource_suite_test.go
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
kminder, line 1: 2021 for new files

tests/e2e/multicluster/verify-install/defaultresource_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
kminder, line 1: 2021 only

ci/multicluster/Jenkinsfile
bhatpmk: I have fixed that now.

tests/e2e/multicluster/verify-install/defaultresource_test.go
bhatpmk: Done

tests/e2e/multicluster/verify-install/defaultresource_suite_test.go
bhatpmk: Done

tests/e2e/config/scripts/create_kind_cluster.sh
bhatpmk: @gilbode I have introduced two new parameters now
1. CLEANUP_KIND_CONTAINERS : When set to true, the script tests/e2e/config/scripts/create_kind_cluster.sh stops and deletes the KIND containers, after attempting to delete the cluster with the given name. The intention is to start the installation in a fresh environment.  When the value is set to false, it just deletes the cluster. In a multi-cluster environment, when trying to create the second cluster onwards, we don't want to delete the running containers created by first (or previous) cluster. The default value for this parameter is true
2. CONNECT_JENKINS_RUNNER_TO_NETWORK : When set to true, the jenkins-runner container will be connected to KIND network. We don't need that from second cluster onwards. Again, the default value for this parameter is false.

ci/multicluster/Jenkinsfile
```diff
@@ -0,0 +1,295 @@
...
+            junit testResults: '**/*test-result.xml', allowEmptyResults: true
+            script {
+                int clusterCount = params.TOTAL_CLUSTERS.toInteger()
+                for(int count=1; count<=clusterCount; count++) {
+                    sh """
+                        kind delete cluster --name ${CLUSTER_NAME_PREFIX}-$count
+                    """
+                }
+            }
+            deleteDir()
```
agv426, line 264: This needs to be in a "cleanup" phase or it will delete things before we capture the build-console-output log

Basically the ordering of post is fixed from what I've read, and "cleanup" will always happen after all of the others:
https://www.jenkins.io/doc/book/pipeline/syntax/#post

The regular build Jenkinsfile has that in cleanup (because of the directory getting blown away too soon)

agv426: ps, this can be fixed in another change after this too. We are capturing that in preparation for the analysis tooling, so nothing is consuming it yet during the build pipeline there and it will only fail there if the build already had failed.

bhatpmk: Done.

agv426 (approved):

markxnelson (approved):

PR #756, opened 2021-03-01
VZ-2117.  Support managed cluster install

michael-cico:
# Description

Support conditional install of VMI components to support managed cluster installs
- Add VMI components to install CRD
- Add hooks to consume those and convert to helm install args for the installer scripts
- Add required settings to the helm values file for the VZ install
- Add a "managed-cluster" profile 
- Align the other VMI component helm variables with the ones for Elasticsearch
- Update the VZ instance info in the CR status field to exclude components that aren't installed
- Some other cleanup and unit tests

Fixes VZ-2117.

# Checklist 


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -222,6 +224,44 @@ type ComponentSpec struct {
 	// Keycloak contains the Keycloak component configuration
 	// +optional
 	Keycloak KeycloakComponent `json:"keycloak,omitempty"`
```
mgianatagh, line 226: I notice the new components were added as pointers to structs, where this existing one is not a pointer.  We should make them all consistent.

michael-cico: Yeah, wasn't sure why.  I'll try updating that.  I also wanted to add the vmi-system-api URL also to the instance info.

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
```diff
@@ -281,7 +281,7 @@ func TestValidVersionWithCertManagerChange(t *testing.T) {
 		Version: "v0.7.0",
 		Profile: "dev",
 		Components: ComponentSpec{
-			CertManager: CertManagerComponent{
+			CertManager: &CertManagerComponent{
```
pfmackin, line 284: Is there some kind of best practice for using a pointer.  Can you check what Kubernetes does for its API?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -16,6 +16,8 @@ const (
 	Dev ProfileType = "dev"
 	// Prod identifies the production install profile
 	Prod ProfileType = "prod"
+	// ManagedCluster identifies the production managed-cluster install profile
+	ManagedCluster ProfileType = "managed-cluster"
```
pfmackin, line 20: This should be managedCluster or ManagedCluster.  If you look at the output of a pod the values are  Cap,lower,...Cap,lower,...  
for example:  reason: PodCompleted.

platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -167,38 +165,49 @@ type MySQL struct {
...
-func GetInstallConfig(vz *installv1alpha1.Verrazzano, log *zap.SugaredLogger) (*InstallConfiguration, error) {
-	if vz.Spec.Components.DNS.External != (installv1alpha1.External{}) {
-		return newExternalDNSInstallConfig(vz, log)
-	}
-
-	if vz.Spec.Components.DNS.OCI != (installv1alpha1.OCI{}) {
-		return newOCIDNSInstallConfig(vz, log)
+func GetInstallConfig(vz *installv1alpha1.Verrazzano) (*InstallConfiguration, error) {
+	dns := vz.Spec.Components.DNS
+	if dns != nil {
```
pfmackin, line 181: this is why it is not ideal to have pointers to structs everywhere, you have to check if the struct exists before you can use it.

platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -417,17 +426,21 @@ func getKeycloak(keycloak installv1alpha1.KeycloakComponent, templates []install
 }
 
 // getIngress returns the json representation for the ingress
```
pfmackin, line 428: Is this comment accurate?  Seems like it is returning struct not json

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
michael-cico: I poked around a bit, the one thing I found so far was in a comment in the Kubebuilder book.  You use pointers to distinguish between explicitly values and not-set.  

https://book.kubebuilder.io/cronjob-tutorial/api-design.html

I do need to do that for some things, for example in the VolumeSource/templates we have; and that follows the pattern for VolumeSource in general.

I also need it for this to distinguish between when the VMI resources have been explicitly enabled/disabled and are just not set by the user and we should default to the chart values.

I think in the other cases, right now it's largely cosmetic; i.e., we have empty structs in the CRs in the clusters for structs that aren't pointers but aren't explicitly set by the user.  If at some point we need to be able to similarly distinguish between explicitly set values and this it may help, but typically I think changing these to pointers down the road should be transparent to the users anyway.

There's other places we might want to change this, but perhaps we should have a discussion with @mgianatagh to see if I should revert this part of the change or follow it through.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: I can see examples of both for values:

>     dnsPolicy: ClusterFirst
>     enableServiceLinks: true
>     nodeName: 10.0.10.2
>     priority: 0
>     restartPolicy: Never
>     schedulerName: default-scheduler

i.e., `dnsPolicy` and `restartPolicy` versus `schedulerName`.   I think we just need to decide on what we prefer.  

Is there some kind of guidline you're aware of for that kind of stuff?  If we go with `ManagedCluster` we'll also have to change `dev` and `prod` to `Dev` and `Prod` to be consistent, and account for the case changes in the code etc.

platform-operator/controllers/verrazzano/installjob/install_config.go
michael-cico: Doesn't seem like it, I'll update it.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: Per the review meeting, we'll leave this as is and have a follow-up later as an API review.

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
michael-cico: Talked this through folks at the review meeting, and we're good with the pointers as defined here.

pfmackin (approved):

platform-operator/controllers/verrazzano/installjob/install_config.go
michael-cico: Addressed at the review meeting, see previous comment.

platform-operator/helm_config/charts/verrazzano/values.managed-cluster.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+name: verrazzano
+
+# Managed cluster profile
+# - ES, Kibana, and Grafana disabled
+
```
desagar, line 7: Does Prometheus default to enabled?

platform-operator/scripts/install/config.sh
```diff
@@ -31,7 +31,6 @@ function get_config_value() {
   fi
   if [ "$config_val" == "null" ]; then
     config_val=""
-    return 0
```
desagar, line 34: why should this case not return 0?

desagar: nvm - looks like it doesn't change the behavior.

pfmackin (approved):

platform-operator/helm_config/charts/verrazzano/values.managed-cluster.yaml
michael-cico: Yes.  Prometheus is always enabled.

desagar (approved):

PR #759, opened 2021-03-02
Change VerrazzanoProject to contain a list of namespace resources

mgianatagh:
# Description

Change the VerrazzanoProject custom resource to contain a list of corev1.Namespace resources (instead of a list of strings of namespace names)

Also fixed a problem in "make manifests" target because it was failing for me.

Fixes VZ-2289

# Checklist 


application-operator/apis/clusters/v1alpha1/verrazzanoproject_types.go
```diff
@@ -4,13 +4,25 @@
...
+
+// ProjectTemplate contains the resources for a project
+type ProjectTemplate struct {
+	Namespaces []NamespaceTemplate `json:"namespaces"`
+}
+
+// VerrazzanoProjectSpec defines the desired state of VerrazzanoProject
 type VerrazzanoProjectSpec struct {
-	Namespaces []string `json:"namespaces"`
+	Template ProjectTemplate `json:"template"`
```
pfmackin, line 25: Why do we need ProjectTemplate? Usually templates are for some other resource that we will be creating.  I would expect to see NamespaceTemplate here.

application-operator/apis/clusters/v1alpha1/verrazzanoproject_types.go
```diff
@@ -4,13 +4,25 @@
...
+// NamespaceTemplate has the metadata and spec of the underlying namespace
+type NamespaceTemplate struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Metadata metav1.ObjectMeta    `json:"metadata"`
+	Spec     corev1.NamespaceSpec `json:"spec,omitempty"`
+}
+
+// ProjectTemplate contains the resources for a project
+type ProjectTemplate struct {
+	Namespaces []NamespaceTemplate `json:"namespaces"`
```
pfmackin, line 20: Shouldn't this be namespaceTemplates?  That is what the user will see in the CR and i think we agreed to use the term template

application-operator/apis/clusters/v1alpha1/verrazzanoproject_types.go
mgianatagh: It is the "Template" section for the VerrazzanoProject resource

application-operator/apis/clusters/v1alpha1/verrazzanoproject_types.go
mgianatagh: I don't think we want to rename this field.  Here is an example of the yaml that would be used:

`spec:
  template:
    namespaces:
      - metadata:
          name: multiclustertest
          labels:
            label1: "test1"
          annotations:
            annot1: "test1"
`

desagar (dismissed):

desagar (approved):

PR #765, opened 2021-03-03
VZ-2108: Make all logging scope fields optional

fredtibbitts:
# Description

All logging scope fields are now optional, except for the workload refs field, because the OAM operator cannot handle a logging scope missing workload refs. The Fluentd image now comes from an environment variable set on the Verrazzano application operator, which is configured by the installer. The Elasticsearch URL and the secret name defaults are still hard-coded (that may change in the future).

Notes:

- I moved the function to fetch the logging scope given labels out of the workload functions file and into the logging scope file. The function has grown in scope, beyond just a simple workload utility function. As a result, the linter complained that the name should drop "LoggingScope" from the front to prevent stuttering when prefixed with the package.
- The defaults apply to multi-cluster logging scopes as well, so now instead of setting MC logging properties on the default logging scope, they can apply to user-defined logging scopes.

# Checklist 

### DIFF
--- a/application-operator/controllers/loggingscope/loggingscope_controller.go
+++ b/application-operator/controllers/loggingscope/loggingscope_controller.go
@@ -22,6 +22,7 @@ import (
 	oamv1alpha1 "github.com/verrazzano/verrazzano/application-operator/apis/oam/v1alpha1"
 
 	vzapi "github.com/verrazzano/verrazzano/application-operator/apis/oam/v1alpha1"
+	"github.com/verrazzano/verrazzano/application-operator/controllers/clusters"
 	vznav "github.com/verrazzano/verrazzano/application-operator/controllers/navigation"
 )
 
@@ -72,6 +73,8 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if scope == nil || err != nil {
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
+
+	applyDefaults(r.Client, scope)
 
 	var errors []string
 	var resources []vzapi.QualifiedResourceRelation
@@ -206,5 +209,56 @@ func toResource(workload *unstructured.Unstructured) vzapi.QualifiedResourceRela
 		Name:       workload.GetName(),
 		Namespace:  workload.GetNamespace(),
 		Kind:       workload.GetKind(),
+	}
+}
+
+// FromWorkloadLabels returns the LoggingScope object associated with the workload or nil if
+// there is no associated logging scope. The workload lookup is done using the OAM labels from the workload metadata.
+func FromWorkloadLabels(ctx context.Context, cli client.Reader, namespace string, labels map[string]string) (*vzapi.LoggingScope, error) {
+	component, err := vznav.ComponentFromWorkloadLabels(ctx, cli, namespace, labels)
+	if err != nil {
+		return nil, err
+	}
+
+	// fetch the first logging scope - do we need to handle multiple logging scopes?
+	for _, s := range component.Scopes {
+		if s.ScopeReference.Kind == vzapi.LoggingScopeKind {
+			scope := vzapi.LoggingScope{}
+			name := types.NamespacedName{
+				Namespace: namespace,
+				Name:      s.ScopeReference.Name,
+			}
+			err = cli.Get(ctx, name, &scope)
+			if err != nil {
+				return nil, err
+			}
+
+			applyDefaults(cli, &scope)
+			return &scope, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// applyDefaults fills in any empty fields in the logging scope - also handle the case
+// where we are running in a managed cluster
+func applyDefaults(cli client.Reader, scope *vzapi.LoggingScope) {
+	if scope.Spec.ElasticSearchURL == "" && scope.Spec.SecretName == "" {
+		// if we're running in a managed cluster, use the multicluster ES URL and secret, and if we're
+		// not the fields will be empty and we will set these fields to defaults below
+		elasticSearchDetails := clusters.FetchManagedClusterElasticSearchDetails(context.TODO(), cli)
+		scope.Spec.ElasticSearchURL = elasticSearchDetails.URL
+		scope.Spec.SecretName = elasticSearchDetails.SecretName
+	}
+
+	if scope.Spec.FluentdImage == "" {
+		scope.Spec.FluentdImage = DefaultFluentdImage
+	}
+	if scope.Spec.ElasticSearchURL == "" {
+		scope.Spec.ElasticSearchURL = DefaultElasticSearchURL
+	}
+	if scope.Spec.SecretName == "" {
+		scope.Spec.SecretName = DefaultSecretName
 	}
 }
--- a/application-operator/controllers/loggingscope/loggingscope_controller.go
+++ b/application-operator/controllers/loggingscope/loggingscope_controller.go
@@ -212,9 +212,9 @@ func toResource(workload *unstructured.Unstructured) vzapi.QualifiedResourceRela
 	}
 }
 
-// FromWorkloadLabels returns the LoggingScope object associated with the workload or nil if
+// FetchLoggingScopeFromWorkloadLabels returns the LoggingScope object associated with the workload or nil if
 // there is no associated logging scope. The workload lookup is done using the OAM labels from the workload metadata.
-func FromWorkloadLabels(ctx context.Context, cli client.Reader, log logr.Logger, namespace string, labels map[string]string) (*vzapi.LoggingScope, error) {
+func FetchLoggingScopeFromWorkloadLabels(ctx context.Context, cli client.Reader, log logr.Logger, namespace string, labels map[string]string) (*vzapi.LoggingScope, error) {
 	component, err := vznav.ComponentFromWorkloadLabels(ctx, cli, namespace, labels)
 	if err != nil {
 		return nil, err
--- a/application-operator/controllers/loggingscope/loggingscope_controller.go
+++ b/application-operator/controllers/loggingscope/loggingscope_controller.go
@@ -74,7 +74,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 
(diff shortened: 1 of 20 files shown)


application-operator/controllers/webhooks/logging_scope_defaulter.go
```diff
@@ -121,12 +110,9 @@ func ensureDefaultLoggingScope(c client.Client, appConfig *oamv1.ApplicationConf
 		var scope *vzapi.LoggingScope
 		scope, err = fetchLoggingScope(context.TODO(), c, namespacedName)
 		if scope == nil && err == nil {
-			// We might be running in a managed cluster - fetch the Elasticsearch Details to use in
-			// that case
-			elasticSearchDetails := clusters.FetchManagedClusterElasticSearchDetails(context.TODO(), c, log)
```
desagar, line 126: why was this removed? This is the code that determines whether to send logs to the managed cluster elasticsearch.

desagar: oh wait, never mind. Let me look at the applyDefaults a bit more closely in the controller

application-operator/controllers/loggingscope/loggingscope_controller.go
```diff
@@ -208,3 +211,54 @@ func toResource(workload *unstructured.Unstructured) vzapi.QualifiedResourceRela
...
+	return nil, nil
+}
+
+// applyDefaults fills in any empty fields in the logging scope - also handle the case
+// where we are running in a managed cluster
+func applyDefaults(cli client.Reader, scope *vzapi.LoggingScope) {
+	if scope.Spec.ElasticSearchURL == "" && scope.Spec.SecretName == "" {
+		// if we're running in a managed cluster, use the multicluster ES URL and secret, and if we're
+		// not the fields will be empty and we will set these fields to defaults below
+		elasticSearchDetails := clusters.FetchManagedClusterElasticSearchDetails(context.TODO(), cli)
```
desagar, line 250: elasticsearchDetails should only be used on an "all-or-nothing" basis. That function (FetchManagedClusterElasticsearchDetails) should return an empty ES details if any one of the fields is not populated, but it would be good to make sure of that here as well (if elasticsearchDetails.URL and secret are BOTH populated, use it, else ignore it)

application-operator/controllers/webhooks/logging_scope_defaulter.go
```diff
@@ -64,27 +61,19 @@ func (d *LoggingScopeDefaulter) Cleanup(appConfig *oamv1.ApplicationConfiguratio
...
-func CreateDefaultLoggingScope(name types.NamespacedName, esDetails clusters.ElasticsearchDetails) *vzapi.LoggingScope {
-	// If Elasticsearch are provided, use them. Otherwise use the defaults.
-	esURL := defaultElasticSearchURL
-	esSecret := defaultSecretName
-	if esDetails.URL != "" && esDetails.SecretName != "" {
-		esURL = esDetails.URL
-		esSecret = esDetails.SecretName
-	}
-	return &vzapi.LoggingScope{
+func createDefaultLoggingScope(name types.NamespacedName) *vzapi.LoggingScope {
```
desagar, line 64: If no logging scope is specified, and it's a managed cluster, I'm not sure if this will still attach the right logging scope. Has this case been tested?

application-operator/controllers/webhooks/logging_scope_defaulter_test.go
```diff
@@ -229,69 +219,6 @@ func TestLoggingScopeDefaulter_Cleanup(t *testing.T) {
 	mocker.Finish()
 }
 
-// TestCreateDefaultLoggingScope tests the behavior of CreateDefaultLoggingScope
```
desagar, line 232: Why were these 2 tests removed? I added one of them to test the managed cluster case and the other to provide a more "pure" unit test for createDefaultLoggingScope.

application-operator/controllers/webhooks/logging_scope_defaulter.go
desagar: looks like that's one of the tests removed from the logging scope defaulter test - let's discuss this to make sure those cases are still being tested.

application-operator/controllers/webhooks/logging_scope_defaulter.go
```diff
@@ -64,27 +61,19 @@ func (d *LoggingScopeDefaulter) Cleanup(appConfig *oamv1.ApplicationConfiguratio
 }
 
 // CreateDefaultLoggingScope creates the default logging scope for the given namespace
```
kminder, line 63: Fix function name in comment to match case.

examples/bobs-books/bobs-books-logging-scope.yaml
```diff
@@ -7,7 +7,6 @@ metadata:
   name: logging-scope
   namespace: bobs-books
 spec:
-  fluentdImage: ghcr.io/verrazzano/fluentd-kubernetes-daemonset:v1.10.4-20201016214205-7f37ac6
```
kminder, line 10: I'm not even sure this file should be included in this example.

application-operator/controllers/loggingscope/loggingscope_controller.go
```diff
@@ -208,3 +211,54 @@ func toResource(workload *unstructured.Unstructured) vzapi.QualifiedResourceRela
 		Kind:       workload.GetKind(),
 	}
 }
+
+// FromWorkloadLabels returns the LoggingScope object associated with the workload or nil if
+// there is no associated logging scope. The workload lookup is done using the OAM labels from the workload metadata.
+func FromWorkloadLabels(ctx context.Context, cli client.Reader, namespace string, labels map[string]string) (*vzapi.LoggingScope, error) {
```
kminder, line 217: Should this be NewFromWorkLoadLabels and if so should it be NewLoggingScopeFromWorkloadLabels because the "New" prefix prevents the repetition with the package name?

application-operator/controllers/webhooks/logging_scope_defaulter.go
fredtibbitts: If no logging scope is specified, then the default logging scope with an empty spec will be attached. When the logging scope is actually used, we check to see if this is a managed cluster and populate the logging scope fields with the MC URL and secret.

application-operator/controllers/webhooks/logging_scope_defaulter_test.go
fredtibbitts: I removed these and added corresponding tests in the logging scope controller test file. I did check the code coverage to make sure all of the MC logging scope code was still tested, but it's certainly possible that I missed cases. We can discuss in the zoom review.

kminder:
I think application-operator/deploy/verrazzano.yaml_template should also be updated to keep it in sync until such time as it is deleted.

kminder:
Doesn't application-operator/config/crd/bases/clusters.verrazzano.io_multiclusterloggingscopes.yaml need to be replicated to the platform operator copy of the helm chart also?

application-operator/controllers/loggingscope/loggingscope_controller.go
fredtibbitts: It's not creating a logging scope, it's fetching one, or is "New" idiomatic go in this situation? I'll rename it if it's the right thing to do.

examples/bobs-books/bobs-books-logging-scope.yaml
fredtibbitts: Yeah the examples use a mix of default and user-supplied logging scopes, but the user-supplied scopes just specify the defaults. It seemed like a good idea to have an example that specifies a logging scope with at least one field that gets defaulted. We can remove this logging scope but other examples do the same thing.

fredtibbitts:
> Doesn't application-operator/config/crd/bases/clusters.verrazzano.io_multiclusterloggingscopes.yaml need to be replicated to the platform operator copy of the helm chart also?

@kminder yes, it is replicated, I see it in the files changed list.

application-operator/controllers/webhooks/logging_scope_defaulter.go
fredtibbitts: Fixed.

fredtibbitts:
> I think application-operator/deploy/verrazzano.yaml_template should also be updated to keep it in sync until such time as it is deleted.

@kminder done. That YAML template is out-of-date, what is it actually used for?

application-operator/controllers/loggingscope/loggingscope_controller.go
fredtibbitts: Added a check and it now logs the condition if the secret name is set but the Elasticsearch URL is empty.

application-operator/controllers/loggingscope/loggingscope_controller.go
fredtibbitts: Changed the name to be more descriptive.

application-operator/controllers/loggingscope/loggingscope_controller.go
```diff
@@ -243,13 +243,18 @@ func FromWorkloadLabels(ctx context.Context, cli client.Reader, namespace string
...
 		// not the fields will be empty and we will set these fields to defaults below
 		elasticSearchDetails := clusters.FetchManagedClusterElasticSearchDetails(context.TODO(), cli)
-		scope.Spec.ElasticSearchURL = elasticSearchDetails.URL
-		scope.Spec.SecretName = elasticSearchDetails.SecretName
+		if elasticSearchDetails.URL != "" && elasticSearchDetails.SecretName != "" {
+			scope.Spec.ElasticSearchURL = elasticSearchDetails.URL
+			scope.Spec.SecretName = elasticSearchDetails.SecretName
+		} else if elasticSearchDetails.SecretName != "" && elasticSearchDetails.URL == "" {
+			// Elasticsearch secret is misconfigured, log it and set defaults below
+			log.Info("Unable to set defaults in logging scope for managed cluster because URL is empty")
```
desagar, line 256: I'd personally prefer a log level of warning here

desagar (approved):

application-operator/controllers/loggingscope/loggingscope_controller.go
fredtibbitts: This logger only has `Error` and `Info`. I think we have a backlog ticket to change the logger, not sure if that one offers more granularity.

PR #778, opened 2021-03-04
Build overhaul step one

markxnelson:
# Description

This PR converts the project into a single Go module, in the root directory, with a single THIRD_PARTY_LICENSES.txt file.  One test that uses cert-manager apis, and some associated helper functions are disabled temporarily as I will need some time to work through the dependency distress.  I will reinstate them in a subsequent PR for this same task.

Contributes to VZ-1973

# Checklist 


pfmackin:
can you try  make manifests and make generate to be sure they didn't break.

tests/e2e/pkg/environment.go
```diff
@@ -137,20 +136,20 @@ func findIstioIngressGatewaySvc(requireLoadBalancer bool) v1.Service {
...
-	}
-	// dump out namespace data to file
-	logData := ""
-	for i := range certs.Items {
-		logData = logData + certs.Items[i].Name + "\n"
-	}
-	CreateLogFile(fmt.Sprintf("%v-certificates", namespace), logData)
-	return certs, err
-}
+//// ListCertificates lists certificates in namespace
```
pfmackin, line 139: did you intend to leave this commented out?

desagar:
is it safe to assume that the go.mod entries are all ones that came from the existing dependencies in the subprojects?

platform-operator/Makefile
```diff
@@ -169,6 +169,8 @@ docker-clean:
 docker-build: go-build-linux
 	@echo Building verrazzano-platform-operator image ${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}
 	@echo using verrazzano-application-operator image ${VERRAZZANO_APPLICATION_OPERATOR_IMAGE}
+	# the TPL file needs to be copied into this dir so it is in the docker build context
+	cp ../THIRD_PARTY_LICENSES.txt .
```
desagar, line 173: we should probably add a .gitignore entry for platform-operator/THIRD_PARTY_LICENSES.txt and for the one in application-operator as well. Some of us routinely do docker-build locally and could end up accidentally checking in these copied TPL files

Jenkinsfile
```diff
@@ -206,16 +206,10 @@ pipeline {
                     echo "copyright scan"
                     cd ${GO_REPO_PATH}/verrazzano
                     time make copyright-check
```
desagar, line 208: how does this work without a top level makefile?

platform-operator/Makefile
gilbode: Can’t we just symlink them?  Seems better than copying files

Jenkinsfile
markxnelson: there is a top level makefile

platform-operator/Makefile
markxnelson: ok

tests/e2e/pkg/environment.go
markxnelson: no, i intend to fix it in a subsequent pr, i said that in the pr description :)

markxnelson (commented):
process feedback

markxnelson:
@desagar git reports one of them as a move, as opposed to a delete. but they are both in fact gone.

platform-operator/Makefile
markxnelson: ok turns our you can't - because docker build does not copy symlinks into the build context afaict - looks like it needs the actual file there

gilbode: Symlinks work but the file it points to has to be in the build context.  Change the build context to top level like it used to?

markxnelson: i'm trying to do this in small manageable chunks, not all in the one pr, to avoid getting so far removed from master that i have to suffer a huge merge

pfmackin (approved):

PR #785, opened 2021-03-06
VZ-2227: Add support for rolebindings in multi-cluster project namespaces

fredtibbitts:
# Description

This PR adds support for rolebindings in multi-cluster project namespaces. The VerrazzanoProject CR supports two optional lists of security subjects, one for admins and one for monitors. If the Project specifies subjects, we create up to four rolebindings per namespace. Admins are bound to the `verrazzano-project-admin` role as well as the cluster `admin` role. Monitor subjects are bound to the `verrazzano-project-monitor` role as well as the cluster `view` role.

I tested this manually by creating admin and managed clusters and applying a VerrazzanoProject YAML containing both admin and monitor subjects. I validated that the rolebindings were created in the project namespaces. For example:

```
$ k get rolebindings -n fred4
NAME                         ROLE                                     AGE
admin                        ClusterRole/admin                        54s
verrazzano-project-admin     ClusterRole/verrazzano-project-admin     55s
verrazzano-project-monitor   ClusterRole/verrazzano-project-monitor   54s
view                         ClusterRole/view                         54s
```

# Checklist 

### DIFF
--- a/application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
+++ b/application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
@@ -10,11 +10,20 @@ import (
 	clustersv1alpha1 "github.com/verrazzano/verrazzano/application-operator/apis/clusters/v1alpha1"
 	"github.com/verrazzano/verrazzano/application-operator/constants"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	projectAdminRole      = "verrazzano-project-admin"
+	projectAdminK8sRole   = "admin"
+	projectMonitorRole    = "verrazzano-project-monitor"
+	projectMonitorK8sRole = "view"
 )
 
 // Reconciler reconciles a VerrazzanoProject object
@@ -46,14 +55,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
-	err = r.createOrUpdateNamespaces(ctx, vp)
+	err = r.createOrUpdateNamespaces(ctx, &vp, logger)
 	return result, err
 }
 
-func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp clustersv1alpha1.VerrazzanoProject) error {
+func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp *clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
 	if vp.Namespace == constants.VerrazzanoMultiClusterNamespace {
 		for _, nsTemplate := range vp.Spec.Template.Namespaces {
-			r.Log.Info("create or update with underlying namespace", "namespace", nsTemplate.Metadata.Name)
+			logger.Info("create or update with underlying namespace", "namespace", nsTemplate.Metadata.Name)
 			var namespace corev1.Namespace
 			namespace.Name = nsTemplate.Metadata.Name
 
@@ -62,7 +71,11 @@ func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp clustersv1
 				return nil
 			})
 			if err != nil {
-				r.Log.Info("create or update namespace %s failed with result %q and error %v", nsTemplate.Metadata.Name, opResult, err)
+				logger.Error(err, "create or update namespace failed", "namespace", nsTemplate.Metadata.Name, "opResult", opResult)
+			}
+
+			if err = r.createOrUpdateRoleBindings(ctx, nsTemplate.Metadata.Name, vp, logger); err != nil {
+				return err
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


markxnelson (approved):

wmhopkins (approved):
Looks good to me, Fred.

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -46,14 +55,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
-	err = r.createOrUpdateNamespaces(ctx, vp)
+	err = r.createOrUpdateNamespaces(ctx, &vp, logger)
```
mgianatagh, line 58: You do not need to pass the logger, it is available in the "Reconcile" object (r.Log)

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -46,14 +55,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	return result, err
 }
 
-func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp clustersv1alpha1.VerrazzanoProject) error {
+func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp *clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
 	if vp.Namespace == constants.VerrazzanoMultiClusterNamespace {
 		for _, nsTemplate := range vp.Spec.Template.Namespaces {
-			r.Log.Info("create or update with underlying namespace", "namespace", nsTemplate.Metadata.Name)
+			logger.Info("create or update with underlying namespace", "namespace", nsTemplate.Metadata.Name)
```
mgianatagh, line 65: These should stay as "r.Log"

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -46,14 +55,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
-	err = r.createOrUpdateNamespaces(ctx, vp)
+	err = r.createOrUpdateNamespaces(ctx, &vp, logger)
 	return result, err
 }
 
-func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp clustersv1alpha1.VerrazzanoProject) error {
+func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp *clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
```
mgianatagh, line 62: I am not sure about the change to passing the VP object by reference.  The pattern all the other controllers are using for createOrUpdateXXXXX is to pass the object.  To be consistent I would put it back to how it was, or have a discussion to change all the other controllers to pass it by pointer.

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -87,3 +100,73 @@ func (r *Reconciler) mutateNamespace(nsTemplate clustersv1alpha1.NamespaceTempla
 		namespace.Labels[label] = value
 	}
 }
+
+// createOrUpdateRoleBindings creates project role bindings if there are security subjects specified in
+// the project spec
+func (r *Reconciler) createOrUpdateRoleBindings(ctx context.Context, namespace string, vp *clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
```
mgianatagh, line 106: Similar comments, the "logger" should not be passed, use r.Log.

To be consistent the VP object should not be passed by reference.

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -87,3 +100,73 @@ func (r *Reconciler) mutateNamespace(nsTemplate clustersv1alpha1.NamespaceTempla
...
+		if err := r.createOrUpdateRoleBinding(ctx, rb, vp, logger); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// createOrUpdateRoleBinding creates or updates a role binding
+func (r *Reconciler) createOrUpdateRoleBinding(ctx context.Context, roleBinding *rbacv1.RoleBinding, vp *clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
```
mgianatagh, line 138: same comments as above regarding parameter list

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
fredtibbitts: Yes you do. This logger has context automatically applied (in this case the namespaced project name). If you use `r.Log` then you lose the context.

application-operator/test/integ/testdata/multi-cluster/verrazzanoproject_sample.yaml
```diff
@@ -20,3 +20,11 @@ spec:
             label2: "test2"
           annotations:
             annot2: "test2"
+    security:
```
mgianatagh, line 23: Can you also update application-operator/test/integ/multi_cluster_test.go (which applies this file) to do some runtime asserts on the role bindings getting created as expected?

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
fredtibbitts: I had made the copy close to where the copy was needed (right before mutating), but I see that's not the case for the namespaces, so I'll change this back. I think it would be more efficient to just copy what's needed instead of passing by value but I'll switch it back.

application-operator/test/integ/testdata/multi-cluster/verrazzanoproject_sample.yaml
fredtibbitts: Yes I'll add that. When I had looked at that test, the `apply` and the validation aren't near each other and I missed the section where the namespaces were validated.

mgianatagh: You may want to add a new "Describe" block that is similar to the section titled "Testing VerrazzanoProject namespace generation".  Maybe create two some new yaml files to test create and update of role bindings.  Or expand the scope of the existing namespace generation test to handle create and update of both namespaces and role bindings.

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
fredtibbitts: I changed this back to pass by value.

application-operator/test/integ/testdata/multi-cluster/verrazzanoproject_sample.yaml
fredtibbitts: I added a new test case to check for the rolebindings.

fredtibbitts: Also added an integ test for the update rolebindings case.

mgianatagh (approved):

PR #800, opened 2021-03-10
Pin the request to delete the Rancher using API to an IP address

bhatpmk:
# Description

This change adds --resolve to pin the request to delete the Rancher using API to an IP address. This is done in other places like the function to get the rancher access token.

Listing of pods from cattle-system in uninstall/uninstall-steps/0-uninstall-applications.sh to relate any future uninstall issue with the status of pods cattle-cluster-agent and cattle-node-agent pods.

The plan is to have this change in and observe the behavior of Verrazzano uninstall by monitoring the uninstall and the install-uninstall in a loop.

Fixes VZ-2192

# Checklist 


ci/uninstall/Jenkinsfile
```diff
@@ -250,8 +250,7 @@ pipeline {
             }
             steps {
                 catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
-                    // Enable Bob's Books demo later
-                    runGinkgo('examples/todo-list')
+                    runGinkgo('examples/todo')
```
guoyongzhang, line 253: what is examples/todo?  isn't it examples/todo-list?

bhatpmk: @guoyongzhang It is changed to examples/todo recently

guoyongzhang: I see.  There is examples/todo-list and there is tests/e2e/examples/todo.  runGinkgo is using tests/e2e.

guoyongzhang (approved):

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -276,7 +276,7 @@ function wait_for_rancher_agent_to_exist {
 }
 
 function patch_rancher_agents() {
-    local rancher_in_cluster_host=$(get_rancher_in_cluster_host ${RANCHER_HOSTNAME})
+    local rancher_in_cluster_host=$(get_nginx_hostip)
```
guoyongzhang, line 280: Now that get_rancher_in_cluster_host is no longer used, please remove it from config.sh.

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -276,7 +276,7 @@ function wait_for_rancher_agent_to_exist {
...
 
     if [ ${RANCHER_HOSTNAME} != ${rancher_in_cluster_host} ]; then
```
guoyongzhang, line 281: This check is no longer necessary since RANCHER_HOSTNAME is never the same as rancher_in_cluster_host.

platform-operator/scripts/install/2-install-system-components.sh
guoyongzhang: patch_rancher_agents() should not be applied unless ".ingress.type" is "NodePort".  Please revert changes in this function.  hostAliases only resolves hostname.  Unlike '--resolve', it can not change the port.  When ".ingress.type" is "LoadBalancer", the port for public address and in-cluster address are completely different.  Simply resolving the public hostname to in-cluster hostip would mess up agents.

bhatpmk: Done

platform-operator/scripts/install/2-install-system-components.sh
bhatpmk: Reverted this.

guoyongzhang (approved):

PR #817, opened 2021-03-15
WIP: Implement more flexible approach to default clusterrolebindings

wmhopkins:
# Description

Changes to align how cluster role bindings for verrazzano-admin and verrazzano-monitor are created at install time with how the project-specific role bindings are created:
- Use the rbacv1.Subject type to specify binding subjects; align field names
- Allow multiple Subjects to be specified in the Verrazzano CR and include all Subjects in the resulting bindings.
- Don't create any default bindings -- user must explicitly request bindings by providing binding Subjects in the VZ CR.
- Create the bindings in the vpo controller instead of the helm template.
- Reconcile the bindings during upgrade as well as install, so that changes to the subjects in the VZ CR result in changes in the subjects in existing bindings; if no subjects, any existing binding is deleted.

Looking for feedback on the approach and implementation; code compiles but has not been tested (holding off on writing tests until I'm certain we want to go this way).

Implements VZ-2393.

# Checklist 


platform-operator/controllers/verrazzano/installjob/clusterrolebinding.go
```diff
@@ -49,3 +63,28 @@ func NewClusterRoleBinding(vz *installv1alpha1.Verrazzano, name string, saName s
 		},
 	}
 }
+
+// NewClusterRoleBindingWithSubjects returns a cluster role binding with the given parameters
+// vz - pointer to verrazzano resource
+// name - name of the clusterrolebinding resource
+// roleName - name of the clusterrole to bind to
+// subjects - slice of subjects to bind to the role
+func NewClusterRoleBindingWithSubjects(vz *installv1alpha1.Verrazzano, bindingName string, roleName string, subjects []rbacv1.Subject) *rbacv1.ClusterRoleBinding {
```
pfmackin, line 72: if this is a generic function (not specific to install) then it should go in platform-operator/internal/k8s/rbac.go or some other go file.   This is  something we can add to a cleanup list since there is probably other generic k8s resource related code here.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -209,6 +213,58 @@ func (r *Reconciler) createClusterRoleBinding(ctx context.Context, log *zap.Suga
...
+	}
+	if err := r.createOrUpdateClusterRoleBinding(ctx, log, vz, "verrazzano-monitor-k8s", "view", vz.Spec.Security.MonitorSubjects); err != nil {
+		return err
+	}
+	return nil
+}
+
+// createOrUpdateClusterRoleBinding creates or updates a cluster role binding
+func (r *Reconciler) createOrUpdateClusterRoleBinding(ctx context.Context, log *zap.SugaredLogger, vz *installv1alpha1.Verrazzano, bindingName string, roleName string, subjects []rbacv1.Subject) error {
+	// Check if the cluster role binding exists
```
pfmackin, line 235: This pattern is very different from what we use in the controller/clusters code where the mutate callback is used.   See any of the _sync.go files.    No saying this won't work but i also don't know if this pattern is used anywhere else in platform operator.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -114,6 +114,10 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	if err := r.createOrUpdateUserClusterRoleBindings(ctx, log, vz); err != nil {
```
pfmackin, line 117: Are you expecting this code to run after verrazzano is installed?  If so you need to do this before upgrade, see line 106

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -209,6 +213,58 @@ func (r *Reconciler) createClusterRoleBinding(ctx context.Context, log *zap.Suga
 	return nil
 }
 
+// createOrUpdateUserClusterRoleBindings creates or updates the user-facing role bindings specified in the vz resource
+func (r *Reconciler) createOrUpdateUserClusterRoleBindings(ctx context.Context, log *zap.SugaredLogger, vz *installv1alpha1.Verrazzano) error {
```
pfmackin, line 217: might be better to put this in sync_clusterrolebindings.go, we are trying to move secondary resource specific logic out of controller.go (see controllers/clusters)

platform-operator/controllers/verrazzano/installjob/clusterrolebinding.go
wmhopkins: It's not completely generic; it takes the VZ CR as input (like the NewClusterRoleBindingMethod that's already in that file).

platform-operator/controllers/verrazzano/controller.go
wmhopkins: It seemed somewhat consistent with the existing platform operator code. I'm happy to look at the mutate callback/_sync.go files, but role bindings have to characteristics that may make them unsuitable for that: Updating the roleRef is not allowed, and attempting to do so will cause an error; and we need to replace the Subjects array, not merge the new ones in with the old ones. If mutate is the best practice and it's not a ton of work to recode, I can look at whether it would work.

wmhopkins: (We also need to delete any existing binding if the subject list is empty.)

platform-operator/controllers/verrazzano/controller.go
wmhopkins: That particular call is intended to do the initial create at install time; there is a second call in the reconcileUpgrade method to cover the case of upgrade -- which seems like all the cases if we don't support update.

Don't really want to solve the update issue just for this, but once we have it this could get placed wherever is appropriate.

platform-operator/controllers/verrazzano/controller.go
wmhopkins: OK. I see there's already a sync_local_registration_secret.go, but you hadn't checked it in yet when I started coding this, so wasn't aware of that model.

wmhopkins: Wondering whether that should also apply to the verrazzano cluster role binding that was already being created?

PR #842, opened 2021-03-18
PR to add status to VerrazzanoProject

desagar:
# Description

A simple status section for per-cluster and overall status for VerrazzanoProject

Fixes VZ-2377

# Checklist 

### DIFF
--- a/application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
+++ b/application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
@@ -58,15 +58,18 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
+	opResult := controllerutil.OperationResultNone
 	if vp.Namespace == constants.VerrazzanoMultiClusterNamespace {
 		err = r.createOrUpdateNamespaces(ctx, vp, logger)
+		if err != nil {
+			// always use OperationResultCreated since we don't really know what happened to individual NS
+			opResult = controllerutil.OperationResultCreated
+		}
 	} else {
 		err = fmt.Errorf("resources of type VerrazzanoProject must be created in the %s namespace", constants.VerrazzanoMultiClusterNamespace)
 	}
 
-	// always use OperationResultCreated since we don't really know what happened to individual NS
-	// The operation result is relevant only if err is nil
-	return r.updateStatus(ctx, &vp, controllerutil.OperationResultCreated, err)
+	return r.updateStatus(ctx, &vp, opResult, err)
 }
 
 func (r *Reconciler) createOrUpdateNamespaces(ctx context.Context, vp clustersv1alpha1.VerrazzanoProject, logger logr.Logger) error {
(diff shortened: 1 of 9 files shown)


application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -55,8 +58,18 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
-	err = r.createOrUpdateNamespaces(ctx, vp, logger)
-	return result, err
+	opResult := controllerutil.OperationResultNone
+	if vp.Namespace == constants.VerrazzanoMultiClusterNamespace {
+		err = r.createOrUpdateNamespaces(ctx, vp, logger)
+		if err != nil {
```
desagar, line 64: == nil

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -55,8 +58,18 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
-	err = r.createOrUpdateNamespaces(ctx, vp, logger)
-	return result, err
+	opResult := controllerutil.OperationResultNone
+	if vp.Namespace == constants.VerrazzanoMultiClusterNamespace {
+		err = r.createOrUpdateNamespaces(ctx, vp, logger)
+		if err == nil {
```
jmaron99, line 64: Looks like we ignore any errors? Is that correct?

Perhaps this should be a check for err and return it and then the opResult below that block?

desagar: we don't ignore errors - we pass the err to the updateStatus function below, and that results in a status update to failure

jmaron99: Never mind - looks like the err does get set and leveraged below in the status update.  Though an "err == nil" seem un-go-like ;)

desagar: The opResult does not get used in the status update if err is non-nil

michael-cico (dismissed):

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
jmaron99: Looks like our comments crossed on the interwebs... :)

mgianatagh (approved):

application-operator/controllers/clusters/verrazzanoproject/verrazzanoproject_controller.go
```diff
@@ -55,8 +57,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, client.IgnoreNotFound(err)
 	}
 
+	// Use OperationResultCreated by default since we don't really know what happened to individual NS
+	opResult := controllerutil.OperationResultCreated
 	err = r.createOrUpdateNamespaces(ctx, vp, logger)
-	return result, err
+	if err != nil {
```
pfmackin, line 63: i think you were going to change this to err == nil

desagar: I made a minor logic change in response to @mgianatagh 's comment, which changed the sense of the comparison from err == nil to err != nil (which Jon suggested was more go-like anyway)

pfmackin (approved):

PR #847, opened 2021-03-18
Application configuration deletion failure prevents namespace deletion

tbeerbower:
# Description

Attempts to delete a namespace containing an Verrazzano OAM based application (e.g. sock shop) are failing. In this use case an attempt is being made to delete the namespace without previously deleting the application.

This occurs when the application uses the default logging scope and the namespace is terminated without deleting the app config first.  When the namespace is deleted, the resources in that namespace are cleaned up including the default logging scope.  During that process the application config defaulter webhook is triggered with an UPDATE notification.  The webhook determines that a default logging scope is required for the app config, so it attempts to create a new one.  Obviously this should be skipped in the case where the namespace is being terminated.

This PR adds code to the app config defaulter webhook to fetch the namespace and check to see if it's terminating.  If terminating then skip the code to create a new default logging scope.

#### Tests
- Update existing unit tests to expect call to get namespace and add case with a terminating namespace.
- Alter the sockshop example tests to uninstall the example by deleting the namespace without deleting the app config and asserting that the namespace is actually deleted. 

Fixes VZ-2332

# Checklist 

### DIFF
--- a/application-operator/controllers/webhooks/logging_scope_defaulter.go
+++ b/application-operator/controllers/webhooks/logging_scope_defaulter.go
@@ -6,6 +6,7 @@ package webhooks
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
 
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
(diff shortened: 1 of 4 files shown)


markxnelson (approved):
this is good, thanks.  i am not sure if this goes far enough to solve the underlying issue, but it will certainly make a big improvement over the status quo, so approving this, and we can open new issues if we find additional conditions that need to be dealt with

PR #849, opened 2021-03-18
VZ-2334.  Update verify-infra and verify-install for profiles, MC AT pipeline

michael-cico:
# Description

Update the verify-infra and verify-install ATs and include them in the MC AT pipelne.  Also cleans up a few things in the MC Pipeline Jenkinsfile.

- Check for VMI endpoints based on profile
- Check for VMI persistent volume configurations based on profile
- Also account in tests for cases where envName is not "default"
- enable the verify-infra and verify-install tests for the MC pipeline
- Jenkinsfile, rename KinD cluster version parameter, and spit out the cluster version before installing VZ
- Jenkinsfile, make profile parameters dropdown lists
- Jenkinsfile, go back to using the same env names in the VZ install CR regardless of cluster type

Fixes VZ-2334

# Checklist 

### DIFF
--- a/tests/e2e/pkg/kubernetes.go
+++ b/tests/e2e/pkg/kubernetes.go
@@ -6,6 +6,8 @@ package pkg
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	vpoClient "github.com/verrazzano/verrazzano/platform-operator/clients/verrazzano/clientset/versioned"
 	"os"
 	"path/filepath"
 	"strings"
@@ -216,7 +218,7 @@ func GetKubernetesClientset() *kubernetes.Clientset {
 	return clientset
 }
 
-// GetVMOClientset returns the Kubernetes clienset for the Verrazzano Monitoring Operator
+// GetVMOClientset returns the Kubernetes clientset for the Verrazzano Monitoring Operator
 func GetVMOClientset() *vmoclient.Clientset {
 	// use the current context in the kubeconfig
 	config := GetKubeConfig()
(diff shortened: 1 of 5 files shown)


tests/e2e/pkg/kubernetes.go
```diff
@@ -230,6 +232,49 @@ func GetVMOClientset() *vmoclient.Clientset {
 	return clientset
 }
 
+// GetPlatformOperatorClientset returns the Kubernetes clientset for the Verrazzano Platform Operator
+func GetPlatformOperatorClientset() *vpoClient.Clientset {
+	client, err := vpoClient.NewForConfig(GetKubeConfig())
+	if err != nil {
+		ginkgo.Fail("Could not get Verrazzano Monitoring Operator clientset")
```
mgianatagh, line 239: Platform Operator

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -66,11 +95,22 @@ var (
 
 var _ = ginkgo.BeforeSuite(func() {
 	var err error
+
+	vzCRD, err = verrazzanoInstallerCRD()
+	if err != nil {
+		ginkgo.Fail(fmt.Sprintf("Error retrieving system VMI CRD: %v", err))
```
mgianatagh, line 101: Verrazzano CRD

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -84,103 +124,167 @@ var _ = ginkgo.BeforeSuite(func() {
...
-		gomega.Eventually(kibanaPodsRunning, waitTimeout, pollingInterval).Should(gomega.BeTrue(), "kibana pods did not all show up")
-		assertOidcIngress("vmi-system-kibana")
+	ginkgo.It("api server should be accessible", func() {
+		assertURLByIngressName("vmi-system-api")
 	})
 
-	ginkgo.It("Prometheus endpoint should be accessible", func() {
-		assertOidcIngress("vmi-system-prometheus")
-	})
+	isManagedClusterProfile := pkg.IsManagedClusterProfile()
```
mgianatagh, line 132: Does the Verrazzano resource get filled out with settings that are defaulted from the profile?  if so could this tests be checking the boolean that indicates if the component is to be installed instead of checking the profile name?

michael-cico: No, it doesn't.  The CR only reflects what was posted.

That being said, we can (and should) have tests that check for overrides in the CR.  But that would mean writing different installer ATs, and those kinds of tests can be time consuming.  Probably for more than just the VMI overrides, etc.  But the profile validation is a good start.

tests/e2e/pkg/kubernetes.go
michael-cico: Fixed and pushed

tests/e2e/verify-infra/vmi/vmi_test.go
michael-cico: Fixed and pushed

tests/e2e/verify-infra/vmi/vmi_test.go
mgianatagh: ok

mgianatagh (approved):

PR #851, opened 2021-03-19
acceptance test that verifies the registration of a managed cluster.

guoyongzhang:
# Description

acceptance test that verifies the registration of a managed cluster.

Fixes VZ-2377

# Checklist 


tests/e2e/config/scripts/register_managed_cluster.sh
```diff
@@ -0,0 +1,88 @@
...
+  prometheusSecret: prometheus-${MANAGED_CLUSTER_NAME}
+EOF
+
+# wait for manifest to be created
+retries=0
+until kubectl --kubeconfig ${ADMIN_KUBECONFIG} -n verrazzano-mc get secret | grep verrazzano-cluster-${MANAGED_CLUSTER_NAME}-manifest; do
+  retries=$(($retries+1))
+  sleep 1
+  if [ "$retries" -ge 10 ] ; then
+    break
```
michael-cico, line 66: Should we exit the script here with a non-zero exit code (fail it basically)?

ci/multicluster/Jenkinsfile
```diff
@@ -346,6 +346,23 @@ pipeline {
...
                 }
+                stage ('Verify Register') {
+                    environment {
+                        ADMIN_KUBECONFIG="${KUBECONFIG_DIR}/1/kube_config"
+                        MANAGED_CLUSTER_NAME="managed1"
+                        MANAGED_KUBECONFIG="${KUBECONFIG_DIR}/2/kube_config"
+                    }
+                    steps {
+                        script {
+                            sh """
```
michael-cico, line 357: Probably don't have to do it in this pass, but likely we should follow the pattern for more than 1 managed cluster elsewhere in the pipeline, e.g., you can iterate over the cluster count (from 2-maxclusters), export the KUBECONFIG and cluster name, and execute the same steps afterwards.

tests/e2e/config/scripts/register_managed_cluster.sh
guoyongzhang: when retries limit is reached without the secret found, the next command would fail and exit.

ci/multicluster/Jenkinsfile
michael-cico: note that this will probably necessary when we have to have 2 managed clusters for the move scenario.

You could also just duplicate it for now, since we only allow a max of 3 (2 mgd) clusters at present in the pipeline), but we can cross that bridge when we get to it.

tests/e2e/multicluster/verify-register/verify_register_test.go
```diff
@@ -0,0 +1,72 @@
...
+		ginkgo.It("admin cluster has the expected filebeat logs", func() {
+			gomega.Eventually(func() bool {
+				return pkg.LogRecordFound("vmo-local-filebeat-"+time.Now().Format("2006.01.02"),
+					time.Now().Add(-24*time.Hour),
+					map[string]string{
+						"fields.verrazzano.cluster.name": managedClusterName})
+			}, waitTimeout, pollingInterval).Should(gomega.BeTrue(), "Expected to find a filebeat log record")
+		})
+
+		ginkgo.It("admin cluster has the expected journal logs", func() {
```
michael-cico, line 48: should we add something for looking for managed cluster metrics (system level)?

michael-cico: Or is that going to be done as a future task?

guoyongzhang: future for sure

michael-cico (approved):

ci/multicluster/Jenkinsfile
guoyongzhang: This suite needs admin and managed clusters.  At minimum, it requires 2 kubeconfig.  
I could iterate like 1+2 and 1+3, etc.  But it is hard to take advantage of runGinkgoRandomize, which takes one kubeconfig.

ci/multicluster/Jenkinsfile
```diff
@@ -346,6 +346,23 @@ pipeline {
...
+                                cd ${GO_REPO_PATH}/verrazzano
+                                ./tests/e2e/config/scripts/register_managed_cluster.sh
```
mgianatagh, line 359: Should the registration be done in a stage that is common to all the tests?  I think we want to have verrazzano installed and the clusters registered before running the tests

guoyongzhang: good point.  combined with Mike C's suggestion, I will register all managed clusters in a common spot.

ci/multicluster/Jenkinsfile
michael-cico: I think we do that in the other stages, including the verify-cluster stage.  We don't have to parallelize the execution of the tests, they could just iterate over each cluster.  There's a helper function in the pipeline that iterates over the clusters and runs a provided test path. 

You might want to add a stage conditional that checks that the cluster count >=2.

ci/multicluster/Jenkinsfile
michael-cico: yeah, there's examples of it in the other stages/helper functions.  We could zoom to go over it if you want, but it's pretty straightforward (aside from the usual Jenkins-isms).

So you could have a stage to do the register, and have another stage to iterate over the clusters and run the verify-registration tests.

PR #862, opened 2021-03-22
JIRA:VZ-2342 Enable MC resource movement between clusters

jmaron99:
# Description

Enables the deletion of MC and associated resources from one cluster and the creation of those resources on another cluster.  Supports resource moves between managed clusters as well as between managed clusters and admin cluster.

Fixes VZ-2342

# Checklist 


application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -307,3 +307,41 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
 	}
 	return statusPtr.State
 }
+
+// DeleteAssociatedResource will retrieve and delete the resource specified by the name
+func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource runtime.Object, finalizerName string, resourceToDelete runtime.Object, name types.NamespacedName) error {
+
```
pfmackin, line 313: we usually don't have a blank line here

application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -307,3 +307,41 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
...
+	mcObj, ok := mcResource.(controllerutil.Object)
```
pfmackin, line 314: not sure what this is doing can you add a comment

application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -307,3 +307,41 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
 	}
 	return statusPtr.State
 }
+
+// DeleteAssociatedResource will retrieve and delete the resource specified by the name
+func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource runtime.Object, finalizerName string, resourceToDelete runtime.Object, name types.NamespacedName) error {
```
pfmackin, line 312: why is the word Associated needed.  Is this more than just deleting a resource?

jmaron99: It's not deleting the MC resource being managed by the controller but rather than wrapped our underlying resource.

application-operator/controllers/clusters/multiclusterapplicationconfiguration/controller.go
```diff
@@ -60,6 +71,11 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		"placement", mcAppConfig.Spec.Placement.Clusters[0].Name)
 	opResult, err := r.createOrUpdateAppConfig(ctx, mcAppConfig)
 
+	// Add our finalizer if not already added
+	if err == nil {
```
pfmackin, line 75: why do we now need a finalizer vs just having the MC own the wrapped appconfig

jmaron99: Since, on the admin cluster, for a change in placement, we will delete the underlying resource.  Setting a controller reference would trigger an endless stream of "the object I owned can't be found" errors, and I imagine the re-appearance of the object if the object is once again placed on the cluster would cause issues as well.

application-operator/controllers/clusters/cluster_utils.go
jmaron99: asserting the type.  I'll add a comment.

application-operator/controllers/clusters/cluster_utils.go
jmaron99: I'll remove it.

application-operator/controllers/clusters/multiclusterapplicationconfiguration/controller.go
```diff
@@ -44,14 +45,24 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, clusters.IgnoreNotFoundWithLog("MultiClusterApplicationConfiguration", err, logger)
 	}
 
+	if !mcAppConfig.ObjectMeta.DeletionTimestamp.IsZero() {
```
mgianatagh, line 48: It may be helpful to add a comment about why we only delete the associated resource.  Put the same comment in each of the MC controller files.

jmaron99: That's ALOT of comments ;)

jmaron99: added.  Will merge once review is closer to completion.  Trying to avoid multiple builds and pushing my luck... ;)

PR #882, opened 2021-03-25
conditionally set ca_file in fluentd conf

guoyongzhang:
# Description

When ca-bundle is empty, do not set ca_file is the fluentd conf.  It applies to the fluentd sidecar in helidon, wls, and coh apps.

Fixes VZ-2447

# Checklist 

### DIFF
--- a/application-operator/controllers/loggingscope/helidon.go
+++ b/application-operator/controllers/loggingscope/helidon.go
@@ -72,8 +72,7 @@ const helidonFluentdContainerConfigurationTemplate = `<source>
 </filter>
 <match {{ .WorkloadName}}-{{ .ContainerName}}>
   @type elasticsearch
-  hosts "#{ENV['ELASTICSEARCH_URL']}"
-  ca_file /fluentd/secret/ca-bundle
+  hosts "#{ENV['ELASTICSEARCH_URL']}"{{ .CAFile}}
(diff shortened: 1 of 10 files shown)


markxnelson (approved):

application-operator/controllers/loggingscope/helidon.go
```diff
@@ -420,38 +419,51 @@ func fluentdConfigMapName(workloadName string) string {
...
 		if err != nil {
-			return err
+			return "", err
 		}
 
 		data := struct {
 			WorkloadName  string
 			ContainerName string
-		}{workloadName, containerName}
+			CAFile        string
```
michael-cico, line 456: So, this will always create a CAFile, just an empty string/file in the case where the bundle is not in the secret right?  Will fluentd be able to handle that?  (i.e., ignore an empty CA bundle file)

guoyongzhang: CAFile is a part of input data for the fluentd conf template.  It has 2 possible values:  "" or "\n  ca_file /fluentd/secret/ca-bundle", depending ca-bundle value in the secret.

markxnelson (approved):

application-operator/controllers/loggingscope/helidon.go
michael-cico: OK, thanks

PR #885, opened 2021-03-25
RBAC Acceptance Tests for new verrazzano cluster roles

ddrechse:
# Description

Create an Acceptance Test suite for RBAC based on new roles, verrazzano-project-admin and verrazzano-project-monitor.
Add helper functions CanI, CanIGroup where the api group is specified and CreateClusterRoleBinding
Verify that users with these rules can and cannot do what the roles grant

# Checklist 

- [ x] Checked that I included or updated copyright and license notices in all files that I altered
- [ x] Added or updated acceptance tests if appropriate

### DIFF
--- a/tests/e2e/pkg/kubernetes.go
+++ b/tests/e2e/pkg/kubernetes.go
@@ -6,7 +6,6 @@ package pkg
 import (
 	"context"
 	"fmt"
-	"k8s.io/api/authorization/v1beta1"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -16,6 +15,7 @@ import (
 	vmcClient "github.com/verrazzano/verrazzano/platform-operator/clients/clusters/clientset/versioned"
 	vpoClient "github.com/verrazzano/verrazzano/platform-operator/clients/verrazzano/clientset/versioned"
 
+	"k8s.io/api/authorization/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 
 	"github.com/onsi/ginkgo"
@@ -539,6 +539,19 @@ func GetConfigMap(configMapName string, namespace string) *corev1.ConfigMap {
 	return configMap
 }
 
+/*
+The following code adds http headers to the kubernetes client invocations.  This is done to emulate the functionality of
+kubectl auth can-i ...
+
+WrapTransport is configured to point to the function
+WrapTransport will be invoked for custom HTTP behavior after the underlying transport is initialized
+(either the transport created from TLSClientConfig, Transport, or http.DefaultTransport).
+The config may layer other RoundTrippers on top of the returned RoundTripper.
+
+WrapperFunc wraps an http.RoundTripper when a new transport is created for a client, allowing per connection behavior to be injected.
+
+RoundTripper is an interface representing the ability to execute a single HTTP transaction, obtaining the Response for a given Request.
+*/
 // headerAdder is an http.RoundTripper that adds additional headers to the request
 type headerAdder struct {
 	headers map[string][]string
--- a/tests/e2e/pkg/kubernetes.go
+++ b/tests/e2e/pkg/kubernetes.go
@@ -36,22 +36,6 @@ import (
 
 const dockerconfigjsonTemplate string = "{\"auths\":{\"%v\":{\"username\":\"%v\",\"password\":\"%v\",\"auth\":\"%v\"}}}"
 
-// headerAdder is an http.RoundTripper that adds additional headers to the request
-type headerAdder struct {
-	headers map[string][]string
-
-	rt http.RoundTripper
-}
-
-func (h *headerAdder) RoundTrip(req *http.Request) (*http.Response, error) {
-	for k, vv := range h.headers {
-		for _, v := range vv {
-			req.Header.Add(k, v)
-		}
-	}
-	return h.rt.RoundTrip(req)
-}
-
 // GetKubeConfig will get the kubeconfig from the environment variable KUBECONFIG, if set, or else from $HOME/.kube/config
 func GetKubeConfig() *restclient.Config {
 	kubeconfig := ""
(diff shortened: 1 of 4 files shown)


tests/e2e/pkg/kubernetes.go
```diff
@@ -499,3 +538,70 @@ func GetConfigMap(configMapName string, namespace string) *corev1.ConfigMap {
...
+		if wt != nil {
+			rt = wt(rt)
+		}
+		return &headerAdder{
+			headers: map[string][]string{"Impersonate-User": {userOCID}},
+			rt:      rt,
+		}
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
```
markxnelson, line 595: can you use the functions we already have in kubernetes.go, like GetClientSet?

tests/e2e/security/rbac/rbac_test.go
```diff
@@ -0,0 +1,250 @@
...
+		ginkgo.Fail(fmt.Sprintf("Failed to create namespace: %v", err))
+	}
+})
+
+var _ = ginkgo.AfterSuite(func() {
+	pkg.Log(pkg.Info, "Delete namespace")
+	if err := pkg.DeleteNamespace(rbacTestNamespace); err != nil {
+		pkg.Log(pkg.Error, fmt.Sprintf("Failed to delete the namespace: %v", err))
+	}
+	gomega.Eventually(func() bool {
```
markxnelson, line 44: you should not assert in the afterSuite

tests/e2e/security/rbac/rbac_test.go
```diff
@@ -0,0 +1,250 @@
...
+		return ns == nil && err != nil && errors.IsNotFound(err)
+	}, 3*time.Minute, 15*time.Second).Should(gomega.BeFalse())
+})
+
+var _ = ginkgo.Describe("Test RBAC Permission", func() {
+	ginkgo.Context("for user with role verrazzano-project-admin", func() {
+
+		ginkgo.It("Fail getting Pods in namespace rbactest", func() {
+			pkg.Log(pkg.Info, "Can User List Pods in NameSpace rbactest?  No")
+			if allowed, reason := pkg.CanI(v80ProjectAdmin, rbacTestNamespace, "list", "pods"); allowed != false {
```
markxnelson, line 55: its better to use an assert than to do this and manually call Fail()

tests/e2e/security/rbac/rbac_test.go
```diff
@@ -0,0 +1,250 @@
...
+		})
+
+		ginkgo.It("Create RoleBinding Admin for verrazzano-project-monitor", func() {
+			pkg.Log(pkg.Info, "Create RoleBinding Admin for verrazzano-project-monitor")
+			if err := pkg.CreateRoleBinding(v80ProjectMonitor, rbacTestNamespace, "monitor-binding", "admin"); err != nil {
+				ginkgo.Fail(fmt.Sprintf("FAIL: RoleBinding creation failed: reason = %s", err))
+			}
+		})
+
+		time.Sleep(sleepDuration)
```
markxnelson, line 168: better to use k8s wait forCondition than to do an arbitrary sleep

markxnelson (changes requested):
each comment applies to all cases, not just the one it is on

tests/e2e/pkg/kubernetes.go
ddrechse: Unless you want to embed the code that adds headers to the rouundtripper but that seemed something particular to the CanI functionality

What happens in the code above is that after getting the config object, the rountripper needs to be modified for the http headers prior to building out the clientset

Like i said, this seemed really specific to the CanI use-case

tests/e2e/security/rbac/rbac_test.go
ddrechse: The condition returns immediately.  What would I wait for?  This only happens on Kind when the full test suite runs under 500 ms

tests/e2e/security/rbac/rbac_test.go
ddrechse: That code was copied from the SpringBoot test.  It appears there and the Auth tests. So the paradigm is embedded in the test suites

tests/e2e/security/rbac/rbac_test.go
ddrechse: why?

tests/e2e/pkg/kubernetes.go
```diff
@@ -6,6 +6,8 @@ package pkg
 import (
 	"context"
 	"fmt"
+	"k8s.io/api/authorization/v1beta1"
```
kminder, line 9: Reorder

tests/e2e/pkg/kubernetes.go
```diff
@@ -499,3 +538,70 @@ func GetConfigMap(configMapName string, namespace string) *corev1.ConfigMap {
...
 }
+
+// headerAdder is an http.RoundTripper that adds additional headers to the request
+type headerAdder struct {
+	headers map[string][]string
+
+	rt http.RoundTripper
+}
+
+func (h *headerAdder) RoundTrip(req *http.Request) (*http.Response, error) {
```
kminder, line 549: How about a comment on this one as it isn't totally clear what it does.

tests/e2e/security/rbac/rbac_test.go
```diff
@@ -0,0 +1,250 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package rbac_test
+
+import (
+	"fmt"
+	"github.com/onsi/gomega"
```
kminder, line 8: reorder

tests/e2e/pkg/kubernetes.go
ddrechse: Done

tests/e2e/security/rbac/rbac_test.go
ddrechse: Done

tests/e2e/pkg/kubernetes.go
ddrechse: Added
The following code adds http headers to the kubernetes client invocations.  This is done to emulate the functionality of
kubectl auth can-i ...

WrapTransport is configured to point to the function
WrapTransport will be invoked for custom HTTP behavior after the underlying transport is initialized
(either the transport created from TLSClientConfig, Transport, or http.DefaultTransport).
The config may layer other RoundTrippers on top of the returned RoundTripper.

WrapperFunc wraps an http.RoundTripper when a new transport is created for a client, allowing per connection behavior to be injected.

RoundTripper is an interface representing the ability to execute a single HTTP transaction, obtaining the Response for a given Request.

tests/e2e/security/rbac/rbac_test.go
markxnelson: gomega.Expect(pkg.CanI(whatever)).To(gomega.BeTrue(), "message if its not")

ddrechse: Done

markxnelson (dismissed):

markxnelson (approved):

PR #901, opened 2021-03-30
Ddrechse/vz2361

ddrechse:
# Description

This removes the Rancher certificate and replaces it with a Verrazzano generated CA cert that will be used by the Verrazzano Cluster Issuer

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/installjob/install_config.go
+++ b/platform-operator/controllers/verrazzano/installjob/install_config.go
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	defaultCAClusterResourceName string = "cattle-system"
-	defaultCASecretName          string = "tls-rancher"
+	defaultCAClusterResourceName string = "verrazzano-install"
+	defaultCASecretName          string = "verrazzano-ca-certificate-secret"
 
 	// Verrazzano Helm chart value names
 	esStorageValueName         string = "elasticSearch.nodes.data.requests.storage"
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -95,6 +95,29 @@ spec:
   elif [ "$CERT_ISSUER_TYPE" == "ca" ]; then
     kubectl apply -f <(echo "
 apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: verrazzano-selfsigned-issuer
+  namespace: verrazzano-install
+spec:
+  selfSigned: {}
+")
+    kubectl apply -f <(echo "
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: verrazzano-ca-certificate
+  namespace: verrazzano-install
+spec:
+  secretName: verrazzano-ca-certificate-secret
+  commonName: verrazzano-root-ca
+  isCA: true
+  issuerRef:
+    name: verrazzano-selfsigned-issuer
+    kind: Issuer
+")
+    kubectl apply -f <(echo "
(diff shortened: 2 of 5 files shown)


platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -95,6 +95,29 @@ spec:
...
+  namespace: verrazzano-install
+spec:
+  selfSigned: {}
+")
+    kubectl apply -f <(echo "
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: verrazzano-ca-certificate
+  namespace: verrazzano-install
```
desagar, line 110: I assume you are going to use the values from the json for the name and namespace instead of these hard-coded ones.

ddrechse: Yeah, of you would ever reply to my query in our slack thread.  Hmmmmm

ddrechse: Done

guoyongzhang (approved):

platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	defaultCAClusterResourceName string = "cattle-system"
-	defaultCASecretName          string = "tls-rancher"
+	defaultCAClusterResourceName string = "verrazzano-install"
```
wmhopkins, line 20: Not sure what the defaultCAClusterResourceName is, but if it maps to a namespace, does it make more sense to put it in verrazzano-system vs. verrazzano-install? Probably doesn't matter unless we support multiple VZ installs in the same cluster, but if we did, each verrazzano should have its own CA.

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -12,6 +12,9 @@ trap 'rc=$?; rm -rf ${TMP_DIR} || true; _logging_exit_handler $rc' EXIT
 
 set -eu
 
+VERRAZZANO_DEFAULT_SECRET_NAMESPACE="verrazzano-install"
```
wmhopkins, line 15: See earlier comment about vz-system vs. vz-install namespace.

wmhopkins (approved):
Looks good.

I think it would be good to create a more descriptive issuer/subject name for the CA. Looks like we're just calling it "CN=verrazzano-root-ca", which is minimally adequate, so I don't think that needs to hold up. I also think we should also be using the Verrazzano cluster CA when we set up the CA bundle for Istio mesh, which happens before this. But neither of those strictly need to be addressed by this Jira, so I'll create another Jira to cover those tasks.

platform-operator/controllers/verrazzano/installjob/install_config.go
ddrechse: The cluster cert issuer needs to be constructed prior to verrazzano ns existing

platform-operator/scripts/install/2-install-system-components.sh
ddrechse: The cluster cert issuer needs to be constructed prior to verrazzano ns existing

platform-operator/controllers/verrazzano/installjob/install_config.go
wmhopkins: OK, but we could theoretically create the namespace sooner, one presumes?

ddrechse: Like multiple verrazzanos on the same node?  I guess. I don't know if thats supported currently

PR #906, opened 2021-03-31
VZ-1343, VZ-2348.  Install/uninstall issues

michael-cico:
Resolve some lingering issues around install/uninstall

- Have install wait for pods in verrazzano-system to reach Ready state before looking for secrets
- Uninstall should be more resilient around Rancher `local` cluster object removal, and general not prevent the uninstall from continuing
- On uninstall, patch Rancher local cluster object to remove any dangling finalizers as a semi-hack to allow it to be deleted
- Do an `ensure-default-admin` with Rancher on install to avoid issues with resetting the password subsequently
- Honor `letsEncrypt` `environment` setting and use the ACME staging env URL when specified
- Updates for `verrazzano-install-loop` pipeline

# Description

Fixes VZ-1343, VZ-2348.

Might also address VZ-1975, VZ-1856, and VZ-2078.

# Checklist 

### DIFF
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -72,6 +72,11 @@ function setup_cluster_issuer() {
         fail "The OCI Configuration Secret $OCI_DNS_CONFIG_SECRET does not exist"
     fi
 
+    acmeURL="https://acme-v02.api.letsencrypt.org/directory"
+    if [ "$(get_acme_environment)" != "production" ]; then
+      acmeURL="https://acme-staging-v02.api.letsencrypt.org/directory"
+    fi
+
     kubectl apply -f <(echo "
 apiVersion: cert-manager.io/v1alpha2
 kind: ClusterIssuer
@@ -80,7 +85,7 @@ metadata:
 spec:
   acme:
     email: $EMAIL_ADDRESS
-    server: "https://acme-v02.api.letsencrypt.org/directory"
+    server: "${acmeURL}"
     privateKeySecretRef:
       name: verrazzano-cert-acme-secret
     solvers:
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -74,6 +74,7 @@ function setup_cluster_issuer() {
 
     acmeURL="https://acme-v02.api.letsencrypt.org/directory"
     if [ "$(get_acme_environment)" != "production" ]; then
+      log "Non-production case, using the ACME staging environment"
       acmeURL="https://acme-staging-v02.api.letsencrypt.org/directory"
     fi
 
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -212,7 +212,21 @@ function install_rancher()
     log "Rollout Rancher"
     kubectl -n cattle-system rollout status -w deploy/rancher || return $?
 
-    log "Create Rancher secrets"
+    log "Ensure default Rancher admin user is present"
+    STDERROR_FILE="/tmp/rancher_ensureadminuser.err"
+    kubectl --kubeconfig $KUBECONFIG -n cattle-system exec $(kubectl --kubeconfig $KUBECONFIG -n cattle-system get pods -l app=rancher | grep '1/1' | head -1 | awk '{ print $1 }') -- ensure-default-admin 2>$STDERROR_FILE
+    RANCHER_ADMIN_USERNAME=$(kubectl get users -l authz.management.cattle.io/bootstrapping=admin-user -o jsonpath={'.items[].username'})
+    if [ -z "${RANCHER_ADMIN_USERNAME}" ]; then
+      echo "Could not detect default Rancher admin user"
+      local std_error_file=$(cat $STDERROR_FILE)
+      log "${std_error_file}"
+      rm "$STDERROR_FILE"
+      return 1
+    else
+      echo "Rancher admin user: ${RANCHER_ADMIN_USERNAME}"
+    fi
+
+    log "Reset Rancher admin password and create secrets"
     STDERROR_FILE="/tmp/rancher_resetpwd.err"
     local max_retries=5
     local retries=0
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -219,7 +219,7 @@ function install_rancher()
     kubectl -n cattle-system rollout status -w deploy/rancher || return $?
 
     log "Ensure default Rancher admin user is present"
-    STDERROR_FILE="/tmp/rancher_ensureadminuser.err"
+    STDERROR_FILE="${TMP_DIR}/rancher_ensureadminuser.err"
     kubectl --kubeconfig $KUBECONFIG -n cattle-system exec $(kubectl --kubeconfig $KUBECONFIG -n cattle-system get pods -l app=rancher | grep '1/1' | head -1 | awk '{ print $1 }') -- ensure-default-admin 2>$STDERROR_FILE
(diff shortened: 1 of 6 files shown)


platform-operator/scripts/install/config.sh
```diff
@@ -284,9 +284,26 @@ function get_verrazzano_ports_spec() {
...
+  fi
+}
+
 function get_acme_environment() {
   if [ -z "$(get_config_value ".certificates.acme.environment")" ]; then
-    echo "production"
+    # Ideally we would handle this by just setting helm chart values, but the Rancher override
+    # and VZ ClusterIssuer resource are done here in the scripts, so switch off the profile value
+    # if the environment is not explicitly specified
+    if [ "$(get_profile)" != "dev" ]; then
```
guoyongzhang, line 302: Did you mean get_install_profile here?  I though using .profile itself as a knob is not something we want to do.

guoyongzhang: To use staging in pipelines, we could append "environment: staging" in tests/e2e/config/scripts/install-verrazzano-ocidns.yaml at line 14.

platform-operator/scripts/install/3-install-verrazzano.sh
```diff
@@ -126,6 +122,9 @@ function install_verrazzano()
       ${PROFILE_VALUES_OVERRIDE} \
       ${EXTRA_V8O_ARGUMENTS} || return $?
 
+  log "Waiting for pods in ${VERRAZZANO_NS} to be reach Ready state"
+  kubectl  wait --for=condition=Ready pod -n verrazzano-system --all=true
```
guoyongzhang, line 126: Does this command wait for all the pods in verrazzano-system are running?  Some of them are really slow, are they all necessary?

guoyongzhang: The secret is generated by VO.  Maybe just wait for VO?

platform-operator/scripts/install/config.sh
michael-cico: It's not, but there's no knob in the charts for what we're trying to do, at least for the `ClusterIssuer` resource.  We're creating that in code in the scripts.  Unless we wanted to create some kind of miscellaneous chart we can invoke to create resources like this before the Verrazzano components install.

platform-operator/scripts/install/3-install-verrazzano.sh
michael-cico: Maybe not, mainly I did all of them to be safe.

platform-operator/scripts/install/config.sh
michael-cico: I think we're going to hit a point where we'd either have to modify 3rd party charts or switch off profile values in the scripts to a small degree, or create a new chart or something like I mentioned in my last comment.

platform-operator/scripts/install/3-install-verrazzano.sh
michael-cico: I'll see if I can scope it down to a specific pod or pods.

michael-cico: Fixed and pushed

platform-operator/scripts/install/config.sh
desagar: If the default should be production, then the onus is on whoever is specifying the config to specify the acme environment setting for staging - in this case the tests can be changed as GZ suggested.
I agree that we want to avoid making assumptions in the scripts about what any of the profiles contain - that reduces our flexibility to change the profile contents later, so I'm hesitant to start the precedent of putting "if profile == xyz" in the scripts.

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -212,7 +218,21 @@ function install_rancher()
     log "Rollout Rancher"
     kubectl -n cattle-system rollout status -w deploy/rancher || return $?
 
-    log "Create Rancher secrets"
+    log "Ensure default Rancher admin user is present"
+    STDERROR_FILE="/tmp/rancher_ensureadminuser.err"
```
bobdonat, line 222: There is variable $TMP_DIR that is already setup and used in this script.  I think you want to use that variable instead of /tmp

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -212,7 +218,21 @@ function install_rancher()
...
+      local std_error_file=$(cat $STDERROR_FILE)
+      log "${std_error_file}"
+      rm "$STDERROR_FILE"
+      return 1
+    else
+      echo "Rancher admin user: ${RANCHER_ADMIN_USERNAME}"
+    fi
+
+    log "Reset Rancher admin password and create secrets"
     STDERROR_FILE="/tmp/rancher_resetpwd.err"
```
bobdonat, line 236: same comment here even though this was already there

platform-operator/scripts/install/config.sh
michael-cico: OK, I'll remove this part.  It just puts the burden on the user to do so, if we want to make defaults for profiles we'll have to come up with a helm-based scheme (special chart(s) for resources like ClusterIssuer, etc) where we can follow the overrides strategy like we do now.

platform-operator/scripts/install/2-install-system-components.sh
michael-cico: yeah, I just copied this line for the one above.

platform-operator/scripts/install/config.sh
michael-cico: Removed dev profile default in latest push.

platform-operator/scripts/install/2-install-system-components.sh
michael-cico: Fixed in latest push.

platform-operator/scripts/install/2-install-system-components.sh
michael-cico: Fixed in latest push.

platform-operator/scripts/uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -53,10 +54,17 @@ function initializing_uninstall {
       fi
       ((retries+=1))
       if [ "$retries" -ge "$max_retries" ] ; then
-        return 0
+        break
       fi
     done
+
+    # Occasionally the 'local' Rancher cluster object is stuck with a dangling finalizer, remove it so
```
bhatpmk, line 61: Can we have this command to delete the finalizers in delete_rancher() of 2-uninstall-system-components.sh so that it will be run even if there is an issue in deleting the local cluster ?

michael-cico: Maybe, this will always run if we can access the rancher admin token though if I understand this correctly.  Does moving it there add much?

bhatpmk: It won't when we fail to delete the local cluster.

michael-cico: Yup, fixed on next push.  I'm verifying it against my own OKE clusters now to make sure it works, then I'll push it with the re-enabled `verify-install` tests if it checks out.

michael-cico: Pushed updates to re-enable the `verify-install` tests and the moved patch command; seems to be working with my local runs.

michael-cico: I'll monitor in the install-loop pipeline and see how it goes.

michael-cico: Pipeline looks rock solid, with the `verify-install` tests added back in since this change.  My own install-loops have been running solid also, so I think this did the trick.

bhatpmk (approved):

PR #910, opened 2021-04-01
JIRA:VZ-2460 create cluster role for application operator

jmaron99:
# Description

Deployed, underplayed, and updated multiple applications and resources to come up with a set of api groups for the cluster role.  The AT tests for VZ and multi cluster revealed some additional api groups and resources that were required for the operator to function properly.

General process:
- Used k8s audit logging to identify resource interactions
- deployed and undeployed all examples
- augmented the above with AT runs to identify other resources utilized by the operator
- Generally, all write verbs, all read verbs, or both were utilized once a need to read or write was identified

Fixes VZ-2460

# Checklist 


platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,171 @@ imagePullSecrets:
...
+  - apiGroups:
+      - clusters.verrazzano.io
+    resources:
+      - multiclusterapplicationconfigurations/status
+      - multiclustercomponents/status
+      - multiclusterconfigmaps/status
+      - multiclusterloggingscopes/status
+      - multiclustersecrets/status
+      - verrazzanoprojects/status
+    verbs:
```
wmhopkins, line 103: This seems like an area where full read/write could be needed for some scenarios, don't see the harm in granting unless the idea is to provide enforcement of a division of labor between the vao and the mco.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,171 @@ imagePullSecrets:
...
+      - get
+      - patch
+      - update
+  - apiGroups:
+      - coherence.oracle.com
+    resources:
+      - coherence
+    verbs:
```
wmhopkins, line 111: Here's an example, I think, of a case where we might need to update or patch a coherence if a customer updated a coherence deployment, but if our test suite doesn't test that scenario, the audit log won't show it. If we grant full read/write, we haven't given any power they don't already have from create/delete, but we avoid problems when the vao can't do what the customer wants because the role isn't complete.

That's theoretically also an issue for resources as well as verbs, but it seems more likely that the test suite would touch all resource types that it deals with during the test run.

wmhopkins (commented):
This mostly looks good to me. I assume you've diff'd the two different files to make sure the roles are identical? I only reviewed one of them.

In general, I do think we should grant all read if we grant any read, and all write if we grant any write. Adding the additional verbs doesn't grant significantly more privilege, and provides insurance against missing a verb that later causes a problem for a customer.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,171 @@ imagePullSecrets:
...
+      - customresourcedefinitions
+    verbs:
+      - '*'
+  - apiGroups:
+      - admissionregistration.k8s.io
+    resources:
+      - mutatingwebhookconfigurations
+      - validatingwebhookconfigurations
+    verbs:
+      - create
```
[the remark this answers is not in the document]
jmaron99: doen

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,171 @@ imagePullSecrets:
...
+    resources:
+      - ingresses
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - clusterroles
```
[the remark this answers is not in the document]
jmaron99: This API Group was exceedingly problematic and puzzling though most of the integration tests and AT tests so I eventually enabled all verbs.  I can try another cycle of testing to see if read is sufficient.  Let me know how strongly you feel about it.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,171 @@ imagePullSecrets:
...
+      - roles
+      - rolebindings
+    verbs:
+      - '*'
+  - apiGroups:
+      - cert-manager.io
+    resources:
+      - certificates
+    verbs:
+      - create
```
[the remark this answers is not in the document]
jmaron99: ok

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
jmaron99: OK. I didn't see much in terms of other interactions but I suppose there's no harm in provided the additional verbs.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
jmaron99: sounds good

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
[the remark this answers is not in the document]
wmhopkins: I do think we should avoid granting write for roles unless we really needed it. Roles we use should be installed, not generated on the fly. If it turns out we are creating some roles dynamically we may have to add write verbs back in, but there would have to be a really good reason for us to keep doing that.

application-operator/deploy/verrazzano.yaml_template
```diff
@@ -8,13 +8,160 @@ metadata:
...
+    resources:
+      - authorizationpolicies
+    verbs:
+      - create
+      - get
+  - apiGroups:
+      - weblogic.oracle
+    resources:
+      - domains
+    verbs:
```
fredtibbitts, line 151: We also update WebLogic domains, must not be an AT for it, but this should be "*".

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -14,13 +14,160 @@ imagePullSecrets:
...
+    resources:
+      - authorizationpolicies
+    verbs:
+      - create
+      - get
+  - apiGroups:
+      - weblogic.oracle
+    resources:
+      - domains
+    verbs:
```
fredtibbitts, line 157: Same comment, we update, should be "*".

wmhopkins (approved):
lgtm

PR #921, opened 2021-04-06
VZ-2385.  Allow retries of install/uninstall jobs

michael-cico:
# Description

Allow job retries for install/uninstall
    - Allow job retryOnFail, and use the default backoff limit
    - Delete any in-progress install job if the VZ resource is deleted

Idempotent install behavior
    - Short-circuit helm chart installs when status is already deployed
    - reset helm chart state on job/pod restart when not in deployed state before continuing
    - Make secret creation idempotent

Fixes VZ-2385.

# Checklist 


platform-operator/scripts/install/common.sh
```diff
@@ -188,6 +188,75 @@ function call_curl {
...
+function reset_chart(){
+  local ns=$2
+  local chartName=$1
+  local chartLocation=${3:-""}
+  # status values: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback
+  local deployment_status=$(get_deployment_status ${chartName} ${ns})
+  log "Deployment status for ${ns}/${chartName}: ${deployment_status}"
+  if [ "${deployment_status}" != "deployed" ]; then
+      log "Resetting chart state for ${ns}/${chartName} if necessary"
+      helm delete -n ${ns} ${chartName} 2>&1 > /dev/null
```
michael-cico, line 233: change to "uninstall" command

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
```diff
@@ -82,7 +82,7 @@ function cleanup_rancher_local_cluster() {
     # patch out any remaining finalizers and check one more time for delete success.
     log "Found 'local' cluster object still present, removing..."
     kubectl delete --wait=false clusters.management.cattle.io local || true
-    kubectl wait --for=delete -n kube-system clusters.management.cattle.io/local --timeout=2m || true
+    kubectl wait --for=delete -n kube-system clusters.management.cattle.io/local --timeout=1m || true
```
bobdonat, line 85: Curious as to why the timeout is being reduced.

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
```diff
@@ -177,6 +177,14 @@ function delete_rancher() {
       | xargsr -I resource kubectl annotate secret resource -n "${namespace}" field.cattle.io/projectId- \
       || err_return $? "Could not delete Annotations from Rancher" || return $? # return on pipefail
   done
+
+  # Remove the Rancher namespace finalizers; do it here since Rancher is not guaranteed to have been installed by VZ
```
bobdonat, line 181: Glad this is being fixed.  I just ran into the problem where the delete of the verrazzano-install ns hung.

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
michael-cico: because we do another 2m wait below after patching out the finalizers as a last-gasp.  I wonder if I should just delete this line, but the thinking generally I had was

- give Rancher every opportunity to clean up
- patch out the finalizers as a last ditch

Maybe I should reverse it, wait 2m after the delete, and then less time after the patch (or don't wait at all).  But not cleaning up Rancher properly seems to make bad things happen over time.

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
michael-cico: Yeah, I hit this early on also.  But this branch has been out so long, this change is ancient.  That's why I want to get it in sooner rather than later.

platform-operator/controllers/verrazzano/installjob/job.go
```diff
@@ -79,7 +79,7 @@ func NewJob(jobConfig *JobConfig) *batchv1.Job {
 							},
 						},
 					}},
-					RestartPolicy:      corev1.RestartPolicyNever,
+					RestartPolicy:      corev1.RestartPolicyOnFailure,
```
pfmackin, line 82: As discussed, as a follow up task, we need some indication that install restarted.  Maybe there is something in the job status indicating that information.  If so, it should be propagated to VZ status.

platform-operator/scripts/install/common.sh
```diff
@@ -188,6 +188,75 @@ function call_curl {
...
+
+  kubectl apply -f <(echo "
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: ${secret_name}
+  namespace: ${ns}
+data:
+  password: $(echo -n "${password_secret}"|base64)
```
pfmackin, line 208: does this show the secret in the install log file?

platform-operator/scripts/install/common.sh
```diff
@@ -188,6 +188,75 @@ function call_curl {
...
+
+# Get the current state of a helm chart
+# $1 chart name
+# $2 namespace
+# $3 chart location
+function get_deployment_status() {
+  echo -n $(helm status $1 -n $2 -o json 2>/dev/null | jq -r .info.status || true)
+}
+
+# Uninstalls a chart if it is stuck in "failed" or "unknown" state
```
pfmackin, line 220: Looks like Uninstalls if not in deployed state

platform-operator/scripts/install/common.sh
```diff
@@ -188,6 +188,75 @@ function call_curl {
...
+  local ns=$2
+  local chartName=$1
+  local chartLocation=${3:-""}
+  # status values: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback
+  local deployment_status=$(get_deployment_status ${chartName} ${ns})
+  log "Deployment status for ${ns}/${chartName}: ${deployment_status}"
+  if [ "${deployment_status}" != "deployed" ]; then
+      log "Resetting chart state for ${ns}/${chartName} if necessary"
+      helm template ${chartName} -n ${ns} ${chartLocation}  |  kubectl delete -f - 2>&1 > /dev/null || true
+      helm uninstall -n ${ns} ${chartName} 2>&1 > /dev/null
```
pfmackin, line 234: Does uninstall work if the chart is in these states states.  (like unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback)

  I assume this is a best try effort and it may not work.   Maybe need an uninstall timeout

michael-cico: Possibly.  We can set a timeout via the `activeSecondsDeadline` attribute on the Jobs.

So far, I haven't hit any issues with uninstall, hardly at all.  The exception being in the uninstall test pipelines where we need to add things to the cleanup, etc.  But so far in my experience the uninstall has been rock solid.

If the helm chart is in a truly unrecoverable state, I'm not sure what we can do.  I think if we implement job retries, so that we continue to reconcile uninstall, that will give users a chance to clear the chart state and the next uninstall try can go through.  But we're somewhat at the mercy of Helm's ability to recover an `uninstall` operation

platform-operator/scripts/install/common.sh
michael-cico: I don't think so, I did look for that.  Let me double-check.

platform-operator/scripts/install/common.sh
michael-cico: Yes, anything other than a `deployed` state is removed and we then retry.

platform-operator/controllers/verrazzano/installjob/job.go
michael-cico: OK.

michael-cico: I'll add that to the follow-up JIRA.

michael-cico: Captured this in VZ-2629.

platform-operator/scripts/install/common.sh
michael-cico: I think I didn't fully answer your question, sorry.  I'm not 100% sure if it will work in all the other states.  But yes, it's a best effort at this point.  I imagine we'll have to revise this as needed.

pfmackin (dismissed):

platform-operator/scripts/install/common.sh
michael-cico: Just looked at the log output from an install to confirm, looks OK.  The `keycloak-http` and `rancher-admin-secret` secrets below are configured using this helper

> [2021-05-05 14:46:34 UTC] secret/keycloak-realm-cacert configured
[2021-05-05 14:46:35 UTC] secret/keycloak-http configured
[2021-05-05 14:46:36 UTC] secret/verrazzano-container-registry created

> [2021-05-05 14:46:07 UTC] Reset Rancher admin password and create secrets
[2021-05-05 14:46:10 UTC] secret/rancher-admin-secret configured
[2021-05-05 14:46:10 UTC] Installing Rancher                                                           [  OK  ]

pfmackin (approved):

PR #923, opened 2021-04-06
Exclude Rancher from an install that uses the managed-cluster profile

bhatpmk:
# Description

This change excludes Rancher from the managed-cluster profile.

When the configuration rancher.enabled is defined in the custom resource, the platform operator uses that to decide whether to install Rancher or not. When the configuration rancher.enabled is not defined in custom resource, the default value is derived from platform-operator/scripts/install/install-overrides/*.json. The install script merges config.json with the json specific for the profile, with profile specific configuration taking the precedence.

The function get_config_value() defined in platform-operator/scripts/install/config.sh is updated to read platform-operator/scripts/install/install-overrides/*.json, when it fails to find the configuration in the config.json defined in config map.

Fixes VZ-2356

# Checklist 


platform-operator/config/samples/install-managed-cluster.yaml
```diff
@@ -19,3 +21,6 @@ metadata:
   name: my-verrazzano-managed-cluster
 spec:
   profile: managed-cluster
+  components:
+    rancher:
+      enabled: false
```
mgianatagh, line 26: The other components that are disabled by the managed-cluster profile are not declared explicitly.  I feel for consistency this declaration should be removed.

bhatpmk: Done.

platform-operator/scripts/install/config.sh
```diff
@@ -329,6 +337,44 @@ function get_nginx_nodeport() {
...
 }
 
+# Merge the default config json, install-overrides.json with the corresponding override for the profile
+function compute_effective_override() {
+  set -o pipefail
+  local profile=$(get_install_profile)
+  local default_config="${INSTALL_OVERRIDES_DIR}/config.json"
+  local profile_config_override="${INSTALL_OVERRIDES_DIR}/config.${profile}.json"
+  if [ ! -f "${profile_config_override}" ]; then
+    error "The file ${profile_config_override} does not exist"
```
desagar, line 347: It's possible we might not have overrides for a config. Does this have to be an error condition?

platform-operator/scripts/uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -14,11 +14,14 @@ UNINSTALL_DIR=$SCRIPT_DIR/..
 set -o pipefail
 
 function initializing_uninstall {
+  # Check whether rancher is installed for the given profile
+  rancher_exists=$(kubectl get namespace cattle-system) || return 0
+  local rancher_exists=$(kubectl get namespace verrazzano-install --ignore-not-found)
```
desagar, line 19: I'm not sure I follow this. Why are you checking the verrazzano-install NS to initialize rancher_exists? And why do we need both line 18 and line 19?

platform-operator/scripts/uninstall/uninstall-steps/0-uninstall-applications.sh
```diff
@@ -14,11 +14,14 @@ UNINSTALL_DIR=$SCRIPT_DIR/..
 set -o pipefail
 
 function initializing_uninstall {
```
desagar, line 16: this seems like a bad name for this function. All it's doing is Rancher stuff, so maybe we should rename it to uninstall_rancher, and also update the message printed at the bottom of this script where we call this function?

platform-operator/scripts/uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -86,6 +86,11 @@ function finalize() {
       || err_return $? "Could delete Helm Repos" || return $? # return on pipefail
   fi
 
+  local rancher_exists=$(kubectl get namespace cattle-system --ignore-not-found)
+  if [ -z "$rancher_exists" ] ; then
+    return 0
+  fi
+
   log "Removing Namespace Finalizers"
```
desagar, line 94: While we are here, another cleanup item....this log should probably say "Removing Rancher Namespace Finalizers"

tests/e2e/verify-install/kubernetes/kubernetes_test.go
```diff
@@ -58,13 +58,19 @@ var _ = ginkgo.Describe("Kubernetes Cluster",
 		ginkgo.It("has the expected namespaces",
 			func() {
 				namespaces := pkg.ListNamespaces()
-				gomega.Expect(nsListContains(namespaces.Items, "cattle-global-data")).To(gomega.Equal(true))
-				gomega.Expect(nsListContains(namespaces.Items, "cattle-global-nt")).To(gomega.Equal(true))
-				gomega.Expect(nsListContains(namespaces.Items, "cattle-system")).To(gomega.Equal(true))
+				if isManagedClusterProfile {
+					gomega.Expect(nsListContains(namespaces.Items, "cattle-global-data")).To(gomega.BeFalse())
+					gomega.Expect(nsListContains(namespaces.Items, "cattle-global-nt")).To(gomega.BeFalse())
+					gomega.Expect(nsListContains(namespaces.Items, "cattle-system")).To(gomega.BeFalse())
```
desagar, line 64: do we also need to check that the "local" NS doesn't exist in isManagedClusterProfile, if that's something rancher creates?

platform-operator/scripts/install/config.sh
```diff
@@ -329,6 +337,44 @@ function get_nginx_nodeport() {
...
+    exit 1
+  fi
+  if [ -f "$EFFECTIVE_CONFIG_VALUES" ]; then
+   rm "$EFFECTIVE_CONFIG_VALUES"
+  fi
+  jq -rs 'reduce .[] as $item ({}; . * $item)' $default_config $profile_config_override > $EFFECTIVE_CONFIG_VALUES
```
desagar, line 353: Neat! I had no idea you could do this with jq. I tried this jq command and tried to understand reduce but I still don't follow exactly how it is working, but it does work ;-)

platform-operator/scripts/install/config.sh
```diff
@@ -329,6 +337,44 @@ function get_nginx_nodeport() {
...
 
+# Merge the default config json, install-overrides.json with the corresponding override for the profile
+function compute_effective_override() {
+  set -o pipefail
+  local profile=$(get_install_profile)
+  local default_config="${INSTALL_OVERRIDES_DIR}/config.json"
+  local profile_config_override="${INSTALL_OVERRIDES_DIR}/config.${profile}.json"
+  if [ ! -f "${profile_config_override}" ]; then
+    error "The file ${profile_config_override} does not exist"
+    exit 1
```
desagar, line 348: if there is an error condition we should return 1 rather than exit 1 from a function. Returning 1 will cause the control to return to the caller and the stuff to the right of the "||" will be executed

platform-operator/scripts/uninstall/uninstall-steps/0-uninstall-applications.sh
bhatpmk: This is mistake from my side, a leftover change. I will correct this, thanks for noticing.

bhatpmk: The issue is fixed now.

platform-operator/scripts/install/config.sh
bhatpmk: Thank you, it is fixed now.

tests/e2e/verify-install/kubernetes/kubernetes_test.go
bhatpmk: Done

platform-operator/scripts/uninstall/uninstall-steps/1-uninstall-istio.sh
bhatpmk: I have updated the log message now

platform-operator/scripts/uninstall/uninstall-steps/0-uninstall-applications.sh
bhatpmk: It is actually deleting the local cluster, the name of the function and the message is updated.

platform-operator/scripts/install/config.sh
bhatpmk: This is my understanding from the jq man page of Reduce, for each result that .[] produces, $item is run to accumulate key-values, starting from {}

1. jq -rs 'reduce .[] as $item ({}; . * $item)' a.json --> fills empty json {} in the filter with the contents from a.json and outputs the raw string. 
2. jq -rs 'reduce .[] as $item ({"abc":"def"}; . * $item)' a.json --> fills a json with "abc":"def" by the contents of a.json, so the output will be something like -
{
  "abc": "def",
  <contents of a.json>
}

platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -353,6 +372,16 @@ func getInstallArgs(args []installv1alpha1.InstallArgs) []InstallArg {
 	return installArgs
 }
 
+func getRancher(rancher *installv1alpha1.RancherComponent) (Rancher, error) {
```
michael-cico, line 375: Looks like an error is never returned here?  If so, you can just have it return *Rancher and inline the calls, to reduce some of the repetitive code.

platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -172,6 +176,7 @@ type InstallConfiguration struct {
 	Ingress         Ingress                     `json:"ingress"`
 	Certificates    Certificate                 `json:"certificates"`
 	Keycloak        Keycloak                    `json:"keycloak"`
+	Rancher         Rancher                     `json:"rancher"`
```
michael-cico, line 179: Should this be a *Rancher, so that we can tell when it hasn't been set?  Or does that matter?

platform-operator/scripts/install/config.sh
michael-cico: Nice...there's our helm merge capability...

I did have an earlier question, my concern is that in the config.json struct in install_config.go, if we don't use *Rancher in the struct we won't be able to discern between Rancher being explicitly disabled, and the value not being set, because we always populate the field.  Does this take care of that?

platform-operator/controllers/verrazzano/installjob/install_config.go
bhatpmk: A value not set is taken care while reading the json configuration, in install scripts.

platform-operator/scripts/install/config.sh
bhatpmk: The updated get_config_value takes care of reading the value from the overrides.

    # The configuration is not defined in CONFIG_JSON, check if it is defined in the json files under install-overrides
    config_val=$(get_override_config_value "$jq_expr")
  fi

platform-operator/controllers/verrazzano/installjob/install_config.go
bhatpmk: I will fix this

platform-operator/scripts/install/config.sh
michael-cico: But, I think because we aren't using a pointer to the struct, there will always be a Rancher value in the generated config.json in the operator right?  So it will never be null?

bhatpmk: Right, it will be "rancher": {} and we read ".rancher.enabled" from the json in is_rancher_enabled()

platform-operator/controllers/verrazzano/installjob/install_config.go
michael-cico: Ah, never mind...I think because it's a string instead of a bool, it will be empty if not set.  I missed that on the first pass.

platform-operator/scripts/install/config.sh
michael-cico: I posted above, I missed that the Enabled flag is a string and not a bool, so it will be empty if not set...so that should work then.  Although I'm not sure if that will show up as "null" or "" in the default case in the generated config.

bhatpmk: It comes as   "rancher": {}

platform-operator/controllers/verrazzano/installjob/install_config.go
bhatpmk: Done

michael-cico (approved):

platform-operator/scripts/install/config.sh
desagar: let me add to the general confusion ;-) IIRC, the "null" is a string produced by jq when there isn't a value for the query, which is why get_config_value checks for that string. But I think you were asking about the final JSON, which Bhat responded to.

platform-operator/scripts/install/config.sh
```diff
@@ -329,6 +337,44 @@ function get_nginx_nodeport() {
   echo ${nodePort}
 }
 
+# Merge the default config json under install/install-overrides with the corresponding override for the profile
+function compute_effective_override() {
+  set -o pipefail
```
desagar, line 342: We don't need to set pipefail here since there are no pipes used. Same with get_override_config_value. Not urgent to fix with this PR, but if you happen to make another push, it would be good to remove it.

desagar (approved):

platform-operator/scripts/install/config.sh
bhatpmk: I will make this change in a follow up change, @desagar

PR #924, opened 2021-04-06
Tvlatas/multicache

agv426:
# Description

This will switch the CI over from using pre-created clusters to using pre-created caches instead. While we will lose some of the time gains we see with pre-created clusters (a minute or so for each cluster created in a run), it does not have the trade-off that the cluster config is known to our VM images (which requires more vigilance to ensure they are aligned).

This relies on a new CI image which has the pre-created clusters removed and the caches added into it. So this effectively reverts the pre-created cluster changes, and also consolidates the cluster creation setup logic as well on top of that to clean things up a bit where we had a duplicated scripting embedded in various Jenkinsfile's

IMPORTANT:

- We have to coordinate changing the switch over to the new CI and changing the labels in this PR from "largeexperimental". So we will switch in the new CI image, then change the PR to use "Standard2.8". This will be DISRUPTIVE and needs to be scheduled off-hours (may see some breaks during this time).
- Once this is all in, folks with outstanding branches will need to REBASE them to pickup these changes for their branches to be clean in the CI again (ie: once this is in there will not be pre-created clusters present).

# Checklist 

- [ X Checked that I included or updated copyright and license notices in all files that I altered


application-operator/Makefile
```diff
@@ -185,27 +179,28 @@ cleanup-cluster:
...
         echo 'Create cluster...' ;\
         echo 'KUBECONFIG=${KUBECONFIG}' ;\
         HTTP_PROXY="" HTTPS_PROXY="" http_proxy="" https_proxy="" time kind create cluster -v 1 \
             --name ${CLUSTER_NAME} \
             --wait 5m \
-            --config=test/kind-config.yaml ;\
+            --config=test/${KIND_CONFIG} ;\
         kubectl config set-context kind-${CLUSTER_NAME} ;\
     fi
+ifdef JENKINS_URL
```
agv426, line 192: Looking to see if we can get rid of the JENKINS_URL that came back along with the revert of pre-created. Want to get that out of here if I can with these changes (if not in a followup).

markxnelson (approved):
LGTM

markxnelson (approved):

PR #938, opened 2021-04-08
 managed cluster uses admin keycloak for auth; remove keycloak from managed-cluster

guoyongzhang:
# Description

1. VMO redeploy to update the OIDC sidecars, upon cluster registration changes.
2. managed cluster prometheus secret no longer includes password.
3. prometheus scraper uses admin verrazzano password
4. remove keycloak in managed-cluster profile
5. uptake VMO in which managed cluster prometheus re-routes auth requests to admin keycloak, proxy-body-size and reduced log

Fixes VZ-2357

# Checklist 

### DIFF
--- a/application-operator/mcagent/mcagent.go
+++ b/application-operator/mcagent/mcagent.go
@@ -53,7 +53,8 @@ func StartAgent(client client.Client, statusUpdateChannel chan clusters.StatusUp
 		if err != nil {
 			s.Log.Error(err, "error processing multi-cluster resources")
 		}
-		s.configureBeats()
+		s.updateDeployment("verrazzano-operator")
+		s.updateDeployment("verrazzano-monitoring-operator")
 		if !s.AgentReadyToSync() {
 			// there is no admin cluster we are connected to, so nowhere to send any status updates
 			// received - discard them
@@ -220,10 +221,10 @@ func getAdminClient(secret *corev1.Secret) (client.Client, error) {
 	return clientset, nil
 }
 
-// reconfigure beats by restarting Verrazzano Operator deployment if ManagedClusterName has been changed
-func (s *Syncer) configureBeats() {
-	// Get the verrazzano-operator deployment
-	deploymentName := types.NamespacedName{Name: "verrazzano-operator", Namespace: constants.VerrazzanoSystemNamespace}
+// reconfigure deployment if cluster registration has been changed
+func (s *Syncer) updateDeployment(name string) {
+	// Get the deployment
+	deploymentName := types.NamespacedName{Name: name, Namespace: constants.VerrazzanoSystemNamespace}
 	deployment := appsv1.Deployment{}
 	err := s.LocalClient.Get(context.TODO(), deploymentName, &deployment)
 	if err != nil {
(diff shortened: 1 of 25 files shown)


bhatpmk (approved):
The change to remove keycloak in managed-cluster profile looks good to me.

platform-operator/scripts/install/4-install-keycloak.sh
```diff
@@ -154,16 +154,27 @@ data:
...
+      "$SCRIPT_DIR"/k8s-dump-objects.sh -o "jobs" -n "${KEYCLOAK_NS}" -m "install_mysql"
+      "$SCRIPT_DIR"/k8s-dump-objects.sh -o "nodes" -n "default" -m "install_mysql"
+      log "For additional detailed information on the cluster at the time of this error, please check the diagnostics log file"
+      fail "Installation of MySQL failed"
+    fi
 
-action "Installing Keycloak" install_keycloak || exit 1
+  action "Installing Keycloak" install_keycloak || exit 1
+else
+  log "Skip Keycloak installation on managed cluster"
```
michael-cico, line 176: Just a nit, if there's a general knob for this in the CRD then it can be applied via any VZ CR... probably just log "Skipping Keycloak installation, disabled" or something like that.

application-operator/mcagent/mcagent.go
```diff
@@ -248,7 +249,7 @@ func (s *Syncer) configureBeats() {
 	}
 	secretVersionEnv := getEnvValue(deployment.Spec.Template.Spec.Containers[0].Env, registrationSecretVersion)
```
shihchang, line 250: is it safe to assume only one container?

wmhopkins (commented):
Looks good to me.

application-operator/mcagent/mcagent.go
guoyongzhang: I think it is safe to assume there is at least one container.  When there is more than one container, maybe it is good enough since updating template restarts the pod; maybe it is not enough.  We will deal with that when it comes to that.
But for now, VO, VMO, even your fluentd are single container pod.

platform-operator/scripts/install/4-install-keycloak.sh
michael-cico: Actually, probably not worth another build cycle, if you want to just merge this I can change the message since I'm poking around the installer scripts anyway.

guoyongzhang: done

tests/e2e/verify-infra/restapi/vmi_urls_test.go
```diff
@@ -14,52 +14,51 @@ import (
...
-			}
-
-			if isPrometheusEnabled {
-				gomega.Expect(verifySystemVMIComponent(sysVmiHTTPClient, vmiCredentials, "vmi-system-prometheus", "https://prometheus.vmi.system")).To(gomega.BeTrue(), "Unable to access Prometheus VMI url")
-			}
-
-			if isGrafanaEnabled {
-				gomega.Expect(verifySystemVMIComponent(sysVmiHTTPClient, vmiCredentials, "vmi-system-grafana", "https://grafana.vmi.system")).To(gomega.BeTrue(), "Unable to access Garafana VMI url")
+			isManagedClusterProfile := pkg.IsManagedClusterProfile()
+			if !isManagedClusterProfile {
```
desagar, line 20: lf the VMI will contain the enabled flags even on the managed cluster, then we can just get rid of the check for isManagedClusterProfile and just check the individual components based on whether they are enabled or disabled.

michael-cico: I don't think they will, I think with a profile the user won't have to explicitly set the flag for certain things.  The only indicator with a profile is the profile name most of the time.

Although, it is possible for someone to override the profile to enable something that's disabled by default, and vice-versa...so we might have to check for both or something.

michael-cico: If the user always has to set the flag even with the mgd cluster profile, then it's probably not right...

desagar: ok so we don't "fill out" the vmi with the full complement of flags - only what user sets.

desagar (approved):

PR #954, opened 2021-04-11
VZ-2482: Add Rancher network policies

fredtibbitts:
# Description

This PR adds network policies for Rancher. Since we are not use many of Rancher's features, these policies are fairly restrictive. As we take advantage of more of Rancher's features in the future, we will likely need to relax these policies.

I tested by installing admin and managed clusters and validating that I could still register managed clusters and they show up properly in the Rancher UI. I also verified that I could inspect managed cluster events, services, etc. in the UI.

NOTE: We create the `cattle-system` namespace and network policy even if the `rancher.enabled` flag is false. That way when the user applies the managed cluster registration YAML at some point in the future, the network policy will already exist. 

Implements VZ-2482

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
@@ -95,36 +95,6 @@ spec:
           protocol: UDP
         - port: 53
           protocol: TCP
-# Network policy for Rancher cluster agent
-# Ingress: deny all
-# Egress: Kubernetes API server, DNS for Rancher host lookups, and port 443 to talk to Rancher on the admin cluster
-apiVersion: networking.k8s.io/v1
-kind: NetworkPolicy
-metadata:
-  name: cattle-cluster-agent
-  namespace: cattle-system
-spec:
-  podSelector:
-    matchLabels:
-      app: cattle-cluster-agent
-  policyTypes:
-    - Ingress
-    - Egress
-  egress:
-    - ports:
-      - port: 443
-        protocol: TCP
-    - ports:
-        - port: {{ .Values.kubernetes.service.endpoint.port }}
-          protocol: TCP
-      to:
-        - ipBlock:
-            cidr: {{ .Values.kubernetes.service.endpoint.ip }}/32
-    - ports:
-      - port: 53
-        protocol: UDP
-      - port: 53
-        protocol: TCP
       to:
         - namespaceSelector:
             matchLabels:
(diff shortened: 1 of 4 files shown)


platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -392,9 +399,16 @@ RANCHER_HOSTNAME=rancher.${NAME}.${DNS_SUFFIX}
...
+
 if [ $(is_rancher_enabled) == "true" ]; then
   action "Installing Rancher" install_rancher || exit 1
   action "Setting Rancher Server URL" set_rancher_server_url || true
   action "Patching Rancher Agents" patch_rancher_agents || true
 fi
 
+# Label the kube-system namespace so that we can apply network policies
+log "Adding label needed by network policies to kube-system namespace"
+kubectl label namespace kube-system "verrazzano.io/namespace=kube-system" || true
```
fredtibbitts, line 414: I will remove these lines once Bob's 948 PR is merged (he also labels the kube-system namespace, just in a different file).

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -212,18 +212,25 @@ function install_external_dns()
...
+}
 
+function install_rancher()
+{
+    local RANCHER_CHART_DIR=${CHARTS_DIR}/rancher
     local INGRESS_TLS_SOURCE=""
     local EXTRA_RANCHER_ARGUMENTS=""
     local RANCHER_PATCH_DATA=""
+
+    # Create the rancher-operator-system namespace so we can create network policies
```
pfmackin, line 229: Why are we creating a namespace?  Wouldn't we just skip creating network policies if the namespace doesn't exist?

fredtibbitts: The namespace does eventually get created by Rancher, but there's a timing issue. If we don't create it, it may or may not exist when we apply the network policy. I tried adding a wait for the namespace in this script but it didn't consistently work and I'm not sure why. I'll check with Kevin and see if he knows why Rancher creates the namespace after the installation.

bobdonat: Is this a new namespace for the new version of Rancher?

platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
```diff
@@ -53,3 +53,93 @@ spec:
...
+  namespace: cattle-system
+spec:
+  podSelector:
+    matchLabels:
+      app: cattle-cluster-agent
+  policyTypes:
+    - Ingress
+    - Egress
+  egress:
+    - ports:
```
bobdonat, line 73: should this have selectors for Rancher pods

fredtibbitts: The Rancher pods are handled by the "rancher" network policy (conditionally applied based on `rancher.enabled`).

platform-operator/scripts/install/2-install-system-components.sh
fredtibbitts: Yes, new namespace in upgraded Rancher.

fredtibbitts: @pfmackin apparently Rancher does its own helm installs asynchronously, so either we create the rancher-operator-system namespace so we can add the network policy or we wait for Rancher to create the namespace, which can potentially add minutes to the Verrazzano installation.

platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
bobdonat: Not sure I understand why selectors cannot be specified

fredtibbitts: Might be easier to discuss this in a zoom, but the rules are different for cattle-cluster-agent and rancher pods. Maybe I'm misunderstanding what you're asking for.

fredtibbitts: Clarified the egress comment so it's clear that 443 egress is to Rancher on the admin cluster.

platform-operator/scripts/install/2-install-system-components.sh
fredtibbitts: Removed.

fredtibbitts:
I did confirm that the `rancher-operator-system` ends up with the proper annotations, labels, etc. even though we pre-create the namespace in the install script, so that should not cause any problems.

pfmackin (approved):

PR #956, opened 2021-04-12
Tvlatas/multicache

agv426:
# Description

This will switch the CI over from using pre-created clusters to using pre-created caches instead. While we will lose some of the time gains we see with pre-created clusters (a minute or so for each cluster created in a run), it does not have the trade-off that the cluster config is known to our VM images (which requires more vigilance to ensure they are aligned).

This relies on a new CI image which has the pre-created clusters removed and the caches added into it. So this effectively reverts the pre-created cluster changes, and also consolidates the cluster creation setup logic as well on top of that to clean things up a bit where we had a duplicated scripting embedded in various Jenkinsfile's

This is the same as the original PR which hit a snag in multi-cluster tests, we needed to install Calico in the Multi-cluster setup

IMPORTANT:

- We have to coordinate changing the switch over to the new CI and changing the labels in this PR from "largeexperimental". So we will switch in the new CI image, then change the PR to use "Standard2.8". This will be DISRUPTIVE and needs to be scheduled off-hours (may see some breaks during this time).
- Once this is all in, folks with outstanding branches will need to REBASE them to pickup these changes for their branches to be clean in the CI again (ie: once this is in there will not be pre-created clusters present).

# Checklist 


application-operator/Makefile
```diff
@@ -185,27 +179,27 @@ cleanup-cluster:
...
         echo 'Create cluster...' ;\
         echo 'KUBECONFIG=${KUBECONFIG}' ;\
         HTTP_PROXY="" HTTPS_PROXY="" http_proxy="" https_proxy="" time kind create cluster -v 1 \
             --name ${CLUSTER_NAME} \
             --wait 5m \
-            --config=test/kind-config.yaml ;\
+            --config=test/${KIND_CONFIG} ;\
         kubectl config set-context kind-${CLUSTER_NAME} ;\
     fi
+ifdef JENKINS_URL
```
gilbode, line 192: Why is this needed?  There is a script that does all of this that can be called from Jenkins without then condition in the makefile- look through the git history and you’ll probably find ot

application-operator/test/kind-config-ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+          extraArgs:
+            "service-account-issuer": "kubernetes.default.svc"
+            "service-account-signing-key-file": "/etc/kubernetes/pki/sa.key"
+      - |
+        kind: InitConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            node-labels: "ingress-ready=true"
+            authorization-mode: "AlwaysAllow"
+    extraPortMappings:
```
gilbode, line 21: Why are we not using the same config ee document?  https://verrazzano.io/docs/setup/platforms/kind/kind/#create-the-kind-cluster.  Particularly we shouldn’t need any extra port mappings

gilbode (approved):

michael-cico (dismissed):

application-operator/test/kind-config-ci.yaml
agv426: This yaml was just the one they had already but with the cache added. The port stuff is in the one they had there, not sure if there was a reason they had those in there, but I can remove them and see if anything falls over and respin testing for this

application-operator/Makefile
agv426: Yeah it is on the list to separately clean that up later this sprint, basically the way the app and platform operator integ-test targets work is different currently. So I will need to change up the way the app operator integ-test target works so it does things similarly (that way we can have the call to that script be done from the Jenkinsfile like it is for platform-operator)

application-operator/test/kind-config-ci.yaml
agv426: local integ-test without it works fine. Checked with folks, our best guess is that this was needed before metal-lb was being used there, and it old cruft

agv426: integ tests are clean with it gone

michael-cico (approved):

PR #967, opened 2021-04-13
Run verrazzano-api pod as non root user

desagar:
# Description

Run verrazzano-api pod as non-root user. This required changing the default prefix path to a location that the nginx user has permissions to, and changing the permissions on some of the startup shell scripts, since those are owned by root. 

**Note**: I tried to fix fluentd to not run as root, but from what I can gather, it needs root permission to read /var/log. Found some info [here](https://github.com/fluent/fluentd-kubernetes-daemonset/issues/420) and [here](https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/)

Partially Fixes VZ-2504

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
@@ -279,8 +279,9 @@ data:
 
     export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
     ln -s /etc/ssl/certs/ca-bundle.trust.crt /etc/nginx/ca-bundle
-    /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf -t --prefix /etc/nginx
-    /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf --prefix /etc/nginx
+
+    /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf -t -p /etc/nginx
+    /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf -p /etc/nginx
 
     sh -c "`dirname $0`/reload.sh &"
     tail -f /etc/nginx/logs/error.log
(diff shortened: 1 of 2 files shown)


platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -367,16 +365,16 @@ spec:
                 items:
                   - key: startup.sh
                     path: startup.sh
-                    mode: 0744
+                    mode: 0755
```
wmhopkins, line 368: Why are we adding execute permission for group+world?

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -342,9 +343,6 @@ spec:
       labels:
         app: {{ .Values.api.name }}
     spec:
-      securityContext:
```
wmhopkins, line 345: If we delete the "run as user/group 0", what uid/gid does nginx run as?

wmhopkins (commented):
Looks good to me, but had a couple questions.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
desagar: It runs by default as user id 101 which is `www-data:www-data` - here is their [Dockerfile]([Dockerfile](https://github.com/kubernetes/ingress-nginx/blob/master/rootfs/Dockerfile))

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
desagar: when the config map containing these scripts is mounted at `/api-config`, everything is owned by root. I tried specifying a securityContext section with runAsUser id 101 (the nginx default user `www-data`) but still everything mounted is owned by root:root. The `www-data` user is not a member of root group but needs to execute the scripts mounted from the config map.

wmhopkins (approved):
lgtm

PR #981, opened 2021-04-15
add debugging to MC verify permissions test

desagar:
# Description

add debugging to MC verify permissions test to figure out intermittent issue

# Checklist 

### DIFF
--- a/tests/e2e/multicluster/verify-permissions/verify_permissions_test.go
+++ b/tests/e2e/multicluster/verify-permissions/verify_permissions_test.go
@@ -57,6 +57,7 @@ var _ = ginkgo.Describe("Multi Cluster Verify Kubeconfig Permissions", func() {
 				return findMultiClusterConfigMap(testNamespace, "mymcconfigmap")
 			}, waitTimeout, pollingInterval).Should(gomega.BeTrue(), "Expected to find mc configmap")
 
+			eventuallyIterations := 0
 			gomega.Eventually(func() bool {
 				// Verify we have the expected status update
 				configMap := clustersv1alpha1.MultiClusterConfigMap{}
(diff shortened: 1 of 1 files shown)


tests/e2e/multicluster/verify-permissions/verify_permissions_test.go
```diff
@@ -66,6 +67,13 @@ var _ = ginkgo.Describe("Multi Cluster Verify Kubeconfig Permissions", func() {
 					pkg.Log(pkg.Debug, string("cluster reported status: "+configMap.Status.Clusters[0].State))
 					pkg.Log(pkg.Debug, "cluster reported name: "+configMap.Status.Clusters[0].Name)
 				}
+				eventuallyIterations++
+				if eventuallyIterations >= 30 && eventuallyIterations % 10 == 0 {
```
jmaron99, line 71: So you're starting about half way through the full 600 second wait?

desagar: I didn't want to do that because you'd get a huge dump of these things on _every_ iteration, even when the test succeeds. I wanted to minimize the debug output and keep it to only the cases where we encounter the issue that it never updates the status after waiting for a long time.

agv426 (approved):

tests/e2e/multicluster/verify-permissions/verify_permissions_test.go
jmaron99: that makes sense.  Perhaps we can print a message out when it succeeds?  I'd just like to have a baseline or average to compare to.

desagar: Yeah I think that's a good idea. But since I have an approval and a green test, I'll merge this and do that in the next PR. I expect I'll have another one tomorrow.

PR #990, opened 2021-04-16
Add support for ensuring the test jenkinsfiles are refreshed in downs…

agv426:
…tream tests for master/release*/branches where full tests are triggered

# Description

This adds in a REFRESH_JENKINS capability for the master, release*, and any feature branch that is running the full triggered set of tests.

When REFRESH_JENKINS is true, the test job will just execute the init stage then nothing else (except for some post stuff, most of that is not executed too)

For master, release*, and branches that are set to trigger all tests. the stage after the init will start a triggered tests build but giving it the REFRESH_JENKINS==true, and it will not wait for that. That will fan out and call all triggered tests telling them to also REFRESH_JENKINS. So when the actual jobs gets around to triggering the tests for real, they will have a current Jenkinsfile there.

This allows master/release* to automatically run successfully if someone changes a test Jenkinsfile to add/change parameters. This also allows devs to run their branch triggering all tests when they need to do so without starting and killing jobs manually.

# Checklist 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -139,6 +139,28 @@ pipeline {
             }
         }
 
+        stage('Refresh Downstream Jenkinsfiles') {
+            when {
+                allOf {
+                    not { buildingTag() }
+                    anyOf {
+                        branch 'master';
+                        branch 'release-*';
+                        expression {params.TRIGGER_FULL_TESTS == true};
+                    }
+                }
+            }
+            steps {
+                script {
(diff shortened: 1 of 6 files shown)


michael-cico (approved):

agv426:
We decided not to do this so closing it out:
1. For new parameters added to the pipeline that will cause issues, the developer checking them in must watch/fix master (ie: kick a new build off after it fails, don’t replay it, kick new build)
2. For branches that aren’t adding a parameter, the root cause looks like it may be that groovy interpolation and when that happens relative to when Jenkins loads things, it may be something we can fix separately

PR #1000, opened 2021-04-19
Fix jq error complaining about non-existing effective.config.json in install/uninstall logs

bhatpmk:
# Description

This PR fixes the jq error, complaining about non-existing effective.config.json when config.sh reads the value from CONFIG_JSON from function validate_config_json() before the step which generates "effective.config.json" and during undeploy.

As part of this change, the Jenkinsfile used by multi-cluster test is updated to archive the install and uninstall logs.

Fixes VZ-2564

# Checklist 

### DIFF
--- a/platform-operator/scripts/install/config.sh
+++ b/platform-operator/scripts/install/config.sh
@@ -14,6 +14,9 @@ EFFECTIVE_CONFIG_VALUES="${INSTALL_OVERRIDES_DIR}/effective.config.json"
 # The max length of the environment name passed in by the user.
 ENV_NAME_LENGTH_LIMIT=10
 
+# Flag to differentiate install from uninstall
+INSTALL_PATH=true
+
 # Read a JSON installation config file and output the JSON to stdout
 function read_config() {
   local config_file=$1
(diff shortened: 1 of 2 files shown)


platform-operator/scripts/install/config.sh
```diff
@@ -39,9 +42,11 @@ function get_config_value() {
   if [ "$config_val" == "null" ]; then
     config_val=""
   fi
+
   # check if it is defined in the json files under install-overrides
-  if [ ! $(get_override_config_value "$jq_expr") == "" ]; then
-    config_val=$(get_override_config_value "$jq_expr")
+  local config_override_value=$(get_override_config_value "$jq_expr")
+  if [ ! "$config_override_value" == "" ]; then
```
desagar, line 48: you can use != in shell i.e. `"$config_override_value" != ""`

michael-cico (approved):

platform-operator/scripts/install/config.sh
bhatpmk: Right, I just re-used the existing syntax.

platform-operator/scripts/install/config.sh
```diff
@@ -159,7 +164,12 @@ function validate_config_json {
   set -o pipefail
   local jsonToValidate=$1
   echo "$jsonToValidate" | jq . > /dev/null || fail "Failed to read installation config file contents. Make sure it is valid JSON"
+}
 
+# Make sure the environmentName, dns entries and certificates are valid in CONFIG_JSON
```
bhatpmk, line 169: We don't need a separate function validate_config_json_entries() now, will revert to the original function validate_config_json()

bhatpmk: Reverted the earlier change to validate_config_json()

michael-cico (approved):

desagar (approved):

gilbode (approved):

PR #1013, opened 2021-04-20
Tvlatas/calico version and mccaching

agv426:
# Description

This includes:

-  Added multi-cluster caching support (it is using the caches we have now, we can follow up with specific caches for multi-cluster). The intention is that this can help with rate limiting on Calico related image pulls. The images were already on the VM in docker, the suspicion is that we needed to use the cache as well. This is working and it looks like install time is down, but we won't really know related to rate limiting if it will really help or if something else with Calico could be forcing the images to always be pulled that I'm not seeing under the covers.
-  Changed the Calico yaml we are using to be as specific as they allow for version, and which yaml we use: https://docs.projectcalico.org/v3.18/manifests/calico.yaml
-  Fixed problem where apo-integ tests were not deleting the cluster in Jenkins, we had hoped it was related to the long cluster time with Calico but it isn't so this is more just a cleanup

# Checklist 


bobdonat (approved):
Looks ok to me

PR #1020, opened 2021-04-21
Potential fix for MC change placement test failure

desagar:
# Description

Potential fix for MC change placement test failure - does the following things:
* Changes the Helidon and ChangePlacement example tests to explicitly delete apps in managed cluster and wait for namespace deletion to complete before proceeding
* Changes the finalizer removal logic for MC resources to remove the finalizer _after_ deleting the underlying resource
* Re-enables the disabled previously failing assertions in Change Placement test
* Only perform a cluster dump for the 2 examples e2e tests if they fail

Partially fixes VZ-2559

# Checklist 

### DIFF
--- a/application-operator/controllers/clusters/cluster_utils.go
+++ b/application-operator/controllers/clusters/cluster_utils.go
@@ -330,7 +330,9 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
 	return statusPtr.State
 }
 
-// DeleteAssociatedResource will retrieve and delete the resource specified by the name
+// DeleteAssociatedResource will retrieve and delete the resource specified by the name. It is used to delete
+// the underlying resource corresponding to a MultiClusterxxx wrapper resource (e.g. the OAM app config corresponding
+// to a MultiClusterApplicationConfiguration)
 func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource runtime.Object, finalizerName string, resourceToDelete runtime.Object, name types.NamespacedName) error {
 	// Get and delete the associated with the name specified by resourceToDelete
 	err := c.Get(ctx, name, resourceToDelete)
--- a/application-operator/controllers/clusters/cluster_utils.go
+++ b/application-operator/controllers/clusters/cluster_utils.go
@@ -332,15 +332,7 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
 
 // DeleteAssociatedResource will retrieve and delete the resource specified by the name
 func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource runtime.Object, finalizerName string, resourceToDelete runtime.Object, name types.NamespacedName) error {
-	// assert the MC object is a controller util Object that can be processed by controller util convenience methods
-	mcObj, ok := mcResource.(controllerutil.Object)
-	if ok {
-		controllerutil.RemoveFinalizer(mcObj, finalizerName)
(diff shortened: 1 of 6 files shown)


application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -332,15 +332,7 @@ func SetEffectiveStateIfChanged(placement clustersv1alpha1.Placement,
 
 // DeleteAssociatedResource will retrieve and delete the resource specified by the name
```
pfmackin, line 333: might be good to explain the associated resource is the underlying resource MCSecret -> Secret, and that there is always exactly one.

tests/e2e/pkg/kubernetes.go
```diff
@@ -110,10 +110,15 @@ func DoesCRDExist(crdName string) bool {
 	return false
 }
 
-// DoesNamespaceExist returns whether a namespace with the given name exists for the cluster
+// DoesNamespaceExist returns whether a namespace with the given name exists for the cluster set in the environment
 func DoesNamespaceExist(name string) bool {
+	return DoesNamespaceExistInCluster(name, GetKubeConfigPathFromEnv())
```
pfmackin, line 115: it is better to inject the kube config.  How will this work for MC where there are 2 kube configs?

desagar: The MC tests can call DoesNamespaceExistInCluster and pass in the kubeconfig of the cluster they want to check for the namespace.

The existing DoesNamespaceExist _only_ works if you do os.Setenv of the kubeconfig, so only works for one at a time.

I have been slowly adding support over time to do things without os.Setenv, so that eventually, we might be able to run some of the MC tests in parallel.

application-operator/controllers/clusters/cluster_utils.go
desagar: added a comment and pushed.

application-operator/controllers/clusters/cluster_utils_test.go
```diff
@@ -437,3 +439,81 @@ func TestSetEffectiveStateIfChanged(t *testing.T) {
...
+	// THEN the finalizer should be removed
+
+	// expect get and delete the app config with name resourceToDeleteName, mocking successful deletion
+	expectGetAndDeleteAppConfig(t, cli, resourceToDeleteName, nil)
+
+	// The finalizer should be removed
+	cli.EXPECT().
+		Update(gomock.Any(), gomock.Not(gomock.Nil())).
+		DoAndReturn(func(ctx context.Context, mcApp *clustersv1alpha1.MultiClusterApplicationConfiguration, opts ...client.UpdateOption) error {
+			asserts.NotContains(t, mcApp.GetFinalizers(), finalizerToDelete)
```
pfmackin, line 481: not following how one finalizer is deleted and the other not.

tests/e2e/pkg/kubernetes.go
pfmackin: ok, but in general injection is usually best if we can do it.

application-operator/controllers/clusters/cluster_utils_test.go
desagar: This is a unit test of the method, which requires the finalizer name to be passed in. I'm passing in the finalizer name and asserting that it deletes what was passed in. 

It's probably unnecessary to add one more finalizer and test that it's not deleted, since that's never a real use case. If you feel strongly I can simply remove that test case.

pfmackin (approved):

PR #1028, opened 2021-04-23
Changes to allow magic DNS provider to be specified (xip.io, nip.io, sslip.io)

bobdonat:
# Description

This pull request allows the magic DNS provider to be specified when installing Verrazzano.  

The verrazzano CRD has been changed to replace dns.XIPIO with dns.wildcard.domain where the DNS provider can be specified. For example:

    dns:
      wildcard:
        domain: nip.io

The install continues to default to xip.io for now.  That can be changed later if we want a different default.

Fixes VZ-2566

# Checklist 


pfmackin (approved):

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -733,19 +733,28 @@ func buildNamespacedDomainName(cli client.Reader, trait *vzapi.IngressTrait) (st
 
 	domain := externalDNSAnno[len(constants.VzConsoleIngress)+1:]
 
-	// If this is xip.io then build the domain name using Istio info
+	suffix := ""
 	if strings.HasSuffix(domain, "xip.io") {
-		domain, err = buildDomainNameForXIPIO(cli, trait)
+		suffix = "xip.io"
```
michael-cico, line 738: Don't have to do this on this pass, but maybe we should just validate this on the VZ CR in the webhook?  Then we can use the value without the if-branch.

bobdonat: yes, we can validate in the webhook but I think this check would still be needed.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -733,19 +733,28 @@ func buildNamespacedDomainName(cli client.Reader, trait *vzapi.IngressTrait) (st
...
+	} else if strings.HasSuffix(domain, "nip.io") {
```
gilbode, line 739: This should not be hardcoded to a predefined set of domains

bobdonat: working on fix in follow on PR

PR #1029, opened 2021-04-23
Revert "Revert VZ-2346: Update to newer Rancher"

kminder:
Reverts the revert of the Rancher update effectively updating Rancher to 2.5.7.


tests/e2e/verify-install/kubernetes/kubernetes_test.go
```diff
@@ -137,17 +137,13 @@ var _ = ginkgo.Describe("Kubernetes Cluster",
 					gomega.Expect(vzComponentPresent(name, "cattle-system")).To(gomega.BeFalse())
 				},
 				ginkgoExt.Entry("includes rancher", "rancher", false),
-				ginkgoExt.Entry("includes rancher-agent", "cattle-node-agent", false),
```
desagar, line 140: is there no longer a node agent and cluster agent? What is in the managed cluster when the registration is completed?

desagar:
another question - are the changes in the third-party/charts dir all part of the vanilla Rancher helm chart or did we have any edits on top of it which need to be documented?

desagar (approved):

tests/e2e/verify-install/kubernetes/kubernetes_test.go
kminder: In the multi cluster case I believe there ends up being a rancher-agent on the managed cluster.  On the admin cluster there is no longer a separate rancher-agent.

kminder:
In response to
> another question - are the changes in the third-party/charts dir all part of the vanilla Rancher helm chart or did we have any edits on top of it which need to be documented?
All of the helm chart changes are from the upstream Rancher.

fredtibbitts (approved):

PR #1043, opened 2021-04-28
Add test for api-proxy serviceaccount impersonation privilege

lake-of-dreams:
# Description

Verifies that the serviceaccount verrazzano-api which is used in api-proxy is not able to impersonate any other service account by invoking equivalent of 
```bash
kubectl auth can-i impersonate serviceaccounts --token=<token-of-service-account-of-api-proxy>
```
using the kuberenets golang client. On kind clusters, the result returned by the golang clientset may not be accurate and therefore the test retries by making a REST API call to verify the privilege on k8s api server directly.

Fixes VZ-2465

# Checklist 


wmhopkins (commented):
This looks OK to me, but it seems odd to me that we can't rely on the results of kubectl (or a k8s API call), and have to resort to using curl to call the API. Isn't kubectl calling the same REST API as curl? If so, why does it work with curl and not kubectl? I'm concerned that we're coding a workaround for a problem we shouldn't be having in the first place.

lake-of-dreams:
@wmhopkins please if you can re-review?

wmhopkins (approved):
lgtm

PR #1057, opened 2021-04-30
Changes to Istio authorization policy to allow traffic between apps in a project

bobdonat:
# Description

This pull request includes changes to Istio authorization policies to allow traffic between all apps in a project.  If projects are not being used, then the Istio authorization policy remains as is and traffic is limited to pods in an application.

Istio authorization policies are created/updated via the istio-defaulter webhook and authorizations polices are updated via the appconfig-defaulter webhook when an appconfig resource is deleted. 

Fixes VZ-2565

# Checklist 


application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+	// Walk the list of projects looking for a project namespace that matches the given namespace
+	for _, project := range projectsList.Items {
+		namespaceFound := false
+		for _, ns := range project.Spec.Template.Namespaces {
+			if ns.Metadata.Name == namespace {
+				namespaceFound = true
+				break
+			}
+		}
+
```
pfmackin, line 54: better to return here if namespace not found then get rid if the if block block

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+			if err != nil {
+				return err
+			}
+
+			// get the list of service accounts remaining in the namespace of where an appconfig delete is taking place
+			saList := []string{}
+			for _, pod := range podList.Items {
+				if value, ok := pod.Labels[verrazzanoIstioLabel]; ok {
+					if value != appConfigName {
+						saList = append(saList, pod.Spec.ServiceAccountName)
```
pfmackin, line 89: won't this end up with potential duplicate service accounts on the list?

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+	projectsList := &cluv1alpha1.VerrazzanoProjectList{}
+	listOptions := &client.ListOptions{Namespace: constants.VerrazzanoMultiClusterNamespace}
+	err := ap.Client.List(context.TODO(), projectsList, listOptions)
+	if err != nil {
+		return err
+	}
+
+	// Walk the list of projects looking for a project namespace that matches the given namespace
+	for _, project := range projectsList.Items {
+		namespaceFound := false
```
pfmackin, line 138: this is dup, might want common function

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: correct

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: if namespace is not found we need to continue looking in the next project.  Not sure I understand what a return would do.

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+		for _, ns := range project.Spec.Template.Namespaces {
+			if ns.Metadata.Name == namespace {
+				namespaceFound = true
+				break
+			}
+		}
+
+		// Project has a namespace that matches the given namespace
+		if namespaceFound {
```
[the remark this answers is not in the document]
bobdonat: see response to previous comment

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: Some of the code is duplicated but there are differences so I had two functions.

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+				}
+			}
+
+			// Create list of unique principals for all authorization policies in a project.
+			uniquePrincipals := make(map[string]bool)
+			for _, authzPolicy := range authzPolicyList {
+				for _, principal := range authzPolicy.Spec.Rules[0].From[0].Source.Principals {
+					// For the namespace passed, only include the service accounts remaining for the namespace
+					split := strings.Split(principal, "/")
+					if split[2] == namespace {
```
pfmackin, line 100: are we sure there will alway be at least 3 segments?

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+
+		// Project has a namespace that matches the given namespace
+		if namespaceFound {
+			// Get the authorization policies for all the namespaces in a project.
+			tempAuthzPolicyList, err := ap.getAuthorizationPoliciesForProject(project.Spec.Template.Namespaces)
+			if err != nil {
+				return err
+			}
+
+			// Filter the authorization policies we retrieved
```
pfmackin, line 63: Is this filtering for the specific app?  If so can you add that comment

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+
+			// Create list of unique principals for all authorization policies in a project.
+			uniquePrincipals := make(map[string]bool)
+			for _, authzPolicy := range authzPolicyList {
+				for _, principal := range authzPolicy.Spec.Rules[0].From[0].Source.Principals {
+					// For the namespace passed, only include the service accounts remaining for the namespace
+					split := strings.Split(principal, "/")
+					if split[2] == namespace {
+						for _, sa := range saList {
+							if split[4] == sa {
```
pfmackin, line 102: same question as before, we are assuming at least 5 segments

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+				return nil
+			}
+
+			// get a list of pods for the given namespace
+			podList, err := ap.KubeClient.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{})
+			if err != nil {
+				return err
+			}
+
+			// get the list of service accounts remaining in the namespace of where an appconfig delete is taking place
```
pfmackin, line 84: can you add "This service account list does not include the service account used by the app being deleted"

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+			for _, pod := range podList.Items {
+				if value, ok := pod.Labels[verrazzanoIstioLabel]; ok {
+					if value != appConfigName {
+						saList = append(saList, pod.Spec.ServiceAccountName)
+					}
+				}
+			}
+
+			// Create list of unique principals for all authorization policies in a project.
+			uniquePrincipals := make(map[string]bool)
```
pfmackin, line 95: Can you add "... this doesn't include principals used by the app being deleted"

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: there will always be 5 segments

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: yes, always 5

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: fixed

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: fixed

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: comment added.

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: we always create 5 segments.  I guess the user could edit an authorization policy and include a bad principal.  The code should probably generate an error if there are not 5 segments.

bobdonat: the principals can be something like

principals: ["admin", "dev"]

but I am not sure what that means

application-operator/controllers/webhooks/project_authorization_policy.go
pfmackin: right, it would need a continue.  Anyway it is fine, just a lot in one function and less indentation is better.

application-operator/controllers/webhooks/project_authorization_policy.go
bobdonat: added error check for number of segments to avoid invalid index panic

application-operator/controllers/webhooks/appconfig_defaulter.go
```diff
@@ -104,10 +106,22 @@ func (a *AppConfigWebhook) Handle(ctx context.Context, req admission.Request) ad
...
+
+	err = a.cleanupSecret(appConfig)
+	if err != nil {
+		return
+	}
+
+	// Fixup Istio Authorization policies within a project
+	ap := &AuthorizationPolicy{
+		Client:      a.Client,
+		KubeClient:  a.KubeClient,
```
pfmackin, line 121: Since we are using kubeclient directly (no controller runtime cache), we should minimize calls got the API server if possible, like getting a list of namespaces.  Especially since this is a webhook and the user is waiting for the kubecutl (or whatever) response.  Just a suggestion for future improvement (if possible)

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+	for element := range uniquePrincipals {
+		if !contains(element, principals) {
+			return false
+		}
+	}
+	return true
+}
+
+// contains checks that a given string is in an array of strings
+func contains(given string, list []string) bool {
```
pfmackin, line 251: You should use verrazzano/pkg/string for this

application-operator/controllers/webhooks/project_authorization_policy.go
```diff
@@ -0,0 +1,252 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// unorderedEqual checks if a map and array have the same string elements
+func unorderedEqual(uniquePrincipals map[string]bool, principals []string) bool {
```
pfmackin, line 238: can you put this verrazzano/pgk/string/slice.go

pfmackin (approved):

PR #1058, opened 2021-04-30
VZ-2484: Add support for Calico on OKE

fredtibbitts:
# Description

This PR adds support for installing Calico on OKE when running OKE tests.

Notes:

- The OKE documentation for Calico recommends installing Calico in the cluster before the node pool is created. I'm doing it this way in the script that creates the OKE cluster and it seems to solve the issues with coredns pods restarting and getting into a bad state.
- For some reason I can't explain, the network policy for webhook ingress was causing problems with our operator webhooks. Removing the source IP from the rule and just specifying the 9443 ingress port solved the problem. I don't think we saw this problem with Kind, so not sure what's going on, but this seems like a reasonable solution (at least for now).
- The `create_oke_cluster.sh` script was using an optional arg to setup the cluster with private access. None of our other scripts use optional args, we use positional args, so I changed it to be consistent and replaced the "-p" with a boolean positional arg.

Implements VZ-2484

# Checklist 

### DIFF
--- a/tests/e2e/config/scripts/install_calico_oke.sh
+++ b/tests/e2e/config/scripts/install_calico_oke.sh
@@ -4,13 +4,17 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 #
 
-echo "Install Calico..."
+CALICO_VERSION=${1:-"3.18.1"}
 
-curl -sS https://docs.projectcalico.org/v3.18/manifests/calico-policy-only.yaml -o /tmp/calico.yaml
+echo "Install Calico ${CALICO_VERSION}..."
+
+# CALICO_HOME must be set to the location of the downloaded Calico bundle
+CALICO_YAML=${CALICO_HOME}/${CALICO_VERSION}/k8s-manifests/calico-policy-only.yaml
 
 # uncomment the pod CIDR setting and replace the default pod CIDR with the OKE pod CIDR default
-sed -i -e "s?# - name: CALICO_IPV4POOL_CIDR?- name: CALICO_IPV4POOL_CIDR?" /tmp/calico.yaml
-sed -i -e "s?#   value: \"192.168.0.0/16\"?  value: \"10.244.0.0/16\"?" /tmp/calico.yaml
-grep -B 1 -A 2 POOL_CIDR /tmp/calico.yaml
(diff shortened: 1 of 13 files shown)


tests/e2e/config/scripts/install_calico_oke.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+# Copyright (c) 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+
+echo "Install Calico..."
+
+curl -sS https://docs.projectcalico.org/v3.18/manifests/calico-policy-only.yaml -o /tmp/calico.yaml
```
bobdonat, line 9: for Kind I believe we are using 3.18.1.  should we do the same and be locked to 3.18.1 instead of getting the latest 3.18 version.

bobdonat: for kind it looks like we are using calico.yaml instead of calico-policy-only.yaml.  we should probably be consistent one way or the other

bobdonat:
"The OKE documentation for Calico recommends installing Calico in the cluster before the node pool is created".  Is this something we will need to document?  Is there a way to do that if you create an OKE cluster with the OCI console?

tests/e2e/config/scripts/install_calico_oke.sh
fredtibbitts: They don't provide links that are that granular. I would have to do what Bhat did and download the tar file and pull out the YAML. I suppose we should be using the exact same patch version though, so let me look at doing that.

fredtibbitts:
@bobdonat you can do it if you use the OCI Console to create the cluster. The cluster is accessible while the node pool is coming up, and there's enough time to install Calico, but it's much easier to do if the cluster is provisioned with Terraform.

tests/e2e/config/scripts/install_calico_oke.sh
fredtibbitts: I'm not sure what the difference is, but maybe it's platform dependent? The OKE instructions specifically call out using calico-policy-only.yaml.

bobdonat: I downloaded both and they were the same

bobdonat (approved):

pfmackin (dismissed):

pfmackin (approved):

PR #1070, opened 2021-05-04
Automated tests for network security policies

tbeerbower:
# Description

Add automated tests for the layer 3/4 network security policies. 
For the expected collection of network policies...

- examine each ingress rule
  - from each pod specified in the ingress rule attempt to connect to the policy's pod on the specified port(s) and verify connection
  - from at least one pod NOT specified in the ingress rule attempt to connect to the policy's pod on the specified port(s) and verify that connection can not be established
- examine each egress rule
  - from the policy's pod attempt to connect to each pod specified in the egress rule on the specified port(s) and verify connection
  - from the policy's pod attempt to connect to at least one pod NOT specified in the egress rule on the specified port(s) and verify that connection can not be established

Fixes VZ-2561

# Checklist 


tests/e2e/multicluster/verify-install/defaultresource_test.go
```diff
@@ -5,6 +5,7 @@ package defaultresource_test
 
 import (
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
kminder, line 8: import order

tests/e2e/pkg/kubernetes.go
```diff
@@ -128,12 +133,12 @@ func DoesNamespaceExistInCluster(name string, kubeconfigPath string) bool {
...
 
-// ListNamespaces returns whether a namespace with the given name exists for the cluster
-func ListNamespaces() *corev1.NamespaceList {
+// ListNamespaces returns a namespace list for the given list options
+func ListNamespaces(opts metav1.ListOptions) *corev1.NamespaceList {
 	// Get the Kubernetes clientset
 	clientset := GetKubernetesClientset()
 
-	namespaces, err := clientset.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	namespaces, err := clientset.CoreV1().Namespaces().List(context.TODO(), opts)
```
kminder, line 141: Not part of your change but calling ginkgo.Fail from within a utility function causes issues with things like retry.

tests/e2e/pkg/kubernetes.go
```diff
@@ -178,6 +183,38 @@ func ListNodes() *corev1.NodeList {
 	return nodes
 }
 
+// GetPodsFromSelector returns a collection of pods for the given namespace and selector
+func GetPodsFromSelector(selector *metav1.LabelSelector, namespace string) []corev1.Pod {
```
kminder, line 187: Should these be checking for an returning errors?

tests/e2e/pkg/kubernetes.go
```diff
@@ -582,6 +619,47 @@ func DoesRoleBindingExist(name string, namespace string) bool {
...
 }
 
+// GetNetworkPolicy returns the the network policy with the given name
+func GetNetworkPolicy(name string, namespace string) *v1.NetworkPolicy {
+	// Get the Kubernetes clientset
+	clientset := GetKubernetesClientset()
+
+	policy, err := clientset.NetworkingV1().NetworkPolicies(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		ginkgo.Fail(fmt.Sprintf("Failed to get network policy: %v", err))
```
kminder, line 629: Not a good idea to fail in a utility function.

tests/e2e/security/network-policies/network_policy_test.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// validatePolicyRule validates ingress or egress rules by checking that a connection can be (or can't be for
+// negative test) established between the network policy pod and the pods specified in the given network policy peers
+// for the given network policy ports
+func validatePolicyRule(policyPod corev1.Pod, fn orderPods, peers []v1.NetworkPolicyPeer, ports []v1.NetworkPolicyPort, podMap map[string]corev1.Pod) error {
```
pfmackin, line 109: Negative tests are going to time out, which will make the test take a very long time to run if they are done serially.    We should run all negative tests in parallel for n seconds (say 15 secs) and if any successfully connect then that would be considered a failure.

tests/e2e/security/network-policies/network_policy_test.go
```diff
@@ -0,0 +1,339 @@
...
+			} else if !connected && expectConnect {
+				return fmt.Errorf(fmt.Sprintf("Should be able to connect from pod %s to pod %son port %d", fromPod.Name, toPod.Name, port))
+			}
+		}
+	}
+	return nil
+}
+
+// attemptConnection attempts to establish a connection from a given pod to another pod on a given port
+func attemptConnection(fromPod, toPod *corev1.Pod, port, steps int, duration time.Duration) bool {
```
pfmackin, line 324: How long does this function take to execute in the case where access is denied and it will be a timeout.

tbeerbower: It uses a backoff backoff.  It tries 3 times for a total of 7 seconds.  When it connects, I've never seen it go past the first attempt.

tests/e2e/security/network-policies/network_policy_test.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+})
+
+var _ = ginkgo.Describe("Test Network Policies", func() {
+
+	// GIVEN a Verrazzano deployment
+	// WHEN connections are attempted between pods within the ingress/egress rules of the Verrazzano network policies
+	// THEN the connections should be successfully established
+	ginkgo.It("Test NetworkPolicy Rules", func() {
+		pkg.Concurrently(
```
pfmackin, line 43: Ideally, we should get a list network policies then generically do these tests based on what is in the policy vs hard coding these policies in the list below.   Furthermore, this won't test project policies and app policies.  With that said this is a good first step.

tests/e2e/security/network-policies/network_policy_test.go
```diff
@@ -0,0 +1,339 @@
...
+			func() {
+				pkg.Log(pkg.Info, "Test cert-manager ingress rules")
```
pfmackin, line 45: rancher and istio-system are missing

tests/e2e/security/network-policies/network_policy_test.go
```diff
@@ -0,0 +1,339 @@
...
+		)
+	})
+
+	// GIVEN a Verrazzano deployment
+	// WHEN connections are attempted between pods that violate the ingress/egress rules of the Verrazzano network policies
+	// THEN the connections should NOT be established
+	ginkgo.It("Negative Test NetworkPolicy Rules", func() {
+		pkg.Concurrently(
+			func() {
+				pkg.Log(pkg.Info, "Negative test cert-manager ingress rules")
```
pfmackin, line 173: rancher and istio-system are missing

tests/e2e/security/network-policies/network_policy_test.go
tbeerbower: Were they just recently added?

tests/e2e/security/network-policies/network_policy_test.go
tbeerbower: Hey @pfmackin.  Thanks for reviewing.
I started down that road but it's not really that easy.  You can't just look at a single policy and come up with a valid test.  For example, you can't generate a valid test for an ingress rule unless you know the corresponding egress rules in the other policies.  Also, it's hard to write negative tests because a connection that is not allowed by one policy may be allowed by another.
Also, it kind of feels like not having explicit validations for policies that we wrote isn't really testing verrazzano, but is instead testing calico.  If someone broke verrazzano and some or all of the policies either didn't get created or created rules that prevented verrazzano from operating correctly, the generic test code wouldn't care.  As long as the tests could do what was specified in the policy the generic tests would pass even if that didn't allow verrazzano to function.

pfmackin (approved):
Asking Fred to double check the rancher tests since the the netpols recently got changed.  Other than that looks ok

PR #1071, opened 2021-05-04
Remove --wait from the kind create cluster command

bhatpmk:
# Description

This change resolves the issue of KIND cluster creation taking more than 5 minutes with Calico (networking configuration in KIND configuration file).

When the flag --wait is specified for the 'kind create cluster' command, it waits till the duration specified, for control plane node to be ready. With the KIND configuration setting "networking.disableDefaultCNI to true", the node will be READY only after we apply calico.yaml.

The --wait 5m is removed from the "kind create cluster" command. A separate check is added to ensure that the node is ready after installing Calico, which waits for 300 seconds, for the node to reach Ready status.

Fixes VZ-2562

# Checklist 


ci/scripts/wait_for_control_plane.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# This script waits for the KIND control plane node to reach Ready status, with a default wait period of 300 seconds.
+# The script assumes the cluster access configuration is stored in ${HOME}/.kube/config or environment variable
+# KUBECONFIG is set appropriately.
+
+SCRIPT_DIR=$(cd $(dirname "$0"); pwd -P)
+CLUSTER_NAME=${1:-"kind"}
+MAX_WAIT=${2:-300}
+POLLING_INTERVAL=${3:-10}
+LIST_PODS=${4:-"true"}
```
agv426, line 15: We should use boolean for LIST_PODS and the isNodeReady rather than strings (the strings used here are not consistent casing now, but can just make them boolean true/false)

bhatpmk: @agv426 I have fixed the script now.

markxnelson (dismissed):

agv426 (approved):

ci/multicluster/Jenkinsfile
```diff
@@ -583,6 +583,10 @@ def installKindCluster(count, metallbAddressRange, cleanupKindContainers, connec
                     echo "Install Calico"
                     cd ${GO_REPO_PATH}/verrazzano
                     ./ci/scripts/install_calico.sh "${CLUSTER_NAME_PREFIX}-$count" "${CALICO_VERSION}"
+                    kubectl wait --for=condition=ready nodes/${CLUSTER_NAME_PREFIX}-$count-control-plane --timeout=5m
```
agv426, line 586: We want to always wait after creating the cluster right ? ie: even if calico wasn't installed, after this if block

bhatpmk: I think that is better, as the --wait is removed while creating the cluster.

agv426: But this only happens when someone specifies to use Calico right (ie: in this if block), are we waiting somewhere else for the cluster to come up when Calico is not specified?

agv426: ie: the --wait was removed in general right?

bhatpmk: Please ignore my previous comment, I have fixed the Jenkins script now.

agv426 (approved):

PR #1073, opened 2021-05-04
WVZ-2626: Change default wildcard DNS from xip.io to nip.io

wmhopkins:
# Description

Change the default wildcard DNS from xip.io to nip.io

Fixes VZ-2626

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/installjob/install_config_test.go
+++ b/platform-operator/controllers/verrazzano/installjob/install_config_test.go
@@ -16,11 +16,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// TestXipIoInstallDefaults tests the creation of an xip.io install default configuration
+// TestNipIoInstallDefaults tests the creation of an xip.io install default configuration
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/installjob/install_config_test.go
```diff
@@ -16,11 +16,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
```
pfmackin, line 18: There are a bunch of unit tests that have xip.io as default.   I guess that is ok for this commit, but we should change the default in all our unit tests at some point.  Also there are some comments saying xip.io is default, just grep for xip.io in the go code.

wmhopkins: Yeah, I saw that a lot of files had "xip.io" in them, but figured a bunch were probably comments, and if the tests passed this change would be OK to submit, since we're trying to get it in ASAP.

pfmackin (approved):

PR #1078, opened 2021-05-05
Rdonat/vz 2628

bobdonat:
# Description

This pull request fixes a bug where the tls-rancher-ingress cert would not get to a ready state because a ClusterIssuer kind was not being used but instead an Issuer kind was being used.  This problem was introduced with the upgrade to Rancher and happens during an OCI DNS install with the ACME certificate type.

Fixes VZ-2628

# Checklist 


wmhopkins (approved):
Not sure how this actually works, but the code looks syntactically correct.

PR #1083, opened 2021-05-06
Remove premature Fail in acceptance tests

guoyongzhang:
# Description

Remove some premature Fail in acceptance tests.
For example, instead of Fail, we want to return false in case of IsNotFound when getting a resource.  So, Eventually block in tests would have a chance to come back later to retry.

# Checklist 

### DIFF
--- a/tests/e2e/istio/authz/authpolicy_test.go
+++ b/tests/e2e/istio/authz/authpolicy_test.go
@@ -478,6 +478,9 @@ var _ = ginkgo.Describe("Verify Auth Policy Prometheus Scrape Targets", func() {
 			var httpsFound bool = false
 
 			configMap := pkg.GetConfigMap(vmiPromConfigName, verrazzanoNamespace)
+			if configMap == nil {
+				return false
+			}
 			dataMap := configMap.Data
 			v := dataMap[prometheusConfigMapName]
 			rdr := strings.NewReader(v)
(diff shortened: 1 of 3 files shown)


tests/e2e/istio/authz/authpolicy_test.go
```diff
@@ -478,6 +478,9 @@ var _ = ginkgo.Describe("Verify Auth Policy Prometheus Scrape Targets", func() {
 			var httpsFound bool = false
 
 			configMap := pkg.GetConfigMap(vmiPromConfigName, verrazzanoNamespace)
+			if configMap == nil {
```
shihchang, line 481: Does the pkg.GetConfigMap return nil, error, or just ginkgo.Fail when notFound or encountering an error?

tests/e2e/security/rbac/rbac_test.go
```diff
@@ -253,6 +253,9 @@ var _ = ginkgo.Describe("Test Verrazzano API Service Account", func() {
 	ginkgo.Context("for serviceaccount verrazzano-api", func() {
 		var apiProxy corev1.Pod
 		sa := pkg.GetServiceAccount(verrazzanoSystemNS, verrazzanoAPI)
+		if sa == nil {
```
shihchang, line 256: Does pkg.GetServiceAccount return nil, error, or ginkgo.Fail when notfound?

tests/e2e/istio/authz/authpolicy_test.go
desagar: I checked - it does ginkgo.Fail. Same is true for the other method call.

desagar (approved):

PR #1086, opened 2021-05-06
HelidonWorkload CRD Validation

arjav-desai:
# Description

Add CRD validation tags and a small check for metadata field which is open ended in k8s.

Fixes VZ-2625

# Checklist 


application-operator/controllers/helidonworkload/helidonworkload_controller_test.go
```diff
@@ -126,6 +126,66 @@ func TestReconcileFetchWorkloadError(t *testing.T) {
...
+		},
+	}
+	deploymentTemplate := &vzapi.DeploymentTemplate{
+		Metadata: metav1.ObjectMeta{
+			Namespace: namespace,
+			Labels: map[string]string{
+				"app": "hello-helidon-deploy-new",
+			},
+		},
+		PodSpec: v1.PodSpec{
```
fredtibbitts, line 160: Is populating the PodSpec necessary in this test? I think you could remove the PodSpec and associated container structs and the test would still pass? It's minor so I think it's fine to leave it.

fredtibbitts (approved):
LGTM

application-operator/controllers/helidonworkload/helidonworkload_controller_test.go
arjav-desai: Since its unit tests, its not required but if it was real scenario than CRD validation will fail during application config apply (since podSpec is required) and not even hit our controller.

PR #1088, opened 2021-05-06
VZ-2542: Add network policies for Istio

fredtibbitts:
# Description

This PR adds ingress network policies for things running in the `istio-system` namespace. There are static policies that get created at Verrazzano install time (in `08-thirdparty-networkpolicy.yaml`) as well as dynamic policies created by the Verrazzano Application Operator by the app config mutating webhook. The dynamic policies allow the Istio proxy sidecars in pods in the application namespaces to talk to iostiod.

**Notes**:

- I ran into intermittent problems labeling the application namespace because other things were updating the namespace at the same time. This resulted in sporadic "the resource has already been updated, please retry..." errors that resulted in test failures. I added retries around labeling the namespace to handle that case.
- The application operator integration tests failed because the `istio-system` namespace did not exist. I updated the Makefile target that sets up the integration test cluster to create the namespace.
- This PR also includes cleanup of duplicate imports in several files, as well as removing an unused variable, and refactoring to use shared constants to remove code duplication (not related to this work).

Implements VZ-2542

# Checklist 

### DIFF
--- a/application-operator/controllers/webhooks/net_policy_defaulter.go
+++ b/application-operator/controllers/webhooks/net_policy_defaulter.go
@@ -10,6 +10,7 @@ import (
 	"github.com/verrazzano/verrazzano/application-operator/constants"
 	corev1 "k8s.io/api/core/v1"
 	netv1 "k8s.io/api/networking/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -36,13 +37,16 @@ func (n *NetPolicyDefaulter) Default(appConfig *oamv1.ApplicationConfiguration, 
 
 	if !dryRun {
 		// label the namespace
-		n.ensureNamespaceLabel(appConfig.Namespace)
+		err := n.ensureNamespaceLabel(appConfig.Namespace)
+		if err != nil {
+			return err
+		}
 
 		// create/update the network policy
 		logger.Info("Ensuring Istio network policy exists", "namespace", appConfig.Namespace, "app config name", appConfig.Name)
 		netpol := newNetworkPolicy(appConfig)
 
-		_, err := controllerutil.CreateOrUpdate(context.TODO(), n.Client, &netpol, func() error {
+		_, err = controllerutil.CreateOrUpdate(context.TODO(), n.Client, &netpol, func() error {
 			netpol.Spec = newNetworkPolicySpec(appConfig.Namespace)
 			return nil
 		})
@@ -63,9 +67,9 @@ func (n *NetPolicyDefaulter) Cleanup(appConfig *oamv1.ApplicationConfiguration, 
 		logger.Info("Deleting Istio network policy", "namespace", appConfig.Namespace, "app config name", appConfig.Name)
 		netpol := newNetworkPolicy(appConfig)
 		err := n.Client.Delete(context.TODO(), &netpol, &client.DeleteOptions{})
-		if err != nil {
+		if err != nil && !k8serrors.IsNotFound(err) {
 			// log the error but don't return it so we continue processing
-			logger.Error(err, "Unable to delete network policy")
+			logger.Error(err, "Unable to delete network policy, ignoring")
 		}
 	}
 
@@ -81,6 +85,10 @@ func (n *NetPolicyDefaulter) ensureNamespaceLabel(namespace string) error {
 	if err != nil {
 		logger.Error(err, "Unable to fetch namespace", "namespace", namespace)
 		return err
+	}
+
+	if ns.ObjectMeta.Labels == nil {
+		ns.ObjectMeta.Labels = make(map[string]string)
 	}
 
 	val, exists := ns.ObjectMeta.Labels[constants.LabelVerrazzanoNamespace]
--- a/application-operator/controllers/webhooks/net_policy_defaulter.go
+++ b/application-operator/controllers/webhooks/net_policy_defaulter.go
@@ -36,17 +36,23 @@ func (n *NetPolicyDefaulter) Default(appConfig *oamv1.ApplicationConfiguration, 
 	}
 
 	if !dryRun {
-		// label the namespace
-		err := n.ensureNamespaceLabel(appConfig.Namespace)
-		if err != nil {
-			return err
+		// label the namespace - retry if we get update conflicts
+		for retryCount := 0; retryCount < 5; retryCount++ {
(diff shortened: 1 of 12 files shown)


application-operator/controllers/webhooks/net_policy_defaulter.go
```diff
@@ -0,0 +1,154 @@
...
+		logger.Info("App config is being deleted, nothing to do", "namespace", appConfig.Namespace, "app config name", appConfig.Name)
+		return nil
+	}
+
+	if !dryRun {
+		// label the namespace - retry if we get update conflicts
+		for retryCount := 0; retryCount < 5; retryCount++ {
+			err := n.ensureNamespaceLabel(appConfig.Namespace)
+			if err != nil {
+				if k8serrors.IsConflict(err) {
```
bobdonat, line 43: Do you know why there are conflicts labeling the namespaces?

bobdonat: Is this because the cache client is being used.

platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
```diff
@@ -580,3 +580,110 @@ spec:
...
+  name: istio-ingressgateway
+  namespace: istio-system
+spec:
+  podSelector:
+    matchLabels:
+      app: istio-ingressgateway
+  policyTypes:
+    - Ingress
+  ingress:
+    - ports:
```
bobdonat, line 613: I noticed that the ingressgateway and egressgateway have other container ports.
For example, ingressgateway has these ports 15021/TCP, 8080/TCP, 8443/TCP, 15443/TCP, 15090/TCP.  Are use of those other ports going to be problematic.

fredtibbitts: I'm not sure, all of the tests pass without those other ports open. I'm not sure how else to check?

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: I don't know, something else is updating the namespace though. I've seen this conflict error in other controller logs in the past, but because they use reconciling, eventually the operations complete. Since we don't have reconciling here, I think retrying on conflict is reasonable.

platform-operator/helm_config/charts/verrazzano/templates/08-thirdparty-networkpolicy.yaml
```diff
@@ -580,3 +580,110 @@ spec:
       ports:
         - port: 8080
           protocol: TCP
+---
```
pfmackin, line 583: This file is starting to get gigantic.  I would prefer to break it up, or at least put this in an istio specific file.

application-operator/controllers/webhooks/net_policy_defaulter.go
pfmackin: I think this is a cache client, see main.go where mgr.GetClient is used.

application-operator/controllers/webhooks/net_policy_defaulter.go
```diff
@@ -0,0 +1,154 @@
...
+	if appConfig.DeletionTimestamp != nil {
+		logger.Info("App config is being deleted, nothing to do", "namespace", appConfig.Namespace, "app config name", appConfig.Name)
+		return nil
+	}
+
+	if !dryRun {
+		// label the namespace - retry if we get update conflicts
+		for retryCount := 0; retryCount < 5; retryCount++ {
+			err := n.ensureNamespaceLabel(appConfig.Namespace)
+			if err != nil {
```
pfmackin, line 42: better to break immediately, reduces indentation
if err == nil {
  break
}

application-operator/controllers/webhooks/net_policy_defaulter.go
```diff
@@ -0,0 +1,154 @@
...
+			Name:      appConfig.Namespace + "-" + appConfig.Name,
+		},
+	}
+}
+
+// newNetworkPolicySpec returns a NetworkPolicySpec struct populated with the istiod
+// ingress rule that allows ingress to istiod from the application namespace.
+func newNetworkPolicySpec(namespace string) netv1.NetworkPolicySpec {
+	tcpProtocol := corev1.ProtocolTCP
+	istiodPort := intstr.FromInt(15012)
```
pfmackin, line 129: would be good if this was a constant in this file (or function) to make it clear what the port is.

pfmackin: What port is used for Coherence where TLS is disabled?  We will add the rule correct?  I assume the coherence pod doesn't have the proxy sidecar.

application-operator/controllers/webhooks/net_policy_defaulter.go
```diff
@@ -0,0 +1,154 @@
...
+			logger.Error(err, "Unable to create or update network policy", "namespace", appConfig.Namespace, "app config name", appConfig.Name)
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Cleanup deletes the Istio network policy associated with the app config.
+func (n *NetPolicyDefaulter) Cleanup(appConfig *oamv1.ApplicationConfiguration, dryRun bool) error {
```
pfmackin, line 70: when is this called?

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: The client used here is the one returned from `mgr.GetClient()` in `main.go`. Are you saying I should use a different client? Where do I get the other client from?

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: I have no idea. I'll need more info on this.

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: Done

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: Sounds like this isn't an issue.

application-operator/controllers/webhooks/net_policy_defaulter.go
fredtibbitts: Per conversation, I changed this and added a kubeclient namespace client for getting/updating the namespace. That should help with potential update conflicts since the namespace won't be cached in that client.

pfmackin (approved):

PR #1118, opened 2021-05-12
VZ-2603: STaaS issue

kminder:
# Description

Prevents Istio Envoy from generating HTTP Server header.  Also adds a test for the ingress-nginx which is currently disabled but can be enabled when VZ-2599 is completed.

Fixes VZ-2603: STaaS issue

# Checklist 


tests/e2e/examples/helidon/helidon_example_test.go
```diff
@@ -177,8 +179,37 @@ func helloHelidonPodsRunning() bool {
...
+	}
+	req.Host = hostname
+	httpClient := pkg.GetVerrazzanoHTTPClient()
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		pkg.Log(pkg.Error, fmt.Sprintf("Unexpected error=%v", err))
+		return false
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
```
pfmackin, line 195: You have to always read the body or there will be handle leak.   See https://github.com/golang/go/blob/master/src/net/http/response.go. Response.Body comments.

platform-operator/thirdparty/charts/istio/gateways/istio-ingress/templates/preconfigured.yaml
```diff
@@ -70,3 +70,26 @@ spec:
       mode: ISTIO_MUTUAL
 ---
 {{- end }}
+## Prevent generation of HTTP Server header.
```
bobdonat, line 73: My understanding is we should not be changing the thirdparty helm charts.  Can this be applied in the istio install script instead.

kminder: This is a good point.  For some reason I saw the multi-cluster stuff and thought that was "our stuff".  I'll figure out a different way to apply this.

tests/e2e/examples/helidon/helidon_example_test.go
kminder: Ok, I'll fix.

pfmackin (approved):

PR #1124, opened 2021-05-13
Add simplified diffing utility, with unit tests

dansimone:
# Description

This PR introduces a greatly simplified k8s object diffing function over what we have currently in use, which is duplicated across several projects:
https://github.com/verrazzano/verrazzano-operator/blob/master/pkg/util/diff/diff.go
https://github.com/verrazzano/verrazzano-cluster-operator/blob/master/pkg/util/diff/diff.go
https://github.com/verrazzano/verrazzano-monitoring-operator/blob/master/pkg/util/diff/diff.go

The simplified code here uses github.com/google/go-cmp/cmp with a custom filter to achieve the same functionality as before: compare 2 Golang structs, but ignore any fields that are not specified in the “source” object.

The plan after this is to update the 3 repos referenced above to make use of this utility.

Fixes VZ-2649

# Checklist 

- [x] Addressed the requirement and meets the acceptance criteria
- [x] Does not introduce unrelated or spurious changes
- [x] Does not introduce any unapproved dependency
- [x] Makes sense and it easy to understand, and/or difficult areas of code are clearly documented so that they can be understood

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/gertd/go-pluralize v0.1.7
 	github.com/go-logr/logr v0.1.0
 	github.com/golang/mock v1.4.4
+	github.com/google/go-cmp v0.5.0
 	github.com/gordonklaus/ineffassign v0.0.0-20210104184537-8eed68eb605f
 	github.com/hashicorp/go-retryablehttp v0.6.8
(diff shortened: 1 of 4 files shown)


dansimone:
We decided to put this in https://github.com/verrazzano/pkg instead.

PR #1140, opened 2021-05-17
Tvlatas/periodic tests

agv426:
# Description

This is setting up a periodic test job. The periodic job will run only against master and the intention is that it could be used both to give us additional data points for test runs if desired (ie: same configurations that master runs after a PR merge), but it also can run various configurations which are not being run against master after PR merges.

When the job executes periodically, it will look in object store to get the latest stable commit tested in master. This is saved in object store by the Master runs. So when this wakes up it will be testing against the last clean master commit level (not the tip of master).

This is the first checkin to get the basic support into place with a few test configurations which are not normally tested, there will be more to fine tune things, add more test cases to cover, etc... Once this is stable, we can enable alerting when we have failures.

# Checklist 


ci/JenkinsfilePeriodicTests
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
bhatpmk, line 1: Minor comment : Do we still need to use 2020 ?

bhatpmk (approved):

ci/JenkinsfilePeriodicTests
agv426: not on this one, I'll clean it up in the next pass

bhatpmk: sure

PR #1149, opened 2021-05-20
Fix the issue with Prometheus scraper for managed cluster

bhatpmk:
# Description

This change has a fix to scrape Prometheus metrics from managed cluster, by utilizing the _**/federate**_ endpoint. The match[] argument is set to federate any metric name. The configuration **_honor_labels_** is set to true, to not overwrite any labels exposed by the source server.

Fixes VZ-2448

# Checklist 


michael-cico (dismissed):

platform-operator/controllers/clusters/sync_prometheus.go
```diff
@@ -28,6 +28,11 @@ const (
 scrape_interval: 20s
 scrape_timeout: 15s
 scheme: https
+honor_labels: true
```
pfmackin, line 31: How is conflict avoided when metrics on the master and admin clusters have the same name?

jmaron99: The scraped metrics have an additional label indicating the managed cluster from which they were scraped.

tests/e2e/pkg/common.go
```diff
@@ -255,7 +255,7 @@ func MetricsExist(metricsName, key, value string) bool {
 
 // MetricsExist validates the availability of a given metric in the given cluster
 func MetricsExistInCluster(metricsName, key, value, kubeconfigPath string) bool {
-	metric, err := QueryMetric(metricsName, getPrometheusIngressHost(kubeconfigPath))
+	metric, err := QueryMetric(metricsName, getPrometheusIngressHost(kubeconfigPath), kubeconfigPath)
```
jmaron99, line 258: just a minor comment, but it seems we could keep the arguments to two:  metricName and kubeConfig, and then leverage getPrometheusIngressHost within QueryMetric to obtain the host there

jmaron99 (approved):

tests/e2e/pkg/common.go
bhatpmk: @jmaron99 I have addressed the review comment.

PR #1151, opened 2021-05-20
Fix for api server error on node restarts

lake-of-dreams:
# Description

An issue was reported where the console is unable to fetch data from the api server after k8s node(s) is restarted. The api server proxies the request for `kubernetes.default.svc` in `nginx.conf` and there could be an issue with kube dns (as described [here](https://github.com/kubernetes/kubernetes/issues/66924)) that might prevent the nginx to perform dns resolution of name  `kubernetes.default.svc`  and hence it will fail. Since we already have `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` as environment variables in the api server pod, we don't need to use `kubernetes.default.svc` name.

Fixes VZ-2662

# Checklist 


mgianatagh (approved):

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -182,14 +184,17 @@ data:
             server_name  apiserver;
             root     /opt/nginx/html;
             rewrite_by_lua_file /etc/nginx/conf.lua;
+            set_by_lua $kubernetes_server_host 'return os.getenv("KUBERNETES_SERVICE_HOST")';
```
wmhopkins, line 187: Where do these env variables get set? Are they already in the docker image or the deployment CR for the api-proxy?

wmhopkins (commented):
Lgtm - I did have a question about how/where the KUBERNETES_SERVICE_XXXX variables are set.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: these exist in the container - https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment

PR #1159, opened 2021-05-21
VZ-2552: Fail creating resources placed in non-existent cluster

kminder:
# Description

When a MultiClusterXXX or VerrazzanoProject resource is created with a placement value that does not correspond to an existing registered VerrazzanoManagedCluster, the resource creation should not be allowed.

Fixes VZ-2552: Fail creating resources placed in non-existent cluster.

# Checklist 


application-operator/controllers/webhooks/multicluster_validation.go
```diff
@@ -0,0 +1,64 @@
...
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// isLocalClusterAdminCluster determines if the local cluster is the admin cluster.
+func isLocalClusterAdminCluster(c client.Client) bool {
+	s := core.Secret{}
+	k := client.ObjectKey{Name: "verrazzano-cluster-registration", Namespace: constants.VerrazzanoSystemNamespace}
```
desagar, line 23: I know we discussed this, but I don't think this check actually establishes that we are an admin cluster. This will reliably tell you that we are NOT a managed cluster (either because we are an admin cluster, or because we have not YET started to be a managed cluster).

application-operator/controllers/webhooks/multicluster_validation.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// validateTargetClustersExist determines if all of the target clusters of the project have
+// corresponding managed cluster resources.  The results are only valid when this
+// is executed against an admin cluster.
+func validateTargetClustersExist(c client.Client, p clusters.Placement) error {
+	for _, cluster := range p.Clusters {
+		targetClusterName := cluster.Name
+		// If the target cluster name is local then assume it is valid.
+		if targetClusterName != "local" {
```
desagar, line 38: we should probably have a constant for "local" if we don't already somewhere.

application-operator/controllers/webhooks/multicluster_validation.go
kminder: Fixed based on review discussion.

application-operator/controllers/webhooks/multicluster_validation.go
kminder: Fixed based on review discussion.

desagar (dismissed):

mgianatagh (dismissed):

desagar (approved):

PR #1165, opened 2021-05-21
rancher set privateCA when using CA cert issuer

guoyongzhang:
# Description

When CA is the cert issuer type and the default secret is used, set "privateCA=true" when installing rancher.
If customers provide ca secret in verrrazzano install CR, skip "privateCA=true" since we have no idea whether or not their CA is private.

# Checklist 

### DIFF
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -311,6 +311,10 @@ function install_rancher()
         RANCHER_PATCH_DATA="{\"metadata\":{\"annotations\":{\"kubernetes.io/tls-acme\":\"true\",\"nginx.ingress.kubernetes.io/auth-realm\":\"${DNS_SUFFIX} auth\",\"external-dns.alpha.kubernetes.io/target\":\"verrazzano-ingress.${NAME}.${DNS_SUFFIX}\",\"cert-manager.io/issuer\":null,\"cert-manager.io/issuer-kind\":null,\"external-dns.alpha.kubernetes.io/ttl\":\"60\"}}}"
       elif [ "$CERT_ISSUER_TYPE" == "ca" ]; then
         INGRESS_TLS_SOURCE="secret"
+        if [ $(get_config_value ".certificates.ca.secretName") == "$VERRAZZANO_DEFAULT_SECRET_NAME" ] &&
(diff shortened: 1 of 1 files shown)


fredtibbitts (dismissed):

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -311,6 +311,12 @@ function install_rancher()
         RANCHER_PATCH_DATA="{\"metadata\":{\"annotations\":{\"kubernetes.io/tls-acme\":\"true\",\"nginx.ingress.kubernetes.io/auth-realm\":\"${DNS_SUFFIX} auth\",\"external-dns.alpha.kubernetes.io/target\":\"verrazzano-ingress.${NAME}.${DNS_SUFFIX}\",\"cert-manager.io/issuer\":null,\"cert-manager.io/issuer-kind\":null,\"external-dns.alpha.kubernetes.io/ttl\":\"60\"}}}"
       elif [ "$CERT_ISSUER_TYPE" == "ca" ]; then
         INGRESS_TLS_SOURCE="secret"
+        if [ $(get_config_value ".certificates.ca.secretName") == "$VERRAZZANO_DEFAULT_SECRET_NAME" ] &&
+           [ $(get_config_value ".certificates.ca.clusterResourceNamespace") == "$VERRAZZANO_DEFAULT_SECRET_NAMESPACE" ]; then
```
fredtibbitts, line 315: What happens if this if condition is false? Will the Rancher installation be in a bad state at that point?

fredtibbitts (approved):

platform-operator/scripts/install/2-install-system-components.sh
guoyongzhang: If condition is false, it means customer is providing their own CA cert and it is specified in the installation CR.  In this case, we have no idea if the CA cert they provided is private or public.   We could add a field like `private` in `certificates.ca` section in the CRD.  But, this is for the future.

guoyongzhang: With this PR as is, in the case of customer providing their own CA, if CA is public CA, it will still work.  If CA is private, rancher agent will fail.  But, I can not think of a reason any customer would want to supply their own self-signed CA, instead of using default Verrazzano self-signed CA.

PR #1168, opened 2021-05-25
Explicitly pre-create the Verrazzano system namespace

desagar:
# Description

Explicitly pre-create the Verrazzano system namespace before kicking off install, so that sync_local_registration_secret does not have failures due to missing namespace.
Refactored the verrazzano controller_test to reduce duplication of the similar mocks in several tests.

Fixes VZ-2404

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -135,7 +135,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// since it is needed for the subsequent step to syncLocalRegistration secret.
 	err = r.createVerrazzanoSystemNamespace(ctx, log)
 	if err != nil {
-		log.Errorf("Failed to create or update namespace %v: %v", constants.VerrazzanoSystemNamespace, err)
+		log.Errorf("Failed to create namespace %v: %v", constants.VerrazzanoSystemNamespace, err)
 		return reconcile.Result{}, err
 	}
 
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/verrazzano/verrazzano/application-operator/constants"
 	installv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/installjob"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/uninstalljob"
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -130,6 +131,15 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	}
 
+	// Pre-create the Verrazzano System namespace if it doesn't already exist, before kicking off the install job,
+	// since it is needed for the subsequent step to syncLocalRegistration secret.
+	err = r.createVerrazzanoSystemNamespace(ctx)
+	if err != nil {
+		log.Errorf("Failed to create or update namespace %v: %v", constants.VerrazzanoSystemNamespace, err)
+		return reconcile.Result{}, err
+	}
+	log.Infof("Namespace %v was successfully created", constants.VerrazzanoSystemNamespace)
```
bobdonat, line 141: I think this log message should be moved down into createVerrazzanoSystemNamespace only when the namespace is created

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -632,6 +642,20 @@ func (r *Reconciler) getInternalConfigMap(ctx context.Context, vz *installv1alph
 	return installConfig, err
 }
 
+// createVerrazzanoSystemNamespace creates the verrazzano system namespace if it does not already exist
```
bobdonat, line 645: Is there a need to create the verrazzano-system namespace in the install scripts now that the platform operator is creating this namespace

desagar: I guess there is no longer a need, but that create is gated by a "get". I would prefer to err on the side of defensiveness unless you feel strongly we should just remove it.

platform-operator/controllers/verrazzano/controller.go
desagar: moved it.

platform-operator/controllers/verrazzano/controller.go
bobdonat: that is fine with me

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -130,6 +131,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	// Pre-create the Verrazzano System namespace if it doesn't already exist, before kicking off the install job,
+	// since it is needed for the subsequent step to syncLocalRegistration secret.
+	err = r.createVerrazzanoSystemNamespace(ctx, log)
+	if err != nil {
+		log.Errorf("Failed to create or update namespace %v: %v", constants.VerrazzanoSystemNamespace, err)
```
bobdonat, line 138: Just noticed this.  The log message is saying "... create or update namespace" but this code does not update the namespace

desagar: good catch, fixed. that was leftover from when I initially used createOrUpdate and then decided that was a waste.

pfmackin (approved):

bobdonat (approved):

PR #1170, opened 2021-05-25
Upgrade support of private registry

pfmackin:
Upgrade must be able to upgrade using the image information specified in the new bom JSON file.  This file contains registry, repository, image names, and image tags for all of the Verrazzano components.  Upgrade needs to still pass the helm_config/overrides files to the helm upgrade command, but new additional override image values must also be passed.  Furthermore, there are exception cases, like Keycloak themes, where the override value must be generated at runtime, and those generated values must be included in the override file passed to helm.

Two additional special cases: 
1. The platform operator component are not include in upgrade, so those images are ignored
2. The application operator image:tag in the BOM file is generated and updated during build time.

Here is a summary of the changes:
1. Add bom processing via bom.go
2. Change helm component upgrade to call hooks to append additional helm overrides
3. Move custom Verrazzano component to table based component (see register.go)
4. Generate a new helm override file that contains image overrides along with additional component specific overrides (like Keycloak theme mentioned above).  This file is only generated if the KV array is not empty.
5. Generate an istio override file with global.hub defined
6. Fix updateStatus to return an error if update conflict error returned

# Checklist 


platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -24,11 +31,22 @@ type helmComponent struct {
 	// Upgrade is ignored
 	ignoreNamespaceOverride bool
 
+	// ignoreImageOverrides bool indicates that the image overrides processing should be ignored
+	// This should only be set to true if the component doesn't have images (like istio-base) in
+	// which case it is not in the bom
+	ignoreImageOverrides bool
```
michael-cico, line 37: This is just for the istio-base case at present, right?

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -24,11 +31,22 @@ type helmComponent struct {
...
 	valuesFile string
 
 	// preUpgradeFunc is an optional function to run before upgrading
 	preUpgradeFunc preUpgradeFuncSig
+
+	// appendOverridesFunc is an optional function get additional override values
+	appendOverridesFunc appendOverridesSig
+
+	// resolveNamespaceFunc is an optional function to process the namespace name
+	resolveNamespaceFunc resolveNamespaceSig
```
michael-cico, line 49: Not sure what this is for?

platform-operator/controllers/verrazzano/component/helm_component.go
pfmackin: yes

platform-operator/controllers/verrazzano/component/helm_component.go
pfmackin: needed by verrazzano to preprocess namespace, it is only used there

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -0,0 +1,285 @@
...
+// Each component has one or more subcomponents.
+type BomComponent struct {
+	// The name of the component, for example: Istio
+	Name string `json:"name"`
+
+	// SubComponents is the array of subcomponents in the component
+	SubComponents []BomSubComponent `json:"subcomponents"`
+}
+
+// BomSubComponent contains the bom information for a single helm chart.
```
michael-cico, line 62: This isn't actually 100% true; there are cases where we group images that we don't install into a subcomponent, or that have a different registry or repo than other images in the parent component.  

Generally speaking yeah, you have a correlation between a subcomponent and a chart, but it's not 100%.  The recent CertManager 1.2 chart is a good example, or the Keycloak theme image.  

Not going to change any comments here for this pass, but just clarifying.

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -0,0 +1,285 @@
...
+// Initialize the BomInfo.  Load the Bom from the JSON file and build
+// a map of subcomponents
+func (b *Bom) init(jsonBom string) error {
+	// Convert the json into a to bom
+	if err := json.Unmarshal([]byte(jsonBom), &b.bomDoc); err != nil {
+		return err
+	}
+
+	// Build a map of subcomponents
+	for _, comp := range b.bomDoc.Components {
```
michael-cico, line 147: It's  possible you could have have subcomponents with the same name in different components.  Right now there's no collisions, but it's possible there could be in the future.   But for now flattening is fine.

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -77,11 +105,62 @@ func (h helmComponent) Upgrade(log *zap.SugaredLogger, client clipkg.Client, ns
 		}
 	}
 
+	// Optionally create a second override file.  This will contain both image overrides and any additional
+	// overrides required by a component.
+	// Get image overrides unless opt out
+	var kvs []keyValue
+	if !h.ignoreImageOverrides {
```
michael-cico, line 112: I'm thinking we can look for an empty list of images here and ignore, rather than a flag (next pass).

michael-cico (approved):

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -0,0 +1,285 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package component
+
+import (
+	"encoding/json"
```
mgianatagh, line 7: Do you have goimrport configured to run on your changes?  The sorting order of the imports does not look correct.

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -0,0 +1,285 @@
...
+// BomImage describes a single image used by one of the helm charts.  This structure
+// is needed to render the helm chart manifest, so the image fields are correctly populated
+// in the resulting YAML.  There is no helm standard which specifies the keys for
+// image information used by a template, each product usually has a custom way to do this.
+// The helm keys fields in this structure specify those custom keys.
+type BomImage struct {
+	// ImageName specifies the name of the image tag, such as `nginx-ingress-controller`
+	ImageName string `json:"image"`
+
+	// ImageName specifies the name of the image tag, such as `0.46.0-20210510134749-abc2d2088`
```
mgianatagh, line 85: ImageTag

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// BomImage describes a single image used by one of the helm charts.  This structure
+// is needed to render the helm chart manifest, so the image fields are correctly populated
+// in the resulting YAML.  There is no helm standard which specifies the keys for
+// image information used by a template, each product usually has a custom way to do this.
+// The helm keys fields in this structure specify those custom keys.
+type BomImage struct {
+	// ImageName specifies the name of the image tag, such as `nginx-ingress-controller`
+	ImageName string `json:"image"`
```
mgianatagh, line 83: We usually have the json tags be the same as the struct field (e.g. imageName)

platform-operator/controllers/verrazzano/component/bom.go
pfmackin: we can fix as follow up

platform-operator/controllers/verrazzano/component/bom.go
pfmackin: I agree that is ideal but in this case i don't think tag is clear enough, same for a few other fields.

platform-operator/controllers/verrazzano/component/helm_component.go
pfmackin: I wanted to distinguish between explicitly indicating there are no images vs not finding any because someone messed up the bom.  But we can remove this if you feel strongly about it.

platform-operator/controllers/verrazzano/component/bom.go
mgianatagh: I was pointing out a typo on line 85.  ImageName should be ImageTag.  Was your comment meant to respond to a different issue?

platform-operator/controllers/verrazzano/component/bom.go
michael-cico: I think I'll have to take that up as follow-on work, that would force a change on all the consuming sides as well.  We have another follow-up JIRA for BOM enhancements I can add this to.

platform-operator/controllers/verrazzano/component/bom.go
michael-cico: I'll check, usually the IDE does that for me.

michael-cico: fixed on next push.

platform-operator/controllers/verrazzano/component/helm_component.go
michael-cico: Tracked in VZ-2703

platform-operator/controllers/verrazzano/component/bom.go
michael-cico: Tracked in VZ-2703

platform-operator/controllers/verrazzano/component/bom.go
michael-cico: Fixed on next push

fredtibbitts (approved):

PR #1181, opened 2021-05-27
VZ-2688: generate config for both proxies using same source

wmhopkins:
# Description

This PR creates a new package, pkg/proxy, that contains templates for nginx proxy configuration files, including files containing the lua code that implements the proxy logic. This common code replaces two separate bits of code -- the `oidc_proxy.go` code in the verrazzano-monitoring-operator repo, which is used to generate configmaps for the oauth-proxy in VMI deployments; and the configmap in the verrazzano helm chart template `06-verrazzano-api.yaml`, which is used to configure the api-proxy.

The code has been refactored to:
- support both proxies with the same code
- make the proxy logic more secure and more maintainable

The PR also adds a script to generate the api-proxy helm template from the go code, and updates the platform-operator Makefile to re-generate the api-proxy configmap when the `manifests` target is invoked. Changes to the new templates may result in changes to the generated configmap, which is a checked-in file, and will therefore show up as `modified` in git if changed.

Once this change is in, the VMO code can be updated to consume this config instead of the existing config.

Fixes VZ-2688

# Checklist 


tests/e2e/pkg/kubernetes.go
```diff
@@ -210,17 +210,17 @@ func ListPodsInCluster(namespace string, clientset *kubernetes.Clientset) *corev
...
-	// Get the Kubernetes clientset
-	clientset := GetVMOClientset()
-
-	vmi, err := clientset.VerrazzanoV1().VerrazzanoMonitoringInstances(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-	if err != nil && !errors.IsNotFound(err) {
-		ginkgo.Fail(fmt.Sprintf("Failed to get Verrazzano monitoring instance %s in namespace %s with error: %v", name, namespace, err))
-	}
-	return vmi, err
-}
+//  // GetVerrazzanoMonitoringInstance returns the a Verrazzano monitoring instance in a given namespace for the cluster
```
tbeerbower, line 213: Any reason that we need to hang on to the commented out code?

wmhopkins: No, just left there while I was debugging. Planning to delete before merging.

pkg/proxy/conf_lua_template.go
```diff
@@ -0,0 +1,80 @@
...
+
+// OidcConfLuaFilename defines the conf.lua file name in OIDC proxy ConfigMap
+const OidcConfLuaFilename = "conf.lua"
+
+// OidcConfLuaFileTemplate is the template of conf.lua file in OIDC proxy ConfigMap
+const OidcConfLuaFileTemplate = `|
+    local ingressUri = 'https://'..'{{ .Ingress }}'
+    local callbackPath = "{{ .OidcCallbackPath }}"
+    local logoutCallbackPath = "{{ .OidcLogoutCallbackPath }}"
+
```
shihchang, line 14: Is it going to change VMO to create the oidc-proxy sidecar of vmi components, elasticsearch, kibana, grafana, and prometheus form reusing these?

I guess you can create the configmaps of those vmi oidc-proxy sidecars ahead of vmo. Then in vmo just let those oidc-proxy sidecars use those 'pre'-created configmaps instead of creating the configmaps in vmo. Or do you have better approach to share these in VMO?

pkg/proxy/conf_lua_template.go
```diff
@@ -0,0 +1,80 @@
...
+    local authHeader = ngx.req.get_headers()["authorization"]
+    local token = nil
+    if authHeader then
+{{ if eq .Mode "api-proxy" }}
+        -- console sent access token with k8s api request (not cached)
+        token = auth.handleBearerToken(authHeader)
+{{ else if eq .Mode "oauth-proxy" }}
+        -- vz component calling vmi component using basic auth (cached locally)
+        token = auth.handleBasicAuth(authHeader)
```
shihchang, line 37: Isn't the oauth-proxy able to handle either the bearer type of basic type? It can use the "type" to chose how to validate the auth header - 
```
Authorization: <type> <credentials>
Authorization: Basic|Bearer <basicAuth>|<bearerToken>
```

pkg/proxy/proxy_configmap.go
```diff
@@ -0,0 +1,149 @@
...
+
+package proxy
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// ProxyModeAPI mode in which twhere the proxy accepts only bearer tokens (from console) and impersonates the token's subject to Kubenetes
+const ProxyModeAPI = "api-proxy"
```
tbeerbower, line 12: Looks like this ProxyModeAPI const isn't used.  Was it intended for this line below? ...
```
	if config.Mode == "api-proxy" {
```

pkg/proxy/proxy_configmap.go
```diff
@@ -0,0 +1,149 @@
...
+
+// ProxyModeOauth mode in which the proxy supports both Password Grant and PKCE OIDC flows, and provides authentication/sso for VMI components
+const ProxyModeOauth = "oauth-proxy"
```
tbeerbower, line 15: const ProxyModeOauth not used.

pkg/proxy/proxy_configmap.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+	startup, err := oidcStartup(config)
+	if err != nil {
+		return nil, err
+	}
+
+	var cm map[string]string
+
+	if config.Mode == "api-proxy" {
```
tbeerbower, line 127: use const ProxyModeAPI here

pkg/proxy/proxy_configmap_test.go
```diff
@@ -0,0 +1,91 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testConfigMap",
+			Namespace: "testConfigMapNamespace",
+		},
+		Data: data,
+	}
+	t.Logf("\n%s", configMap)
+}
+
+func TestCreateAPIProxyConfigMap(t *testing.T) {
```
tbeerbower, line 72: It's not clear to me... do the tests in this file actually assert that a config map is created as expected or is this just a utility to generate config map yaml for manual testing?

pkg/proxy/conf_lua_template.go
wmhopkins: No, the pkg/vmo/configmap.go code will be changed to call a new method, GetOidcProxyConfigMapData, in the pkg/proxy package in the verrazzano repo. I will continue to create config maps as usual, but getting the OIDC one from the new package.

pkg/proxy/conf_lua_template.go
wmhopkins: I left it out on the assumption that we don't use it like that. The only Bearer case I'm aware of is when the console sends an access token to the api-proxy. As such, I didn't want to expose an interface that increases our attack surface if we don't actually support that use case.

pkg/proxy/proxy_configmap.go
wmhopkins: Is mostly intended for use by calling code (i.e., v-m-o/pkg/vmo/configmap.go, but you're right that we should probably use it internally as well.

pkg/proxy/proxy_configmap.go
wmhopkins: See comment re: ProxyModeAPI above.

pkg/proxy/proxy_configmap_test.go
wmhopkins: It's a little of both at this point, but eventually these are meant to be unit tests that test that a config map is created as expected.

pkg/proxy/auth_lua_template.go
```diff
@@ -0,0 +1,576 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package proxy
+
+// OidcAuthLuaFilename defines the auth.lua file name in OIDC proxy ConfigMap
+const OidcAuthLuaFilename = "auth.lua"
+
+// OidcAuthLuaFileTemplate is the content of auth.lua file in OIDC proxy ConfigMap
+const OidcAuthLuaFileTemplate = `|
+    local me = {}
```
lake-of-dreams, line 11: I guess I do not understand how this is making the code maintainable by putting lua code inside a go string and then using it to create helm template programatically. If the requirement was to be able to reuse the code from one template, could we have not used something like https://helm.sh/docs/topics/library_charts/ ? Same is also true for adding nginx.conf and then programatically creating helm template.

pkg/proxy/auth_lua_template.go
```diff
@@ -0,0 +1,576 @@
...
+         if not i then
+             i = 0
+         else
+             i = i + 1
+         end
+         return ngx.decode_args(req_uri:sub(i), 0)
+    end
+
+    function me.query(req_uri, name)
+        local i = req_uri:find("&"..name.."=")
```
lake-of-dreams, line 105: this function and the one above should use https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxreqget_uri_args

pkg/proxy/auth_lua_template.go
```diff
@@ -0,0 +1,576 @@
...
+            me.unauthorized("Invalid BasicAuth authorization header")
+        end
+        me.info("Found basic auth credentials in authorization header")
+        local now = ngx.time()
+        local basicAuth = basicCache[basicCred]
+        if basicAuth and (now < basicAuth.expiry) then
+            me.info("Returning cached token")
+            return basicAuth.id_token
+        end
+        local decode = ngx.decode_base64(basicCred)
```
lake-of-dreams, line 194: `decode_base64` returns error too

pkg/proxy/auth_lua_template.go
```diff
@@ -0,0 +1,576 @@
...
+        cookiePairs.refresh_expiry = now + refresh_expires_in - skew - expiryBuffer
+        me.setCookie("authn", cookiePairs, tonumber(tokenRes.refresh_expires_in)-expiryBuffer)
+    end
+
+    function me.setCookie(ckName, cookiePairs, expiresInSec)
+        local expires = ngx.cookie_time(ngx.time() + expiresInSec)
+        local attributes = "; Path=/; Secure; HttpOnly; Expires="..expires
+        local encrypted = me.aes256:encrypt(cjson.encode(cookiePairs))
+        local cookie = base64.encode_base64url(encrypted)
+        ngx.header["Set-Cookie"] = ckName..'='..cookie..attributes
```
lake-of-dreams, line 512: where is this code getting called from, all ngx.header[<header-name>] calls are going to add header twice if they are not being called from `header_filter_by_lua_*` in case a request is being proxied from admin to managed cluster

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
```diff
@@ -0,0 +1,822 @@
...
+
+{{ if eq .Mode "api-proxy" }}
+    -- impersonate the user
+    auth.impersonateKubernetesUser(token)
+{{ else if eq .Mode "oauth-proxy" }}
+    -- set the oidc_user
+    ngx.var.oidc_user = auth.usernameFromIdToken(idToken)
+{{ end }}
+  auth.lua: |
+    local me = {}
```
lake-of-dreams, line 85: Is this code same as the one in auth_lua_template.go? I mean I don't understand why it is being repeated..

lake-of-dreams:
Overall I am finding that it is going to be a bit difficult to maintain this kind of template generation on the fly and changing code in scripts/conf files based on the template values. for example

```lua
if authHeader then
{{ if eq .Mode "api-proxy" }}        
-- console sent access token with k8s api request (not cached)        
token = auth.handleBearerToken(authHeader)
{{ else if eq .Mode "oauth-proxy" }}        
-- vz component calling vmi component using basic auth (cached locally)     
   token = auth.handleBasicAuth(authHeader)
```

instead of instrumenting this code based on template value, why can't this pod just have an environment variable PROXY_MODE and handle the same thing

```lua
if authHeader then
local proxyMode = os.getenv("PROXY_MODE")
if proxyMode == "api-proxy" 
-- console sent access token with k8s api request (not cached)        
token = auth.handleBearerToken(authHeader)
else if proxyMode == "api-proxy"      
-- vz component calling vmi component using basic auth (cached locally)     
   token = auth.handleBasicAuth(authHeader)
```

or we just move common code to separate lua files and then `require`  those where-ever needed. Similarly common code from `nginx.conf` can be separate and used using http://nginx.org/en/docs/ngx_core_module.html#include instead of generating that code using helm template. Helm template itself has https://helm.sh/docs/topics/library_charts/ to support modularisation. Maybe this needs a review on approach from @markxnelson @gilbode..

pkg/proxy/proxy_configmap.go
```diff
@@ -0,0 +1,141 @@
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: remove 2020

pkg/proxy/proxy_configmap.go
```diff
@@ -0,0 +1,141 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package proxy
```
pfmackin, line 4: I expected to see something related to Kubernetes ConfigMap, which this is not.  Would it be better just to call this proxy_config.go

pkg/proxy/proxy_configmap.go
pfmackin: Can you add a comment that the other repo uses this, it is the only case i know of where code from this repo is used by another.  Also a unit test that used those contstants would make them look like they are being used, not sure that really helps though.

tests/e2e/pkg/kubernetes.go
```diff
@@ -377,14 +349,14 @@ func APIExtensionsClientSet() *apixv1beta1client.ApiextensionsV1beta1Client {
...
 //func CertManagerClient() *certclientv1alpha2.CertmanagerV1alpha2Client {
-//	config := GetKubeConfig()
+//     config := GetKubeConfig()
 //
-//	client, err := certclientv1alpha2.NewForConfig(config)
-//	if err != nil {
-//		ginkgo.Fail(fmt.Sprintf("Failed to create cert-manager client: %v", err))
-//	}
+//     client, err := certclientv1alpha2.NewForConfig(config)
+//     if err != nil {
```
lake-of-dreams, line 355: we should probably remove the commented code?

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
```diff
@@ -0,0 +1,986 @@
...
+         local i = req_uri:find("?")
+         if not i then
+             i = 0
+         else
+             i = i + 1
+         end
+         return ngx.decode_args(req_uri:sub(i), 0)
+    end
+
+    function me.query(req_uri, name)
```
lake-of-dreams, line 185: not sure what this method is doing but should be able to do this using `req.get_uri_args()`?

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
```diff
@@ -0,0 +1,986 @@
...
+                me.oidcValidateToken(token)
+                return token
+            else
+                me.unauthorized("Missing token in authorization header")
+            end
+        end
+        return nil
+    end
+
+    local basicCache = {}
```
lake-of-dreams, line 260: do we want this to persist between requests? not sure what is the purpose of this?

tests/e2e/pkg/kubernetes.go
wmhopkins: This has been removed.

pkg/proxy/proxy_configmap.go
wmhopkins: Done. Used the constants variables rather than the raw string constants elsewhere in the file, added a comment to the file to indicate it's primarily used by the VMO code.

pkg/proxy/proxy_configmap.go
wmhopkins: Fixed.

pkg/proxy/proxy_configmap.go
wmhopkins: Done.

pkg/proxy/auth_lua_template.go
wmhopkins: That's because the VMO code in verrazzano-monitoring-operator repo currently creates oauth-proxy deployments dynamically in go code. The existing implementation in the v-m-o repo used Sprintf to substitute values in the constant strings; this code modifies the approach to use go templates so that the same code can also be used as a helm template. We could theoretically change how the VMO creates the configmaps, but changing all that code seemed like it would be a big job that didn't need to be done to achieve the desired result. We could change how VMO works in another iteration if we want.

pkg/proxy/auth_lua_template.go
wmhopkins: Don't disagree, but since this was existing code that I didn't change, I'm planning to leave it as is. We can do incremental code cleanup going forward, but need to get these changes merged before making more changes.

pkg/proxy/auth_lua_template.go
wmhopkins: Added a check for error return there.

pkg/proxy/auth_lua_template.go
wmhopkins: This code is called only in the oauth-proxy case, so won't be called for api-proxy. It's called when we've validated a token and want to save it so the user doesn't have to reauthenticate on the next call, as long as the cookie contents (token) isn't expired. I'll try to catch you later to understand better if this is something that should be done differently.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
wmhopkins: The SoT is the go code. The verrazzano-api-configmap.yaml file is generated from the go code.

wmhopkins:
> Overall I am finding that it is going to be a bit difficult to maintain this kind of template generation on the fly and changing code in scripts/conf files based on the template values. for example

I agree with the high-level concern, and if there's a way to use lua libraries, for example, that might be a good approach. It's not clear to me that the {{ if / else / end }} code is really any better or worse that using Lua code do implement the same logic, though, and ideally there would be few differences. Many of the existing differences are because I've been focused on getting the shared code working rather than minimizing the differences; if we take the time to align, for example, the names/locations where secrets and config maps are mounted for example, between api-proxy and oauth-proxy, a lot of the conditional inclusion code would go away. At this point, I'm just trying to get this merged, but it can definitely be cleaned up a bit even if the overall structure stays the same.

pkg/proxy/proxy_configmap.go
wmhopkins: I had that in there because even though the file is new, some of the code it's based on existed previously. It's derived from code in a file marked copyright 2020, 2021.

pkg/proxy/proxy_configmap.go
wmhopkins: Renamed it.

tests/e2e/pkg/kubernetes.go
wmhopkins: This is not a change from what's currently on master, except for whitespace differences that apparently crept in during a merge from master. I've adjusted the whitespace so it no longer shows up as a difference in my change.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
wmhopkins: Probably so, but leaving as is for now as that was existing code that I did not change (other than moving it around).

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api-configmap.yaml
wmhopkins: Yes. It's a cache of tokens keyed by basic auth credentials. That's bad, but it's intentional (legacy code). I plan to fix, but not in this change.

pkg/proxy/conf_lua_template.go
```diff
@@ -0,0 +1,86 @@
...
+{{ end }}
+        if not token then
+            auth.info("No recognized credentials in authorization header")
+        end
+    else
+{{ if eq .Mode "api-proxy" }}
+        auth.info("No authorization header")
+{{ else if eq .Mode "oauth-proxy" }}
+        if string.find(ngx.var.request_uri, callbackPath) then
```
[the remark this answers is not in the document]
wmhopkins: The link just takes me to the PR itself, so not sure where you think it's duplicated. If you're talking about verrazzano-api-proxy-configmap.yaml, that file is generated from this one.

gilbode (approved):

PR #1182, opened 2021-05-27
JIRA:VZ-2414 allow deletion of namespaces with deployed MC resources

jmaron99:
# Description

- Fixed logic that prevented the removal of the resource finalizer if the associated/wrapped resource was no longer found
- Added integration test to validate namespace with MC resources can be deleted
- Added MC AT test to ensure application can be removed by deleted project and namespace

Fixes VZ-2414

# Checklist 

### DIFF
--- a/application-operator/controllers/clusters/cluster_utils.go
+++ b/application-operator/controllers/clusters/cluster_utils.go
@@ -342,11 +342,11 @@ func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource r
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		return nil
-	}
-	err = c.Delete(ctx, resourceToDelete)
-	if err != nil {
-		return err
+	} else {
+		err = c.Delete(ctx, resourceToDelete)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Deletion succeeded, now we can remove the finalizer
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -454,7 +454,7 @@ pipeline {
                         }
                     }
                 }
-                stage ('Examples Delete Helidon NS') {
+                stage ('Delete Deployed App NS') {
                     steps {
                         script {
                             int clusterCount = params.TOTAL_CLUSTERS.toInteger()
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -464,15 +464,7 @@ pipeline {
                                         # create modified versions of the hello helidon MC example
                                         export NS_SOURCE_DIR=${GO_REPO_PATH}/verrazzano/examples/multicluster/hello-helidon-ns
                                         export HH_SOURCE_DIR=${GO_REPO_PATH}/verrazzano/examples/multicluster/hello-helidon
-                                        mkdir -p \$NS_SOURCE_DIR
-                                        # create project file
-                                        yq eval ".metadata.name"=\"hello-helidon-ns\" \$HH_SOURCE_DIR/verrazzano-project.yaml > \$NS_SOURCE_DIR/verrazzano-project.yaml
-                                        yq -i eval ".spec.template.namespaces[0].metadata.name"=\"hello-helidon-ns\" \$NS_SOURCE_DIR/verrazzano-project.yaml
-                                        # create component file
-                                        yq eval ".spec.template.spec.workload.metadata.namespace"=\"hello-helidon-ns\" \$HH_SOURCE_DIR/mc-hello-helidon-comp.yaml > \$NS_SOURCE_DIR/mc-hello-helidon-comp.yaml
-                                        yq -i eval ".metadata.namespace"=\"hello-helidon-ns\" \$NS_SOURCE_DIR/mc-hello-helidon-comp.yaml
-                                        # create app file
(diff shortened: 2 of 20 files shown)


application-operator/controllers/clusters/cluster_utils.go
```diff
@@ -342,11 +342,11 @@ func DeleteAssociatedResource(ctx context.Context, c client.Client, mcResource r
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		return nil
```
pfmackin, line 345: i think you need this return for the not found case

tests/e2e/multicluster/examples/helidon-ns-ops/helidon_example_delete_ns_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+var _ = ginkgo.AfterEach(func() {
+	// set failed to true if any of the tests has failed
+	failed = failed || ginkgo.CurrentGinkgoTestDescription().Failed
+})
+
+// set the kubeconfig to use the admin cluster kubeconfig and deploy the example resources
+var _ = ginkgo.BeforeSuite(func() {
+	// deploy the VerrazzanoProject
+	err := examples.DeployHelloHelidonProject(adminKubeconfig)
```
mgianatagh, line 36: Is this project used by any other tests?  Asking to make sure there is not possibility of namespace collision between tests.

jmaron99: It is, but it is underplayed in the previous test.  In addition, this is a separate suite/stage to ensure that the before and after operations are unique to each stage.

application-operator/controllers/clusters/cluster_utils.go
jmaron99: Actually I don't.  In this case it means that the underlying resource is no longer found and I want to proceed to removing the finalizer.  This is precisely the change that fixes the issue.

tests/e2e/multicluster/examples/helidon-ns-ops/helidon_example_delete_ns_test.go
mgianatagh: I feel it would be better if those tests could run in parallel.  We are extending the time to run the pipeline.

jmaron99: OK - will address that...

ci/multicluster/Jenkinsfile
```diff
@@ -438,19 +438,51 @@ pipeline {
...
+                                        export CLUSTER_COUNT=$clusterCount
+                                        cd ${GO_REPO_PATH}/verrazzano/tests/e2e
+                                        export DUMP_KUBECONFIG="${KUBECONFIG_DIR}/2/kube_config"
+                                        export DUMP_DIRECTORY="${TEST_DUMP_ROOT}/examples-helidon"
+                                        ginkgo -v -keepGoing --noColor multicluster/examples/helidon/...
+                                    """
+                                }
+                            }
+                        }
+                        stage ('Delete Deployed App NS') {
```
desagar, line 459: why can't this be part of the previous test?

tests/e2e/multicluster/examples/helidon-ns-ops/helidon_example_delete_ns_test.go
```diff
@@ -0,0 +1,139 @@
...
+		// WHEN the multi-cluster example application has been created on admin cluster and placed in managed cluster
+		// THEN expect that the app is deployed to the managed cluster
+		ginkgo.It("Has application placed", func() {
+			gomega.Eventually(func() bool {
+				return examples.VerifyHelloHelidonInCluster(managedKubeconfig, false, true, testProjectName, testNamespace)
+			}, waitTimeout, pollingInterval).Should(gomega.BeTrue())
+		})
+	})
+
+	ginkgo.Context("Delete resources", func() {
```
desagar, line 93: I commented on this in the Jenkinsfile as well - is there any reason we need to duplicate all this? Can't we alter the existing hello helidon example test to verify the deletion of the namespace?

jmaron99: Not easily.  That test undeploys the tests via 'kubectl delete' and validates undeployment.  So what would be required is modifying the cleanup to be in line with the other test sequence blocks, and then redeploy, validate, and Undeploy, which would one again re-introduce the time delay @mgianatagh is looking to minimize.

ci/multicluster/Jenkinsfile
```diff
@@ -438,19 +438,51 @@ pipeline {
...
+                            steps {
+                                script {
+                                    int clusterCount = params.TOTAL_CLUSTERS.toInteger()
+                                    sh """
+                                        # create modified versions of the hello helidon MC example
+                                        export NS_SOURCE_DIR=${GO_REPO_PATH}/verrazzano/examples/multicluster/hello-helidon-ns
+                                        export HH_SOURCE_DIR=${GO_REPO_PATH}/verrazzano/examples/multicluster/hello-helidon
+                                        mkdir -p ${NS_SOURCE_DIR}
+                                        # create project file
+                                        yq eval ".metadata.name"=\"hello-helidon-ns\" ${HH_SOURCE_DIR}/verrazzano-project.yaml > ${NS_SOURCE_DIR}/verrazzano-project.yaml
```
mgianatagh, line 469: I wonder if at some point we should change the examples to use helm charts, and then use helm template to create the deployment files

desagar (approved):

PR #1183, opened 2021-05-27
Fix the Prometheus metrics from the default scraping jobs with network policy

bhatpmk:
# Description

This PR updates the Verrazzano Monitoring Operator image, to include the fix for the NGINX ingress controller. During the course of the testing, it is observed that the Prometheus metrics for the default scraping jobs were broken when the network policy is applied on the cluster.  Ingress policy to allow scraping metrics from Prometheus was missing in 08-thirdparty-networkpolicy.yaml. This change fixed all the scraping jobs, except Istio envoy stats, which will be done in a follow-up change. 

New tests are added to acceptance suite as appropriate, to validate the sample metrics from the default scraping jobs.

Fixes VZ-2351

# Checklist 


bobdonat (dismissed):

tests/e2e/verify-install/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	longPollingInterval = 20 * time.Second
+	longWaitTimeout     = 10 * time.Minute
+)
+
+// Query Prometheus for the sample metrics from the default scraping jobs
+var _ = ginkgo.Describe("Verify default component metrics", func() {
+	isManagedClusterProfile := pkg.IsManagedClusterProfile()
+	if !isManagedClusterProfile {
```
pfmackin, line 22: prometheus runs on managed cluster also, why are we only running this test on admin cluster?

bhatpmk: @pfmackin Prometheus on the managed cluster is there to be scraped by the admin cluster. The prometheus URL is not set in the webui of Keycloak, so is not accessible.

pfmackin (approved):

tests/e2e/verify-install/prometheus/prometheus_test.go
bhatpmk: New tests will be added to ensure that the admin cluster scrapes the metrics of system components from managed cluster(s).

PR #1188, opened 2021-05-28
Fix operator.yaml generation in local builds

fredtibbitts:
# Description

This PR fixes a problem introduced by the new bill of materials-based installation. The developer `make create-test-deploy` scenario resulted in an operator.yaml that when applied, failed to install the Verrazzano Application Operator.

The fix is to add a `--set` argument to the `helm template` command that sets the helm value to override the application operator image. I searched through the makefiles and jenkinsfiles to make sure this only gets applied when building locally, and as best I can tell, that is the case.

# Checklist 

### DIFF
--- a/platform-operator/Makefile
+++ b/platform-operator/Makefile
@@ -162,7 +162,7 @@ integ-test: create-cluster generate-operator-yaml
 generate-operator-yaml:
 	OPERATOR_DIR=$$(dirname ${OPERATOR_YAML}) ; \
 	mkdir -p $${OPERATOR_DIR} ; \
-	env DOCKER_IMAGE=${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG} IMAGE_PULL_SECRETS=${IMAGE_PULL_SECRETS} ../tools/scripts/generate_operator_yaml.sh > ${OPERATOR_YAML}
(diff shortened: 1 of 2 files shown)


platform-operator/Makefile
```diff
@@ -162,7 +162,7 @@ integ-test: create-cluster generate-operator-yaml
 generate-operator-yaml:
 	OPERATOR_DIR=$$(dirname ${OPERATOR_YAML}) ; \
 	mkdir -p $${OPERATOR_DIR} ; \
-	env DOCKER_IMAGE=${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG} IMAGE_PULL_SECRETS=${IMAGE_PULL_SECRETS} ../tools/scripts/generate_operator_yaml.sh > ${OPERATOR_YAML}
+	env DOCKER_IMAGE=${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG} IMAGE_PULL_SECRETS=${IMAGE_PULL_SECRETS} APP_OPERATOR_IMAGE=${VERRAZZANO_APPLICATION_OPERATOR_IMAGE} ../tools/scripts/generate_operator_yaml.sh > ${OPERATOR_YAML}
```
michael-cico, line 165: Isn't VERRAZZANO_APPLICATION_OPERATOR_IMAGE always passed in by `docker-push` at the top-level?  I would think this is always going to set the override?

fredtibbitts: In the Jenkins pipelines the app operator image always seems to be just the image name, without the tag, which is why I added the ":" check so that the override is only set when an image:tag is provided. As best I can tell, none of the automation flows will result in the override being set. I think this is OK.

The generated operator.yaml in my branch build does not have the override.

PR #1189, opened 2021-05-28
Proxy calls to managed cluster api server from admin cluster api server

lake-of-dreams:
# Description

- To access resources in admin or managed cluster - any api client will always access the admin cluster, for managed cluster resources will pass the cluster name in the url param (`?cluster=managed1`)
- The api server will read the vmc resource and associated secret containing the ca cert and the api server address for this managed cluster api server to make the connection and will proxy the requests to managed cluster api server.
- added test that reads verrazzano resource for admin and managed cluster from admin cluster api server
- all tests passed on kind/oci+nip.io/oci_oci dns
- Fixed an issue where the nginx conf was being reloaded even when the admin ca crt in managed cluster was not changing u sing
- Related console PR https://github.com/verrazzano/console/pull/74
Fixes VZ-9999

# Checklist 


platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -50,13 +52,95 @@ data:
...
+
+    local function getK8SResource(resourcePath)
+      local http = require "resty.http"
+      local httpc = http.new()
+      local res, err = httpc:request_uri("https://" .. vzApiHost .. "/" .. vzApiVersion .. resourcePath,{
+          headers = {
+              ["Authorization"] = ngx.req.get_headers()["authorization"],
+          },
+      })
+      if err then
```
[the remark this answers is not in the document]
lake-of-dreams: Thats direction from Will, we always need to send back 401 to client on all errors and just log the actual ones

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -194,7 +317,20 @@ data:
...
+              lua_ssl_verify_depth 2;
+              lua_ssl_trusted_certificate /etc/nginx/upstream.pem;
+              set $kubernetes_server_url "";
+              rewrite_by_lua_file /etc/nginx/conf.lua;
+              proxy_pass $kubernetes_server_url;
+              proxy_ssl_trusted_certificate /etc/nginx/upstream.pem;
+              header_filter_by_lua_block {
+                local h, _ = ngx.req.get_headers()["origin"]
+                if h and h ~= "*" and  h ~= "null" then
+                  ngx.header["Access-Control-Allow-Origin"] = h
```
desagar, line 329: I am not a fan of this because it just bypasses CORS altogether. I think the idea you suggested of having an optional env var for ALLOWED_ORIGINS is a better approach in order to allow local testing. (Until then, for local testing, just editing the config map in the cluster you are using will work).

I don't think this header is required for any other case besides local testing, since the requests to managed cluster api server are no longer required, unless I am missing something.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -133,6 +207,58 @@ data:
...
+    if args.cluster then
+      logJsonMessage(ngx.INFO, "Read vmc resource for " .. args.cluster)
+      local vmc = getVMC(args.cluster)
+      if not(vmc) or not(vmc.status) or not(vmc.status.apiUrl) then
+        ngx.status = 401
+        logJsonMessage(ngx.ERR, "Unable to fetch vmc api url for vmc " .. args.cluster)
+        ngx.exit(ngx.HTTP_UNAUTHORIZED)
+      end
+
+      serverUrl = vmc.status.apiUrl .. "/" .. vzApiVersion
```
desagar, line 222: There should be a comment here indicating that we are using the Prometheus secret to get the ca cert for talking to the managed cluster. Otherwise it is confusing to read since the secret is not named more generally. I knew what you were doing and it even took me a minute to remember and put it together.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -133,6 +207,58 @@ data:
...
+        ngx.exit(ngx.HTTP_UNAUTHORIZED)
+      end
+
+      local startIndex, endIndex = string.find(decodedSecretYaml, "cacrt: |")
+      if startIndex >= 1 and endIndex > startIndex then
+        local sub = string.sub(decodedSecretYaml, endIndex+1)
+        local startIndex, _ = string.find(sub, "-----BEGIN CERTIFICATE-----")
+        local _, endIndex = string.find(sub, "-----END CERTIFICATE-----")
+        if startIndex >= 1 and endIndex > startIndex then
+          write_file("/etc/nginx/upstream.pem", string.sub(sub, startIndex, endIndex))
```
desagar, line 249: This code runs for every request, correct? If multiple requests are coming in at the same time for different clusters (or even the same cluster), won't this upstream.pem file get overwritten? 

I'm thinking that at a minimum, we need to have filenames for individual clusters, and if you want it to be really safe, it may be better to have unique filenames per request.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -194,7 +317,20 @@ data:
             add_header        Cache-Control private;
 
             location / {
-              proxy_pass https://$kubernetes_server_host:$kubernetes_server_port;
+              lua_ssl_verify_depth 2;
+              lua_ssl_trusted_certificate /etc/nginx/upstream.pem;
```
desagar, line 321: From this line here, I see why you had the single file upstream.pem. But that does not seem safe to me since there could be simultaneous requests. Can there be a wildcard and a list of trusted certs?

desagar: Ok I looked it up. If I'm reading this right, that lua_ssl_trusted_certificate can be a directory: 

https://github.com/openresty/lua-nginx-module#lua_ssl_trusted_certificate

`Specifies a file path with trusted CA certificates in the PEM format used to verify the certificate of the SSL/TLS server in the tcpsock:sslhandshake method.`

desagar: I read a bit closer - I see that you are using "a+" which is probably append mode, so you are adding to the same file. _sigh_ sorry, this is complicated so it is taking me a while to follow the logic. 

Then you can ignore my comments about using a single file above. The only concern I would have then is, we may keep appending to that file and it may grow unbounded.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
desagar: Ignore this - I see this is an append. See my later comment below.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: This is based on suggestion from Will that we should allow all `Origin` s apart from  * and null. The CORS will still be there unless the client provides the Origin header. @wmhopkins Can you please suggest if we specify the ALLOWED_ORIGINS parameter or this approach is fine?

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: added

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -133,6 +207,58 @@ data:
...
+      if not(vmc) or not(vmc.status) or not(vmc.status.apiUrl) then
+        ngx.status = 401
+        logJsonMessage(ngx.ERR, "Unable to fetch vmc api url for vmc " .. args.cluster)
+        ngx.exit(ngx.HTTP_UNAUTHORIZED)
+      end
+
+      serverUrl = vmc.status.apiUrl .. "/" .. vzApiVersion
+      if not(vmc.spec) or not(vmc.spec.prometheusSecret) then
+        ngx.status = 401
+        logJsonMessage(ngx.ERR, "Unable to fetch prometheus secret name for vmc " .. args.cluster)
```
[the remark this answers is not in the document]
lake-of-dreams: fixed

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: I have added logic to refresh this file if it gets bigger than 10M in `reload.sh` section of configmap

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
wmhopkins: I think the goal is to honor the origins that have been set by the browser based on the origin of the code that triggered the api call. Note that keycloak sends a list of allowed origins, from the client configuration, that we probably want to add as well.

On the other hand, origins are probably moot for the api-proxy -- it's strictly proxying rest calls to the k8s server, so other origins should be irrelevant. This seems like an issue mostly for the oauth-proxy.

All that said, I'm far from expert on this stuff -- do we have an HTTP / web / brower expert on the team, with whom we could walk through this question?

wmhopkins (approved):

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
desagar: @wmhopkins I have enough knowledge of this area to discuss with you, although I am far from an expert in HTTP

desagar (approved):
Approving with the caveat that I still think we need to fix CORS handling, but can do it in a follow up.

PR #1215, opened 2021-06-03
Fail explicitly if we fail waiting for the platform operator to deplo…

agv426:
…y, have master runs only use PHX, other branches can use PHX or LHR

# Description

From VZ-2738 investigation and other issues we are seeing slow image pulls. The suspicion is that we are getting but with slow ghcr.io pulls (very slow) in LHR.

This change:

1. Have master jobs that pull images for Kind run in PHX only, other branches will continue to use either PHX or LHR.
2. For VZ-2738 specifically, it kept going after the platform operator timed out, having it fail explicitly there now (more obvious failure instead of continuing and failing later)

I didn't try to see if we can also drive this from a parameter as well, we can investigate that separately later, this was intended to be a quick way to get master locked to PHX so we can spin runs there to see how much of that really is slow image pulls from ghcr.io in LHR

# Checklist 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,8 @@ def DOCKER_IMAGE_TAG
 def SKIP_ACCEPTANCE_TESTS = false
 def SUSPECT_LIST = ""
 
+def agentLabel = env.JOB_NAME.contains('master') ? "phxlarge" : "VM.Standard2.8"
+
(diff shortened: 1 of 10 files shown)


bobdonat (approved):
Looks good to me

pfmackin (approved):

PR #1216, opened 2021-06-03
VZ-2737.  Make ca.crt TLS secret field mount optional in the verrazzano-api deployment

michael-cico:
# Description

Fixes behavior change in CM 1.2 by marking the `ca.crt` secret mount optional, as CM no longer creates an empty `ca.crt` field in the cert secret in the case where no CA is provided when the `ClusterIssuer` is created.

Also
* Fix the `platform-operator/scripts/create_managed_cluster_secret.sh` and `tests/e2e/config/scripts/register_managed_cluster.sh` scripts to handle the case where `ca.crt` is not present in the secret
* Fix Multicluster pipeline OCI DNS/ACME configurations
* Remove a duplicate ACME_ENVIRONMENT parameter from the OCI DNS Jenkinsfile
* Add Mutating/Validating webhooks to cluster dump script

Note that there is still an issue with managed cluster registration in the OCI-DNS/ACME configuration, which we will resolve through a subsequent PR.

Fixes VZ-2737

# Checklist 

### DIFF
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -301,7 +301,7 @@ pipeline {
                                     }
                                     steps {
                                         script {
-                                            configureVerrazzanoInstallers(env.INSTALL_CONFIG_FILE_OCIDNS, "./tests/e2e/config/scripts/process_oci_dns_install_yaml.sh", $ACME_ENVIRONMENT)
+                                            configureVerrazzanoInstallers(env.INSTALL_CONFIG_FILE_OCIDNS, "./tests/e2e/config/scripts/process_oci_dns_install_yaml.sh", params.ACME_ENVIRONMENT)
                                         }
                                     }
                                 }
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -128,11 +128,12 @@ pipeline {
         OCI_CLI_KEY_FILE = credentials('oci-api-key')
         OCI_CLI_REGION = "${params.OKE_CLUSTER_REGION}"
         OCI_CLI_SUPPRESS_FILE_PERMISSIONS_WARNING = 'True'
+
         INSTALL_CONFIG_FILE_OCIDNS = "${GO_REPO_PATH}/verrazzano/tests/e2e/config/scripts/install-verrazzano-ocidns.yaml"
         INSTALL_CONFIG_FILE_NIPIO = "${GO_REPO_PATH}/verrazzano/tests/e2e/config/scripts/install-verrazzano-nipio.yaml"
-        OCI_DNS_ZONE_SUFFIX=credentials('oci-dns-zone')
-        OCI_DNS_ZONE_NAME="z${zoneId}." + "${OCI_DNS_ZONE_SUFFIX}"
+        OCI_DNS_ZONE_NAME="z${zoneId}.v8o.io"
         ACME_ENVIRONMENT="${params.ACME_ENVIRONMENT}"
+
         TIMESTAMP = sh(returnStdout: true, script: "date +%Y%m%d%H%M%S").trim()
         SHORT_TIME_STAMP = sh(returnStdout: true, script: "date +%m%d%H%M%S").trim()
         TEST_SCRIPTS_DIR = "${GO_REPO_PATH}/verrazzano/tests/e2e/config/scripts"
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -301,7 +301,7 @@ pipeline {
                                     }
                                     steps {
                                         script {
-                                            configureVerrazzanoInstallers(env.INSTALL_CONFIG_FILE_OCIDNS, "./tests/e2e/config/scripts/process_oci_dns_install_yaml.sh", params.ACME_ENVIRONMENT)
+                                            configureVerrazzanoInstallers(env.INSTALL_CONFIG_FILE_OCIDNS, "./tests/e2e/config/scripts/process_oci_dns_install_yaml.sh", "acme", params.ACME_ENVIRONMENT)
                                         }
                                     }
                                 }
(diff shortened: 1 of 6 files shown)


platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
```diff
@@ -511,6 +511,7 @@ spec:
                   - key: ca.crt
                     path: default-ca-bundle
                 name: system-tls
+                optional: true
```
lake-of-dreams, line 514: this change is of as far as api server is concerned, however we may want to check its impact on [create_managed_cluster_secret.sh](https://github.com/verrazzano/verrazzano/blob/master/platform-operator/scripts/create_managed_cluster_secret.sh)

michael-cico: OK.  @guoyongzhang also had concerns about the managed cluster scenario, but I've been having trouble getting the MC pipeline running with OCI DNS/ACME.  But I will look at this script, thanks.

michael-cico: Good catch, I did need to update this script to handle the `null` case when ca.crt isn't present.  Pushed fix.

ci/multicluster/Jenkinsfile
```diff
@@ -128,11 +128,12 @@ pipeline {
         OCI_CLI_KEY_FILE = credentials('oci-api-key')
         OCI_CLI_REGION = "${params.OKE_CLUSTER_REGION}"
         OCI_CLI_SUPPRESS_FILE_PERMISSIONS_WARNING = 'True'
+
         INSTALL_CONFIG_FILE_OCIDNS = "${GO_REPO_PATH}/verrazzano/tests/e2e/config/scripts/install-verrazzano-ocidns.yaml"
         INSTALL_CONFIG_FILE_NIPIO = "${GO_REPO_PATH}/verrazzano/tests/e2e/config/scripts/install-verrazzano-nipio.yaml"
-        OCI_DNS_ZONE_SUFFIX=credentials('oci-dns-zone')
-        OCI_DNS_ZONE_NAME="z${zoneId}." + "${OCI_DNS_ZONE_SUFFIX}"
+        OCI_DNS_ZONE_NAME="z${zoneId}.v8o.io"
```
desagar, line 134: Do we no longer use the OCI_DNS_ZONE_SUFFIX anywhere? In that case, we should delete it from Jenkins once this is merged.

ci/multicluster/Jenkinsfile
```diff
@@ -128,11 +128,12 @@ pipeline {
...
         ACME_ENVIRONMENT="${params.ACME_ENVIRONMENT}"
```
desagar, line 135: probably can remove this ACME_ENVIRONMENT variable

ci/multicluster/Jenkinsfile
michael-cico: You mean the `oci-dns-zone` credentials?  I'm not sure, I'd have to check with Tony.

desagar (approved):

ci/multicluster/Jenkinsfile
michael-cico: Good catch, I'll fix in the next PR.

ci/multicluster/Jenkinsfile
agv426: Not sure but we can tell from Jenkins itself if it is being referenced by jobs

ci/multicluster/Jenkinsfile
michael-cico: Fixed in next branch to address remaining MC/OCI-DNS issues.

tests/e2e/config/scripts/register_managed_cluster.sh
```diff
@@ -35,7 +35,7 @@ fi
 # create managed cluster prometheus secret yaml on managed
 PROMETHEUS_SECRET_FILE=${MANAGED_CLUSTER_NAME}.yaml
 TLS_SECRET=$(kubectl --kubeconfig ${MANAGED_KUBECONFIG} -n verrazzano-system get secret system-tls -o json | jq -r '.data."ca.crt"')
-if [ ! -z "${TLS_SECRET%%*( )}" ] ; then
+if [ ! -z "${TLS_SECRET%%*( )}" ] && [ "null" != "${TLS_SECRET}" ] ; then
```
lake-of-dreams, line 38: I am not sure if I understand correctly, does this mean that now certificates will be signed by a public CA? If not then how managed cluster and admin cluster are going to communicate over https in absence of ca certificate? it is critical for api server to use a trust mechanism while calling the api server and as well as keycloak..

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
lake-of-dreams: But in that case will it mean that certificate will be signed by public ca? I mean prometheus scraping will require trust between managed cluster prometheus and admin cluster..

michael-cico: When ACME is in play, the certificates are signed by them using their CAs.  When ACME `production` is used, those are signed by a valid public CA.  When ACME `staging` is used, those are signed by their own "staging" CAs, and those bundles have to be loaded by clients.  So we may have to make some adjustments to the product to handle the ACME `staging` cases.

tests/e2e/config/scripts/register_managed_cluster.sh
michael-cico: See my comment on the other thread that's similar to this question.

platform-operator/helm_config/charts/verrazzano/templates/06-verrazzano-api.yaml
michael-cico: There are still problems to fix for the Multi-cluster scenarios with ACME in play, I have a separate JIRA open for the remaining items.  It's possible we might have a few other areas to fix around ACME.

I think (at least based on our current testing) we're good for the self-signed issuer/magic-dns cases

PR #1224, opened 2021-06-04
Display console URLs in the install log, only if the component is installed in the given profile

bhatpmk:
# Description

This PR contains the change to display the console URLs for Grafana, Kibana, Prometheus, Elasticsearch and Verrazzano Console in the install logs, only when the component is deployed. By default, these URLs will no longer be there in the install logs of managed clusters. If the CR for the managed cluster enables one or more of these components, the install log will display the URLs only for those components.

The automated test case included to validate the change, reads the install log set in / derived by the environment variable(s), captured by the CI pipelines. The test is enabled only for the verrazzano and the multi-cluster pipeline.

Fixes VZ-2664

# Checklist 


platform-operator/scripts/install/install-overrides/config.managed-cluster.json
```diff
@@ -4,5 +4,20 @@
...
+  "elasticSearch": {
+    "enabled": false
+  },
+  "grafana": {
+    "enabled": false
+  },
+  "kibana": {
+    "enabled": false
+  },
+  "prometheus": {
```
guoyongzhang, line 17: prometheus is installed in managed cluster

platform-operator/scripts/install/4-install-keycloak.sh
```diff
@@ -767,18 +767,40 @@ rm -rf $TMP_DIR
...
+    consoleout "You will need the credentials to access the preceding user interface. The user interface can be accessed by the username/password."
+  else
+    consoleout "You will need the credentials to access the preceding user interfaces. They are all accessed by the same username/password."
+  fi
+  consoleout "User: verrazzano"
+  consoleout "Password: kubectl get secret --namespace verrazzano-system verrazzano -o jsonpath={.data.password} | base64 --decode; echo"
+  consoleout
+fi
+
 if [ $(is_rancher_enabled) == "true" ]; then
```
guoyongzhang, line 804: Can't we just query ingresses to populate this log?  instead of adding to config-overrides?

mgianatagh (approved):

desagar (approved):

platform-operator/scripts/install/install-overrides/config.managed-cluster.json
bhatpmk: Yes, but the intention is not not show print the Prometheus URL by default.

platform-operator/scripts/install/4-install-keycloak.sh
bhatpmk: Based on the subsequent discussion, we have decided to go with config overrides.

mgianatagh (approved):

desagar (approved):

PR #1239, opened 2021-06-09
fix api server error for Multi cluster due to certificate order in upstream certs file

lake-of-dreams:
# Description

The  lua code in api server maintains a file `/etc/nginx/upstream.pem` in which it appends the CA certificates of local cluster and as well as of managed clusters as and when the cluster is registered as managed/is admin and are required for cross cluster communication. For example the api server in managed cluster would have contents of this file like..

```bash
<certs from in pod /etc/ssl/certs/ca-bundle.crt bundle>
-----BEGIN CERTIFICATE-----
<local cluster CA certificate>
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
<admin cluster CA certificate when cluster is registered>
-----END CERTIFICATE-----
```

The managed cluster would make a call to keycloak in admin cluster using this file and because the admin cluster CA is appended , the call would work.

However it was observed that after cert-manager 1.2 uptake, with the same structure of file, the calls started failing with `certificate signature issue` error. When same was tried using curl, it gave following error

```bash
error:0407008A:rsa routines:RSA_padding_check_PKCS1_type_1:invalid
```

It was also observed that if we just change the order of certificates, it will work
```bash
<certs from in pod /etc/ssl/certs/ca-bundle.crt bundle>
-----BEGIN CERTIFICATE-----
<admin cluster CA certificate when cluster is registered>
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
<local cluster CA certificate>
-----END CERTIFICATE-----
```

So this change was done to assure that certificate ordering is such that admin cluster CA will be before local cluster CA. Also the test to verify API in case of multi cluster is fixed and included in Jenkinsfile to test this.
A separate Jira vz-2803 logged to investigate this different behaviour after cert-manager 1.2.

Fixes VZ-2800

# Checklist 


lake-of-dreams:
The test started to fail after recent [merge](https://github.com/verrazzano/verrazzano/pull/1181) from @wmhopkins, making WIP

lake-of-dreams:
> The test started to fail on rebase after recent [merge](https://github.com/verrazzano/verrazzano/pull/1181) from @wmhopkins, making WIP

Fixed the issue, removed WIP

wmhopkins (approved):
Looks good to me.

PR #1244, opened 2021-06-11
OKE multi-cluster AT Calico installation fix

fredtibbitts:
# Description

This PR fixes a broken build pipeline when running the multi-cluster acceptance tests configured to use OKE with Calico enabled. This configuration is not one that we routinely run and likely has never worked.

## Problem:
When running Calico on OKE clusters, Calico should be installed before the node pool is ready and any system pods (e.g. coredns) have started. If Calico is installed after system pods are running, those pods get into a bad state that then leads to fatal errors. The CI scripts that create the OKE clusters and install Calico were run in such a way that Calico was installed after system pods were already running. This resulted in the Verrazzano installation failing.

## Solution:
To create the clusters, the CI scripts were generating a single Terraform file with the resources for all of the clusters. Calico was then installed afterwards. The first cluster created by Terraform had sufficient time to add the node pool and start pods while the second cluster was being created, all before we had the opportunity to install Calico.

To fix this, I modified the scripts to create separate Terraform files for each cluster, and then use a separate Terraform workspace per cluster for state management. This way, we can apply the Terraform for a single cluster, then install Calico, and repeat until all of the clusters are created.

# Checklist 


tests/e2e/config/scripts/create_oke_multi_cluster.sh
```diff
@@ -51,65 +51,69 @@ if [ -f /tmp/main.tf ]; then
...
+    mkdir -p "${KUBECONFIG_DIR}/$i"
+    oci ce cluster create-kubeconfig --cluster-id ${CLUSTER_OCID} --file "${KUBECONFIG_DIR}/$i/kube_config" --region "${TF_VAR_region}" --token-version 2.0.0
+    export KUBECONFIG="${KUBECONFIG_DIR}/$i/kube_config"
+    # Adding a Service Account Authentication Token to kubeconfig
+    # https://docs.cloud.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengaddingserviceaccttoken.htm
+    ${SCRIPT_DIR}/update_oke_kubeconfig.sh
+
+    # Calico needs to be installed before the node pool and workers are ready, otherwise the kube-system
+    # pods get into a bad state
+    if [ $INSTALL_CALICO == true ] ; then
```
michael-cico, line 98: I think that this is probably still somewhat of a race condition, but that can probably only be fixed by separating out the node pool creation from the cluster creation.  But we probably won't hit that given that the node pools probably take about 10 minutes to become ready.

Probably not much to worry about now though, but something to keep in mind if we start hitting issues again at some point.

michael-cico (approved):

tests/e2e/config/scripts/create_oke_multi_cluster.sh
fredtibbitts: Yes, that's true. We've been using this approach for the single cluster OCI DNS tests for some time and as far as I know we haven't seen an instance where the node pool comes up before Calico is installed. If we do see issues then yes, there are probably a couple of ways we could improve this, including splitting out the node pool creation.

PR #1264, opened 2021-06-16
Re-introduce managed cluster registration fixes with OCI DNS+ACME

michael-cico:
* Fix rancher mgd cluster registration for OCI DNS/ACME scenarios
* Skip ATs when mgd cluster registration fails
* Install rancher with the ACME staging CAs loaded into tls-ca secret with privateCA=true to allow cattle agent to work  in
  managed cluster
* Update create_managed_cluster_secret to include staging CAs if set
* Update cluster registration and mgd cluster secret sync to use staging CAs if available
* Update MC AT mgd cluster registration script to use ACME staging CAs when necessary

# Description

Fixes VZ-9999

# Checklist 


tests/e2e/config/scripts/register_managed_cluster.sh
```diff
@@ -39,9 +39,9 @@ if [ ! -z "${TLS_SECRET%%*( )}" ] && [ "null" != "${TLS_SECRET}" ] ; then
   CA_CERT=$(kubectl --kubeconfig ${MANAGED_KUBECONFIG} -n verrazzano-system get secret system-tls -o json | jq -r '.data."ca.crt"' | base64 --decode)
 else
   # For Let's Encrypt staging, we need to use the ACME staging CAs
-  ACME_SECRET=$(kubectl -n cattle-system get secret tls-ca-additional -o json | jq -r '.data."ca-additional.pem"')
+  ACME_SECRET=$(kubectl --kubeconfig ${ADMIN_KUBECONFIG} -n cattle-system get secret tls-ca-additional -o json | jq -r '.data."ca-additional.pem"')
```
michael-cico, line 42: We should make the same change in the script under platform-operator/scripts...good catch.

michael-cico: Although, I'm wondering if this is supposed to be the secret for the scrape target for the admin to scrape the managed prometheus.  Might be a little more complicated if they aren't both using ACME staging, but in practice it's hard to see why they wouldn't be the same.

bhatpmk: Rancher is installed only for admin cluster, so don't think we need any change. Also the pipeline sets the kube config properly while installing Verrazzano/

michael-cico: For now, and for this situation, this will work.  And I think it's scoped to a particular scenario.  But generally speaking, the mgd cluster could be using anything (self-signed and not staging), in which case this will be wrong.

Since it's a bundle, I'm wondering if maybe we should just glom whatever is there into the secret; the managed cluster's CA if present, and if the staging CA bundle is present, add that.  In practice, it covers all the bases right?

What I don't like about how I initially handled the bundle stuff is that in both cases we're using secrets either generated by CM, or a secret in the Rancher NS to get the staging bundle.  We control the secrets in both cases, I'm thinking we should create one we know we manage, and put any bundles there, and just use that for all scenarios (self-signed and ACME staging).  Does that make sense?  But perhaps as follow-on work from this, I think this needs to go in sooner rather than later, as soon as we think things are working enough.

platform-operator/controllers/clusters/rancher.go
```diff
@@ -362,12 +363,12 @@ func getRancherTLSRootCA(rdr client.Reader) ([]byte, error) {
...
 	nsName := types.NamespacedName{
 		Namespace: rancherNamespace,
-		Name:      rancherTLSSecret}
+		Name:      rancherTLSCaSecret}
 
 	if err := rdr.Get(context.TODO(), nsName, secret); err != nil {
 		return nil, client.IgnoreNotFound(err)
 	}
-	return secret.Data["ca.crt"], nil
+	return secret.Data[tlsCaSecretKey], nil
```
desagar, line 371: was the secret's key wrong? Looks like you changed it from "ca.crt" to "cacerts.pem".

michael-cico: We were changing the secret/key to be consistent between the self-signed case, to load the CA bundle from a well known location regardless of which kinds of certs we were using. 

Most of the code in the product that is trying to load a CA bundle is getting it from the CM-generated certs, which doesn't work for the ACME staging case.  And, I didn't see anything on the CM side of things to make that happen for the ACME case.

PR #1267, opened 2021-06-17
Default metrics traits sometimes not injected into appconfig

tbeerbower:
# Description

Fixes an issue where the default metrics traits do not always get created.  The check that was being done in the webhook metrics trait defaulter relied on getting the workload kind from each component.  It is possible that the app config can exist and the webhook called before the components are created.  In this case the check failed and the default metrics traits were not created.
Since it is not possible to do a reliable check in the webhook, this change checks to see if the metric type is a supported kind in the metrics trait reconciler.  If it is not then the trait is deleted.

#### Tests

- Added unit test for above scenario.  
- Changed order of app config / component in bobs books acceptance test to create the conditions above. 

Fixes VZ-2848

# Checklist 

### DIFF
--- a/application-operator/controllers/metricstrait/metricstrait_controller.go
+++ b/application-operator/controllers/metricstrait/metricstrait_controller.go
@@ -205,7 +205,16 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	if trait.DeletionTimestamp.IsZero() {
-		return r.reconcileTraitCreateOrUpdate(ctx, trait)
+		result, supported, err := r.reconcileTraitCreateOrUpdate(ctx, trait)
+		if err != nil {
+			return result, err
+		}
+		if !supported {
+			// If the workload kind is not supported then delete the trait
+			r.Log.V(1).Info(fmt.Sprintf("deleting trait %s because workload is not supported", trait.Name))
+			err = r.Client.Delete(context.TODO(), trait, &client.DeleteOptions{})
+		}
+		return result, err
 	}
 	return r.reconcileTraitDelete(ctx, trait)
 }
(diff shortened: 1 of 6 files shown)


application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -221,39 +230,40 @@ func (r *Reconciler) reconcileTraitDelete(ctx context.Context, trait *vzapi.Metr
...
 	var traitDefaults *vzapi.MetricsTraitSpec
-	traitDefaults, err = r.fetchTraitDefaults(ctx, workload)
+	var supported bool
+	traitDefaults, supported, err = r.fetchTraitDefaults(ctx, workload)
 	if err != nil {
-		return reconcile.Result{}, err
-	} else if traitDefaults == nil {
-		return reconcile.Result{}, nil
+		return reconcile.Result{}, supported, err
+	} else if traitDefaults == nil || !supported {
```
pfmackin, line 253: please remove this second else leaving the if

application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -205,7 +205,16 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	if trait.DeletionTimestamp.IsZero() {
-		return r.reconcileTraitCreateOrUpdate(ctx, trait)
+		result, supported, err := r.reconcileTraitCreateOrUpdate(ctx, trait)
+		if err != nil {
+			return result, err
+		}
+		if !supported {
+			// If the workload kind is not supported then delete the trait
+			r.Log.V(1).Info(fmt.Sprintf("deleting trait %s because workload is not supported", trait.Name))
+			err = r.Client.Delete(context.TODO(), trait, &client.DeleteOptions{})
```
pfmackin, line 215: so we are deleting the trait but the appconfig still has the trait right?  If so that is a little strange.

tbeerbower: I guess that could happen.  So far I've only ever seen this case for default metrics traits which are not specified in the app config.  But a user could create a component with an unsupported workload type and explicitly add the metric trait for it into the app config.

pfmackin: We are injecting the default traits for all components in our metrics defaulter.  So every component will have a metrics trait in the app config.

application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -221,39 +230,40 @@ func (r *Reconciler) reconcileTraitDelete(ctx context.Context, trait *vzapi.Metr
 }
 
 // reconcileTraitCreateOrUpdate reconciles a metrics trait that is being created or updated.
-func (r *Reconciler) reconcileTraitCreateOrUpdate(ctx context.Context, trait *vzapi.MetricsTrait) (ctrl.Result, error) {
+func (r *Reconciler) reconcileTraitCreateOrUpdate(ctx context.Context, trait *vzapi.MetricsTrait) (ctrl.Result, bool, error) {
 	var err error
```
pfmackin, line 234: This could be broken up where the first section determines if the component is supported then returns immediately if not.  Not required for 1.0

pfmackin (approved):

PR #1273, opened 2021-06-18
Remove Istio multi-cluster support

pfmackin:
Remove Istio multi-cluster support since we don't use or need it and it causes problems if the CoreDNS config map doesn't have the .global section.  Following is some background information:

Pods contain an /etc/resolv.conf file, which are configured with the IP of the CoreDNS DNS name server.  That file also has a set of strings in a search  field, such as svc.cluster.local, cluster.local, etc.  During name resolution on the pod, the resolver appends each search string to the hostname being resolved, then calls CoreDNS to resolve the name.  If none of the names+suffix resolve, then the resolver tries to resolve the original name.  Since the hostname has cluster.local suffix, CoreDNS will delegate the resolution to the Kubernetes CoreDNS plugin, and the name will be resolved.   In the case of multicluster Istio, that /etc/resolv.conf search string set also includes the string "global".  Verrazzano install replaces the CoreDNS configmap to include a .global section, telling CoreDNS to forward all .global requests to the Istio CoreDNS server.   If that .global section is missing, CoreDNS, forwards the .global name  to the default resolver, which returns some IP 46.x.x.x, which is not the correct IP of the service.

At some point, over the course of a week, the CoreDNS configmap got replaced and the .global section got removed.  We don't know what product or code did this, we just know it was a Java client (per the configmap metadata).  It is an open issue to determine what overwrote the configmap.  Regardless, Verrazzano doesn't use the multi-cluster Istio feature, so removing the MC configuration causes the "global" string is removed from the /etc/resolv.conf search path and name resolution works properly.  

In summary, this PR removes multi-cluster Istio config to prevent DNS name resolution problems.  I am leaving the Istio CoreDNS at this time since I am not sure if is used for other purposes.

# Checklist 


platform-operator/helm_config/overrides/istio-values.yaml
```diff
@@ -9,10 +9,8 @@ global:
     s390x: 2
     ppc64le: 2
   # Provides dns resolution for global services
```
bobdonat, line 11: this comment should be removed

pfmackin: fixed

bobdonat (approved):

PR #1274, opened 2021-06-21
Cluster commands for CLI

saketmht:
# Description

- Added following commands for vz cluster:

    - register                            - Registers a managed cluster, takes caSecret and description(optional) as flags 
    - list                                   - Lists all the managed cluster, output options as flags
    - deregister                        - De-registers a managed cluster
    - get-registration-manifest - Fetches the registration manifest to be applied on the manged cluster, output options as flags
    - get                                   - Gets the info on the managed cluster given as argument

- Added tests for each commands and PrintFlags utility class for output options for list and get-registration-manifest commands.
- Minor changes in other files to fix output to the streams being passed.
- Fixed the debug info/warnings

Changes after review:
- Added/edited comments for godoc
- List command now lists all cluster
- Alias 'ls' for list command
- Root command now implements the Kubernetes Interface instead of cluster command so that other commands can use it as well

Additional Changes:
- Change to the multicluster pipeline
                      - Added 'CLI Tests' stage which builds the CLI and runs cluster commands to register a managed cluster
                      - The cluster used for CLI is created in the initial stages itself and is deleted in the post action stage at the end
                      - This cluster is not used for any of the existing stages
                      - Changed the verify_register_test to use the variable for managedcluster name instead of using a fixed name "managed1"

# Checklist 


tools/cli/vz/cmd/cluster/get_registration_manifest.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	//output option was specified
+	if len(*o.PrintFlags.OutputFormat) != 0 {
+		printer, err := o.PrintFlags.ToPrinter()
+		if err != nil {
+			return err
+		}
+		//set group and kind
+		secret.APIVersion = "v1"
```
lake-of-dreams, line 84: why do we have to set it explicitly here? does the secret being returned in line 71 not already have it?

tools/cli/vz/cmd/root.go
```diff
@@ -25,13 +25,13 @@ func NewRootOptions(streams genericclioptions.IOStreams) *RootOptions {
 }
 
 func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
-	o := NewRootOptions(streams)
+	_ = NewRootOptions(streams)
```
lake-of-dreams, line 28: what does `NewRootOptions` return? If we do not use the returned value, why do we call this at all?

tools/cli/vz/cmd/root.go
```diff
@@ -25,13 +25,13 @@ func NewRootOptions(streams genericclioptions.IOStreams) *RootOptions {
...
 	cmd := &cobra.Command{
 		Use:   "vz",
 		Short: "Verrazzano CLI",
 		Long:  "Verrazzano CLI",
 	}
-	o.configFlags.AddFlags(cmd.Flags())
+	//o.configFlags.AddFlags(cmd.Flags())
```
lake-of-dreams, line 34: why have we removed this?

tools/cli/vz/pkg/helpers/kubernetes.go
```diff
@@ -0,0 +1,14 @@
...
+package helpers
+
+import (
+	clientset "github.com/verrazzano/verrazzano/platform-operator/clients/clusters/clientset/versioned"
+	"k8s.io/client-go/kubernetes"
+)
+
+type Kubernetes interface {
+	NewClientSet() (clientset.Interface, error)
+	NewKubernetesClientSet() kubernetes.Interface
```
lake-of-dreams, line 13: it should rather be `NewKubernetesClient` ? I find it a bit confusing to have both `NewClientSet` and `NewKubernetesClientSet` in a interface that is itself named `Kubernetes`

tools/cli/vz/cmd/cluster/get_registration_manifest.go
saketmht: No, the secret being returned has empty strings for APIversion and Kind.

tools/cli/vz/cmd/root.go
saketmht: This line adds the top level config flags used for kubectl (comes from the genericcli library) which is not being used by the vz cli currently.

tools/cli/vz/cmd/root.go
saketmht: Every command has a ***Options struct associated to it which defines all the flags and values it will be using. As of now the root command "vz" itself is not using any flags or fields so the struct is redundant but maybe later on we might want to add fields like "usage" or "examples" or some other flags so I have left it as an option. For now we could remove this line.

tools/cli/vz/pkg/helpers/kubernetes.go
saketmht: Okay, will change that accordingly.

tools/cli/vz/cmd/cluster/list.go
```diff
@@ -20,35 +20,38 @@ type ClusterListOptions struct {
...
 }
 
-func NewCmdClusterList(streams genericclioptions.IOStreams) *cobra.Command {
+func NewCmdClusterList(streams genericclioptions.IOStreams, kubernetesInterface helpers.Kubernetes) *cobra.Command {
 	o := NewClusterListOptions(streams)
 	cmd := &cobra.Command{
 		Use:   "list",
-		Short: "List the clusters",
-		Long:  "List the clusters",
+		Short: "List the managed clusters",
```
markxnelson, line 38: i think it should list all the clusters, not just the managed clusters

tools/cli/vz/cmd/cluster/list.go
```diff
@@ -60,23 +63,67 @@ func listClusters(args []string) error {
...
+		if conditionType.Type == "Ready" {
+			switch conditionType.Status {
+			case "True":
+				return "Ready"
+			case "False":
+				return "Not Ready"
+			case "Unknown":
+				return ""
+			default:
+				panic("shouldn't reach here")
```
markxnelson, line 124: almost certainly should not be a panic

tools/cli/vz/cmd/root.go
markxnelson: remove the `_ =` and just call the function

tools/cli/vz/pkg/helpers/print_flags.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
+	"strings"
+)
+
+type PrintFlags struct {
```
markxnelson, line 14: add godoc to all of these exposed funcs

tests/e2e/pkg/acme_ca_certs.go
```diff
@@ -16,13 +16,10 @@ type ACMEStagingCerts struct {
...
 }
 
-var stagingCerts = ACMEStagingCerts{
-	letsEncryptStagingIntE1CA: loadStagingCA(newSimpleHTTPClient(), letsEncryptStagingIntE1, "E1"),
-	letsEncryptStagingIntR3CA: loadStagingCA(newSimpleHTTPClient(), letsEncryptStagingIntR3, "R3"),
-}
-
 func getACMEStagingCAs() [][]byte {
-	return [][]byte{stagingCerts.letsEncryptStagingIntE1CA, stagingCerts.letsEncryptStagingIntR3CA}
+	letsEncryptStagingIntE1CA := loadStagingCA(newSimpleHTTPClient(), letsEncryptStagingIntE1, "E1")
```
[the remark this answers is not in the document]
saketmht: So I only moved the declaration to the function to avoid the debug info we got every time we ran the CLI. @michael-cico might have an idea what they are.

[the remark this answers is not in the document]
michael-cico: These are the staging CAs for Let's Encrypt/ACME staging certificates.

tools/cli/vz/pkg/helpers/printer.go
```diff
@@ -24,8 +25,9 @@ func PrintTable(headings []string, data [][]string) error {
 		return err
 	}
 
-	fmt.Println(output)
-	return nil
+	//fmt.Println(output)
```
lake-of-dreams, line 28: don't comment, just remove

markxnelson (dismissed):

lake-of-dreams (approved):
This was previously approved by Mark, Re-approving after rebase and tests have passed in all the pipelines

PR #1286, opened 2021-06-23
Tvlatas/dump on success global

agv426:
# Description

This adds support for a global boolean setting (FORCE_DUMP_K8S_CLUSTER_ON_SUCCESS) that will enable the dump cluster on success to be performed on master. The intention from discussions is that we can enable that setting to true when we have instabilities in master where we need to capture dumps even during successful runs from master runs that are normally being run (giving us some recent baselines to compare with the failures).

We normally will have FORCE_DUMP_K8S_CLUSTER_ON_SUCCESS be false as dumping on success adds quite a bit of time to some tests (ie: uninstall tests take 2.5 hours with that enabled).

We can fine tune the jobs, but this adds the main global support.

# Checklist 


jmaron99:
I guess in a subsequent PR we should ensure that we can generate dumps from all tests prior to the tear down of test namespaces etc.

jmaron99 (approved):

agv426:
> I guess in a subsequent PR we should ensure that we can generate dumps from all tests prior to the tear down of test namespaces etc.

I would create a task in the test infra stuff for that.

1. We do not capture the test related dumps on success currently at all, so that would need to be in place to handle that
2. There needs to be a review overall to ensure that tests are all doing that (Not sure all tests handle the pre-cleanup dumping yet, most do though now I think, but that needs review)
3. That may need a different setting to manage that (ie: independent setting), that could add quite a long time dumping after every successful test

Issue #1240, opened 2021-06-10
Verrazzano Keycloak issue


antonionappi88:
Hello,
sorry for the title but the final issue is with Keycloak but I think there are various steps that aren't working. I find a workaround  (at the end) but maybe it would be good to fix docs/deployment.

I'm trying to install on-premise K8s cluster a simple Verrazzano. I don't want to install it with PVC. I don't have support on the cluster and I'd like just use an empty dir. Then following the docs:
```yaml
apiVersion: install.verrazzano.io/v1alpha1
kind: Verrazzano
metadata:
  name: my-verrazzano
spec:
  environmentName: env
  profile: dev
  defaultVolumeSource:
    emptyDir: {} # Use ephemeral storage for dev mode for all Components
  components:
    ingress:
      type: LoadBalancer
```

I added this piece of yaml to the CRD to tell Verrazzano to use emptyDir, as describe [here](https://verrazzano.io/docs/setup/platforms/generic/generic/), storage tab.

```
defaultVolumeSource:
    emptyDir: {} # Use ephemeral storage for dev mode for all Components
```

When it gets to installing mysql, that line is completely ignored and tries to install using as volume a PVC. 
The pod goes in Pending status because of that:

```
vents:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  42s (x10 over 10m)  default-scheduler  0/11 nodes are available: 11 pod has unbound immediate PersistentVolumeClaims.
```

The installation of mysql fails and also of Keycloak but Verrazzano continues to install the other components. Should this be a critical error ? Without keycloak I cannot access any dashboard...

Plus I tried to modify deployment of the Mysql deployment replacing the PVC with an emptydir. It worked and the mysql pod started but Verrazzano installation didn't realize and didn't create the Keycloak. 

The only way to make it working was:

```yaml
apiVersion: install.verrazzano.io/v1alpha1
kind: Verrazzano
metadata:
  name: my-verrazzano
spec:
  environmentName: env
  profile: dev
  defaultVolumeSource:
    emptyDir: {}
  components:
    keycloak:
      mysql:
        volumeSource: 
          emptyDir: {}
    ingress:
      type: LoadBalancer
```

Even though logs are a bit strange. Looks like that tries to install twice mysql ?!
```bash
2021-06-10 10:22:25 UTC] Installing MySQL                                                             [ .... ]
[...]

PR #1287, opened 2021-06-23
Ddrechse/vz2816

ddrechse:
# Description

This fix addresses an issue identified by
https://github.com/verrazzano/verrazzano/issues/1240
Fixes turning off storage for KeyCloak MySQL using dev profile
Fixes overriding the dev profile default of no storage by allocating a pvc for MySQL
Updates Unit Tests
Adds Acceptance Tests
Create a new testdata directory for capturing valid override install yamls

# Checklist 

- [ x] Added or updated integration tests if appropriate
- [x ] Added or updated acceptance tests if appropriate

### DIFF
--- a/platform-operator/scripts/install/4-install-keycloak.sh
+++ b/platform-operator/scripts/install/4-install-keycloak.sh
@@ -75,6 +75,14 @@ function install_mysql {
   build_image_overrides mysql oraclelinux
   HELM_IMAGE_ARGS="${HELM_IMAGE_ARGS} ${image_args}"
 
+  local PROFILE_VALUES_OVERRIDE=""
+  local profile=$(get_install_profile)
+  if [ "$profile" == "dev" ]; then
+    local PROFILE_VALUES_OVERRIDE=" -f ${VZ_CHARTS_DIR}/verrazzano/mysql.${profile}.yaml"
+  fi
+
+  log "PROFILE VALUES OVERRIDE = ${PROFILE_VALUES_OVERRIDE}"
+
   helm upgrade ${chart_name} ${MYSQL_CHART_DIR} \
       --install \
       --namespace ${KEYCLOAK_NS} \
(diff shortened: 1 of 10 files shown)


platform-operator/scripts/install/4-install-keycloak.sh
```diff
@@ -75,6 +75,14 @@ function install_mysql {
   build_image_overrides mysql oraclelinux
   HELM_IMAGE_ARGS="${HELM_IMAGE_ARGS} ${image_args}"
 
+  local PROFILE_VALUES_OVERRIDE=""
+  local profile=$(get_install_profile)
+  if [ "$profile" == "dev" ]; then
```
michael-cico, line 80: I wonder if this can be generalized into a fn and shared with the VZ chart, so the pattern duplication can be minimized.

I also think that the VZ chart version will just always look for a file like `<chart-name>.<profile-name>.yaml`, and not conditionalize on the profile name.  So you could check if the file exists and add the override that way also.

tests/testdata/install-configurations/install-dev-pvc-keycloak-override.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
michael-cico, line 1: These are just for the future ATs right? and to provide override examples?

michael-cico (approved):

platform-operator/scripts/install/4-install-keycloak.sh
ddrechse: we spoke and this approach is fine for now

tests/testdata/install-configurations/install-dev-pvc-keycloak-override.yaml
ddrechse: Yes these valid config overrides are being checked in for future AT runs

PR #1298, opened 2021-06-25
OKE fluentd data volume mount

shihchang:
# Description

When process a Verrazzano install request, platform-operator walks /var/log dir and traverse symbolic links to find out log file location to configure system fluentd with required extra volume mounts

Fixes VZ-2837

# Checklist 


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -740,3 +742,87 @@ func getIngressIP(c client.Client) (string, error) {
 	}
 	return "", fmt.Errorf("Unsupported service type %s for Nginx ingress", string(nginxService.Spec.Type))
 }
+
+func addFluentdExtraVolumeMounts(vz *installv1alpha1.Verrazzano) *installv1alpha1.Verrazzano {
+	varLog := "/var/log/containers/"
+	var files []string
+	filepath.Walk(varLog, func(path string, info os.FileInfo, err error) error {
```
markxnelson, line 749: what happens if /var/log/containers does not exist?

shihchang: It does nothing is /var/log/containers/ does not exist or all the log files are in /var/log/containers/ which fluentd always has the volume mount. If /var/log/containers does not exist, fluentd however will not be able to do anything.

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -2151,3 +2134,107 @@ func Test_isParentDir(t *testing.T) {
...
+			vz: &vzapi.Verrazzano{
+				Spec: vzapi.VerrazzanoSpec{Components: vzapi.ComponentSpec{Fluentd: &vzapi.FluentdComponent{
+					ExtraVolumeMounts: []vzapi.VolumeMount{{
+						Source: "/u0/x",
+					}, {
+						Source: "/u01/",
+					}},
+				}}},
+			},
+			want: []string{"/u0/", "/u01/", "/u0/x"},
```
desagar, line 2204: 2 questions: 
- should /u0 and /u0/x both be in the list per your expectation? Isn't /u0 enough?
- If /u0/x should be in the list, why not /u0/y, since that's also outside /var/log?

shihchang: "/u0/" should be sufficient. but "/u0/x" was in the Verrazzano CR (user explicitly specify it), so we don't want to remove it.

mgianatagh (approved):

PR #1337, opened 2021-07-08
Add build string to version

mbenmclean:
From work on VZ 2916

Adds version string to the bom file. Reworks platform operator version validating code to look at the bom file and and remove references to Chart.yaml. Modifies app operator to allow for upgrade. Reworks unit test code to handle new method of validating versions. Build BOM file version at build at populates it into the BOM file.

# Checklist 


platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -99,7 +88,7 @@ func ValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec
 		}
 	}
 
-	// If any other field has changed from the stored spec return false
+	// If any other field has changed from the stored spec return falsec
```
michael-cico, line 91: Nit, typo in the comment, `falsec`

platform-operator/controllers/verrazzano/component/app_operator.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
michael-cico, line 1: You might need to rebase/merge from master still, this change should be there.

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
```diff
@@ -5,45 +5,37 @@ package v1alpha1
 
 import (
 	"context"
-	"errors"
-	"io/ioutil"
-	"testing"
-
 	"github.com/stretchr/testify/assert"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component"
```
michael-cico, line 9: General testing comment, do we want to add some tests here for the `-dev-<commit-hash>` style versioning?  I know we are implementing that in the build as the second pass of this work, but we can unit test for it now, unless you feel it's ok just testing it at the `semver` level.

platform-operator/controllers/verrazzano/component/bom.go
```diff
@@ -46,6 +46,9 @@ type BomDoc struct {
 	// An example is ghcr.io
 	Registry string `json:"registry"`
 
+	// Version is the verrazzano version corresponding to the build
+	Version string `json:"version"`
```
michael-cico, line 50: I wonder if we should make this an object, instead of just the version string.  For example, 

```
struct VersionMetaData {
   Version String // semver version
   Build String // Commit ID  
}
```

Or something along those lines.  Reason being that we may want to have more complex version information down the road, and with this approach we can always have a commit-ID associated with the build, even if it's a released version that does not have it in the version string (i.e., `v0.18.0`).

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -337,11 +337,13 @@ func (r *Reconciler) createInstallJob(ctx context.Context, log *zap.SugaredLogge
...
 
 	// Set the version in the status.  This will be updated when the starting install condition is updated.
-	chartSemVer, err := installv1alpha1.GetCurrentChartVersion()
-	if err != nil {
-		return err
-	}
-	vz.Status.Version = chartSemVer.ToString()
+	// Not working right now
+
+	//bomSemVer, err := installv1alpha1.GetCurrentBomVersion()
```
michael-cico, line 342: Dead code?  Or do we need to add this back in?  It looks like this is not setting the version in the status field in the VZ object.

platform-operator/internal/util/semver/semver.go
```diff
@@ -103,8 +104,9 @@ func (v *SemVersion) CompareTo(from *SemVersion) int {
 	var result int
 	if result = compareVersion(from.Major, v.Major); result == 0 {
 		if result = compareVersion(from.Minor, v.Minor); result == 0 {
-			result = compareVersion(from.Patch, v.Patch)
-			// Ignore pre-release/buildver fields for now
+			if result = compareVersion(from.Patch, v.Patch); result == 0 {
+				result = compareVersionBuild(from.Build, v.Build)
```
michael-cico, line 108: Do we need a pre-release compare here?  It's probably not necessary now, but if we ever do use it for a value beyond `dev` we'll wonder why it's not working.  A few more unit test cases, but might be worth it to make this complete.

tools/scripts/generate_bom.sh
```diff
@@ -13,28 +13,34 @@ fi
 BOM_FILE=$1
 
 if [ -z "$2" ]; then
-  echo "You must specify the Application Operator Image"
+  echo "You must specify the Version"
   exit 1
 fi
-VERRAZZANO_APPLICATION_OPERATOR_IMAGE=$2
+VERRAZZANO_VERSION=$2
```
michael-cico, line 19: I wonder if we should move the generating of the build version string here, and not be in the Jenkinsfile.  Although it might just be splitting hairs.  My thinking is to try to make the Jenkinsfile simpler, but it's probably not important for this pass.  We can probably do that on the next one when we make the build changes.

platform-operator/Makefile
```diff
@@ -36,6 +36,8 @@ endif
 
 DOCKER_IMAGE_TAG ?= local-$(shell git rev-parse --short HEAD)
 
+VERRAZZANO_MAJOR_MINOR_PATCH = $(shell grep verrazzano-development-version ../.verrazzano-development-version | sed -e 's/verrazzano-development-version=//')
```
michael-cico, line 39: So, is this just scaffolding until we pass in a real version from Jenkins?

platform-operator/controllers/verrazzano/controller.go
mbenmclean: This has been fixed. This is uncommented and the unit testing for this now works

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
mbenmclean: I have moved my changes to semver.go to another branch that I will use for the next step where we are including a commit hash in the version. This PR doesn't contain any code to add a commit hash to the version so while the handling of the commit hashes in the versions shouldn't break anything there is no reason to include it in this PR

platform-operator/apis/verrazzano/v1alpha1/validate.go
mbenmclean: Fixed

platform-operator/controllers/verrazzano/component/bom.go
mbenmclean: This PR doesn't add commit hashes to the version string anymore so probably not necessary to implement that as part of this PR but will definitely look into for the step

tools/scripts/generate_bom.sh
mbenmclean: Ok I have a running list of things to add for the next task. I will add this.

platform-operator/Makefile
mbenmclean: No right now the source of truth for the version is in .verrazzano-development-version. That gets updated manually every time we release. Right now the version is build in the Makefile and passed into generate_bom.sh as a parameter. This can be changed though either for this PR or for the next task.

mbenmclean:
I created a task for follow-on work from this story (https://jira.oraclecorp.com/jira/browse/VZ-2916) and this PR. I put some of the comments that have been made that I think are best handled as part of the next pass in the description for the subsequent task (https://jira.oraclecorp.com/jira/browse/VZ-3024).

platform-operator/Makefile
bobdonat, line 39: Is there a need for this intermediate variable (VERRAZZANO_MAJOR_MINOR_PATCH)

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -7,36 +7,25 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component"
```
bobdonat, line 10: imports are not ordered.  With GoLand and the goimports file watcher the imports are ordered when the file is saved.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_webhook_test.go
```diff
@@ -4,34 +4,23 @@
 package v1alpha1
 
 import (
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component"
```
bobdonat, line 7: same comment about imports

platform-operator/Makefile
mbenmclean: Well I initially had a variable that I stored the short commit hash in and then concatenated that with VERRAZZANO_MAJOR_MINOR_PATCH to create the full verrazzano version. However for this PR I removed the short commit hash so this is still structured from that. I could go back in and clean up the intermediary variable but I will need the intermediary variable for the next step so I would recommend just leaving it for now it won't stay like this forever.

platform-operator/apis/verrazzano/v1alpha1/validate.go
mbenmclean: This is how goimports with Intellij sorted the imports

platform-operator/apis/verrazzano/v1alpha1/verrazzano_webhook_test.go
mbenmclean: This is how goimports with Intellij sorted the imports

platform-operator/Makefile
pfmackin: When do we run generate_bom.sh?  Not sure why we need  .verrazzano-development-version since the BOM is the source of truth for the version.

platform-operator/apis/verrazzano/v1alpha1/testdata/test_bom.json
```diff
@@ -0,0 +1,448 @@
+{
```
pfmackin, line 1: We already have a sample bom used for bom_test.go.  We don't want duplicate test boms for unit tests unless there is good reason for it.  If not, this should go in platform-operator/testdata/bom and the bom_test.go should be modified to use it.

mbenmclean: Ok I can refactor all of the test boms into that locations

platform-operator/apis/verrazzano/v1alpha1/verrazzano_webhook_test.go
bobdonat: I pulled your branch and this is what I get

	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component"
	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/client/fake"

platform-operator/apis/verrazzano/v1alpha1/validate.go
bobdonat: I pulled your branch and when I save the file I get

	"context"
	"errors"
	"fmt"
	"reflect"

	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component"
	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
	"github.com/verrazzano/verrazzano/platform-operator/internal/util/semver"
	"go.uber.org/zap"
	corev1 "k8s.io/api/core/v1"
	k8serrors "k8s.io/apimachinery/pkg/api/errors"
	"sigs.k8s.io/controller-runtime/pkg/client"

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
```diff
@@ -5,45 +5,37 @@ package v1alpha1
...
 
-const validChartYAML = `
-apiVersion: v1
-description: A Helm chart for Verrazzano
-name: verrazzano
-version: 0.7.0
-appVersion: 0.7.0
-`
+// For unit testing
+const testBomFilePath = "../../../controllers//verrazzano/testdata/test_bom.json"
```
bobdonat, line 21: extra / in paths 
//verrazzano instead of /verrazzano

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -118,6 +122,12 @@ func TestSuccessfulInstall(t *testing.T) {
 		Labels:    labels}
 	verrazzanoToUse.Spec.Components.DNS = &vzapi.DNSComponent{External: &vzapi.External{Suffix: "mydomain.com"}}
 
+	// Sample bom file for version validation functions
+	component.SetUnitTestBomFilePath(testBomFilePath)
+	defer func() {
+		component.SetUnitTestBomFilePath("")
```
bobdonat, line 128: Is there a need to set the bomFilePath to an empty string after a test is run.  This seems to done in each test.

mbenmclean: Yes when the platform operator tries to access the bom file it first checks the unit test bom file path. If it is not empty it reads the bom file in that path. If it is an empty string then it read from the default bom file location where the real one is stored. So that is the mechanism we are using for unit testing.

platform-operator/apis/verrazzano/v1alpha1/validate_test.go
mbenmclean: Oh good catch will fix that

platform-operator/Makefile
fredtibbitts: I was also wondering if we could get rid of `.verrazzano-development-version` with these changes. Looks like lots of Jenkinsfile references to that file though. It doesn't have to be part of this PR, but we should change the Jenkinsfiles to fetch the version from the BOM, and then we can remove `.verrazzano-development-version`, and back out the changes to the Makefile and `generate_bom.sh`.

platform-operator/apis/verrazzano/v1alpha1/validate.go
fredtibbitts: @mbenmclean I think your IDE isn't configured correctly. I see the same as Bob. If you run `goimports -d` on these files you'll see the diffs.

mbenmclean: Oh so I was digging through the intellij preferences and the import sorting drop down had gofmt instead of goimports which was the issue. Just changed that I can commit and push those changes

bobdonat: you should have both gofmt and goimports enabled

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -7,36 +7,25 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/yaml"
 )
 
-// For unit test purposes
-var readFileFunction = ioutil.ReadFile
-
-// GetCurrentChartVersion Load the current Chart.yaml into a chartVersion struct
-func GetCurrentChartVersion() (*semver.SemVersion, error) {
```
fredtibbitts, line 28: Now that we're not getting the version from the helm chart, should or can we remove the versions from the Verrazzano helm charts? I see `version` and `appVersion` in each of these charts, and they aren't even in sync. Can we remove these or are they required?

`platform_operator/helm_config/charts/verrazzano/Chart.yaml`
`platform_operator/helm_config/charts/verrazzano-platform-operator/Chart.yaml`
`platform_operator/helm_config/charts/verrazzano-application-operator/Chart.yaml`

@bobdonat @pfmackin thoughts?

platform-operator/Makefile
mbenmclean: Well we're treating .verrazzano-development-version as the source of truth for the major minor patch version. So the bom file gets populated with the version specified in .verrazzano-development-version. Right now we are just changing .verrazzano-development-version manually at each release. So unless were automating incrementing the version it has to be hardcoded somewhere and right now that's in .verrazzano-development-version.  The change this is making is it was also hardcoded in the Chart.yaml files and now that version is obsolete it all comes from the BOM via .verrazzano-development-version

fredtibbitts: I get that, my point was that if we change the Jenkinsfiles to pull the version from the BOM, then we just hardcode the version in the BOM and we can get rid of `.verrazzano-development-version`. The changes to update the placeholder in the BOM become unnecessary.

mbenmclean: Oh I suppose we could do that. I think I personally prefer having it just stored in a hardcoded file by itself and populating it into the bom at build because we have to populate fields in the bom anyways. It feels a little cleaner but yes definitely more moving pieces and more potential to break.

The other thing is down the road we're going to be putting the short commit hash into the version string so we're going to be modifying that field at build regardless

platform-operator/apis/verrazzano/v1alpha1/validate.go
mbenmclean: Yes I think these might be obsolete now.

fredtibbitts (approved):

PR #1340, opened 2021-07-09
VZ2835: Login and Logout commands for CLI

viswanathtadi:
VZ2835: Login and Logout commands

Added implementation for

vz login VerrazzanoAPIURL
vz logout
Respective test files

Helper files
url_gen.go
browser.go
code_gen.go
kube_config.go

Checklist

- Checked that I included or updated copyright and license notices in all files that I altered
- Added or updated unit tests for any new functions I added

- [ ]  Addressed the requirement and meets the acceptance criteria
- [ ]  Does not introduce unrelated or spurious changes


tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+type LoginOptions struct {
```
lake-of-dreams, line 25: all these exported types/functions need to have comments so that godoc is generated

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	// Check if the user is already logged out
+	if helpers.LoggedIn() {
+		fmt.Fprintln(streams.Out, "Already Logged in")
+		return nil
+	}
+
+	var verrazzanoAPIURL string
+
+	// Extract parameters from args
+	for index, element := range args {
```
lake-of-dreams, line 67: why looping entire array when you just need 0th ele?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		if index == 0 {
+			verrazzanoAPIURL = element
+			continue
+		}
+	}
+
+	// Obtain the certificate authority data in the form of byte stream
+	caData, err := extractCAData(kubernetesInterface)
+	if err != nil {
+		fmt.Println("Unable to obtain certificate authority data")
```
lake-of-dreams, line 77: will this be printed on user screen?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Add the verrazzano cluser into config
+	helpers.SetClusterInKubeConfig("verrazzano",
+		verrazzanoAPIURL,
+		caData,
+	)
+
+	// Add the logged-in user with nickname verrazzano
+	helpers.SetUserInKubeConfig("verrazzano",
+		jwtData["access_token"].(string),
```
lake-of-dreams, line 97: what happens if one of these in not in `jwtData` , you should first check if these exists

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		jwtData["refresh_token"].(string),
+		int64(jwtData["expires_in"].(float64))+time.Now().Unix()-10,
```
lake-of-dreams, line 99: why this random logic of `-10` ?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		int64(jwtData["refresh_expires_in"].(float64))+time.Now().Unix()-10,
+	)
+
+	// Add new context with name verrazzano@oldcontext
+	// This context uses verrazzano cluster and logged-in user
+	// We need oldcontext to fall back after logout
+	helpers.SetContextInKubeConfig("verrazzano"+"@"+helpers.GetCurrentContextFromKubeConfig(),
+		"verrazzano",
```
lake-of-dreams, line 107: make `"verrazzano"` a constant rather as is being used in multiple places

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		"verrazzano",
+	)
+
+	// Switch over to new context
+	helpers.SetCurrentContextInKubeConfig("verrazzano" + "@" + helpers.GetCurrentContextFromKubeConfig())
```
lake-of-dreams, line 112: use `fmt.SPrintf` than concatenating

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+
+	fmt.Fprintln(streams.Out, "Login successful!")
+	return nil
+}
+
+var authCode = "" //	Http handle fills this after keycloak authentication
```
lake-of-dreams, line 118: why is this declared globally like this?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		listener.Close()
+	}()
+
+	// Make sure the go routine is running
+	time.Sleep(time.Second)
+
+	// Open the generated keycloak login url in the browser
+	err := helpers.OpenURLInBrowser(loginURL)
+	if err != nil {
```
lake-of-dreams, line 152: return error rather than continuing

tools/cli/vz/cmd/login/login.go
viswanathtadi: Previously I also implemented direct login which needed additional parameters.
I deleted that approach afterwards.
I will extract the first one directly.

tools/cli/vz/cmd/login/login.go
viswanathtadi: I assumed that the command request will be sent to vz api in the next 10 seconds.
The token may be valid at current moment but may expire when the request is sent to vz api.

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	kclientset := kubernetesInterface.NewClientSet()
+	secret, err := kclientset.CoreV1().Secrets("verrazzano-system").Get(context.Background(),
+		"system-tls",
+		metav1.GetOptions{},
+	)
+
+	if err != nil {
+		return cert, err
+	}
+	cert = (*secret).Data["ca.crt"]
```
lake-of-dreams, line 192: what if `ca.crt` does not exist in secret data?

tools/cli/vz/cmd/login/login.go
viswanathtadi: The http handle function cant return anything.
So I used the global variable.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	return cert, nil
+}
+
+// Handler function for http server
+// The server page's html,js,etc code is embedded here.
+func handle(w http.ResponseWriter, r *http.Request) {
+	u, _ := url.Parse(r.URL.String())
```
lake-of-dreams, line 199: do not suppress errors..

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	authCode = m["code"][0]
+	fmt.Fprintln(w, "<p>You can close this tab now</p>")
+}
+
+// Fetch an available port
+// Return in the form of listener
+func getFreePort() net.Listener {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		panic(err)
```
lake-of-dreams, line 211: we never `panic` , return error and handle it gracefully in calling code

tools/cli/vz/cmd/login/login.go
viswanathtadi: I ll implement checks before handling

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay will change that.

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
viswanathtadi: Will implement checks

tools/cli/vz/cmd/login/login.go
viswanathtadi: yes

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// The response is going to be filled in this
+	var jsonData map[string]interface{}
+
+	// Get the keycloak url to obtain tokens
+	tokenURL := helpers.GenerateKeycloakTokenURL()
+
+	// Create new http POST request to obtain token as response
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caData)
```
lake-of-dreams, line 254: in cases such as for oci dns, ca.crt will be empty

viswanathtadi: Should I use the insecure option for post requests in those cases?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: caCertPool,
+			},
+		},
+	}
+
+	request, err := http.NewRequest(http.MethodPost, tokenURL, strings.NewReader(grantParams.Encode()))
+	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
lake-of-dreams, line 265: this line should come after checking error condition

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		fmt.Println("Unable to create POST request")
```
lake-of-dreams, line 267: print actual error

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+		return jsonData, err
+	}
+	responseBody, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		fmt.Println("Unable to read the response body")
+		return jsonData, err
+	}
+
+	// Convert the response into a map
+	json.Unmarshal([]byte(responseBody), &jsonData)
```
lake-of-dreams, line 284: does this not return error?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+
+	return jsonData, nil
+}
+
+func RefreshToken() error {
```
lake-of-dreams, line 289: Add proper documentation to these functions

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Nothing to do when the user is not logged in
+	if !helpers.LoggedIn() {
+		return nil
+	}
+
+	accessTokenExpTime, refreshTokenExpTime, _, refreshToken := helpers.GetAuthDetails()
```
lake-of-dreams, line 296: instead of returning 4 values from a function return a struct

tools/cli/vz/cmd/login/login.go
viswanathtadi: It returns an error.
Apparently I suppressed it.
Will change that

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	grantParams.Set("redirect_uri", "http://localhost:8080")
+	grantParams.Set("scope", "openid")
+
+	caData, err := base64.StdEncoding.DecodeString(string(helpers.GetCAData()))
+	if err != nil {
+		return err
+	}
+
+	// Execute the request
+	jwtData, err := executeRequestForJWT(grantParams, []byte(caData))
```
lake-of-dreams, line 322: there will be cases when there will not be any `caData` such that for well known CAs

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
viswanathtadi: okay

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		fmt.Println("Request failed")
```
lake-of-dreams, line 324: log actual error

tools/cli/vz/cmd/login/login_test.go
```diff
@@ -0,0 +1,288 @@
...
+func (o *TestKubernetes) NewProjectClientSet() (projectclientset.Interface, error) {
+	return o.fakeProjectClient, nil
+}
+
+func (o *TestKubernetes) NewClientSet() kubernetes.Interface {
+	return o.fakek8sClient
+}
+
+func authHandle(w http.ResponseWriter, r *http.Request) {
+	u, _ := url.Parse(r.URL.String())
```
lake-of-dreams, line 93: why suppress errors?

tools/cli/vz/pkg/helpers/auth_status.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package helpers
+
+import (
+	"strings"
+)
+
+func LoggedIn() bool {
```
lake-of-dreams, line 10: document all these exported funcs

tools/cli/vz/pkg/helpers/auth_status.go
```diff
@@ -0,0 +1,31 @@
...
+	return strings.Split(GetCurrentContextFromKubeConfig(), "@")[0] == "verrazzano"
+}
+
+func LoggedOut() bool {
+	return strings.Split(GetCurrentContextFromKubeConfig(), "@")[0] != "verrazzano"
+}
+
+func RemoveAllAuthData() {
+	// Remove the cluster with nickname verrazzano
+	RemoveClusterFromKubeConfig("verrazzano")
```
lake-of-dreams, line 20: and none of these returns any error?

viswanathtadi: They actually panic inside.

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+
+	if len(kubeConfigEnvVar) > 0 {
+		// Find using environment variables
+		kubeConfig = kubeConfigEnvVar
+	} else if home := homedir.HomeDir(); home != "" {
+		// Find in the ~/.kube/ directory
+		kubeConfig = filepath.Join(home, ".kube", "config")
+	} else {
+		// give up
+		panic("Unable to find kubeconfig")
```
lake-of-dreams, line 29: handle this error gracefully rather than panic

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+	kubeConfig := ReadKubeConfig()
+	pos := -1
+	for i := 0; i < len(kubeConfig["contexts"].([]interface{})); i++ {
+		if kubeConfig["contexts"].([]interface{})[i].(map[string]interface{})["name"] == name {
+			pos = i
+			break
+		}
+	}
+	if pos != -1 {
+		kubeConfig["contexts"] = append(kubeConfig["contexts"].([]interface{})[:pos], kubeConfig["contexts"].([]interface{})[pos+1:]...)
```
lake-of-dreams, line 44: I am sure there will be some way to write this code in simpler way

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+	kubeConfig := ReadKubeConfig()
+	kubeConfig["current-context"] = name
+	WriteToKubeConfig(kubeConfig)
+}
+
+func SetClusterInKubeConfig(name string, serverURL string, caData []byte) {
+	RemoveClusterFromKubeConfig(name)
+	kubeConfig := ReadKubeConfig()
+	currentCluster := make(map[string]interface{})
+	currentCluster["name"] = name
```
lake-of-dreams, line 89: don't we already have structs defined for things that are in kubeconfig, if not we should create structs instead of using maps

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+	kubeConfigLoc := GetKubeConfigLocation()
+	byteSliceKubeconfig, err := ioutil.ReadFile(kubeConfigLoc)
+	if err != nil {
+		panic("Unable to read kubeconfig")
+	}
+	kubeConfig := make(map[string]interface{})
+	// Load the default kubeconfig's configuration into clientcmdapi object
+	err = yaml.Unmarshal(byteSliceKubeconfig, &kubeConfig)
+	if err != nil {
+		panic("Unable to unmarshall kubeconfig")
```
lake-of-dreams, line 137: no panic

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return kubeConfig
+}
+
+func WriteToKubeConfig(kubeConfig map[string]interface{}) {
+	// Write the new configuration into the default kubeconfig file
+	kubeConfigLoc := GetKubeConfigLocation()
+	byteSliceKubeConfig, err := yaml.Marshal(kubeConfig)
+	if err != nil {
+		panic("Unable to marshall kubeconfig")
```
lake-of-dreams, line 147: remove all these `panic` and handle the error gracefully

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,190 @@
...
+		if kubeConfig["users"].([]interface{})[i].(map[string]interface{})["name"] == "verrazzano" {
+			pos = i
+			break
+		}
+	}
+	if pos == -1 {
+		panic("No user with nickname verrazzano")
+	}
+	currentUserInfo := kubeConfig["users"].([]interface{})[pos].(map[string]interface{})["user"].(map[string]interface{})
+	return int64(currentUserInfo["accessTokenExpTime"].(float64)), int64(currentUserInfo["refreshTokenExpTime"].(float64)), currentUserInfo["token"].(string), currentUserInfo["refreshToken"].(string)
```
lake-of-dreams, line 173: simplify this

tools/cli/vz/pkg/helpers/kube_config.go
viswanathtadi: We are deleting a element from a slice.
I felt this was the easiest thing to do.
Another approach is loop through all elements and shift them left.

tools/cli/vz/pkg/helpers/kube_config.go
viswanathtadi: I ll use a struct to return these 4 values

tools/cli/vz/pkg/helpers/url_gen.go
```diff
@@ -0,0 +1,83 @@
...
+	return realmName
+}
+
+// Generates the keycloak api url to login
+// Return string of the form `https://keycloak.xyz.io:123/auth/realms/verrazzano-system/protocol/openid-connect/auth?redirect_uri=abc&state=xyz...`
+func GenerateKeycloakAPIURL(codeChallenge string, redirectURI string) string {
+	urlParams := map[string]string{
+		"client_id":             GetClientID(),
+		"response_type":         "code",
+		"state":                 "fj8o3n7bdy1op5",
```
lake-of-dreams, line 63: why this random string?

viswanathtadi: Supposed to be random.
Have to change that

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
lake-of-dreams, line 1: remove 2020 from copyright

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+
+// Creates a LoginOptions struct to run the login command
+func NewLoginOptions(streams genericclioptions.IOStreams) *LoginOptions {
+	return &LoginOptions{
+		configFlags: genericclioptions.NewConfigFlags(true),
+		IOStreams:   streams,
+	}
+}
+
+// root.go returns this function for login
```
lake-of-dreams, line 41: `root.go returns this function for login` - remove this line

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+			return nil
+		},
+	}
+	o.configFlags.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func login(streams genericclioptions.IOStreams, args []string, kubernetesInterface helpers.Kubernetes) error {
+	// Check if the user is already logged out
+	loggedIn, err := helpers.LoggedIn()
```
lake-of-dreams, line 63: Change `LoggedIn` to `IsLoggedIn` or something better

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	if err != nil {
+		return err
+	}
+	if loggedIn {
+		_, err := fmt.Fprintln(streams.Out, "Already Logged in")
```
lake-of-dreams, line 68: why read a value `_` when you are not using it, also why does the next line return the `err` always?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+		return err
+	}
+
+	// Extract parameters from args
+	verrazzanoAPIURL := args[0]
+
+	// Obtain the certificate authority data in the form of byte stream
+	caData, err := extractCAData(kubernetesInterface)
+	if err != nil {
+		// Assuming well-known authority and proceeding
```
lake-of-dreams, line 78: it is not wise to ignore all the errors retuned from this function call, handle the cases with empty ca data explicitly rather than relying in returned errors

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	err = helpers.SetCurrentContextInKubeConfig(fmt.Sprintf("%v@%v", helpers.Verrazzano, currentContext))
+	if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fprintln(streams.Out, "Login successful!")
+	return err
+}
+
+var authCode = ""          //	Http handle fills this after keycloak authentication
```
lake-of-dreams, line 151: why can't you defined this in the function below?

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	cert, ok := (*secret).Data["ca.crt"]
+	if !ok {
+		return cert, errors.New("ca.crt not found")
+	}
+	return cert, nil
+}
+
+// Handler function for http server
+// The server page's html,js,etc code is embedded here.
+func handle(w http.ResponseWriter, r *http.Request) {
```
lake-of-dreams, line 249: Give some meaningful name to this function than a general name `handle`

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	u, serverErr := url.Parse(r.URL.String())
```
lake-of-dreams, line 250: can't you use a channel to populate server error than global variable

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	return jwtData, nil
+}
+
+// Creates and executes the POST request
+// Returns the key-value pairs obtained from server in the form of a map
+func executeRequestForJWT(grantParams url.Values, caData []byte) (map[string]interface{}, error) {
+
+	// The response is going to be filled in this
```
lake-of-dreams, line 301: No need of obvious comments like these

tools/cli/vz/cmd/login/login.go
viswanathtadi: There is no way to get second return value without reading the first return value.
When the user tries lo login when he is already logged in, the login function simply ignores.
So If the fprintln is good, everything is good and vice versa.
So it makes sense to return the err always.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	request, err := http.NewRequest(http.MethodPost, tokenURL, strings.NewReader(grantParams.Encode()))
+	if err != nil {
+		return jwtData, err
+	}
+	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	// Send the request and get response
+	response, err := client.Do(request)
```
lake-of-dreams, line 333: there should be a retry here with timeouts

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	// Execute the request
+	jwtData, err := executeRequestForJWT(grantParams, []byte(caData))
+	if err != nil {
+		return err
+	}
+
+	// Set new auth details in kubeconfig
+	accessToken, ok := jwtData["access_token"]
```
lake-of-dreams, line 402: the lines below are repeated on line 99, move these to a func instead of repetition

tools/cli/vz/pkg/helpers/auth_status.go
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package helpers
+
+import (
+	"strings"
+)
+
+// To be used as nick name for verrazzano related clusters,contexts,users,etc in kubeconfig
+const Verrazzano = "verrazzano"
```
lake-of-dreams, line 11: change this constant name to something more meaningful than Verrazzano

tools/cli/vz/pkg/helpers/browser.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
lake-of-dreams, line 1: remove 2020

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.16.0
 	golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5
+	gopkg.in/yaml.v2 v2.3.0
```
desagar, line 26: is this PLS approved for use? @lake-of-dreams

tools/cli/vz/cmd/logout/logout.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/spf13/cobra"
+	"github.com/verrazzano/verrazzano/tools/cli/vz/pkg/helpers"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// Struct to store Logout-command related data. eg.flags,streams,args..
+type LogoutOptions struct {
+	configFlags *genericclioptions.ConfigFlags
+	args        []string
+	genericclioptions.IOStreams
```
desagar, line 17: it seems that because of an accident of the way they are declared, configFlags and args are package private, whereas the IOStreams field is public (since capitalized = public in Golang). I would suggest that unless you need to make IOStreams field of LogoutOptions available to other packages, make that private by giving it an explicit lower-case name, the way you have done for the configFlags field.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Struct to store Login-command related data. eg.flags,streams,args..
+type LoginOptions struct {
+	configFlags *genericclioptions.ConfigFlags
+	args        []string
+	genericclioptions.IOStreams
```
desagar, line 29: see my comment below on LogoutOptions struct. Same applies here too.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// Obtain the certificate authority data
+// certificate authority data is stored as a secret named system-tls in verrazzano-system namespace
+// Use client cmd to extract the secret
+func extractCAData(kubernetesInterface helpers.Kubernetes) ([]byte, error) {
+	var cert []byte
+
+	kclientset := kubernetesInterface.NewClientSet()
+	secret, err := kclientset.CoreV1().Secrets("verrazzano-system").Get(context.Background(),
```
desagar, line 231: The system-tls secret will not exist if we use a well-known CA. If the error is a "not found" error, that is not considered an error condition.

go.mod
viswanathtadi: The IDE added it when I was trying something else.
Removed it now.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+		fmt.Fprintln(streams.Out, "Already Logged in")
+		return nil
+	}
+
+	// Extract parameters from args
+	verrazzanoAPIURL := args[0]
+
+	// Obtain the certificate authority data in the form of byte stream
+	caData, err := extractCAData(kubernetesInterface)
+	if err != nil && err.Error() == "ca.crt not found" {
```
desagar, line 75: We need to handle well-known CA here. I made a comment inside the extractCAData function as well. There are cases where it is ok for ca.crt to be absent. 

To test that use case, you'll need access to an OKE cluster using a "real" CA instead of self-signed certificates. If Ravi is successful to set up a cluster in the new tenancy with OCI DNS-based install, you might be able to use that to test.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+	// Make sure the go routine is running
+	time.Sleep(time.Second)
+
+	// Open the generated keycloak login url in the browser
+	err = helpers.OpenURLInBrowser(loginURL)
+	if err != nil {
+		return jwtData, err
+	}
+
+	urlParamChannel := make(chan keycloakRedirectionURLParams)
```
desagar, line 169: keycloakRedirectionURLParams is a type, correct? It should be capitalized by convention unless you have a reason not to

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+	go func() {
+		http.HandleFunc("/",
+			handleKeycloakRedirection(urlParamChannel),
+		)
+		http.Serve(listener,
+			nil,
+		)
+	}()
+
+	keycloakRedirectionURLParams := <-urlParamChannel
```
desagar, line 181: this variable has the same name as the keycloakRedirectionURLParams datatype - change the variable name to something other than the type name.

tools/cli/vz/cmd/login/login.go
```diff
@@ -0,0 +1,434 @@
...
+	// Open the generated keycloak login url in the browser
+	err = helpers.OpenURLInBrowser(loginURL)
+	if err != nil {
+		return jwtData, err
+	}
+
+	urlParamChannel := make(chan keycloakRedirectionURLParams)
+
+	// Set the handle function and start the http server
+	go func() {
```
desagar, line 172: Can you explain how this works? It might be worth actually explaining it in a comment in the code as well.

Keycloak will send a redirect to your generated redirectURI only if that's in the approved list in Keycloak settings, correct? So how do you get it to send a redirect to this URI?

viswanathtadi: We add http://localhost to keycloak trusted redirect uris.
We send our redirect-uri when asking for authcode to the keycloak api.

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	pos := -1
+	for i := 0; i < len(kubeConfig.Clusters); i++ {
+		if kubeConfig.Clusters[i].Name == name {
+			pos = i
+			break
+		}
+	}
+	if pos == -1 {
+		return caData, errors.New("Unable to find cluster with nick name verrazzano")
```
desagar, line 333: nickname = one word :-)

tools/cli/vz/cmd/login/login.go
viswanathtadi: I dont want to export the type.
I am guessing if I make it capitalised, it would export it

tools/cli/vz/cmd/login/login.go
viswanathtadi: So if I cannot find system-tls, I can assume well known authority

tools/cli/vz/cmd/logout/logout_test.go
```diff
@@ -0,0 +1,165 @@
...
+	fakeVerrazzanoAPIURL := "verrazzano.fake.nip.io/12345"
+	fakeCAData := []byte("LS0tCmFwaVZlcnNpb246IHYxCmRhdGE6CiAgYWRtaW4ta3ViZWNvbmZpZzogWTJ4MWMzUmxjbk02Q2kwZ1kyeDFjM1JsY2pvS0lDQWdJR05sY25ScFptbGpZWFJsTFdGMWRHaHZjbWwwZVMxa1lYUmhPaUJNVXpCMFRGTXhRMUpWWkVwVWFVSkVVbFpL")
+	fakeAccessToken := "fhuiewhfbudsefbiewbfewofnhoewnfoiewhfouewhbfgonewoifnewohfgoewnfgouewbugoewhfgojhew"
+	fakeRefreshToken := "fhuiewhfbudsefbiewbfewofnhoewnfoiewhfouewhbfgonewoifnewohfgoewnfgouewbugoewhfgojhew"
+
+	// Set environment variable for kubeconfig
+	err = os.Setenv("KUBECONFIG", currentDirectory+"/fakekubeconfig")
+	asserts.NoError(err)
+
+	err = helpers.SetClusterInKubeConfig(helpers.KubeConfigKeywordVerrazzano,
```
desagar, line 73: looks like you are taking the KUBECONFIG specified in env var, copying it and editing the copy to prepare for the logout test. 

For a test to work reliably, it is better to have a fixed input that already looks like how you want. e.g. you would have a test kubeconfig file checked in which looks like a "logged in" file (with fake cert data etc of course)

tools/cli/vz/pkg/helpers/kube_config.go
```diff
@@ -0,0 +1,338 @@
...
+	TokenInfo TokenInfo `json:"tokenInfo"`
+}
+
+type Config struct {
+	*clientcmdapi.Config `json:",inline"`
+	KeycloakTokenInfos   []NamedKeycloakTokenInfo `json:"keycloakTokenInfo,omitempty"`
+}
+
+// Helper function to obtain the default kubeConfig location
+func GetKubeConfigLocation() (string, error) {
```
desagar, line 41: The functions in this helper package look complex enough that it would be good to have unit tests for these functions.

tools/cli/vz/cmd/logout/logout_test.go
viswanathtadi: I thought we want to test on the current platform.
The way I wrote the tests will ensure that if the tests pass with fake-kubeconfig, they will pass with the original-kubeconfig.
That is the reason why I cloned the original kubeconfig and modified it so that it looks like it is logged in.
I can change if we have different expectations

tools/cli/vz/pkg/helpers/kube_config.go
viswanathtadi: Added new test files.

tools/cli/vz/cmd/login/login.go
viswanathtadi: Handling the "not found" error here.
If the error is "secret system-tls not found" , I am assuming well known CA
For any other error, exiting

tools/cli/vz/cmd/login/login.go
viswanathtadi: Added new comments above the authFlowLogin function explaining the wholw process

tools/cli/vz/cmd/logout/logout.go
viswanathtadi: I get that but they are declared in a imported package genericclioptions.

tools/cli/vz/cmd/logout/logout_test.go
viswanathtadi: And I am not assuming that kubeconfig will present in KUBECONFIG env var.
I clone the original kubeconfig and set a env var pointing to the clone so that the original is not used for testing.
This will ensure that tests will also work on original as well.

tools/cli/vz/cmd/logout/logout.go
desagar: The fact that the datatype is in a different, imported package doesn't mean the field in LogoutOptions needs to be public.
e.g. "args" field is private, and its datatype is []string, which is from the Golang default package. 
Similarly you could declare as: `iostreams genericclioptions.IOStreams` - the field would become private.

tools/cli/vz/cmd/logout/logout_test.go
desagar: Perhaps I'm not following this correctly but it looks like you are calling `createFakeKubeConfig`, which is then calling `helpers.GetKubeConfigLocation()`, which is using the KUBECONFIG env var to find the original Kubeconfig, and then cloning that to a new file.

Your goal in this test, for example, is to start with a "logged in" kubeconfig and make sure that logout works. 

For that purpose, you are calling a bunch of functions to take the cloned kubeconfig, make it into a "logged in" one, and then testing the logout. 

In general, a better approach would be to start with a fixed kubeconfig file that you check in alongside the test (instead of cloning the one provided in env var), which is already a "logged in" one, and just test the logout.

The advantages of this are:
1) The test will _always_ test the same input, so it will be repeatable no matter the environment, and will not depend on what the passed in KUBECONFIG that is cloned looks like
2) If there is some bug in the set up functions that you are using to simulate "logging in", that will not affect this test - it will only affect the tests specific to those functions.

Hope this makes sense.

desagar (approved):

tools/cli/vz/cmd/logout/logout_test.go
lake-of-dreams: As was discussed, To be implemented in a subsequent PR

tools/cli/vz/cmd/logout/logout.go
lake-of-dreams: As was discussed, To be implemented in a subsequent PR

PR #1350, opened 2021-07-14
Fix the script to register the managed cluster for the ACME production environment

bhatpmk:
# Description

The issue with registering the managed cluster with ACME production certs is fixed by this change. When the CA is publicly available, ca.crt will be empty in system-tls. This change sets an empty string to ca.crt in the secret file, which is used to create a secret on the admin cluster containing the certificate for the managed cluster.

Fixes VZ-2952

# Checklist 


markxnelson (commented):
is there a test for this?

The script is used by the CI pipeline for multi-cluster AT. The change here will be invoked when production is selected for ACME_ENVIRONMENT

tests/e2e/config/scripts/register_managed_cluster.sh
```diff
@@ -38,8 +44,14 @@ TLS_SECRET=$(kubectl --kubeconfig ${MANAGED_KUBECONFIG} -n verrazzano-system get
 if [ ! -z "${TLS_SECRET%%*( )}" ] && [ "null" != "${TLS_SECRET}" ] ; then
   CA_CERT=$(kubectl --kubeconfig ${MANAGED_KUBECONFIG} -n verrazzano-system get secret system-tls -o json | jq -r '.data."ca.crt"' | base64 --decode)
 fi
+
 if [ ! -z "${CA_CERT}" ] ; then
-   kubectl create secret generic "ca-secret-${MANAGED_CLUSTER_NAME}" -n verrazzano-mc --from-literal=cacrt="$CA_CERT" --dry-run=client -o yaml >> ${CA_SECRET_FILE}
+  kubectl create secret generic "ca-secret-${MANAGED_CLUSTER_NAME}" -n verrazzano-mc --from-literal=cacrt="$CA_CERT" --dry-run=client -o yaml >> ${CA_SECRET_FILE}
+else
+  # When the CA is publicly available/accessible, ca.crt would be empty in system-tls on the admin cluster. So, set an empty string for cacrt
+  if [ "production" == "${ACME_ENVIRONMENT}" ] ; then
```
desagar, line 52: should the build fail in an else here? i.e. CA_CERT is empty, and ACME_ENVIRONMENT is not production. This situation is unlikely to happen, but it _could_ happen if there was some issue in installation. I think we should have an else and exit with non-zero status so that it fails the script and therefore the stage.

desagar: Otherwise, it will fail cryptically during VMC creation.

bhatpmk: Done

desagar (dismissed):

desagar (approved):

PR #1354, opened 2021-07-15
Fixes for installing optional components

bobdonat:
# Description

This pull request contains a number of fixes for the Verrazzano install.  The fixes include:

1. The optional install of Keycloak now works when disabling the component in the vz resource
2. The optional install of Rancher now works when disabling the component in the vz resource
3. The optional install of Elasticsearch now works when disabling the component in the vz resource
4. Network policies were being installed if component was disabled. This has been fixed.

Fixes VZ-3010

# Checklist 

### DIFF
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -342,7 +342,7 @@ type ConsoleComponent struct {
 
 type RancherComponent struct {
 	// +optional
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled *bool `json:"enabled,omitempty"`
 }
 
 // FluentdComponent specifies the Fluentd DaemonSet configuration
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -320,7 +320,7 @@ type KeycloakComponent struct {
 	// +optional
 	MySQL MySQLComponent `json:"mysql,omitempty"`
 	// +optional
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled *bool `json:"enabled,omitempty"`
 }
 
 // MySQLComponent specifies the MySQL configuration
--- a/platform-operator/helm_config/charts/verrazzano/templates/thirdparty-networkpolicy.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/thirdparty-networkpolicy.yaml
@@ -52,6 +52,7 @@ spec:
           protocol: TCP
         - port: 8000
           protocol: TCP
+{{- if .Values.grafana.enabled}}
 ---
 # Network policy for VMI System Grafana
 # Ingress: allow connect from the ingress controller to oidc port 8775
@@ -89,6 +90,8 @@ spec:
       ports:
         - port: 15090
           protocol: TCP
+{{- end }}
+{{- if .Values.prometheus.enabled}}
 ---
 # Network policy for VMI System Prometheus
 # Ingress: allow connect from the ingress controller to oidc port 8775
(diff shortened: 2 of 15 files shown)


platform-operator/internal/vzinstance/verrazzano_instance.go
```diff
@@ -30,8 +30,15 @@ func GetInstanceInfo(client client.Client) *v1alpha1.InstanceInfo {
 		return nil
 	}
 
+	// Console ingress always exist. Only show console URL if the console was enabled during install.
+	var consoleURL *string
+	if cr.Spec.Components.Console == nil || cr.Spec.Components.Console.Enabled {
```
pfmackin, line 35: shouldn't this be ...
if ..Console != nil && ...

bobdonat: Console URL is included when the console component is enabled.  The console component is enabled when not specified or explicitly enabled.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -342,7 +342,7 @@ type ConsoleComponent struct {
 
 type RancherComponent struct {
 	// +optional
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled *bool `json:"enabled,omitempty"`
```
mgianatagh, line 345: I did a search on the file and looks like there is one more instance to maybe fix while you are at it?

type MonitoringComponent struct {
	// +optional
	// +kubebuilder:default:=true
	Enabled bool `json:"enabled,omitempty"`
}

bobdonat: Yes,  I have another branch I am working on to fix that.  The +kubebuilder:default:=true is also problematic.

mgianatagh: ok

mgianatagh (approved):

platform-operator/internal/vzinstance/verrazzano_instance.go
pfmackin: do you mean explicity disabled

bobdonat: no

bobdonat: These are the two flavors of a vz resource where the console URL would be included

spec:
  profile: dev
  components:
    console:
      enabled: true

spec:
  profile: dev

platform-operator/helm_config/charts/verrazzano/templates/thirdparty-networkpolicy.yaml
```diff
@@ -288,7 +293,6 @@ spec:
       app: cattle-cluster-agent
   policyTypes:
     - Ingress
-{{- if .Values.rancher.enabled }}
```
fredtibbitts, line 291: The reason why we always create the network policy for the cluster agent is because the Rancher installation is disabled on managed clusters, but Rancher may eventually be installed by the user as a side effect of applying the generated registration YAML. With this change, the network policy for cattle-cluster-agent will never be installed on the managed cluster.

I should have put a comment in here to make it clear why that network policy is always installed.

bobdonat: Ok, I will undo that change

bobdonat: fixed

fredtibbitts (approved):

platform-operator/internal/vzinstance/verrazzano_instance.go
pfmackin: ok, not sure why i had a hard time understanding that.  Maybe added this to your comment
// Console ingress always exist. Only show console URL if the console component was not 
// specified (the default is enabled) or if it was specified and the value of Enabled is true

pfmackin (approved):

PR #1359, opened 2021-07-19
Remove +kubebuilder:default:=true annotation from MonitoringComponent in vz CRD

bobdonat:
# Description

This pull request removes the **+kubebuilder:default:=true** annotation from the MonitoringComponent type in the Verrazzano CRD.  With the default value annotation, the enabled value for the VMI components in a Verrazzano custom resource would always show a value of true once applied.

For example, the following Verrazzano custom resource spec:
```
spec:
  components:
    elasticsearch:
      enabled: false
```
when applied would end up as follows with enabled having a value of true
```
spec:
  components:
    elasticsearch:
      enabled: true
```

# Checklist 


platform-operator/controllers/verrazzano/installjob/install_config.go
```diff
@@ -673,10 +674,12 @@ func getVMIInstallArgs(vzSpec *installv1alpha1.VerrazzanoSpec) []InstallArg {
...
-		vmiArgs = append(vmiArgs, InstallArg{
-			Name:  esEnabledValueName,
-			Value: strconv.FormatBool(vzSpec.Components.Elasticsearch.Enabled),
-		})
+		if vzSpec.Components.Elasticsearch.Enabled != nil {
+			vmiArgs = append(vmiArgs, InstallArg{
+				Name:  esEnabledValueName,
+				Value: strconv.FormatBool(*vzSpec.Components.Elasticsearch.Enabled),
+			})
+		}
```
pfmackin, line 682: We don't want to append the install args if ES is not enabled, do we?

pfmackin: The enabled for ES should default to true.  Is that still the case?  We don't wan the user to have to explicitly specify true if they added install args.

bobdonat: The install arg will be appended if the enabled flag is specified in the vz CR. 
The default is still true for ES unless overridden by profile settings.

bobdonat: The install behavior is unchanged with these changes.

pfmackin (approved):

PR #1360, opened 2021-07-19
Acceptance test fixes

fredtibbitts:
# Description

This PR updates two acceptance tests:

1. Sock shop intermittently fails in `AfterSuite` because the sockshop namespace is not deleting properly. I added calls to delete the test resources, which should not make a difference, but all of the other tests clean up resources they create so I made this consistent. I also added a cluster dump when the namespace doesn't delete, so hopefully when this problem occurs the cluster dump will give us more to go on.
2. The network policy test, when testing access from one pod to another, had two issues:
    - When expecting pod-to-pod access to work, it was only running the check once. A transient network issue could make the check fail, so I wrapped it in `Eventually`.
    - When expecting pod-to-pod access to NOT work, again it was only doing the check once. A transient network issue could cause the test to pass even if pod-to-pod access would otherwise work. I wrapped that check in `Consistently` to help minimize the risk of the test passing by accident.

# Checklist 

### DIFF
--- a/tests/e2e/examples/socks/sock_shop_example_test.go
+++ b/tests/e2e/examples/socks/sock_shop_example_test.go
@@ -278,13 +278,23 @@ var _ = AfterSuite(func() {
 	}
 
 	Eventually(func() error {
+		return pkg.DeleteResourceFromFile("examples/sock-shop/sock-shop-app.yaml")
+	}, shortWaitTimeout, shortPollingInterval).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
+		return pkg.DeleteResourceFromFile("examples/sock-shop/sock-shop-comp.yaml")
+	}, shortWaitTimeout, shortPollingInterval).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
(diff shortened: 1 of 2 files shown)


markxnelson (dismissed):

tests/e2e/examples/socks/sock_shop_example_test.go
```diff
@@ -277,14 +277,37 @@ var _ = AfterSuite(func() {
 		pkg.ExecuteClusterDumpWithEnvVarConfig()
 	}
 
+	Eventually(func() error {
```
bobdonat, line 280: I thought the theory was that the delete namespace was failing because the appconfig and components were not deleted before deleting the namespace.  This change seems to workaround the delete namespace problem which I think still needs to be resolved.

fredtibbitts: Good point, I've removed the two calls to delete the test resources, and I have also added a dump of the sockshop namespace. I did get this to fail in my branch but looking at the cluster dump, I'm not seeing anything left hanging. Will need more investigation.

bobdonat (approved):

PR #1374, opened 2021-07-22
Mbenmclean/version for lre

mbenmclean:
This PR contains the changes necessary to upgrade verrazzano on every build. Adds support for version string with prerelease and build fields in the platform operator. It also adds unit testing for these changes. The PR also modifies the Jenkinsfile to version the BOM file with major.minor.patch-prerelease+build fields for branches that don't begin with "release-". For release branches it versions the BOM with major.minor.patch. For local builds, it uses "local" instead of the branch name for the prerelease value.

Release version strings:           1.1.0
Branch version strings:            1.1.0-mbenmclean+160d7d24
Locally build version strings:   1.1.0-local+160d7d24


platform-operator/Makefile
```diff
@@ -36,8 +36,9 @@ endif
 
 DOCKER_IMAGE_TAG ?= local-$(shell git rev-parse --short HEAD)
 
+SHORT_COMMIT_HASH ?= $(shell git rev-parse --short=8 HEAD)
 VERRAZZANO_MAJOR_MINOR_PATCH = $(shell grep verrazzano-development-version ../.verrazzano-development-version | sed -e 's/verrazzano-development-version=//')
-VERRAZZANO_VERSION=${VERRAZZANO_MAJOR_MINOR_PATCH}
+VERRAZZANO_VERSION=${VERRAZZANO_MAJOR_MINOR_PATCH}-dev+${SHORT_COMMIT_HASH}
```
michael-cico, line 41: So, before we do any merge, if we go with this we'll have to switch the string generation to be in the Jenkinsfile right?

michael-cico: Reason being that every build will have the `dev` version, not just non-release branches it looks like.

platform-operator/internal/util/semver/semver_test.go
```diff
@@ -252,3 +268,50 @@ func TestIsGreatherThan(t *testing.T) {
...
+}
+
+// TestCompareVersionSubstring Tests CompareVersionSubstring function which converts a Semver object to its string representation
+// GIVEN a call to CompareVersionSubstring with two strings
+// WHEN the Semver field is equivalent to the passed in string
+// THEN return Equal
+func TestCompareVersionSubstring(t *testing.T) {
+	v010, _ := NewSemVersion("v0.1.0")
+	v010dev, _ := NewSemVersion("v0.1.0-dev")
+	v010a, _ := NewSemVersion("v0.1.0+aaaaaaaa")
```
michael-cico, line 299: Forgot my semver stuff, but it looks like we can just have a BuildID and not a prerelease value?  Just curious.

platform-operator/Makefile
mbenmclean: Yes but I think that falls under the build logic category which I'm leaving for now

platform-operator/internal/util/semver/semver_test.go
mbenmclean: Semver allows for just a build ID or just a prerelease value. Don't think we'd ever use it but this handles it correctly

platform-operator/Makefile
michael-cico: OK, just pointing it out so it doesn't get lost, but we'd figure it out in a hurry anyway I guess ;-).

Jenkinsfile
```diff
@@ -153,7 +153,12 @@ pipeline {
                     def props = readProperties file: '.verrazzano-development-version'
                     VERRAZZANO_DEV_VERSION = props['verrazzano-development-version']
                     TIMESTAMP = sh(returnStdout: true, script: "date +%Y%m%d%H%M%S").trim()
-                    SHORT_COMMIT_HASH = sh(returnStdout: true, script: "git rev-parse --short=8 HEAD").trim()
+                    SHORT_COMMIT_HASH = sh(returnStdout: true, script: "echo $env.GIT_COMMIT | head -c 8")
+                    env.VERRAZZANO_VERSION = "${VERRAZZANO_DEV_VERSION}"
+                    if (!"${env.GIT_BRANCH}".startsWith("jenkins-")) {
+                        STRIPPED_BRANCH_NAME = sh(returnStdout: true, script: "echo $env.GIT_BRANCH | sed 's/[^a-zA-Z0-9]//g' | head -c 10")
```
michael-cico, line 159: Is 16 chars too long for this?  Just wondering if 10 will be enough visually for developer branches, but I also don't want it being enormous.  curious what others think here.

platform-operator/Makefile
```diff
@@ -36,8 +36,9 @@ endif
 
 DOCKER_IMAGE_TAG ?= local-$(shell git rev-parse --short HEAD)
 
-VERRAZZANO_MAJOR_MINOR_PATCH = $(shell grep verrazzano-development-version ../.verrazzano-development-version | sed -e 's/verrazzano-development-version=//')
-VERRAZZANO_VERSION=${VERRAZZANO_MAJOR_MINOR_PATCH}
+SHORT_COMMIT_HASH ?= $(shell git rev-parse --short=8 HEAD)
+VERRAZZANO_DEV_VERSION ?= $(shell grep verrazzano-development-version ../.verrazzano-development-version | sed -e 's/verrazzano-development-version=//')
+VERRAZZANO_VERSION ?= ${VERRAZZANO_MAJOR_MINOR_PATCH}-local+${SHORT_COMMIT_HASH}
```
michael-cico, line 41: You tested the local build right? for the default cases?

platform-operator/Makefile
mbenmclean: Fixed in the latest commit

Jenkinsfile
mbenmclean: Lets see what others think. For reference, these are the two options:
1.1.0-mbenmclean+160d7d24
1.1.0-mbenmcleanversio+160d7d24

kminder: Selfishly I'd prefer 16 to 10.  A typical branch name for me is kminder/vz1234 yielding
1.2.3-kmindervz1234+ABCDEFGH
1.2.3-kmindervz+ABCDEFGH

platform-operator/internal/util/semver/semver.go
```diff
@@ -127,8 +131,17 @@ func (v *SemVersion) IsLessThan(from *SemVersion) bool {
 }
 
 // ToString Convert to a valid semver string representation
+// TODO: unit test this
```
kminder, line 134: Add the tests and remove the TODO.

Jenkinsfile
michael-cico: Yeah, I agree.  I too am prone to long-ish branch names.

gilbode: why do we need the branch name at all?  past experience says truncating anything will not turn out well if we don't also add some randomness, we'll hit conflicts somehow.  why can't we just use the branch name as-is?

michael-cico: I thought it would be useful, as you can have builds at the same commit ID for different branches.  We could just use the  user GUID instead.  But if you merge to master from a branch, and there's no branch ID or something else unique in the version, can you have 2 builds from different branches with the same version string?  I think this is also making the case for a full branchname string.  

Or, we could just use a truncated branch + some random string appended to it, but seems like it's not worth the effort if we can just use the full branch name.

mbenmclean: @gilbode so you are advocating for using the full branch name instead of truncating. In theory I think a full branch name with a commit hash should be unique. Edge case if one branch was called ab/cd and another was called ab-cd, they would be the same when stripped of all alphanumerics and could theoretically have the same version number if the commit hash were the same. Even in that case I don't think anything would break as this version string is only really being used as a trigger to upgrade on each build. The customer won't see it

platform-operator/internal/util/semver/semver.go
mbenmclean: Units tests are in the PR just forgot to remove the TODO. Just removed

Jenkinsfile
```diff
@@ -153,7 +153,12 @@ pipeline {
                     def props = readProperties file: '.verrazzano-development-version'
                     VERRAZZANO_DEV_VERSION = props['verrazzano-development-version']
                     TIMESTAMP = sh(returnStdout: true, script: "date +%Y%m%d%H%M%S").trim()
-                    SHORT_COMMIT_HASH = sh(returnStdout: true, script: "git rev-parse --short=8 HEAD").trim()
+                    SHORT_COMMIT_HASH = sh(returnStdout: true, script: "echo $env.GIT_COMMIT | head -c 8")
+                    env.VERRAZZANO_VERSION = "${VERRAZZANO_DEV_VERSION}"
+                    if (!"${env.GIT_BRANCH}".startsWith("jenkins-")) {
```
agv426, line 158: I'm assuming the "jenkins-" is just temporary for testing and will be changed to "release-" before this is merged

mbenmclean: It was changed

agv426 (approved):

PR #1393, opened 2021-07-28
Fix the issue with function kubectl_apply_with_retry

bhatpmk:
# Description

This change addresses the issue with the function kubectl_apply_with_retry in the install script 2-install-system-components.sh. The script will attempt to run the kubectl apply till it gets the return code of 0. 
When the return code is non-zero, it attempts a maximum of 60 times, with a delay of 5 seconds between the retry. An unsuccessful apply after 60 attempts will result in the installation failure.

The uninstall script is updated to delete the secret rancher-admin-secret.

Contributes to VZ-2993

# Checklist 


platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -335,8 +335,8 @@ function reset_rancher_admin_password() {
     fi
     log "Retry Rancher admin password reset..."
   done
-
   update_secret_from_literal rancher-admin-secret cattle-system "$ADMIN_PW"
+  kubectl get secret rancher-admin-secret -n cattle-system
```
michael-cico, line 339: so, this will fail the install if the secret doesn't exist right?

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -547,23 +547,22 @@ function patch_rancher_agents() {
...
   local ret=0
-  until kubectl apply -f <(echo "$1") "${@:2}"; do
-    ret=$?
-    count=$((count+1))
-    if [[ "$count" -lt 60 ]]; then
-      echo "kubectl apply failed, waiting for 5 seconds and trying again"
+  local max_retries=60
+  while true ; do
+    kubectl apply -f <(echo "$1") "${@:2}"
+	  ret=$?
```
michael-cico, line 553: Just a nit, looks like some white space/formatting issues.

michael-cico (approved):

platform-operator/scripts/install/2-install-system-components.sh
bhatpmk: I don't think so, it should just say "Error from server (NotFound): secrets "rancher-admin-secret" not found. Let me confirm that.

michael-cico: I think you have to give it an option to ignore not found errors.  Otherwise I think the `-e` setting on the scripts will fail the run on any non-zero exit code.

Which maybe we want?  If the admin secret isn't there should we fail?

PR #1399, opened 2021-07-29
Added support for DryRun feature of WIT

muhilaru:
# Description

With this change, the user may now specify if they would like to run the image job as a DryRun. Setting the DryRun flag to true will cause WIT to print the Dockerfile to stdout instead of building the image.

Fixes VZ-3115

# Checklist 

### DIFF
--- a/image-patch-operator/README.md
+++ b/image-patch-operator/README.md
@@ -71,6 +71,7 @@ NAME                                               READY   STATUS         RESTAR
 verrazzano-image-patch-operator-7477f65ccf-rjmsc   1/1     Running        0          6m42s
 ```
 Please note that the created Deployment will by default set the IBR_DRY_RUN flag to false. If you would like to run an image job as a dry run, then you may edit the Deployment and set the environment variable IBR_DRY_RUN to true. This will result in the WebLogic Image Tool script to print the Dockerfile to stdout instead of building the image.
+
 ### Create a Secret with Credentials for Pushing the Image
 The Secret can be manually created using `kubectl`.<br>
 ```bash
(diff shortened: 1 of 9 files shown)


image-patch-operator/README.md
```diff
@@ -82,6 +82,7 @@ This will show an output similar to the following:
 NAME                                               READY   STATUS         RESTARTS   AGE
 verrazzano-image-patch-operator-7477f65ccf-rjmsc   1/1     Running        0          6m42s
 ```
+Please note that the created Deployment will by default set the IBR_DRY_RUN flag to false. If you would like to run an image job as a dry run, then you may edit the Deployment and set the environment variable IBR_DRY_RUN to true. This will result in the WebLogic Image Tool script to print the Dockerfile to stdout instead of building the image.
```
mgianatagh, line 85: This is more of a debug feature.  Maybe we should create a new section titled something like "Troubleshooting" and have these instructions under here.  Also include instructions on how to view the output of the dry run.

I notice you also have this value settable via the helm chart.  It would also be helpful to show how to deploy with the dry run enabled.

image-patch-operator/main.go
```diff
@@ -65,7 +66,8 @@ func main() {
 	}
 
 	// Setup the reconciler
-	_, dryRun := os.LookupEnv("VZ_DRY_RUN") // If this var is set, the image jobs are no-ops
+	// If dryRun is set to true, then the image job will print the Dockerfile to stdout instead of building the image
+	dryRun, _ := strconv.ParseBool(os.Getenv("IBR_DRY_RUN"))
```
mgianatagh, line 70: Does this get used?  I think it can be removed.

muhilaru: This gets used inside of imagebuildrequest_controller.go to determine what types of status updates to display based on if a DryRun is ongoing or not.

image-patch-operator/README.md
muhilaru: Created Troubleshooting section in Readme which lists both ways of running ImageJobs as Dry Runs and explains how to view the output.

mgianatagh (approved):

PR #1402, opened 2021-08-02
Rdonat/vz 3088

bobdonat:
# Description

This pull request allows for Fluentd to optionally be installed and changes upgrade to not use the --reuse-values argument during helm upgrade.  Instead of using the --reuse-values argument, the output from `helm get values` is used as input to helm upgrade.  The change in upgrade resolves a nil reference to a new chart value that contains a default value in values.yaml.

Fixes VZ-3088

# Checklist 


platform-operator/internal/util/helm/helmcli.go
```diff
@@ -7,16 +7,39 @@ import (
...
-	vz_os "github.com/verrazzano/verrazzano/platform-operator/internal/util/os"
+	vzos "github.com/verrazzano/verrazzano/platform-operator/internal/util/os"
 	"go.uber.org/zap"
 )
 
 // cmdRunner needed for unit tests
-var runner vz_os.CmdRunner = vz_os.DefaultRunner{}
+var runner vzos.CmdRunner = vzos.DefaultRunner{}
+
+func GetValues(log *zap.SugaredLogger, releaseName string, namespace string) ([]byte, error) {
```
michael-cico, line 17: Nit, but do we need a fn comment here?  Thought the linter or compiler would catch that kind of stuff.

michael-cico (dismissed):

platform-operator/internal/util/helm/helmcli.go
bobdonat: Added a comment.  @michael-cico please approve again.

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -136,8 +139,35 @@ func (h helmComponent) Upgrade(log *zap.SugaredLogger, client clipkg.Client, ns
...
+	if err := tmpFile.Close(); err != nil {
+		log.Errorf("Failed to close temporary file: %v", err)
+		return err
+	}
+
+	log.Infof("Created values file: %s", tmpFile.Name())
+
 	// Do the upgrade
-	_, _, err = upgradeFunc(log, h.releaseName, namespace, h.chartDir, h.valuesFile, overrides)
+	_, _, err = upgradeFunc(log, h.releaseName, namespace, h.chartDir, h.valuesFile, overrides, tmpFile.Name())
```
pfmackin, line 170: This tmpfile should come before the overrides, since we want the overrides to have precedence

pfmackin (approved):

PR #1407, opened 2021-08-03
Eventually linter bug fix and enable build gate

fredtibbitts:
# Description

This PR fixes a bug in the Eventually linter that resulted in false positives. The problem occurred when the parser encountered generic decls after a function decl, and the calls inside the generic decl were incorrectly associated with the previous function decl. For example:

```
func foo() {
 ...
}

var _ = Describe("Tests", func() {
   It("aSpec", func() {
      Expect(...)
   })
})
```

Prior to the fix, the `Expect` was incorrectly associated with the `foo` function.

To fix the bug, I first added a generic decl to the test `main.go` and reproduced the false positive. The fix is to keep track of the token end position of the current function decl. Once the parser encounters a node after that end position, it resets the current func decl name and end position. Then we only record the function calls when the current function decl is defined.

This PR also removes the `-report` flag when running the linter so that builds are now gated on a clean linter run.

# Checklist 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,10 +102,9 @@ copyright-check-branch: copyright-check
 .PHONY: check-tests
 check-tests: check-eventually
 
-# using "-report" here displays results but does not exit with a non-zero code, doing this until we address all of the existing complaints
 .PHONY: check-eventually
 check-eventually: check-eventually-test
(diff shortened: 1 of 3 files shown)


bobdonat (approved):
Looks good to me

PR #1412, opened 2021-08-03
Pass base image to Image Tool

marcogelle:
# Description

Image Patch Service.
Includes the `--fromImage` parameter to specify the base image in the call to `imagetool create`.
Adds resource limits and requests to the Pod running the Image Tool in order for this to work. Values for the resource limits and requests are specified as Helm values.

Fixes VZ-3185
# Checklist 

### DIFF
--- a/image-patch-operator/controllers/imagebuildrequest_controller_test.go
+++ b/image-patch-operator/controllers/imagebuildrequest_controller_test.go
@@ -22,6 +22,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	k8score "k8s.io/api/core/v1"
 	k8net "k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -114,6 +115,20 @@ func TestNewImageBuildRequest(t *testing.T) {
 
 	// Verify istio-injection disabled for the job
 	assert.Equal("false", jb.Labels["sidecar.istio.io/inject"])
+
+	// Default values for resource limits and requests
+	cpuValue := resource.NewMilliQuantity(1100, resource.DecimalSI)
+	memoryValue := resource.NewQuantity(1*1024*1024*1024, resource.BinarySI)
+
+	// Verify that the Job has the expected values for resource limits and requests
+	cpuLimit := jb.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"]
+	memoryLimit := jb.Spec.Template.Spec.Containers[0].Resources.Limits["memory"]
+	cpuRequest := jb.Spec.Template.Spec.Containers[0].Resources.Requests["cpu"]
+	memoryRequest := jb.Spec.Template.Spec.Containers[0].Resources.Requests["memory"]
+	assert.Zero(cpuLimit.Cmp(*cpuValue))
+	assert.Zero(memoryLimit.Cmp(*memoryValue))
+	assert.Zero(cpuRequest.Cmp(*cpuValue))
+	assert.Zero(memoryRequest.Cmp(*memoryValue))
 
 	// Testing that the spec fields of the IBR propagate to the environmental variables of the ImageJob
 	assert.Equal("test-build", jb.Spec.Template.Spec.Containers[0].Env[0].Value)
--- a/image-patch-operator/controllers/imagebuildrequest_controller_test.go
+++ b/image-patch-operator/controllers/imagebuildrequest_controller_test.go
@@ -88,6 +88,14 @@ func TestNewImageBuildRequest(t *testing.T) {
 	// Creating an ImageBuildRequest resource
 	assert.NoError(createResourceFromTemplate(cli, "test/templates/imagebuildrequest_instance.yaml", params))
 
+	// Set test values for resource limits and requests
+	cpuValueString := "1100m"
+	memoryValueString := "1Gi"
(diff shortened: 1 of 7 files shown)


image-patch-operator/helm_config/charts/image-patch-operator/values.yaml
```diff
@@ -11,6 +11,14 @@ imageTool:
   name: verrazzano-image-build-job
   # image value will be filled in during the build pipeline
   image:
+  # Specifies the resource limits for the Pod running WebLogic Image Tool
+  resourceLimits:
+    cpu: "1100" # Units: millicores
+    memory: "1073741824" # = 2^30. Units: Gibibytes
```
mgianatagh, line 17: this unit looks like it would be cumbersome to set.  Can this unit instead be represented as the number of GB?  For example "1" or "2.5"?

mgianatagh:
The unit tests should also be updated to test that the contents of the job object contain the expected values for the limits and resources.  The tests are in image-patch-operator/controllers/imagebuildrequest_controller_test.go.  The function `TestNewImageBuildRequest` contains an example of checks against the job object created for an image build request.

image-patch-operator/helm_config/charts/image-patch-operator/values.yaml
marcogelle: Ok I can change that.
Also, I just noticed the description of the units I put was wrong according to that version. Making this change should make the comment accurate.

image-patch-operator/controllers/imagejob/job.go
```diff
@@ -48,6 +65,16 @@ func NewJob(jobConfig *JobConfig) *batchv1.Job {
 				},
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{{
+						Resources: corev1.ResourceRequirements{
+							Limits: corev1.ResourceList{
+								"cpu":    resource.MustParse(os.Getenv("WIT_POD_RESOURCE_LIMIT_CPU")),
```
mgianatagh, line 70: I do not feel that MustParse is meant to be used in production code.  If the parse fails, it does a panic instead of throwing an error.  From looking at the comment header to the function, it appears it is meant for tests.

// MustParse turns the given string into a quantity or panics; for tests
// or others cases where you know the string is valid.

image-patch-operator/controllers/imagejob/job.go
```diff
@@ -30,6 +31,22 @@ func NewJob(jobConfig *JobConfig) *batchv1.Job {
 		annotations[k8s.DryRunAnnotationName] = strconv.FormatBool(jobConfig.DryRun)
 	}
 
+	// If resource limit and request environment variables are not set, then use some default values
```
mgianatagh, line 34: I am not sure we want to default the values of these parameters.  If we did, a couple of things I would suggest to do differently are to have the default values be stored in image-patch-operator/constants/constants.go.  You probably wouldn't set environment variables here either.  Instead you could have a function inline in the creation of the batch job that defaulted the value.  I suggest removing these two for loops for now.

image-patch-operator/controllers/imagebuildrequest_controller_test.go
```diff
@@ -115,6 +116,20 @@ func TestNewImageBuildRequest(t *testing.T) {
 	// Verify istio-injection disabled for the job
 	assert.Equal("false", jb.Labels["sidecar.istio.io/inject"])
 
+	// Default values for resource limits and requests
```
mgianatagh, line 119: As I looked as this more I think I want to rework the job.go code a bit more than we have time for at the moment.  Perhaps use a resource template like we do for some of the tests to generate the job struct.  But for now I think the quickest thing to do for testing is to do the os.SetEnv of the resource and limit values here (to any test value you want) and verify the job struct gets created with those values.  My concern in the long run of setting the environment variable is that may not work well for running tests in parallel.

mgianatagh (changes requested):

image-patch-operator/controllers/imagejob/job.go
marcogelle: Ok. I am not defaulting these and just removed the two for loops.

image-patch-operator/controllers/imagebuildrequest_controller_test.go
marcogelle: As you suggested, I first did the quick change of moving the os.SetEnv into imagebuildrequest_controller_test.go.

image-patch-operator/controllers/imagebuildrequest_controller_test.go
```diff
@@ -115,6 +124,22 @@ func TestNewImageBuildRequest(t *testing.T) {
...
+	memoryValue, _ := resource.ParseQuantity(memoryValueString)
+
+	// Get resource limits and requests on the Pod
+	cpuLimit := jb.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"]
+	memoryLimit := jb.Spec.Template.Spec.Containers[0].Resources.Limits["memory"]
+	cpuRequest := jb.Spec.Template.Spec.Containers[0].Resources.Requests["cpu"]
+	memoryRequest := jb.Spec.Template.Spec.Containers[0].Resources.Requests["memory"]
+
+	// Verify that the resource limits and requests are the expected test values
+	assert.Zero(cpuLimit.Cmp(cpuValue))
```
mgianatagh, line 138: the assert.Equal function will handle the comparison.  So these can be rewritten to be like below.  I feel that will be more readable.

assert.Equal(cpuValue, jb.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"])

marcogelle: Changed

mgianatagh (approved):

PR #1415, opened 2021-08-04
Changes to enable installing calico directly from tf module in OKE

lake-of-dreams:
# Description

We use `install_calico_oke.sh` to install calico in oke cluster and that has a race condition with workers being ready and adds unexpected behavior. This change to use `3.3.0` version of oke tf module that has feature to install calico.

Fixes VZ-2847

# Checklist 


tests/e2e/config/scripts/delete_oke_cluster.sh
```diff
@@ -72,5 +73,9 @@ do
     ./terraform workspace select $workspace -no-color
   fi
 
+  # Set Calico related mandatory variables
```
fredtibbitts, line 76: Do these really need to be set when deleting the cluster?

fredtibbitts (approved):

fredtibbitts:
One other thing, we need this bug fix: https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/307
Are we using the OCI OKE TF provider that has that fix merged?

tests/e2e/config/scripts/delete_oke_cluster.sh
lake-of-dreams: `calico_enabled` and `calico_version` are mandatory inputs to oci terraform modules while creating, I am not sure if we can omit them while deleting because I don't know if oci terraform modules will need to take some separate action to remove calico and for that it requires the version, thats why I kept it as is

PR #1430, opened 2021-08-12
emit jenkins metrics

shihchang:
# Description

emit metrics from jenkins build and tests stages

Fixes VZ-3289

# Checklist 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,6 +43,7 @@ pipeline {
                 defaultValue: 'master',
                 description: 'The branch to check out after cloning the console repository.',
                 trim: true)
+        booleanParam (description: 'Whether to emit pipeline metrics', name: 'EMIT_METRICS', defaultValue: true)
     }
 
     environment {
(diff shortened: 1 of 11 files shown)


bhatpmk:
@shihchang I think we should pass the value of the parameter EMIT_METRICS to the triggered test (JenkinsfileTestTrigger) and pass the value to the triggered jobs. This helps in re-running master by disabling the flag in one place.

Jenkinsfile
```diff
@@ -219,16 +222,24 @@ pipeline {
         stage('Build') {
             when { not { buildingTag() } }
             steps {
-                buildImages("${DOCKER_IMAGE_TAG}")
+                script {
```
bhatpmk, line 225: I am going through some of the Blue Ocean APIs, which seems to be a clean approach where we can gather the result of the stage and the time information. We can use it at the end of the pipeline, rather than updating each and every stage we are interested in collecting the metrics.

shihchang: Some metrics scope multiple stages, and using an environment variable seems to be simpler than making an api call and process the response. Will investigate the  Blue Ocean APIs for improvement of next PR.

Jenkinsfile
```diff
@@ -766,3 +784,36 @@ def getSuspectList(commitList, userMappings) {
     echo "returning suspect list: ${retValue}"
     return retValue
 }
+
+def metricJobName(stageName) {
```
bhatpmk, line 788: Can we move these duplicating methods to a script ?

shihchang: i did try moving to shell scripts but found shell scripts cannot set a pipeline scope env var, and cannot do simple math like $((3 - 1)) or expr 3 - 1

bhatpmk (approved):

PR #1438, opened 2021-08-16
Rdonat/helm retry

bobdonat:
# Description

The pull request changes the install scripts to detect slow image pulls after a helm install fails.  If slow image pulls are detected then the helm install will be retried after cleaning up the failed helm install.

Fixes VZ-3263

# Checklist 


platform-operator/scripts/install/common.sh
```diff
@@ -373,6 +375,52 @@ function generate_password() {
...
+  local max_retries=1
+  while true ; do
+    log "Installing ${ns}/${chart_name}"
+    helm upgrade ${chart_name} ${chart_dir} \
+      --install --namespace ${ns} \
+      --wait --timeout 10m \
+      "$@" && break
+    local helm_status=$?
+    if [ "$retries" -eq "$max_retries" ] ; then
+fix      return $helm_status
```
bhatpmk, line 415: @bobdonat Is "fix" something your forgot to remove ?

bobdonat: not sure what that is doing there.  I will remove that.

bobdonat: pushed change

bobdonat: thanks for catching that

platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -406,14 +396,26 @@ function install_rancher()
...
 
-      # Do not add --wait since helm install will not fully work in OLCNE until MKNOD is added in the next command
-      helm upgrade ${chart_name} ${RANCHER_CHART_DIR} \
-        --install --namespace cattle-system \
-        --set hostname=rancher.${NAME}.${DNS_SUFFIX} \
-        --set ingress.tls.source=${INGRESS_TLS_SOURCE} \
-        ${HELM_IMAGE_ARGS} \
-        ${IMAGE_PULL_SECRETS_ARGUMENT} \
-        ${EXTRA_RANCHER_ARGUMENTS}
+      if [ $(is_external_dns) == "true" ]; then
```
fredtibbitts, line 399: Does external DNS always mean OLCNE? When I first looked at this I was confused but then it made more sense when I looked at the `is_external_dns` function. Maybe a comment would help?

platform-operator/scripts/install/1-install-istio.sh
```diff
@@ -84,24 +84,20 @@ function install_istio()
     fi
 
     if ! is_chart_deployed istio-base istio-system ${ISTIO_CHART_DIR}/base ; then
-      log "Installing Istio base"
+      log "Installing istio-system/istio-base"
       helm upgrade istio-base ${ISTIO_CHART_DIR}/base \
```
fredtibbitts, line 88: A comment explaining why we don't need to use `helm_install_retry` here might be good, but I wouldn't add it unless you have other changes to push.

platform-operator/scripts/install/2-install-system-components.sh
bobdonat: I think external DNS means OLCNE today.

platform-operator/scripts/install/1-install-istio.sh
bobdonat: Do you think we should use helm_install_retry here for consistency?

bobdonat: even though there are no images in the istio-base helm chart

fredtibbitts: Actually yes, it's probably best to just make it consistent.

fredtibbitts (approved):

PR #1440, opened 2021-08-16
VZ-3280. Internal word linter

stevshap:
# Description

Adds a "linter" to "make check" to look for forbidden/bad words. Fails build if any target words are detected or if curl command does return status=200. The list of words can be updated by modifying the file in object store.

Changes were validated by placing "bad words" in different files at various directory levels to ensure the build failed when the words were present.

Fixes VZ-3280

# Checklist 

### DIFF
--- a/make/quality.mk
+++ b/make/quality.mk
@@ -26,7 +26,7 @@ endif
 .PHONY: word-linter
 word-linter:
 	words=$(curl -L https://bit.ly/3iIUcdL | grep -v '^\s*\(#\|$\)')
-	grep $words -r --exclude-dir=vendor --exclude-dir=word-checker ./../*
+	grep $words -r --exclude-dir=vendor --exclude-dir=word-checker *
 
 .PHONY: coverage
 coverage:
--- a/make/quality.mk
+++ b/make/quality.mk
@@ -23,10 +23,12 @@ else
 endif
 
 # search for internal words that should not be in the repo
+# check fails if res from http req is not successful (200)
 # the actual command being executed in bash is "curl -sL https://bit.ly/3iIUcdL | grep -v '^\s*\(#\|$\)' | ..."
 # additional "$" is to escape literal value in makefile
 .PHONY: word-linter
 word-linter:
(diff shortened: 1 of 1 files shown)


fredtibbitts:
Looks good, although it would be really nice to have a test that validates that the linter is actually catching words. As it is now, it could be broken and we wouldn't know it.

make/quality.mk
```diff
@@ -22,6 +22,11 @@ else
 	$(eval LINTER=$(shell command -v golangci-lint))
 endif
 
+# search for internal words that should not be in the repo
+.PHONY: word-linter
+word-linter:
+	! curl -sL https://bit.ly/3iIUcdL | grep -v '^\s*\(#\|$$\)' | grep -f /dev/stdin -r *
```
gilbode, line 28: Are you sure this will fail if any of the commands in the pipeline fail?  Like what if the curl fails?
Also, don't think you should be using '*', it makes assumptions about current directory, you should explicitly give a directory.

stevshap: "_Are you sure it will fail_" - Yes, I've tested dropping the target words in various files and directories and it finds them and fails accordingly. I went back and forth with @fredtibbitts about better ways to test this/add it to the AT suite but couldn't come to a decision on good ways without adding the actual words to the codebase anywhere

"_What if the curl fails_" - This could be a problem; the curl is publicly accessible so I can't think of specific instances where it will fail unless the instance its' run from isn't connected to the internet, but I'm not sure a way around this without hardcoding values somewhere

"_Don't think you should be using *_ "- * specifies to recursively search every current and subdirectory, and since this will always be called at the root level when the build/make file is executed, I think this is fine. Giving it an explicit directory would just be passing in an additional grep arg

stevshap: If the curl fails the build still succeeds

markxnelson (approved):

stevshap:
@markxnelson added comments clarifying bash grep command

markxnelson (dismissed):

make/quality.mk
gilbode: the build needs to fail if you can't download the word list, otherwise the linter will be broken & we'll never know it

gilbode (approved):

PR #1441, opened 2021-08-17
JIRA:VZ-3324 make txt owner unique for external dns installation

jmaron99:
# Description

Adding a timestamp to the txtOwner fields to ensure specific record ownership for a given installation.  This will hopefully prevent the overwriting of DNS records by a cluster with the same env name in the same DNS zone.

Fixes VZ-3324

# Checklist 

### DIFF
--- a/platform-operator/scripts/install/2-install-system-components.sh
+++ b/platform-operator/scripts/install/2-install-system-components.sh
@@ -256,8 +256,8 @@ function install_external_dns()
         ${HELM_IMAGE_ARGS} \
         --set domainFilters[0]=${DNS_SUFFIX} \
         --set zoneIdFilters[0]=$(get_config_value ".dns.oci.dnsZoneOcid") \
-        --set txtOwnerId=v8o-local-${NAME} \
-        --set txtPrefix=_v8o-local-${NAME}_ \
+        --set txtOwnerId=v8o-local-${NAME}-${TIMESTAMP} \
+        --set txtPrefix=_v8o-local-${NAME}-${TIMESTAMP}_ \
         --set extraVolumes[0].name=config \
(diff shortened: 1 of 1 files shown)


platform-operator/scripts/install/2-install-system-components.sh
```diff
@@ -256,8 +256,8 @@ function install_external_dns()
         ${HELM_IMAGE_ARGS} \
         --set domainFilters[0]=${DNS_SUFFIX} \
         --set zoneIdFilters[0]=$(get_config_value ".dns.oci.dnsZoneOcid") \
-        --set txtOwnerId=v8o-local-${NAME} \
-        --set txtPrefix=_v8o-local-${NAME}_ \
+        --set txtOwnerId=v8o-local-${NAME}-${TIMESTAMP} \
```
mgianatagh, line 259: Does this script run on an upgrade?  If so, would that result in a new set of DNS records getting created?

jmaron99: I don't know if it runs during an upgrade - I'll try to find out.  If it does then the same records will not be allowed to be overwritten and the upgrade will fail.

jmaron99: OK.  Here's what I'v been able to find out:

- The upgrade does not invoke the scripts but rather invokes helm directly from go code
- The upgrade can leverage overrides from the upgrade image or from the VZ CR
- However, I don't feel that an override of txtOwnerId or txtPrefix is likely given that they identify the instance of external DNS. We could augment our documentation with an indication that an override of the values is not a good idea.

mgianatagh: ok, sounds good

mgianatagh (approved):

PR #1450, opened 2021-08-19
WIP:Fix rancher login error after re-install

lake-of-dreams:
# Description
Uninstalltion scripts used to remove rancher crd's but the cr's used to still persist in etcd and when re-install will be done, old cr's will surface and create configuration issues in rancher causing login/cluster status related errors. 
- Added step to remove racher `management.cattle.io` cr's before removing crd
- fixed the helm uninstall command
- moved some of the common code that extracts data from http response and validates the response code to a `httputil` and refactored the calling code in tests and platform-operator
- Enhanced rancher test to also verify rancher local cluster is in active state
- Modified `Jenkinsfile `of uninstall test to include rancher test on install and reinstall
- Fixed an issue with code that adds proxy to `httpClient` for tests, it used to account for http proxy var but not `no_proxy`, replaced with `http. ProxyFromEnvironment` which does both

Fixes VZ-2687

# Checklist 


ci/uninstall/Jenkinsfile
```diff
@@ -316,6 +316,7 @@ pipeline {
             steps {
                 catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
                     runGinkgo('examples/todo')
+                    runGinkgo('verify-infra/restapi')
```
jmaron99, line 319: I'm not sure if I'm the person to decide, but the execution of a test for this pipeline is just to validate the installation.  It may be sufficient to simply replace the todo test with this test rather than prolong the pipeline execution by adding tests.  Might be something to bring up with management?

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
```diff
@@ -109,19 +109,26 @@ function delete_rancher() {
 
   # Deleting rancher components
   log "Deleting rancher"
-  helm ls -n fleet-system | awk '/fleet/ {print $1}' | xargsr helm uninstall -n fleet-system \
+  helm ls -n fleet-system | awk '/fleet/ {print $1}' | xargs -r helm uninstall -n fleet-system \
```
jmaron99, line 112: what's the reason for this change?  is xargsr needed any longer or used by other scripts?

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
```diff
@@ -109,19 +109,26 @@ function delete_rancher() {
...
     || err_return $? "Could not delete rancher-operator-system charts from helm" || return $? # return on pipefail
-  helm ls -n cattle-system | awk '/rancher/ {print $1}' | xargsr helm uninstall -n cattle-system \
+  helm ls -n cattle-system | awk '/rancher/ {print $1}' | xargs -r helm uninstall -n cattle-system \
     || err_return $? "Could not delete cattle-system from helm" || return $? # return on pipefail
 
   log "Delete the additional CA secret for Rancher"
   kubectl -n cattle-system delete secret tls-ca-additional 2>&1 > /dev/null || true
   kubectl -n cattle-system delete secret tls-ca --ignore-not-found=true
 
+  log "Deleting CRs from rancher"
```
jmaron99, line 125: would this code work if no crs are found?  not sure if that scenario could occur (a repeated attempt to uninstall?) but just wondering

lake-of-dreams: yes it will work if no cr's are there. `xargs -r` takes care of that but not processing further if the output from previous command is empty

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
lake-of-dreams: there is no tool named `xargsr` in the image or otherwise, I guess what the original author meant was to use `xargs -r` and thats what I have corrected

ci/uninstall/Jenkinsfile
lake-of-dreams: I don't know why this test was chosen and whether we should keep it. @bhatpmk originally added this so will ask him

bhatpmk: @lake-of-dreams Sorry, I don't remember why did I add examples/todo. Probably we should enable verify-install in this pipeline.

lake-of-dreams: enabled verify-install tests

platform-operator/scripts/uninstall/uninstall-steps/2-uninstall-system-components.sh
jmaron99: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/scripts/uninstall/uninstall-utils.sh#L15

perhaps that function can be fixed?

lake-of-dreams: Thanks for pointing this out @jmaron99 , I was not aware we have created this function as vscode search was not looking in shell scripts, I have reintroduced it and running the tests now

jmaron99 (approved):
LGTM

PR #1451, opened 2021-08-19
JIRA:VZ-3353 exclude kube api server from api istio envoy and add envoy debug logging 

jmaron99:
# Description

This PR is an attempt to address the intermittent 503 issues the VZ API server presents during the KIND tests.  There isn't much in the cluster dumps that indicates the possible cause.  Therefore, this PR is based on the theory that the 503 issues are being generated by an issue in the VZ API server pod (the source) rather than bu an issue at the kube API server.  In addition, I am theorizing that the VZ API container is possibly prevented from accessing the kube API server by the istio proxy due to yet undetermined conditions.  

This PR:
- configures the kube API server's IP as excluded from outbound processing by the VZ API envoy
- adds debug logging for some of the envoy loggers of the VZ API pod.  The loggers selected are the ones primarily involved with outbound connections.  Not all loggers were set to the DEBUG level to prevent unwieldy log files, but additional loggers or all of them can ultimately be enabled in subsequent PRs should that be deemed necessary.

Fixes VZ-3353

# Checklist 

### DIFF
--- a/ci/kind/Jenkinsfile
+++ b/ci/kind/Jenkinsfile
@@ -224,7 +224,7 @@ pipeline {
                             // enable debug logging of verrazzano api istio proxy
                             script {
                                 if (params.ENABLE_API_ENVOY_LOGGING) {
-                                    sh """
+                                    sh '''
                                         vz_api_pod=$(kubectl get pod -n verrazzano-system -l app=verrazzano-api --no-headers -o custom-columns=":metadata.name")
                                         kubectl exec \$vz_api_pod -c istio-proxy -n verrazzano-system -- curl -X POST http://localhost:15000/logging?connection=debug
                                         kubectl exec \$vz_api_pod -c istio-proxy -n verrazzano-system -- curl -X POST http://localhost:15000/logging?filter=debug
@@ -232,7 +232,7 @@ pipeline {
                                         kubectl exec \$vz_api_pod -c istio-proxy -n verrazzano-system -- curl -X POST http://localhost:15000/logging?client=debug
                                         kubectl exec \$vz_api_pod -c istio-proxy -n verrazzano-system -- curl -X POST http://localhost:15000/logging?http=debug
                                         kubectl exec \$vz_api_pod -c istio-proxy -n verrazzano-system -- curl -X POST http://localhost:15000/logging?pool=debug
-                                    """
+                                    '''
                                 }
                             }
                         }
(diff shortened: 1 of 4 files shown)


platform-operator/helm_config/charts/verrazzano/templates/verrazzano-api.yaml
```diff
@@ -43,6 +43,7 @@ spec:
     metadata:
       annotations:
         sidecar.istio.io/inject: "true"
+        traffic.sidecar.istio.io/excludeOutboundIPRanges: {{ .Values.api.kubeServerIPRange }}
```
pfmackin, line 46: I don't think this is needed, we have several operators in the mesh that use the API server with no problem.  Also, if the API server moves to a different node then the API proxy will be broken.

ci/kind/Jenkinsfile
```diff
@@ -220,6 +221,20 @@ pipeline {
                                 VZ_TEST_METRIC = metricJobName('')
                                 metricTimerStart("${VZ_TEST_METRIC}")
                             }
+                            // enable debug logging of verrazzano api istio proxy
+                            script {
+                                if (params.ENABLE_API_ENVOY_LOGGING) {
+                                    sh '''
+                                        vz_api_pod=\$(kubectl get pod -n verrazzano-system -l app=verrazzano-api --no-headers -o custom-columns=\":metadata.name\")
```
pfmackin, line 228: I am wondering if we can enable the Envoy debug logging after we get the first 503.  This will avoid the tons of output that you get with that logging level.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-api.yaml
jmaron99: Yes - but those operators explicitly have no sidecar precisely because we were worried about the communication path with the API server.  The exception is the WebLogic operator, but in that case it appears that you (or someone) excluded port 443, which I assume is specifically for the Kubernetes service port?  Perhaps I should modify the code here to follow the same convention and thus mitigate the possible repositioning issue?

ci/kind/Jenkinsfile
jmaron99: the tests I've conducted show that the amount of logging for the duration of the tests is manageable given the specific loggers we enable.  I think trying to react to the 503s in code is more effort than is warranted here and would unnecessarily prolong this task.

wmhopkins (approved):
lgtm

PR #1462, opened 2021-08-24
Fix upgrade of fluentd daemonset on a managed cluster

bobdonat:
# Description

This pull request fixes the upgrade on a managed cluster when moving to v1.0.1. 

The problem was that the daemonset for fluentd contains a couple of environment variables that are defined as "value" in the helm template.  However, on the managed cluster, the registration process changes those variables to be "valueFrom" in order to obtain the values from the cluster registration secret.  The upgrade was failing because this difference.

Fixes VZ-3378

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/verrazzano.go
+++ b/platform-operator/controllers/verrazzano/component/verrazzano.go
@@ -63,11 +63,11 @@ func fixupFluentdDaemonset(log *zap.SugaredLogger, client client.Client, namespa
 	clusterNameIndex := -1
 	elasticURLIndex := -1
 	for i, env := range daemonSet.Spec.Template.Spec.Containers[fluentdIndex].Env {
-		if env.Name == "CLUSTER_NAME" && env.ValueFrom != nil {
+		if env.Name == constants.ClusterNameEnvVar && env.ValueFrom != nil {
 			clusterNameIndex = i
 			continue
 		}
-		if env.Name == "ELASTICSEARCH_URL" && env.ValueFrom != nil {
+		if env.Name == constants.ElasticsearchURLEnvVar && env.ValueFrom != nil {
 			elasticURLIndex = i
 		}
 	}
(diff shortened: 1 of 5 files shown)


platform-operator/controllers/verrazzano/component/verrazzano.go
```diff
@@ -11,3 +24,90 @@ func resolveVerrazzanoNamespace(ns string) string {
...
+		return fmt.Errorf("the secret named %s in namespace %s is missing the required field %s", secret.Name, secret.Namespace, "managed-cluster-name")
+	}
+
+	// The secret must contain the Elasticsearch endpoint's URL
+	elasticsearchURL, ok := secret.Data["es-url"]
+	if !ok {
+		return fmt.Errorf("the secret named %s in namespace %s is missing the required field %s", secret.Name, secret.Namespace, "es-url")
+	}
+
+	// Update the daemonset to use a value instead of the valueFrom
```
desagar, line 100: should we check if valueFrom is currently nil and if so, avoid calling out to the update?

platform-operator/controllers/verrazzano/component/verrazzano.go
```diff
@@ -11,3 +24,90 @@ func resolveVerrazzanoNamespace(ns string) string {
...
+	// Get the secret containing managed cluster name and Elasticsearch URL
+	secretNamespacedName := types.NamespacedName{Name: constants.MCRegistrationSecret, Namespace: namespace}
+	secret := corev1.Secret{}
+	err = client.Get(context.TODO(), secretNamespacedName, &secret)
+	if err != nil {
+		return err
+	}
+
+	// The secret must contain a cluster name
+	clusterName, ok := secret.Data["managed-cluster-name"]
```
desagar, line 89: these keys ("managed-cluster-name" and "es-url") have constants in controllers/clusters, which this code shouldn't depend on. Do you think there's any point in moving those constants up a level and sharing them across the 2 controllers?

platform-operator/controllers/verrazzano/component/verrazzano.go
```diff
@@ -11,3 +24,90 @@ func resolveVerrazzanoNamespace(ns string) string {
...
+// not fail.  Prior to Verrazzano v1.0.1, the mcagent would change the environment variables CLUSTER_NAME and
+// ELASTICSEARCH_URL on a managed cluster to use valueFrom (from a secret) instead of using a value. The helm chart
+// template for the fluentd daemonset expects a value.
+func fixupFluentdDaemonset(log *zap.SugaredLogger, client client.Client, namespace string) error {
+	// Get the fluentd daemonset resource
+	fluentdNamespacedName := types.NamespacedName{Name: "fluentd", Namespace: namespace}
+	daemonSet := appsv1.DaemonSet{}
+	err := client.Get(context.TODO(), fluentdNamespacedName, &daemonSet)
+	if errors.IsNotFound(err) {
+		return nil
```
pfmackin, line 43: shouldn't you do this test only if err != nil.  Also, shouldn't it be an error if the daemonset is not found?

platform-operator/controllers/verrazzano/component/verrazzano.go
bobdonat: if both env variables are using values then the code would have already returned.  you only get this far if an update is needed.

platform-operator/controllers/verrazzano/component/verrazzano.go
```diff
@@ -11,3 +24,90 @@ func resolveVerrazzanoNamespace(ns string) string {
...
+			clusterNameIndex = i
+			continue
+		}
+		if env.Name == "ELASTICSEARCH_URL" && env.ValueFrom != nil {
+			elasticURLIndex = i
+		}
+	}
+
+	// If valueFrom is not being used then we do not need to fix the env variables
+	if clusterNameIndex == -1 && elasticURLIndex == -1 {
```
pfmackin, line 76: Unlikely but what if one is -1 and the other not?

platform-operator/controllers/verrazzano/component/verrazzano.go
bobdonat: fluentd is optional so it is valid if the daemonset is not found

platform-operator/controllers/verrazzano/component/verrazzano.go
bobdonat: added constants to platform-operator constants.go

pfmackin (approved):

desagar (approved):

PR #1482, opened 2021-08-28
Move install jobs, pods, service accounts, oci secret to verrazzano-install

pfmackin:
Move the following resources to the verrazzano-install namespace from the default namespace
- service account
- config map
- install/uninstall jobs
- install/uninstall pods
- OCI secret (created by user)

In addition
- Cleanup old resources in default namespace from previous install
- Fix Jenkins files to look in verrazzano-install for the pod logs
- This includes Bhat's OCI DNS secret change from  pabhat/oci-secret-vz-nsg (required for this change)

I tried to use a RoleBinding instead of a ClusterRoleBinding but got an error since our install gets nodes, which is cluster level. Also, I had to add a different type of watch for jobs, since the VZ CR no longer owns the jobs because they are in a different namespace.

# Checklist 

### DIFF
--- a/ci/kind/Jenkinsfile
+++ b/ci/kind/Jenkinsfile
@@ -211,8 +211,8 @@ pipeline {
                             sh """
                                 ## dump out install logs
                                 mkdir -p ${VERRAZZANO_INSTALL_LOGS_DIR}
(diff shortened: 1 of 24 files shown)


michael-cico (approved):

agv426:
I'll open a separate task for the analysis tool, it is looking in default for the install pod.

PR #1513, opened 2021-09-07
Tvlatas/vz 3432

agv426:
# Description

Update analysis tool for handling location moves, also reduce some logging noise. This will need to be backported as well

# Checklist 


tools/analysis/internal/util/cluster/verrazzano.go
```diff
@@ -114,10 +114,10 @@ func installationStatus(log *zap.SugaredLogger, clusterRoot string, issueReporte
 
 // IsVerrazzanoInstallJobPod returns true if the pod is an install job related pod for Verrazzano
 func IsVerrazzanoInstallJobPod(pod corev1.Pod) bool {
-	return pod.ObjectMeta.Namespace == "default" && verrazzanoInstallJobPodMatcher.MatchString(pod.ObjectMeta.Name)
+	return verrazzanoInstallJobPodMatcher.MatchString(pod.ObjectMeta.Name) && (pod.ObjectMeta.Namespace == "verrazzano-install" || pod.ObjectMeta.Namespace == "default")
```
bhatpmk, line 117: There is a constant defined for verrazzano-install in platform-operator, if you want to use that.

bhatpmk (approved):

tools/analysis/internal/util/cluster/verrazzano.go
agv426: I don’t think we include any dependencies on the operator code currently (quick grep, maybe we do somewhere). I don’t think we want to add a dependency on operator code here at this point, can look at it generally though when we have cycles to work more on the tool here, ie: common code.

PR #1527, opened 2021-09-08
Add a label to provide an incremental build number with trailing zeroes, to the metrics from CI pipeline

bhatpmk:
Renamed label "number" to "jenkins_build_number" in metrics published from CI pipeline. Also added a new label build_number, to get the metrics in the order in which they are published (that is also order by the label indicating the build number). Without this change, when the X-axis displays the metrics in Series mode in a Grafana panel, the data will be sorted based on the Git commit id. This is nothing but the order in which the Prometheus query returns the result, sorts by the label in the order. With the introduction of the new label "build_number" (which comes before label commit_sha), the result of the prometheus query will be sorted in the order of build number.

The value for the label indicating the status is changed to the first character of the ${currentBuild.currentResult} so that the same can be displayed in legends without consuming too much space, which is required to select a particular color for the metrics based on the status of the build.

Contributes to VZ-3054

# Checklist 


agv426 (dismissed):

shihchang:
To avoid duplicate, most of metrics processing scripts, groovy and shell, could be moved to go code to be shared/reused in jenkinsfiles

shihchang (approved):

adiforluls (dismissed):

adiforluls (approved):

PR #1529, opened 2021-09-08
Initial CheckIn of Validator tool and usage in the Upgrade pipeline

ddrechse:
# Description

 VZ-3445 Upgrade tests should check if new image matches BOM
 Initial checkin of Bom-Validator tool
 Builds in verrazzano pipeline
 Validates in upgrade pipeline
 Modeled after analysis Tool pipeline build and usage semantics

# Checklist 


Jenkinsfile
```diff
@@ -203,6 +203,26 @@ pipeline {
...
+            post {
+                failure {
+                    script {
+                        SKIP_TRIGGERED_TESTS = true
+                    }
+                }
+                success {
+                    sh """
+                        cd ${GO_REPO_PATH}/verrazzano/tools/bom-validator
+                        oci --region us-phoenix-1 os object put --force --namespace ${OCI_OS_NAMESPACE} -bn ${OCI_OS_BUCKET} --name ${env.BRANCH_NAME}/bom-validator --file ./out/linux_amd64/bom-validator
```
michael-cico, line 219: Not necessarily in this pass, but do we need to publish a Mac version of this like the analysis tool?  Probably only if we intend to publish it as a release artifact.

michael-cico: nm, looks like I spoke too soon, I see you build one already.

tools/bom-validator/bom-validator.go
```diff
@@ -0,0 +1,202 @@
...
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Struct based on Verrazzano BOM JSON
+type verrazzanoBom struct {
```
michael-cico, line 17: Also not for this pass, but if you want to use it I moved the VPO BOM code into `pkg/bom` at the top-level in the repo.

michael-cico (approved):

tools/bom-validator/bom-validator.go
ddrechse: Agreed, this task captures that
VZ-3553 BomValidator: write unit tests

Jenkinsfile
ddrechse: Maybe, the makefile builds one so if its needed it is being built

PR #1531, opened 2021-09-08
VZ-3420. Move WKO install to platform operator.

michael-cico:
# Description

Move the WKO install to the platform operator.
- Added hooks for pre-install fn for a `helmComponent`, and added a WKO impl to set up the pre-reqs for that service
- Added the ability to declare dependencies between components in the registry, and validate that those dependencies are `Ready` before allowing install
- Declared dependency on Istio from WKO, and added Istio ready check fn
- Add the correct labels to the `verrazzano-system` namespace
- Added/updated the unit tests

Fixes VZ-3240.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/registry.go
+++ b/platform-operator/controllers/verrazzano/component/registry.go
@@ -4,7 +4,10 @@
 package component
 
 import (
+	"fmt"
+	"go.uber.org/zap"
 	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/verrazzano/verrazzano/application-operator/constants"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
@@ -32,6 +35,7 @@ func GetComponents() []Component {
 			ignoreNamespaceOverride: true,
 			valuesFile:              filepath.Join(overridesDir, "istio-values.yaml"),
 			appendOverridesFunc:     appendIstioOverrides,
+			readyStatusFunc:         istiodReadyCheck,
 		},
 		helmComponent{
 			releaseName:             "istio-ingress",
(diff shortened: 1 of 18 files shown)


platform-operator/controllers/verrazzano/component/registry.go
```diff
@@ -148,3 +158,56 @@ func GetComponents() []Component {
...
+			return false
+		}
+	}
+	return true
+}
+
+// checkDependencies Check the ready state of any dependencies and check for cycles
+func checkDependencies(log *zap.SugaredLogger, client client.Client, c Component, trace map[string]bool) (map[string]bool, error) {
+	for _, dependencyName := range c.GetDependencies() {
+		if trace == nil {
```
pfmackin, line 194: do you plan on leaving the trace in?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -785,15 +791,43 @@ func (r *Reconciler) createVerrazzanoSystemNamespace(ctx context.Context, log *z
...
+		return nil
+	}
+	if err := r.Update(ctx, &vzSystemNS); err != nil {
+		return err
+	}
+	return nil
+}
+
+// mergeMaps Merge one map into another, creating new one if necessary; returns the updated map and true if it was modified
+func mergeMaps(to map[string]string, from map[string]string) (map[string]string, bool) {
```
pfmackin, line 818: this should go in pkg.string maybe.  Not a blocker for merge.

platform-operator/internal/k8s/status/deployment_ready.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: I think this should go in k8s/deployment.  I was sort of confused when i saw status, thinking it was the VZ CR status.  Again, not a blocker, we will be doing some rearranging after you merge as discussed.

pfmackin (dismissed):

pfmackin:
As discussed, we need some follow up work on cleaning up the component struct.

platform-operator/controllers/verrazzano/component/registry.go
michael-cico: The trace is to track cycles in the dependencies.

platform-operator/controllers/verrazzano/controller.go
michael-cico: OK

platform-operator/internal/k8s/status/deployment_ready.go
michael-cico: sounds good...wasn't quite sure where to put it.

pfmackin (approved):

platform-operator/internal/k8s/status/deployment_ready.go
michael-cico: Tracking in VZ-3422

platform-operator/controllers/verrazzano/controller.go
michael-cico: Tracking in VZ-3422

PR #1534, opened 2021-09-09
VZ-3439 mbenmclean/initiate istio upgrade

mbenmclean:
# Description

Fixes VZ-9999

# Checklist 


platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -17,8 +17,11 @@ import (
 
 const vzDefaultNamespace = constants.VerrazzanoSystemNamespace
 
-// helmComponent struct needed to implement a component
-type helmComponent struct {
+// HelmComponent struct needed to implement a component
+type HelmComponent struct {
+
```
mbenmclean, line 22: I need to export type to use it in upgrade.go

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -38,8 +38,27 @@ func (r *Reconciler) reconcileUpgrade(log *zap.SugaredLogger, req ctrl.Request,
...
+				log.Errorf("Error upgrading component %s: %v", c.Name(), err)
+				msg := fmt.Sprintf("Error upgrading component %s - %s\".  Error is %s", comp.Name(),
+					fmtGeneration(cr.Generation), err.Error())
+				err := r.updateStatus(log, cr, msg, installv1alpha1.UpgradeFailed)
+				return ctrl.Result{}, err
+			}
+		case component.IstioComponent:
+			genericComponents = append(genericComponents, c)
+		default:
+			log.Errorf("Unknown component found: %s", c.Name())
```
mbenmclean, line 58: I feel like I should create an err here to return but I don't see any other examples of creating errors in the code. Looks like almost all errs come from other packages. I could do err := errors.New(""Unknown component found: %s", c.Name()). I just don't see examples of that everywhere else. Do you guys think just logging as an Error is fine. Let me know your thoughts.

platform-operator/controllers/verrazzano/component/helm_component.go
michael-cico: You should not be directly accessing `helmComponent`, you should be going through the `Component` interface.

platform-operator/controllers/verrazzano/component/istio_component.go
```diff
@@ -0,0 +1,48 @@
...
+
+package component
+
+import (
+	"github.com/verrazzano/verrazzano/platform-operator/internal/istio"
+	"go.uber.org/zap"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type IstioComponent struct {
```
michael-cico, line 12: So, `IstioComponent` should not be a `HelmComponent`.  It should be a separate implementation of `Component` that implements that interface, IIUC what we're going to be doing with Istioctl.

platform-operator/controllers/verrazzano/component/helm_component.go
mbenmclean: I am but I am using a type switch to determine which components get upgraded first and what go last.

platform-operator/controllers/verrazzano/upgrade.go
michael-cico: So, see earlier comments.  `Component` is the abstraction between the controllers and particular component implementation.  They should follow the same contract, and the controller should have no idea what component it is manipulating.

platform-operator/controllers/verrazzano/component/istio_component.go
mbenmclean: Is that not what this is doing?

platform-operator/controllers/verrazzano/component/helm_component.go
pfmackin: I don't understand the comment about the type.  If you need a new function to check component related items then add a new function to Component which is the public interface.  Do not check the type (as Mike points out)

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -324,6 +324,7 @@ type IstioComponent struct {
 	// Arguments for installing Istio
 	// +optional
 	IstioInstallArgs []InstallArgs `json:"istioInstallArgs,omitempty"`
+	UpgradeApps      *bool         `json:"upgradeApps,omitempty"`
```
pfmackin, line 327: you need to do the following and check in the results:
make generate
make manifest

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -71,31 +74,31 @@ type appendOverridesSig func(log *zap.SugaredLogger, releaseName string, namespa
 // resolveNamespaceSig is an optional function called for special namespace processing
 type resolveNamespaceSig func(ns string) string
 
-// upgradeFuncSig is a function needed for unit test override
-type upgradeFuncSig func(log *zap.SugaredLogger, releaseName string, namespace string, chartDir string, wait bool, dryRun bool, overrides string, overrideFiles ...string) (stdout []byte, stderr []byte, err error)
+// helmUpgradeFuncSig is a function needed for unit test override
```
pfmackin, line 77: This is generic and should have a helm prefix

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -71,31 +74,31 @@ type appendOverridesSig func(log *zap.SugaredLogger, releaseName string, namespa
...
+type helmUpgradeFuncSig func(log *zap.SugaredLogger, releaseName string, namespace string, chartDir string, wait bool, dryRun bool, overrides string, overrideFiles ...string) (stdout []byte, stderr []byte, err error)
 
-// upgradeFunc is the default upgrade function
-var upgradeFunc upgradeFuncSig = helm.Upgrade
+// helmUpgradeFunc is the default upgrade function
```
pfmackin, line 80: This is generic and should have a helm prefix

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -71,31 +74,31 @@ type appendOverridesSig func(log *zap.SugaredLogger, releaseName string, namespa
 // resolveNamespaceSig is an optional function called for special namespace processing
 type resolveNamespaceSig func(ns string) string
 
-// upgradeFuncSig is a function needed for unit test override
-type upgradeFuncSig func(log *zap.SugaredLogger, releaseName string, namespace string, chartDir string, wait bool, dryRun bool, overrides string, overrideFiles ...string) (stdout []byte, stderr []byte, err error)
+// helmUpgradeFuncSig is a function needed for unit test override
+type helmUpgradeFuncSig func(log *zap.SugaredLogger, releaseName string, namespace string, chartDir string, wait bool, dryRun bool, overrides string, overrideFiles ...string) (stdout []byte, stderr []byte, err error)
```
pfmackin, line 78: why are you renaming this?  Same question for others below.

mbenmclean: Because we need two separate upgradeFuncSig because the helm and istio upgrade function signatures take different arguments.

michael-cico: I'm wondering if Istio will even need it.  Given how Golang interfaces work, the istio struct can be completely different, as long as it implements the contract.  We have the function hooks in `helmComponent` so that we can provide component-specific behavior without having a bunch of different component-specific impls, but perhaps migrating to component-specific `Component` impls might make the registry and structures simpler. 

Just riffing right now...we can talk about it more later if/when we get to cleaning up the component registry.

platform-operator/controllers/verrazzano/component/helm_component.go
```diff
@@ -190,7 +190,7 @@ func (h helmComponent) Upgrade(log *zap.SugaredLogger, client clipkg.Client, ns
 	log.Infof("Created values file: %s", tmpFile.Name())
 
 	// Perform a helm upgrade --install
-	_, _, err = upgradeFunc(log, h.releaseName, namespace, h.chartDir, true, dryRun, overridesString, h.valuesFile, tmpFile.Name())
+	_, _, err = upgradeFunc(log, h.releaseName, namespace, h.chartDir, h.waitForInstall, dryRun, overridesString, h.valuesFile, tmpFile.Name())
```
michael-cico, line 193: This should explicitly be `true` for now, to preserve existing behavior.

mbenmclean: Yeah I tried to restore this file but for some reason this changed I will change back

platform-operator/controllers/verrazzano/component/istio_component.go
```diff
@@ -0,0 +1,56 @@
...
+
+// istioUpgradeFunc is the default upgrade function
+var istioUpgradeFunc istioUpgradeFuncSig = istio.Upgrade
+
+// Name returns the component name
+func (i istioComponent) Name() string {
+	return i.componentName
+}
+
+func (i istioComponent) IsOperatorInstallSupported() bool {
```
michael-cico, line 33: This can just always return true or false (false for now).  Eventually this whole method will go away when the scripts go away. This component will always be in-operator.   I added this method as scaffolding while we transition components from the scripts to the operator.

michael-cico: and you can get rid of the struct field also.

michael-cico: BTW, you can do these on the next pass, to avoid an extra push/test cycle.

platform-operator/controllers/verrazzano/component/istio_component.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"github.com/verrazzano/verrazzano/platform-operator/internal/istio"
+	"go.uber.org/zap"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type istioComponent struct {
+	//componentName The name of the component
+	componentName string
```
michael-cico, line 14: For the next pass, do we need this at all?  The component name will be fixed, unless we plan to have multiple types of Istio components.

The Name() method can just return a constant otherwise.  We still need the struct however for the registry declaration.

platform-operator/controllers/verrazzano/component/istio_component.go
mbenmclean: Its ok just fixed.

platform-operator/controllers/verrazzano/component/helm_component.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/istio_component.go
mbenmclean: Ok yeah could probably get changed in next pass. Fine for now though imo

michael-cico: Yeah, next pass.

michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/istio_component.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: remove 2020,

platform-operator/internal/istio/istioctl.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: this should be under component next to istio_component.  Please move it there for now, we are going to create subfolders for each component soon

mbenmclean: This is a the same level as helmcli. They are analogous so I think they should be at the same level. Is that getting moved too?

mbenmclean: I'm going to leave at the same level as helmcli just to keep consistency. When we move them we can move them together

pfmackin (approved):

PR #1545, opened 2021-09-10
VZ-3422.  Move the BOM package so it can be shared

michael-cico:
# Description

Move the BOM package so it can be shared and avoid circular dependencies between API/impl packages
- Export types/functions as needed from the BOM package
- Rename the BOM path override function
- replicate the test BOM where needed for unit tests
- Manage the VPO BOM path through the internal/config package

Fixes VZ-3422 (partial)

# Checklist 


pfmackin:
looks good except i don't think config.go should be in pkg.  This is platform operator config used for main args

michael-cico:
> looks good except i don't think config.go should be in pkg. This is platform operator config used for main args

Took a second pass, moving that config back.  I was just being lazy, I think this pass is better.

pfmackin (dismissed):

pfmackin (dismissed):

pfmackin (approved):

PR #1564, opened 2021-09-14
Ddrechse/vz3345

ddrechse:
# Description

VZ-3445 Upgrade tests should check if new image matches BOM
Initial checkin of Bom-Validator tool
Builds in verrazzano pipeline
Validates in upgrade pipeline
Modeled after analysis Tool pipeline build and usage semantics
# Checklist 


tools/bom-validator/bom-validator.go
```diff
@@ -0,0 +1,220 @@
...
+	if kubeconfig != "" {
+		fmt.Println("USING KUBECONFIG: ", kubeconfig)
+		return true
+	}
+	return false
+}
+
+// Get the BOM from the platform operator in the cluster and build the BOM structure from it
+func getBOM(vBom *verrazzanoBom) {
+	out, err := exec.Command("kubectl", "get", "pod", "-o", "name", "--no-headers=true", "-n", "verrazzano-install").Output()
```
pfmackin, line 96: There are other pods in the verrazzano-install namespace, like the install pod.  How do you know this is getting the platform operator pod?

ddrechse: Because the install pod is in the default ns bit that is a good comment for in the future, I might have to spool through but that is certainly not the current case

michael-cico (approved):

PR #1573, opened 2021-09-16
VZ-3529: Patch Prometheus deployment during upgrade. Epic VZ-3435.

kminder:


platform-operator/controllers/verrazzano/component/prometheus.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: Please put this in component/prometheus dir

platform-operator/controllers/verrazzano/component/prometheus.go
```diff
@@ -0,0 +1,94 @@
...
+	svcObj := k8score.Service{}
+	err = client.Get(ctx, svcKey, &svcObj)
+	if errors.IsNotFound(err) {
+		log.Errorf("Failed to find HTTP Service for Keycloak: %s", err)
+		return err
+	}
+	promObj.Spec.Template.Annotations = setAnnotation(
+		promObj.Spec.Template.Annotations,
+		"traffic.sidecar.istio.io/includeOutboundIPRanges",
+		fmt.Sprintf("%s/32", svcObj.Spec.ClusterIP))
```
michael-cico, line 79: Just curious, do we know the CIDR mask is always `/32`?

michael-cico: Nm, that must just mask it to exactly that IP right, or close to it?

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -49,10 +50,17 @@ func (r *Reconciler) reconcileUpgrade(log *zap.SugaredLogger, req ctrl.Request,
 			return ctrl.Result{}, err
 		}
 	}
+
+	// Invoke the global post upgrade function after all components are upgraded.
+	err := postUpgrade(log, r)
+	if err != nil {
+		return ctrl.Result{}, err
```
michael-cico, line 57: Maybe set `Requeue: true` on this so it will try again?

platform-operator/controllers/verrazzano/component/prometheus.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		log.Errorf("Failed to fetch Prometheus deployment: %s", err)
+		return err
+	}
+
+	// If Keycloak isn't deployed configure Prometheus to avoid the Istio sidecar for metrics scraping.
+	// This is done by adding the traffic.sidecar.istio.io/excludeOutboundIPRanges: 0.0.0.0/0 annotation.
+	// The namespace and name of the Keycloak statefulset should ideally come from the profile or
+	// manifests but that information is not available to upgrade.
+	kcKey := clipkg.ObjectKey{Namespace: "keycloak", Name: "keycloak"}
```
pfmackin, line 43: The component interface has IsInstalled, which is where this code should be (e.g. Keycloak component knows if it is installed).  Also, we currently use helm to check if installed so we need to decide the best way to go.  This is fine the way it is for now, we can rearrange after.

platform-operator/controllers/verrazzano/component/prometheus.go
kminder: All of the other component specific stuff seems to be in the component dir.  Should I be establishing a new pattern here?

platform-operator/controllers/verrazzano/upgrade.go
kminder: Good catch.  Will fix.

platform-operator/controllers/verrazzano/component/prometheus.go
kminder: Confirmed that this is being refactored elsewhere and this should be relocated.

PR #1574, opened 2021-09-16
VZ-3479 generate IstioOperator YAML

pfmackin:
The install and upgrade will be using istioctl and passing in two Istio operator YAML override files.  The first has the hardcoded defaults that are equivalent to the current helm_config/overrides/istio_values.yaml.  This new file is named istio_cr.yaml and uses the IstioOperator CRD schema.  The second override file passed into istioctl, has the contents of the Verrazzano CR Istio component init args, but transformed into the IstioOperator CRD YAML structure.  The Verrazzano CR API has not changed for now.  The  contents of istio_cr.yaml assumes that istioctl is called using the Istio "default" profile. 


platform-operator/helm_config/overrides/istio-cr.yaml
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  components:
```
mbenmclean, line 7: Are you going to add the profile here or should I do it with my implementation jira that I'm about to start. If you could add it here I think it'd be cleaner because that is what the istioctl.go is expecting

mbenmclean (approved):

platform-operator/helm_config/overrides/istio-cr.yaml
pfmackin: you do it please

PR #1575, opened 2021-09-16
VZ-3478 Add istio package with istioctl implementation

mbenmclean:
…d registry.go accordingly

# Description

Fixes VZ-9999

# Checklist 

### DIFF
--- a/platform-operator/internal/istio/istioctl.go
+++ b/platform-operator/internal/istio/istioctl.go
@@ -6,14 +6,45 @@ package istio
 import (
 	vzos "github.com/verrazzano/verrazzano/platform-operator/internal/os"
 	"go.uber.org/zap"
+	"os/exec"
 )
 
 // cmdRunner needed for unit tests
 var runner vzos.CmdRunner = vzos.DefaultRunner{}
 
-func Upgrade(log *zap.SugaredLogger, componentName string) (stdout []byte, stderr []byte, err error) {
-	log.Info("Istio upgrade not implemented yet")
-	return []byte("success"), []byte(""), nil
+// Upgrade function gets called from istio_component to perform istio upgrade
+func Upgrade(log *zap.SugaredLogger, overridesFiles ...string) (stdout []byte, stderr []byte, err error) {
+	args := []string{"install", "-y"}
+
+	for _, overridesFileName := range overridesFiles {
+		args = append(args, "-f")
+		args = append(args, overridesFileName)
+	}
+
+	// Perform istioctl call of type upgrade
+	stdout, stderr, err = runIstioctl(log, args, "upgrade")
+	if err != nil {
+		return stdout, stderr, err
+	}
+
+	return stdout, stderr, nil
+}
+
+// runIstioctl will perform istioctl calls with specified arguments  for operations
+// Note that operation name as of now does not affect the istioctl call (both upgrade and install call istioctl install)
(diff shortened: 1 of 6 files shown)


platform-operator/internal/istio/istioctl.go
```diff
@@ -6,14 +6,45 @@ package istio
...
+func Upgrade(log *zap.SugaredLogger, overridesFiles ...string) (stdout []byte, stderr []byte, err error) {
+	args := []string{"install", "-y"}
+
+	for _, overridesFileName := range overridesFiles {
+		args = append(args, "-f")
+		args = append(args, overridesFileName)
+	}
+
+	// Perform istioctl call of type upgrade
+	stdout, stderr, err = runIstioctl(log, args, "upgrade")
```
pfmackin, line 25: I don't see the canary upgrade unless i am missing it https://istio.io/latest/docs/setup/upgrade/canary/

markxnelson (approved):

platform-operator/internal/istio/istioctl.go
mbenmclean: Oh wow you're totally right. Fixing now

mbenmclean: Fixed

pfmackin: where is the "upgrade" string used by istioctl for upgrade in the istio doc?  I don't see it in the canary doc

mbenmclean: That is not actually used in the istioctl. I just pass it in for clarity in logging. I felt that if the logs for upgrade said install it could be confusing.

mbenmclean: Although looking at it now, the helm install logs as helm upgrade so I guess it's not a big deal. Should I just have it say istioctl install in the logs for both install and upgrade?

platform-operator/internal/istio/istioctl_test.go
```diff
@@ -0,0 +1,74 @@
...
+// GIVEN a set of upgrade parameters and a fake runner that fails
+//  WHEN I call Upgrade
+//  THEN the istioctl upgrade returns an error
+func TestUpgradeFail(t *testing.T) {
+	assert := assert.New(t)
+	SetCmdRunner(badRunner{t: t})
+	defer SetDefaultRunner()
+
+	stdout, stderr, err := Upgrade(zap.S(), "", "")
+	assert.Error(err, "Upgrade should have returned an error")
```
michael-cico, line 53: Probably shouldn't care that stdout/stderr are empty or not.  The main thing is that an error is returned, unless you're just verifying that the fn correctly returns whatever stdout/stderr happen to be?

michael-cico: Actually, it's fine, you are returning both with the error from the fn impl, ignore this.

platform-operator/internal/istio/istioctl.go
michael-cico: The `runIstioCtl` command is internal to the package, I'm fine with it the way it is.  You're using the param to distinguish between the contexts in which the operation is invoked.

The alternative is to move the logging call out to the caller (e.g., `Upgrade()` or `Install()`).  I think at that point we're splitting hairs, and possibly creating unnecessarily redundant code.  I won't go 15 rounds over it either way.

michael-cico (approved):

PR #1591, opened 2021-09-20
VZ-3573.  Upgrade fails with chart stuck in pending-upgrade status

michael-cico:
# Description

Before upgrading, check if a helm-component's chart is stuck in `pending-upgrade` status.  It seems like this can happen when a helm command is interrupted before completing.

Fixes VZ-3573.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/helm/helm_component.go
+++ b/platform-operator/controllers/verrazzano/component/helm/helm_component.go
@@ -244,6 +244,26 @@ func (h HelmComponent) Upgrade(log *zap.SugaredLogger, client clipkg.Client, ns 
 }
 
 func (h HelmComponent) PreUpgrade(log *zap.SugaredLogger, client clipkg.Client, namespace string, dryRun bool) error {
+	resolvedNamespace := resolveNamespace(h, namespace)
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -69,6 +69,34 @@ type HelmComponent struct {
...
+func (h HelmComponent) PreUpgrade(log *zap.SugaredLogger, client clipkg.Client, namespace string, dryRun bool) error {
+	resolvedNamespace := resolveNamespace(h, namespace)
+	chartStatus, err := helm.GetChartStatus(h.ReleaseName, resolvedNamespace)
+	if err != nil {
+		return err
+	}
+	switch chartStatus {
+	case helm.ChartStatusPendingUpgrade:
+		log.Infof("Pre-upgrade, detected pending-upgrade status for %s, rolling back", h.ReleaseName)
+		if stdout, stderr, err := helm.Rollback(log, h.ReleaseName, resolvedNamespace, true, false); err != nil {
```
pfmackin, line 81: I think this is ok, but we should also save the helm version during both install and successful upgrade so we can actually rollback to a know working version.  But this is a good start.

pfmackin (dismissed):

PR #1615, opened 2021-09-23
VZ-3545 enable/disable for Coherence and WebLogic

pfmackin:
Implement disable/ enable component feature for Coherence and WebLogic components.  The first case is covered by our existing e2e tests.   The second and third case is covered by an integration test for now, there is JIRA for future e2e test.

1. If the component.Enabled field is initially set to true then the component will get installed during a new installation
2. If the component.Enabled field is initially set to false then the component will not get installed during a new installation.
3. If the component.Enabled field is changed from false to true, and the component is not installed in an existing installation, then install it.

This includes a new PreInstall component state, where we actually initiate the install.

This change also includes the directory structure for most of the component types.  This was added since I needed it for coherence and WebLogic, plus other folks are starting to write code for various components like keycloak, so i just added the skeleton files where needed.

-------------------------


michael-cico:
Just starting to review, but wondering if we need to implement a sort-of uninstall.  At least just call `helm uninstall`, without worrying about a full cleanup.  A full uninstall should take care of everything else.   But, not for this pass.

platform-operator/test/integ/integ_test.go
```diff
@@ -6,99 +6,166 @@ package integ_test
...
+
+	It("Platform operator pod is eventually running", func() {
+		isPodRunningYet := func() bool {
+			return K8sClient.IsPodRunning(platformOperator, installNamespace)
+		}
+		Eventually(isPodRunningYet, "2m", "5s").Should(BeTrue(),
+			"The verrazzano-platform-operator pod should be in the Running state")
+	})
+
+	It("Verrazzano CR should have disabled components", func() {
```
michael-cico, line 121: I noticed that the install-disabled file disables everything, but the install-enabled file only enables the 2 items.  Does that cause other things like `cert-manager` that are on by default to get installed?  Or is the operator configured for `dry-run` on the install job?

michael-cico (approved):

platform-operator/test/integ/integ_test.go
pfmackin: The reason is that the operator will try a real install and i wanted to minimize what it tried to install, but it just going to fail for all the enabled components.

pfmackin:
> Just starting to review, but wondering if we need to implement a sort-of uninstall. At least just call `helm uninstall`, without worrying about a full cleanup. A full uninstall should take care of everything else. But, not for this pass.

This is an integration test and the kind cluster gets deleted after the integ test run.  But in general i agree that i would be good (and not too hard) to add component based uninstalls.

platform-operator/test/integ/integ_test.go
michael-cico: No, I meant when you apply the 2nd thing to enable them, there's only the 2 components listed, so I was wondering how you stopped the world from being installed, and the install-job.

For the integ tests until the scripts are gone maybe we can have a no-op install-job image so we can do more integ-level testing.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -121,12 +121,21 @@ func ValidateActiveInstall(client client.Client) error {
...
-	if state == Installing || state == Uninstalling || state == Upgrading {
-		return fmt.Errorf("Updates to resource not allowed while install, uninstall or upgrade is in progress")
+func ValidateInProgress(old *Verrazzano, new *Verrazzano) error {
+	if old.Status.State == Ready {
+		return nil
 	}
 
-	return nil
+	// Allow enable component during install
+	if old.Status.State == Installing {
```
michael-cico, line 130: Is it enough to just check that `old.Status.State == Ready` and return, and return an error otherwise?  Basically we don't want any changes (install/upgrade/update/delete) while one is in progress.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -24,49 +26,76 @@ func (r *Reconciler) reconcileComponents(_ context.Context, log *zap.SugaredLogg
...
-	return result, nil
+	if requeue {
+		return newRequeueWithDelay(), nil
+	}
+	return ctrl.Result{}, nil
+}
+
+// IsEnabled returns true if the component spec has enabled set to true
+// Enabled=true is the default
+func isComponentEnabled(cr *vzapi.Verrazzano, componentName string) bool {
```
michael-cico, line 93: so this only affects the components we've moved right?  Probably on the next pass when I move something that depends on profile settings I'll have to change this.

michael-cico (dismissed):

platform-operator/apis/verrazzano/v1alpha1/validate.go
michael-cico: Don't have to address this now anyway, we can modify it on a later pass if needed.

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/validate.go
pfmackin: the unit tests fail unless we check for "", we would have to init all the VZ CRs to ready status but we can do that

platform-operator/controllers/verrazzano/install.go
pfmackin: yes

platform-operator/test/integ/integ_test.go
pfmackin: We don't stop the install job from running but it doesn't really matter since we don't care if it fails.  Once the states of the components reach preInstalling or installing the test will finish and we delete the kind cluster.

michael-cico (approved):

PR #1618, opened 2021-09-23
VZ-3528: Add a secrets list to MultiClusterApplicationConfiguration

bobdonat:
# Description

This pull request adds the following secrets section to the MultiClusterApplicationConfiguration custom resource:
```
spec:
  template:
    ...
  placement:
    ...
  secrets:
    - secret-name1
    - secret-name2
    - ...
```
In addition, webhook validation was added to check that any secrets referenced in the MultiClusterApplicationConfiguration exist.  Unit tests were also added/updated.

# Checklist 


application-operator/controllers/webhooks/multiclusterapplicationconfiguration_webhook.go
```diff
@@ -54,3 +61,35 @@ func (v *MultiClusterApplicationConfigurationValidator) Handle(ctx context.Conte
...
 	return admission.Allowed("")
 }
+
+// Validate that the secrets referenced in the MultiClusterApplicationConfiguration resource exist in the
+// same namespace as the MultiClusterApplicationConfiguration resource.
+func (v *MultiClusterApplicationConfigurationValidator) validateSecrets(mcac *v1alpha1.MultiClusterApplicationConfiguration) error {
+	secrets := corev1.SecretList{}
+	listOptions := &client.ListOptions{Namespace: mcac.Namespace}
+	err := v.client.List(context.TODO(), &secrets, listOptions)
+	if err != nil {
```
mgianatagh, line 71: If it valid for this call to fail with "not found" because there are not any secrets in the namespace?  If so, should this only fail if no secrets in the namespace and there are secrets listed in the app config?

bobdonat: When there are no secrets in the namespace an empty list of secrets is return and err == nil.

bobdonat: I need to add a check up front in the function to return if there are no secrets in the multiclusterapplcationconfiguration

mgianatagh: thanks, I forgot that list would return empty list.  I was thinking of get.

mgianatagh (approved):

application-operator/controllers/webhooks/multiclusterapplicationconfiguration_webhook.go
```diff
@@ -54,3 +61,39 @@ func (v *MultiClusterApplicationConfigurationValidator) Handle(ctx context.Conte
...
+			}
+		}
+		if !found {
+			secretsNotFound = append(secretsNotFound, mcSecret)
+		}
+	}
+
+	if len(secretsNotFound) != 0 {
+		secretsDelimited := strings.Join(secretsNotFound, ",")
+		return fmt.Errorf("secret(s) %s specified in MultiClusterApplicationConfiguration not found in namespace %s", secretsDelimited, mcac.Namespace)
```
desagar, line 95: minor nit - in the case of multiple secrets, I think it might be easier to read if we moved the list of secrets to the end of the error message, something like "the following secret(s) specified ...... not found in namespace %s: %s"

desagar (approved):

PR #1631, opened 2021-09-24
Skumar/vz3531

shashankkumarIITB:
# Description

VZ-3531 
Removes redundant fields from userinfo cookie. 
These changes were not picked up when the branch skumar/vz3452 was merged with shared/vz3256.


wmhopkins (commented):
The only change I see here is that the expiry values for userCookiePairs are removed -- that seems wrong.

PR #1647, opened 2021-09-28
Upgrade/MC test persistence cleanup items

michael-cico:
# Description

Cleanup items for upgrade/MC-upgrade tests
- Fix up uninstall log dumps, update archive patterns
- rename KIND/persistence file to be more general
- Capture install and upgrade YAMLs in upgrade tests

# Checklist 


ci/upgrade/Jenkinsfile
```diff
@@ -208,12 +208,12 @@ pipeline {
                     }
                     post {
                         always {
-                            archiveArtifacts artifacts: "acceptance-test-operator.yaml,downloaded-operator.yaml", allowEmptyArchive: true
```
desagar, line 211: do you think we don't need the operator yamls anymore? Or are those filenames not used anymore?

michael-cico: probably not...in this case we apply an old release manifest directly off the web for install, so they aren't there anyway.

michael-cico: actually this line was moved/replaced by 226.

michael-cico: Sorry, misread the comment/question... so, the fuller, more correct answer is, we don't download those files in this test.  We do a deploy directly from a release so they don't exist on the runner.  So I removed them from the archive statement.

desagar (approved):

ci/upgrade/Jenkinsfile
michael-cico: Now the archive statement on 226 just archives the VZ CR and the install logs

PR #1652, opened 2021-09-29
acceptance tests for external ES

guoyongzhang:
# Description

- create a new pipeline for external ES
- create external ES using elasticsearch-operator
- install VZ with external ES customization
- add fluentd checking in verify register
- add fluentd checking in verify deregister
- extend CLI tests to all managed clusters

Fixes VZ-3494

# Checklist 


ci/external-elasticsearch/Jenkinsfile
```diff
@@ -0,0 +1,1385 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
mgianatagh, line 1: I really don't understand why we are creating pipeline that needs to be run and maintained and keep in sync with changes to the other pipelines.

guoyongzhang: 1) external ES is a new feature in 1.1.  It can not start with 1.0 or 1.0.1 like multiclusters does.  
2) external ES doesn't run upgrade.

desagar: I see the value of testing external ES in the single cluster use case, but it does need to also be tested for managed clusters. 
A couple of questions:
1. Why can't external ES run _after_ the upgrade in the MC pipeline?
2. What do you mean by "external ES doesn't run upgrade"? I am assuming that at some point (maybe not right now) we have to ensure that external ES settings survive the next upgrade

guoyongzhang: 1. Current MC installs with older version and upgrade later.  But the external feature is a part of CRD and has to be installed.  It can not be added as a part of upgrade.
2. We will add upgrade for external ES upgrade where there is an upgrade scenario.

mgianatagh: The current MC pipeline does support starting with the tip of what ever branch you are testing.  You can start with the latest master.  You can control this by passing in which verrazzano-platform-opeator image to use.  I have tested and verified this works.

tests/e2e/pkg/fluentd.go
```diff
@@ -0,0 +1,89 @@
...
+	ds, err := clientset.AppsV1().DaemonSets(verrazzanoNamespace).Get(context.TODO(), fluentdDaemonsetName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return ds, nil
+}
+
+// AssertFluentdURLAndSecret makes sure that expectedURL and expectedSecret are the same as
+// what is configured in the fluentd daemonset.
+// empty string expectedURL means skip comparing URL. Verrazzano prior to 1.1 could configure URL differently.
```
desagar, line 34: is this statement correct? Empty string URL may mean that external es configuration failed so the getexternalESURL function could not find the URL. Would we just pass silently in that case, even if external ES is the case being tested?

guoyongzhang: When 1.0 is installed, the URL in fluentd is set by valueFrom secret.  In that case, I don't want to get too complicated by checking the secret data, so I skip the comparison by having the caller passing empty string.  
The URL checking here is not critical since the log checking can prove the URL is valid.

desagar (approved):

PR #1660, opened 2021-09-29
VZ-3435, VZ-3541: Implement Istio policies to prevent unauthorized access to VMI components

kminder:


platform-operator/helm_config/charts/verrazzano/templates/authorizationpolicy.yaml
```diff
@@ -0,0 +1,221 @@
...
+    - from:
+        - source:
+            namespaces: ["{{ .Release.Namespace }}"]
+            principals: ["cluster.local/ns/{{ .Release.Namespace }}/sa/{{ .Values.monitoringOperator.name }}"]
+      to:
+        - operation:
+            ports: ["15090"]
+
+---
+{{- if .Values.elasticSearch.enabled }}
```
pfmackin, line 44: We are adding enable/disable components.   These policies either need to be in individual files or always apply the policies.

kminder: Discussed with @pfmackin and there isn't really a good solution right now.  Choices are
1. Leave it the way it is and address it as part of the enabled/disable component work going on.
2. Remove all of the .enabled checks so that the authorization policies are applied regardless of the .enabled.

kminder: We decided to leave it the way it is since a number of other similar changes will need to be made to really support enable/disable of components.

wmhopkins (approved):
lgtm

PR #1677, opened 2021-10-01
Support for helm uninstall and uninstall istiocoredns

mbenmclean:
This change will add functionality to uninstall a helm component. It will move the check for uninstall or skip upgrade into HelmComponent methods and modify the Component interface accordingly. It also will modify the registry to reflect these changes

# Checklist 


michael-cico:
So is this just to add the capability of doing an uninstall?  and is the plan to just invoke this at the end of the Istio upgrade to remove the Istio coredns component?

platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -21,9 +21,6 @@ type Component interface {
 	// GetDependencies returns the dependencies of this component
 	GetDependencies() []string
 
-	// GetSkipUpgrade returns the value of the SkipUpgrade field
```
michael-cico, line 24: So, removing the scaffolding from the interface since we only need it for the Istio Helm-based components right?

platform-operator/internal/helm/helmcli.go
```diff
@@ -105,8 +105,7 @@ func Upgrade(log *zap.SugaredLogger, releaseName string, namespace string, chart
 	return stdout, stderr, nil
 }
 
-// Upgrade will upgrade a Helm release with the specified charts.  The overrideFiles array
-// are in order with the first files in the array have lower precedence than latter files.
+// Uninstall will uninstall the release in the specified namespace  using helm uninstall
```
michael-cico, line 108: I think when we get around to doing this, we'll either have to make it blocking (using `--wait`) or have an uninstall lifecycle in the VPO, which we haven't done yet.  The latter would probably be better but maybe a little more work.

platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -203,6 +211,11 @@ func (h HelmComponent) Upgrade(log *zap.SugaredLogger, _ *installv1alpha1.Verraz
 		return nil
 	}
 
+	if h.UninstallComponent {
+		_, _, err = helm.Uninstall(log, h.ReleaseName, namespace, dryRun)
```
michael-cico, line 215: We might want to make this blocking with `--wait` if it's not already; I don't think we're tracking this in the component lifecycle to know if/when it's really gone.

michael-cico: Also, not sure if I like the flag.  It won't apply to most components, this seems like a one-time upgrade thing for Istio as we move to 1.10.x+.

platform-operator/controllers/verrazzano/component/spi/component.go
mbenmclean: Yeah right now skip upgrade is for helm components only so I figured it could stay out of the Component interface. We can leave it in if we think it will be useful though. Although I think it's maybe cleaner to enter the upgrade function for each component in the registry and then have the behavior handled from each components upgrade method rather than handling it from reconcileUpgrade. I think I should have just implemented it that way in the first place

platform-operator/controllers/verrazzano/component/helm/helm_component.go
mbenmclean: I think I think this may be better in post upgrade as well

mbenmclean: We would still need to move the istiocoredns to the end of the registry though right?

platform-operator/controllers/verrazzano/component/spi/component.go
michael-cico: No, it's just scaffolding for now, and I agree it probably doesn't belong in the `Component` interface, it's unique to the situation we're in with the Helm-based components for Istio right now.

platform-operator/controllers/verrazzano/component/helm/helm_component.go
michael-cico: We could probably leverage that helm_component `SkipUpgrade` flag and do it post-upgrade. It leaves that flag in the code for a while though, but there may not be a way around it.  That is, it may live in the code until we hit a point in the life of the product where we're not upgrading from helm-based Istio installs anymore.

Not sure, but as you said offline we should probably meet with the group on it.

mbenmclean: Yeah we certainly could put it in the global post upgrade function. However, that would probably have to be a function call rather than a call on a helm component method which I think would add some complexity

PR #1870, opened 2021-10-29
VZ-3911 add annotation to retry upgrade from failed

mbenmclean:
This change checks for annotations in the VZ CR to see if the upgrade should retry from failed. It also allows entering the reconcile loop when failed and changes the upgrade retries to 5.

# Checklist 


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -256,6 +256,24 @@ func (r *Reconciler) UpgradingState(vz *installv1alpha1.Verrazzano, log *zap.Sug
...
+	retry, err := r.retryUpgrade(ctx, vz)
+	if err != nil {
+		log.Errorf("Failed to update the annotations: %v", err)
+		return newRequeueWithDelay(), err
+	}
+
+	if retry {
+		// Log the retry and set the StateType to ready, then requeue
+		log.Info("Restart Version annotation has changed, retrying upgrade")
+		err = r.updateState(log, vz, installv1alpha1.Ready)
```
michael-cico, line 269: So, this will send us back in, we'll do the version compare, and re-start the regular upgrade path right?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1152,6 +1165,26 @@ func getInstallNamespace() string {
...
+func (r *Reconciler) retryUpgrade(ctx context.Context, vz *installv1alpha1.Verrazzano) (bool, error) {
+	// get the user-specified restart version - if it's missing then there's nothing to do here
+	restartVersion, ok := vz.Annotations[vzconst.RestartVersionAnnotation]
+	if !ok {
+		return false, nil
+	}
+
+	// get the annotation with the previous restart version - if it's missing or the versions do not
+	// match, then return true
+	prevRestartVersion, ok := vz.Annotations[vzconst.ObservedRestartVersionAnnotation]
```
michael-cico, line 1177: I think in GZ's change, this was in the status field right?  Maybe it's good to put it there instead... @pfmackin what do you think?

michael-cico: Actually forget that, that's a CRD change.

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -18,8 +18,8 @@ import (
 	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// The max upgrade failures for a given upgrade attempt is 2
-const failedUpgradeLimit = 2
+// The max upgrade failures for a given upgrade attempt is 5
+const failedUpgradeLimit = 5
```
michael-cico, line 22: So this sets a cap on even user-triggered upgrade retries right?

platform-operator/controllers/verrazzano/controller.go
mbenmclean: yes

platform-operator/controllers/verrazzano/upgrade.go
mbenmclean: Yeah previously it was at 2 I'm upping it to 5. Could do more too I guess. I'm not entirely sure whey we need it.

michael-cico: Me neither, @pfmackin @bobdonat is this needed still?

platform-operator/constants/constants.go
```diff
@@ -101,3 +101,9 @@ const VerrazzanoSystemTLSSecretName = "system-tls"
 
 // The default name for install environment
 const DefaultEnvironmentName = "default"
+
+// RestartVersionAnnotation is the restart version annotation field
+const RestartVersionAnnotation = "verrazzano.io/restart-version"
```
pfmackin, line 106: Shouldn't this be upgrade-retry-version or something?

mbenmclean: This is what weblogic used but I guess that something about upgrade might be more appropriate. I just got a merge conflicting so I'm going to have to rebase anyways so I'll change this

michael-cico (approved):

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -551,7 +562,6 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// The GenerateChangedPredicate will skip update events that have no change in the object's metadata.generation
 		// field.  Any updates to the status or metadata do not cause the metadata.generation to be changed and
 		// therefore the reconciler will not be called.
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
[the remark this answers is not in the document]
pfmackin: This should not cause a problem, if it does then it is a bug in our code that we need to fix.

platform-operator/controllers/verrazzano/upgrade.go
pfmackin: The problem is that the status conditions get written for each failure.  It is unlikely if it fails 2 times that 5 will work.  Furthermore, we retry every component, even if it previously succeeded.  I would leave it at 2 for now.  If it fails 2x on retry the user can modify the annotation to try again.

mbenmclean: I don't think this is limit on how many times it retries automatically. I believe this value looks at previous failures in the status and determines if another upgrade is kicked off. The status only writes once for each time the upgrade is kicked off

pfmackin (approved):

pfmackin (approved):

PR #1890, opened 2021-11-01
VZ-3340 - Kiali Acceptance test

pfmackin:
Initial acceptance test for Kiali.  This also handles the upgrade case, where Kiali is not installed for 1.0.  The PR includes 2 minor Network Policies changes because the kiali label name changed.

### DIFF
--- a/tests/e2e/verify-install/kiali/kiali_test.go
+++ b/tests/e2e/verify-install/kiali/kiali_test.go
@@ -1,0 +1,106 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package kiali
+
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/go-retryablehttp"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/verrazzano/verrazzano/pkg/k8sutil"
+	"github.com/verrazzano/verrazzano/tests/e2e/pkg"
+	networking "k8s.io/api/networking/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"time"
+)
+
+const (
+	systemNamespace = "verrazzano-system"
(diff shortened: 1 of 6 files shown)


tests/e2e/verify-install/kiali/kiali_test.go
```diff
@@ -0,0 +1,117 @@
...
+		client, kialiErr = k8sutil.GetKubernetesClientset()
+		Expect(kialiErr).ToNot(HaveOccurred())
+		httpClient, kialiErr = pkg.GetSystemVmiHTTPClient()
+		Expect(kialiErr).ToNot(HaveOccurred())
+		kialiSupported, kialiErr = pkg.IsVerrazzanoMinVersion("1.1.0")
+		Expect(kialiErr).ToNot(HaveOccurred())
+	})
+
+	// It Wrapper to only run spec if Kiali is supported on the current Verrazzano installation
+	WhenKialiInstalledIt := func(description string, f interface{}) {
```
michael-cico, line 45: This is good, we should generalize this for the flaky test support.

@agv426 @fredtibbitts WDYT?  Might be good way to wrap flaky specs without disabling the whole suite.

michael-cico (approved):

tests/e2e/verify-install/kiali/kiali_test.go
anders-swanson: Yeah, good for flaky tests, make a `WhenRunFlakyTestsIt` that uses some environment variable or flag param at the package level to determine runtime eligibility.

anders-swanson: alt. is to split the specs into separate files.

michael-cico: Fred mentioned also that perhaps 2.0 might have better support for conditionalizing individual specs, but I haven't looked at that yet.

fredtibbitts: Gingko 2 has spec labeling and also a FlakeAttempts decoration, which allows you to specify how many times a spec should be retried if it fails. Labels give you fine-grained control over what suites and specs get run, so we can label individual specs as flaky. I think the plan is to move to Ginkgo 2 but not sure on the timeline.

PR #1899, opened 2021-11-02
VZ-3342 print kiali url if enabled

davisgomes:
# Description

Print Kiali URL in installation complete message if enabled. Prerequisite for updating the documentation.

Fixes VZ-3342

# Checklist 

### DIFF
--- a/platform-operator/scripts/install/config.sh
+++ b/platform-operator/scripts/install/config.sh
@@ -498,6 +498,12 @@ function is_kibana_console_enabled() {
   echo ${kibana_enabled}
 }
 
+# Return whether Kiali console is enabled
+function is_kiali_console_enabled() {
+  local kiali_enabled=$(get_console_enabled 'kiali')
+  echo ${kiali_enabled}
+}
+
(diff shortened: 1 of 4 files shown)


platform-operator/scripts/install/config.sh
```diff
@@ -498,6 +498,12 @@ function is_kibana_console_enabled() {
   echo ${kibana_enabled}
 }
 
+# Return whether Kiali console is enabled
+function is_kiali_console_enabled() {
```
michael-cico, line 502: I think a better approach here might be to look for the existence of the Kiali ingress.  We don't install Kiali via the scripts, and the profile settings may not indicate whether or not it's installed (it can be individually enabled/disabled); and it's more work to pass in the Kiali settings from the operator to the Job I think.

davisgomes: Now, I check the verrazzano instance status to see if the kialiUrl exists. Thanks for the suggestion

michael-cico (approved):

PR #1918, opened 2021-11-04
WIP: hard restart of WebLogic after Istio upgrade to 1.10

guoyongzhang:
# Description

When annotation "restart-version" is added to appconfig, and istio-proxy sidecar of WLS domain is 1.7.3, WLS will be forced to hard start by setting "ServerStartPolicy" to "NEVER" and back.

Fixes VZ-3985

# Checklist 


application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -784,10 +790,110 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+// isDomainForHardRestart determines hard restart or rolling restart based on istio-proxy side car image version
+func (r *Reconciler) isDomainForHardRestart(ctx context.Context, domainName, appName, domainNamespace string, log logr.Logger) bool {
+	podList, err := r.getPodListForDomain(ctx, domainName, appName, domainNamespace, log)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Encnoutered error getting pods for the Weblogic domain %s in namespace %s", domainName, domainNamespace))
+		return false
+	}
+	for _, pod := range podList.Items {
+		for _, container := range pod.Spec.Containers {
+			if container.Name == istioProxyContainerName && container.Image == istioProxyImageForHardRestart {
```
pfmackin, line 825: better to do a startsWith ghcr.io/verrazzano/proxyv2:1.7 in case we ever update 1.0 branch to 1.7.4.  This is not a blocker.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -784,10 +790,110 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+		return err
+	}
+
+	// wait for domain to be deleted
+	r.waitForDomainDeletion(ctx, domainName, appName, domainNamespace, log)
+
+	// set serverStartPolicy back to previousServerStartPolicy
+	existingDomain.Spec.ServerStartPolicy = previousServerStartPolicy
+	log.Info(fmt.Sprintf("Set serverStartPolicy from NEVER to %s in the Weblogic domain %s in namespace %s", previousServerStartPolicy, domainName, domainNamespace))
+	return r.Client.Update(context.TODO(), existingDomain)
```
pfmackin, line 857: sometimes the controller-runtime cache is stale and you might get an error in which case the reconcile will run again.  This should not be a problem, except that you check for NEVER above and return and do nothing.  If the domain was previously set to NEVER then there won't be any pods running for domain and this function should not be called.  So i would remove the NEVER check at 838

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -784,10 +790,110 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+	}
+	return false
+}
+
+func (r *Reconciler) hardRestartDomain(ctx context.Context, existingDomain *wls.Domain, restartVersion, domainName, appName, domainNamespace string, log logr.Logger) error {
+	log.Info(fmt.Sprintf("Restarting the Weblogic domain domain %s in namespace %s by setting serverStartPolicy", domainName, domainNamespace))
+
+	// get previousServerStartPolicy
+	previousServerStartPolicy := existingDomain.Spec.ServerStartPolicy
+	if previousServerStartPolicy == "NEVER" {
```
pfmackin, line 838: Please remove this NEVER check, see comment below.  If it is already NEVER then just don't try to set it to NEVER again.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -784,10 +790,110 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+		log.Info(fmt.Sprintf("serverStartPolicy is already set as NEVER in the Weblogic domain %s in namespace %s", domainName, domainNamespace))
+		return nil
+	}
+
+	// set serverStartPolicy to NEVER
+	existingDomain.Spec.ServerStartPolicy = "NEVER"
```
pfmackin, line 844: Skip if already NEVER

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
guoyongzhang: When serverStartPolicy is initially set to NEVER, the reconciler is triggered again by the fact domain spec is changed while the first reconcile() is still running.  The NEVER block makes the second reconcile no-op.

pfmackin: right but see my comment about the second update failing.  We do see that behavior sometimes with Status

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -784,10 +790,110 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+	r.waitForDomainDeletion(ctx, domainName, appName, domainNamespace, log)
+
+	// set serverStartPolicy back to previousServerStartPolicy
+	existingDomain.Spec.ServerStartPolicy = previousServerStartPolicy
+	log.Info(fmt.Sprintf("Set serverStartPolicy from NEVER to %s in the Weblogic domain %s in namespace %s", previousServerStartPolicy, domainName, domainNamespace))
+	return r.Client.Update(context.TODO(), existingDomain)
+}
+
+// wait for .metadata.deletionTimestamp in all pods
+// returns true if all pods in domain are marked for deletion, otherwise false in 10 seconds
```
michael-cico, line 861: are we sure 10s is enough?  since the check is backgrounded in a go routine, we can probably set the timeout longer.  I'm wondering about the timing of the domain introspector here if the timeout is short, but I might be overthinking it.

PR #1948, opened 2021-11-09
VZ-3978 upgrade changes for new multicluster role bindings

mgianatagh:
# Description

Upgrade changes for new multicluster role bindings.

Fixes VZ-3978

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/upgrade.go
+++ b/platform-operator/controllers/verrazzano/upgrade.go
@@ -153,7 +153,7 @@ func postUpgradeMulticluster(log *zap.SugaredLogger, client clipkg.Client) error
 	for _, vmc := range vmcList.Items {
 		clusterRoleBinding := rbacv1.ClusterRoleBinding{}
 		err := client.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("verrazzano-cluster-%s", vmc.Name)}, &clusterRoleBinding)
-		if err != nil {
+		if err == nil {
 			// Delete the ClusterRoleBinding
 			err = client.Delete(ctx, &clusterRoleBinding)
 			if err != nil {
--- a/platform-operator/controllers/verrazzano/upgrade.go
+++ b/platform-operator/controllers/verrazzano/upgrade.go
@@ -159,7 +159,7 @@ func postUpgradeMulticluster(log *zap.SugaredLogger, client clipkg.Client) error
 			if err != nil {
 				errorList = append(errorList, fmt.Sprintf("failed to delete ClusterRoleBinding %s, error: %s", vmc.Name, err.Error()))
 			} else {
-				log.Info("Deleted ClusterRoleBinding %s", clusterRoleBinding.Name)
+				log.Infof("Deleted ClusterRoleBinding %s", clusterRoleBinding.Name)
 			}
 		}
 	}
--- a/platform-operator/controllers/verrazzano/upgrade.go
+++ b/platform-operator/controllers/verrazzano/upgrade.go
@@ -10,14 +10,12 @@ import (
 	"strconv"
 	"strings"
 
+	vmcv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/clusters/v1alpha1"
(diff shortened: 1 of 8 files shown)


platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -125,6 +130,41 @@ func fmtGeneration(gen int64) string {
 	return "generation:" + s
 }
 
+// postUpgrade - global post upgrade function
+// postUpgrade - global post upgrade function
```
desagar, line 134: minor typo - this is in here twice

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -704,6 +706,141 @@ func TestUpgradeCompleted(t *testing.T) {
...
+	// Expect a call to update the status of the Verrazzano resource
+	mockStatus.EXPECT().
+		Update(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, verrazzano *vzapi.Verrazzano, opts ...client.UpdateOption) error {
+			asserts.Len(verrazzano.Status.Conditions, 3, "Incorrect number of conditions")
+			asserts.Equal(vzapi.UpgradeComplete, verrazzano.Status.Conditions[2].Type, "Incorrect conditions")
+			return nil
+		})
+
+	istiocomp.SetIstioUpgradeFunction(func(log *zap.SugaredLogger, imageOverrideString string, overridesFiles ...string) (stdout []byte, stderr []byte, err error) {
```
desagar, line 726: I don't understand these lines with the istiocom.SetDefaultxxx - why are they added here?

desagar (approved):

platform-operator/controllers/verrazzano/upgrade_test.go
mgianatagh: I don't understand them either, but when I duplicated an upgrade test they were already there.

platform-operator/controllers/verrazzano/upgrade.go
mgianatagh: fixed

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -125,6 +130,41 @@ func fmtGeneration(gen int64) string {
...
 func postUpgrade(log *zap.SugaredLogger, client clipkg.Client) error {
+	// Perform post upgrade steps for multicluster
+	return postUpgradeMulticluster(log, client)
+}
+
+// postUpgradeMulticluster - perform global post upgrade steps for multicluster
+func postUpgradeMulticluster(log *zap.SugaredLogger, client clipkg.Client) error {
```
pfmackin, line 140: This should really go in the application operator post-upgrade.  I am assuming that these bindings are needed by the agent or other app operator code.

mgianatagh: Where is the application operator post-upgrade hook?  I has asked about one because I couldn't find one.  I was directed to use the global post upgrade hook.

desagar (dismissed):

platform-operator/controllers/verrazzano/upgrade.go
michael-cico: You can create one in app_operator_component.go, using the Kiali post-upgrade hooks as an example

https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/kiali/kiali_component.go#L51

I thought you needed this to be in the global post-upgrade hook, but it if it's logically scoped to the app operator then Paul's right, it should probably go there.

mgianatagh: Thank you for the pointers.  I probably thought it needed to be global at the time too.  The ClusterRoleBinding cannot be deleted until the new RoleBindings are created.  But that would occur when the updated platform-operator is deployed.  So moving the post upgrade to the component level does seem appropriate.  I have pushed a change to test this approach and make sure the end to end tests still pass.

desagar (approved):

michael-cico (approved):

PR #1950, opened 2021-11-09
VZ-4047: Release automation updates

fredtibbitts:
# Description

I added a bool param to the Jenkinsfile that allows us to run the stage 1 release pipeline and have it only create the branch. This can be used when creating a major/minor version branch in advance of actually performing a release.

I also removed the check for "master" so that the subsequent release run won't fail when the branch is created ahead of time.

The create_branch.sh script already handles the case where the release branch already exists, so no changes needed there to support this.

Implements VZ-4047

# Checklist 


release/builds/Jenkinsfile
```diff
@@ -93,8 +94,8 @@ pipeline {
         stage('Pipeline inputs validation') {
             steps {
                 script {
```
agv426, line 96: We should probably add a new check here that the params.CREATE_BRANCH_ONLY is only used from master. If someone sets that on a pipeline from a release-* branch we can detect & fail

agv426 (dismissed):

release/builds/Jenkinsfile
fredtibbitts: Done

agv426 (approved):

PR #1952, opened 2021-11-10
VZ-3899.  Update fluend sidecar configmap if necessary

michael-cico:
# Description

In the 1.1 release the fluentd sidecar parse rules for WLS have been updated to accommodate additional log file types that might be available in the server container (nodemanager, access, and domain logs, in addition to the server log).  On upgrade, however, the configmap is not updated.  If the configmap has not changed, the resultant update will be a no-op.  

Also
- renabled the new WLS logging tests, conditionalized for 1.1+
- Added a general utility for conditional execution of a test spec

Fixes VZ-3899.

# Checklist 

### DIFF
--- a/application-operator/controllers/logging/fluentd.go
+++ b/application-operator/controllers/logging/fluentd.go
@@ -165,12 +165,10 @@ func (f *Fluentd) ensureFluentdConfigMapExists(namespace string) error {
 		return err
 	}
 
-	if !configMapExists {
-		if err = f.Create(f.Context, f.createFluentdConfigMap(namespace), &k8sclient.CreateOptions{}); err != nil {
-			return err
-		}
+	if configMapExists {
+		return f.Update(f.Context, f.createFluentdConfigMap(namespace), &k8sclient.UpdateOptions{})
 	}
-	return err
+	return f.Create(f.Context, f.createFluentdConfigMap(namespace), &k8sclient.CreateOptions{})
 }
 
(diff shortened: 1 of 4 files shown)


pfmackin (dismissed):

pfmackin (approved):

guoyongzhang:
Will ensureFluentdConfigMapExists() get called? During normal upgrade,  !upToDate in https://github.com/verrazzano/verrazzano/blob/master/application-operator/controllers/logging/fluentd.go#L68 would be false, right?

guoyongzhang:
Possible to moving f.ensureFluentdConfigMapExists(resource.Namespace) out of the "if" block?  Any updates to configmap will not cause app restart.

pfmackin:
> > Will ensureFluentdConfigMapExists() get called? During normal upgrade, !upToDate in https://github.com/verrazzano/verrazzano/blob/master/application-operator/controllers/logging/fluentd.go#L68 would be false, right?
> 
> Ideally we don't want anything updated until it's intended (app restart, initial deployment). I think @pfmackin 's change will prevent this from being called at an appropriate time.
> 
> @pfmackin do you agree?

Yes, this change is good.  My change will prevent the code from getting this far unless the user modified the restart annotation or the workload spec.

michael-cico:
> Possible to move f.ensureFluentdConfigMapExists(resource.Namespace) out of the "if" block? Any updates to configmap will not cause app restart.

Will do this as either part of Paul's next change, or a separate one.

PR #1957, opened 2021-11-10
VZ-3936 move cert-manager install to go component

davisgomes:
# Description

Migrating all installation code for cert-manager to go code. This is following in suit with other install migrations.

Fixes VZ-3936

# Checklist 


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -334,6 +334,8 @@ type CertManagerComponent struct {
 	// Certificate used for an install
 	// +optional
 	Certificate Certificate `json:"certificate,omitempty"`
+	// +optional
```
michael-cico, line 337: Not sure if we want to do this in this pass, check with @pfmackin .  Reason being, we'd have to have a way to effectively test it.

michael-cico: @pfmackin should we add this Enabled field now?  If so, we probably have to test it in the dynamic config pipeline.

davisgomes: Paul said that he wanted this implemented if possible last I checked.

davisgomes: Let me know what test updates I need to include if I want to add this field

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+// In the case of a CA cert, we install CA resources
+func (c certManagerComponent) PostInstall(compContext spi.ComponentContext) error {
+	// If it is a dry-run, do nothing
+	if compContext.IsDryRun() {
+		compContext.Log().Infof("cert-manager PostInstall dry run")
+		return nil
+	}
+
+	// Verify the certificate information and type
+	certificate := getCertificateConfig(compContext.EffectiveCR())
```
michael-cico, line 128: Might not be necessary to generate a wrapper to check for nils?  You could do a deep-compare with the `CA` field with an empty struct, or in a helper method.

If the profile stuff I just merged is done correctly, only the CA or ACME fields should be non-Empty.  So you could look for 

```
if certificate.ACME != vzapi.ACME{} {
  // do ACME stuff
} else {
  // do CA stuff
}
```

In fact it may be 6 of 1, 1/2 dozen the other, but I think the wrappering just adds another layer that may not be necessary.

Actually, it's a good thing to check with the profiles merge, convert this one branch to the above and run through it in the debugger.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	// Call and execute script for the given DNS type
+	if _, stderr, err := bashFunc(script); err != nil {
+		compContext.Log().Errorf("Failed to apply the cert-manager manifest %s: %s", err, stderr)
+		return err
+	}
+	return nil
+}
+
+// IsCertManagerEnabled returns true if the cert-manager is enabled, which is the default
+func IsCertManagerEnabled(compContext spi.ComponentContext) bool {
```
michael-cico, line 171: Might be good to just embed this in an override of `Component.IsEnabled()` for the `CertManager` component.  Look at the Kiali component as a guide.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+		return true
+	}
+	return *comp.Enabled
+}
+
+// AppendOverrides Build the set of cert-manager overrides for the helm install
+func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	certmanager := compContext.EffectiveCR().Spec.Components.CertManager
+	// Verify that we are using CA certs before appending override
+	if certmanager != nil && (certmanager.Certificate.CA != vzapi.CA{}) {
```
michael-cico, line 183: OK, yeah, this is the kind of check I was referring to in my earlier comment.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+		kvs = append(kvs, bom.KeyValue{Key: "clusterResourceNamespace", Value: namespace})
+	}
+	return kvs, nil
+}
+
+// IsReady checks the state of the expected cert-manager deployments and returns true if they are in a ready state
+func IsReady(context spi.ComponentContext, _ string, namespace string) bool {
```
michael-cico, line 190: Same as the earlier comment, might be good to override the `IsReady` check at the component interface level and get rid of the struct function hook.

I think we're kind of between worlds on this, but probably best to avoid the struct hooks as we can.  For some things (like helm overrides) we still need to use those hooks, but I think here we can go with the override.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+		{Name: certManagerDeploymentName, Namespace: namespace},
+		{Name: cainjectorDeploymentName, Namespace: namespace},
+		{Name: webhookDeploymentName, Namespace: namespace},
+	}
+	return status.DeploymentsReady(context.Log(), context.Client(), deployments, 1)
+}
+
+// getCertificateConfig creates a new Certificate object with pointers and handles the case where CertManager is nil
+func getCertificateConfig(vz *vzapi.Verrazzano) Certificate {
+	if vz.Spec.Components.CertManager != nil && (vz.Spec.Components.CertManager.Certificate.Acme != vzapi.Acme{}) {
```
michael-cico, line 201: Actually, might be good to wrap these kinds of things in an `isACME` or `isCA` kind of check, to save some boiler-plate.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+			ACME: &vzapi.Acme{
+				Provider:     vz.Spec.Components.CertManager.Certificate.Acme.Provider,
+				EmailAddress: vz.Spec.Components.CertManager.Certificate.Acme.EmailAddress,
+				Environment:  vz.Spec.Components.CertManager.Certificate.Acme.Environment,
+			},
+		}
+	}
+	return Certificate{
+		CA: &vzapi.CA{
+			ClusterResourceNamespace: getCAClusterResourceNamespace(vz.Spec.Components.CertManager),
```
michael-cico, line 212: Ideally, as mentioned above with the merge logic you should just be able to rely on these values being present if not overridden by an ACME configuration by the user.

michael-cico: If you want to verify this at the unit-test level, you can use `NewFakeContext` and override the profilesDir param with the relative path to the profiles location.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+			SecretName:               getCASecretName(vz.Spec.Components.CertManager),
+		},
+	}
+}
+
+// getCAClusterResourceNamespace returns the cluster resource name for a CA certificate
+func getCAClusterResourceNamespace(cmConfig *vzapi.CertManagerComponent) string {
+	if cmConfig == nil {
```
michael-cico, line 220: Not to beat a dead horse, but one or the other of either CA or ACME should always be present now.

michael-cico: I think the nil-check of the top level CertManager component is always good, but beyond that ideally just pull a value from the config.

michael-cico: @pfmackin what do you think?  I might flip back on this, on a rethink this may do no harm to have the fall-back.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+
+	// Verify that the secret exists
+	secret := v1.Secret{}
+	if err := compContext.Client().Get(context.TODO(), client.ObjectKey{Name: ociDNSConfigSecret, Namespace: namespace}, &secret); err != nil {
+		compContext.Log().Errorf("Failed to retireve the OCI DNS config secret: %s", err)
+		return err
+	}
+
+	// Verify the acme environment and set the server
+	acmeServer := "https://acme-v02.api.letsencrypt.org/directory"
```
michael-cico, line 265: probably define these as constants

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	if acmeEnvironment := compContext.EffectiveCR().Spec.Components.CertManager.Certificate.Acme.Environment; acmeEnvironment != "" && acmeEnvironment != "production" {
```
michael-cico, line 266: constants for the LetsEncrypt env types, and maybe hide behind `IsProduction` and/or `IsStaging` functions

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	// Create an unstructured object from the template output
+	ciObject := &unstructured.Unstructured{Object: map[string]interface{}{}}
+	if err := yaml.Unmarshal(buff.Bytes(), ciObject); err != nil {
+		compContext.Log().Errorf("Unable to unmarshal yaml: %s", err)
+		return err
+	}
+
+	// Update or create the unstructured object
+	compContext.Log().Info("Applying ClusterIssuer with OCI DNS")
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), compContext.Client(), ciObject, func() error {
```
michael-cico, line 302: So `CreateOrUpdate` handles unstructured data?  Nice.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	return nil
+}
+
+func createCAResources(compContext spi.ComponentContext, certificate Certificate) error {
+	// Create the issuer resource for CA certs
+	compContext.Log().Info("Applying Issuer for CA cert")
+	issuer := certv1.Issuer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "verrazzano-selfsigned-issuer",
```
michael-cico, line 316: constant

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	}); err != nil {
+		compContext.Log().Errorf("Failed to create or update the Issuer: %s", err)
+		return err
+	}
+
+	// Create the certificate resource for CA cert
+	compContext.Log().Info("Applying Certificate for CA cert")
+	certObject := certv1.Certificate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "verrazzano-ca-certificate",
```
michael-cico, line 337: constants for both literals here also?

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), compContext.Client(), &certObject, func() error {
+		return nil
+	}); err != nil {
+		compContext.Log().Errorf("Failed to create or update the Certificate: %s", err)
+		return err
+	}
+
+	// Create the cluster issuer resource for CA cert
+	compContext.Log().Info("Applying ClusterIssuer with OCI DNS")
```
michael-cico, line 358: Probably use Debug for most of these, except of course any direct errors/warnings you need to report?

platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
```diff
@@ -29,7 +31,13 @@ func NewComponent() spi.Component {
 			ChartDir:                filepath.Join(config.GetThirdPartyDir(), "cert-manager"),
 			ChartNamespace:          ComponentName,
 			IgnoreNamespaceOverride: true,
+			SupportsOperatorInstall: true,
+			ImagePullSecretKeyname:  secret.DefaultImagePullSecretKeyName,
 			ValuesFile:              filepath.Join(config.GetHelmOverridesDir(), "cert-manager-values.yaml"),
+			AppendOverridesFunc:     AppendOverrides,
+			IsEnabledFunc:           IsCertManagerEnabled,
```
michael-cico, line 38: Nix the IsEnabled/ReadyStatus func hooks as described earlier.

platform-operator/controllers/verrazzano/component/registry/registry_test.go
```diff
@@ -182,18 +189,10 @@ func TestComponentMultipleDependenciesMet(t *testing.T) {
 		ReadyStatusFunc: nil,
 		Dependencies:    []string{istio.ComponentName, "cert-manager"},
```
michael-cico, line 190: I think I used `cert-manager` here arbitrarily because it worked...might be better to use `fake_component` here in place of real ones.  Maybe something I should take up since I put this here.

platform-operator/scripts/install/apply-cert-manager-manifest.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
michael-cico, line 1: @pfmackin should this go somewhere else, or is this good enough for now?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: we have a dynamic config pipeline, https://build.verrazzano.io/view/Tests/job/verrazzano-dynamic-config-tests/, however, I'm pretty sure if we disable cert-manager on a real install it'll break everything.   Which is why I was thinking we'd avoid it for now.

we probably need to use the integ tests, but I'm not sure how feasible it is to test it in isolation.  Maybe we can check with Paul to see how to best utilize those, as part of a separate task.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -0,0 +1,367 @@
...
+	acmeNotEmpty := components.CertManager.Certificate.Acme != vzapi.Acme{}
+	if caNotEmpty && acmeNotEmpty {
+		return false, errors.New("Certificate object Acme and CA cannot be simultaneously populated")
+	}
+	if caNotEmpty {
+		return true, nil
+	} else if acmeNotEmpty {
+		return false, nil
+	} else {
+		return false, errors.New("Both Acme and CA fields are empty")
```
michael-cico, line 223: I think this message should say something like "Either Acme or CA certificate authorities must be configured", or added to that message.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
davisgomes: Will fix in future PR.

PR #1961, opened 2021-11-11
VZ-3827 Update the DestinationRule fields only if not already set

ayaranga:
Currently, the coherence workload controller code always tries to
override the traffic policy of the destination rule during
reconciliation. This operation results in losing any updates done from
anywhere else.  Made changes to the code to update the destination rule
fields only if it's not already set.

# Checklist 

- [X] Added or updated unit tests for any new functions I added


pfmackin:
The reconciler is not supposed to respect backdoor changes to resources that are being reconciled.  We need a front door mechanism for changing Destination Rules *(via ingress trait or something) unless i am missing something

ayaranga:
Yes, I agree completely that the ticket needs a front door mechanism and this PR is not trying to provide a solution for that. In my testing of various solutions what I noticed was this piece of coherence controller code was overriding any front door changes. This ticket was moved to backlog to be handled by someone with more VZ/Istio experience. Just didn't want my effort to be lost completely. I will discard this PR and whoever takes up this ticket can do everything at one shot.

PR #1964, opened 2021-11-12
VZ-4009 coherence restart

pfmackin:
Prevent modifications  to Coherence resource during reconcile unless the Coherence Workload CR changed or the CR was annotated to restart.  This prevents Coherence cluster from getting automatically restarted after upgrade.  Also, if reconcile is done then inject the new Fluentd image if needed.  There used to be another annotation for that but it is no longer needed.  Reconcile will just fix Coherence CR as required.


application-operator/controllers/controller_utils.go
```diff
@@ -22,11 +26,14 @@ func ConvertAPIVersionToGroupAndVersion(apiVersion string) (string, string) {
...
 
-// IsWorkloadMarkedForUpgrade checks to see if a workload needs to be upgraded to the latest
-// Verrazzano version. Verrazzano defines some resources which are used by applications and when Verrazzano is upgraded,
-// a user can mark an application to indicate that it should use the latest resources defined by Verrazzano.
-// A response of 'true' indicates that reconcile should use the latest resources defined by Verrazzano and a response
-// of 'false' indicates that the application should continue to use the current values.
-func IsWorkloadMarkedForUpgrade(annotations map[string]string, previousUpgrade string) bool {
-	return annotations[constants.AnnotationUpgradeVersion] != previousUpgrade
+// EnsureLastGenerationInStatus ensures that the status has the last generation saved
+func EnsureLastGenerationInStatus(client clipkg.Client, wl *vzapi.VerrazzanoWebLogicWorkload) (ctrl.Result, error) {
```
michael-cico, line 30: Is this used anywhere?  I don't see it called anywhere in these diffs.  Looks like it's just to initialize the status field in the WL workloads?

application-operator/controllers/cohworkload/coherenceworkload_controller.go
```diff
@@ -730,3 +730,29 @@ func (r *Reconciler) addRestartVersionAnnotation(coherence *unstructured.Unstruc
...
 }
+
+// Make sure that the last generation exists in the status
+func (r *Reconciler) ensureLastGeneration(wl *vzapi.VerrazzanoCoherenceWorkload) (ctrl.Result, error) {
+	if len(wl.Status.LastGeneration) > 0 {
+		return ctrl.Result{}, nil
+	}
+
+	// Update the status generation and always requeue
+	wl.Status.LastGeneration = strconv.Itoa(int(wl.Generation))
```
mbenmclean, line 741: Why do we need to put the generation field in the status why can't we just use the wl.Generation field?

application-operator/controllers/cohworkload/coherenceworkload_controller.go
```diff
@@ -730,3 +730,29 @@ func (r *Reconciler) addRestartVersionAnnotation(coherence *unstructured.Unstruc
...
+	err := r.Status().Update(context.TODO(), wl)
+	return ctrl.Result{Requeue: true, RequeueAfter: 1}, err
+}
+
+// Make sure that it is OK to restart Coherence
+func (r *Reconciler) isOkToRestartCoherence(coh *vzapi.VerrazzanoCoherenceWorkload) bool {
+	// Check if user created or changed the restart annotation
+	if coh.Annotations != nil && coh.Annotations[constants.RestartVersionAnnotation] != coh.Status.LastRestartVersion {
```
mbenmclean, line 749: What is the behavior of coh.Annotations[constants.RestartVersionAnnotation] if doesn't find the RestartVersionAnnotation? This will return the value and also a boolean I believe. Have you checked to make sure this works for all cases?

platform-operator/helm_config/charts/verrazzano-application-operator/crds/oam.verrazzano.io_verrazzanocoherenceworkloads.yaml
```diff
@@ -51,9 +51,13 @@ spec:
...
               of VerrazzanoCoherenceWorkload
             properties:
-              currentUpgradeVersion:
-                description: CurrentUpgradeVersion is the version that was specified
-                  when the application was last upgraded with Verrazzano
+              lastGeneration:
+                description: LastGeneration is the last generation of the spec that
+                  was reconciled
+                type: string
+              lastRestartVersion:
```
mbenmclean, line 58: I think we have been using observedRestartVersion for this but I'm not sure. Either way the value should probably be standardized.

application-operator/controllers/controller_utils.go
pfmackin: No it is no longer used.  That was the old annotation that John added to inject the new Fluentd image into the sidecar.  Now that is done during alls reconciles where it is ok to restart Coherence.

application-operator/controllers/cohworkload/coherenceworkload_controller.go
pfmackin: This is the generation of the WebLogicWorkload.

application-operator/controllers/cohworkload/coherenceworkload_controller.go
pfmackin: if the annotation is missing it will return empty string.  So in the case where the annotation is set for the first time, this will return true.

platform-operator/helm_config/charts/verrazzano-application-operator/crds/oam.verrazzano.io_verrazzanocoherenceworkloads.yaml
pfmackin: > I think we have been using observedRestartVersion for this but I'm not sure. Either way the value should probably be standardized.

I don't see observedRestartVersion being used.  If there any inconsistencies please point them out and i will fix them in the next stop app JIRA that I will be doing.

michael-cico (approved):

platform-operator/helm_config/charts/verrazzano-application-operator/crds/oam.verrazzano.io_verrazzanocoherenceworkloads.yaml
mbenmclean: Oh ok that's what I used for the annotation to force an upgrade retry. I saw GZ had used observed in one of his PRs but maybe he switched it. If you want to change it the constants are at the bottom of platform-operator/constants/constants.go you can refactor if you want. Up to you though.

PR #1969, opened 2021-11-15
Mbenmclean/vz 4043 post upgrade test

mbenmclean:
Fixes VZ-4043
This change adds tests to check for the restart annotation and the correct istio proxy image for pods in the platform namespaces. It also adds this check into the upgrade pipeline.

# Checklist 


tests/e2e/pkg/common.go
```diff
@@ -315,6 +315,60 @@ func MetricsExistInCluster(metricsName string, keyMap map[string]string, kubecon
...
+		Log(Error, fmt.Sprintf("Error listing pods in cluster for namespace: %s, error: %v", namespace, err))
+		return false
+	}
+	for _, pod := range pods.Items {
+		_, ok := pod.Labels["istio.io/rev"]
+		if ok {
+			containers := pod.Spec.Containers
+			found := false
+			for _, container := range containers {
+				if strings.Contains(container.Image, "proxyv2:1.10") {
```
michael-cico, line 356: Might be good to generalize this helper, i.e., call it something like `CheckPodsForImage` and pass in the image pattern to look for.  Adds an argument on the invocation, but I think this is a generally useful helper to have.

tests/e2e/upgrade/platform/verify_restart_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"time"
+)
+
+const (
+	waitTimeout     = 3 * time.Minute
+	pollingInterval = 10 * time.Second
+)
+
+var _ = Describe("verify platform pods restarted", func() {
+
```
michael-cico, line 20: Do we need to check other namespaces?  Or are these all the ones where we do injection?

tests/e2e/upgrade/platform/verify_restart_test.go
```diff
@@ -0,0 +1,77 @@
...
+	// GIVEN the verrazzano-system namespace
+	// WHEN the annotations from the pods are retrieved
+	// THEN verify that the have the verrazzano.io/restartedAt annotations
+	It("Verify pods in verrazzano-system restarted post upgrade", func() {
```
michael-cico, line 24: Might want to wrap these in the `pkg.MinVersionSpec` utility to only run these tests if the VZ version is 1.1.0+?

tests/e2e/upgrade/platform/verify_restart_test.go
```diff
@@ -0,0 +1,77 @@
...
+		}, waitTimeout, pollingInterval).Should(BeTrue(), "Expected to find restart annotation in keycloak")
+	})
+})
+
+var _ = Describe("verify platform pods have correct istio proxy image", func() {
+
+	// GIVEN the verrazzano-system namespace
+	// WHEN the container images are retrieved
+	// THEN verify that each pod that uses istio has the correct istio proxy image
+	It("Verify pods in verrazzano-system have correct istio proxy image", func() {
```
michael-cico, line 54: Similar comment, should we only run these if the version is 1.1.0+?

tests/e2e/upgrade/platform/verify_restart_test.go
mbenmclean: It's just these 3 right now

tests/e2e/pkg/common.go
mbenmclean: Ok I'll change that

tests/e2e/upgrade/platform/verify_restart_test.go
mbenmclean: Ok

tests/e2e/pkg/common.go
mbenmclean: Actually the function is a little more complicated than that because we're only doing the check if the pod also has the label istio.io/rev. Going to leave it as is for now.

tests/e2e/pkg/common.go
```diff
@@ -315,6 +315,59 @@ func MetricsExistInCluster(metricsName string, keyMap map[string]string, kubecon
...
+	for _, pod := range pods.Items {
+		_, hasAnnotation := pod.Annotations[annotation]
+		if !hasAnnotation && !strings.Contains(pod.Name, "vmi-system-kiali") {
+			return false
+		}
+	}
+	return true
+}
+
+func CheckPodsForImage(namespace string) bool {
```
pfmackin, line 338: This func is checking for istio 1.10  and istio labels.  It needs to have Istio in the name (CheckPodsForIstio) or something

ci/upgrade/Jenkinsfile
```diff
@@ -531,6 +515,11 @@ pipeline {
                         runGinkgoRandomize('verify-infra/vmi')
                     }
                 }
+                stage('platform upgrade') {
+                    steps {
+                        runGinkgoRandomize('upgrade/platform')
```
pfmackin, line 520: Are these pre or post upgrade tests?  You should have them in a subdir like upgrade/pre_upgrade or upgrade/post_upgrade.  Not sure we need to distinguish platform or apps at the directory level, if you do the have upgrade/post-upgrade/platform or something.

tests/e2e/pkg/common.go
mbenmclean: changed to CheckPodsForIstioImage

ci/upgrade/Jenkinsfile
mbenmclean: Well there's currently nothing we do pre upgrade. Do think there's some value in breaking things up between the platform upgrade tests and the app upgrade tests. I think this may start to take shape later on when we build out our testing. I can add a postupgrade directory

michael-cico (approved):

pfmackin (dismissed):

pfmackin (approved):

PR #1978, opened 2021-11-16
VZ-3985 stop/start WebLogic

pfmackin:
Use annotations to stop a WebLogic domain, then to start it. A new annotation is used,  verrazzano.io/lifecycle-action with values "start" and "stop".  Add VerrazzanoWebLogicWorkload status fields to keep track of the last lifecycle operation.

This is subtask VZ-4060

### DIFF
--- a/application-operator/constants/constants.go
+++ b/application-operator/constants/constants.go
@@ -102,3 +102,12 @@ const VerrazzanoPasswordData = "password"
 
 // RestartVersionAnnotation - the annotation used by user to tell Verrazzano applicaton to restart its components
 const RestartVersionAnnotation = "verrazzano.io/restart-version"
+
+// LifecycleAnnotation - the annotation perform lifecycle actions on a workload
+const LifecycleActionAnnotation = "verrazzano.io/lifecycle-action"
+
+// LifecycleActionStop - the annotation value used to stop a workload
+const LifecycleActionStop = "stop"
+
+// LifecycleActionStart - the annotation value used to start a workload
+const LifecycleActionStart = "start"
--- a/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
+++ b/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
@@ -39,16 +39,16 @@ import (
 )
 
 const (
-	specField                       = "spec"
-	destinationRuleAPIVersion       = "networking.istio.io/v1alpha3"
-	destinationRuleKind             = "DestinationRule"
-	loggingNamePart                 = "logging-stdout"
-	loggingMountPath                = "/fluentd/etc/custom.conf"
-	loggingKey                      = "custom.conf"
-	defaultMode               int32 = 400
-	serverStartPolicyAnnotation     = "verrazzano-io/last-server-start-policy"
- 	Never 							= "NEVER"
- 	IfNeeded 						= "IF_NEEDED"
+	specField                         = "spec"
+	destinationRuleAPIVersion         = "networking.istio.io/v1alpha3"
+	destinationRuleKind               = "DestinationRule"
+	loggingNamePart                   = "logging-stdout"
+	loggingMountPath                  = "/fluentd/etc/custom.conf"
+	loggingKey                        = "custom.conf"
+	defaultMode                 int32 = 400
+	serverStartPolicyAnnotation       = "verrazzano-io/last-server-start-policy"
+	Never                             = "NEVER"
+	IfNeeded                          = "IF_NEEDED"
 )
 
 const defaultMonitoringExporterData = `
@@ -173,6 +173,7 @@ var specConfigurationIstioEnabledFields = []string{specField, "configuration", "
 var specConfigurationRuntimeEncryptionSecret = []string{specField, "configuration", "model", "runtimeEncryptionSecret"}
 var specMonitoringExporterFields = []string{specField, "monitoringExporter"}
 var specRestartVersionFields = []string{specField, "restartVersion"}
+var specServerStartPolicyFields = []string{specField, "serverStartPolicy"}
 
 // this struct allows us to extract information from the unstructured WebLogic spec
 // so we can interface with the FLUENTD code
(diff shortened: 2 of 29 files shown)


application-operator/constants/constants.go
```diff
@@ -102,3 +102,12 @@ const VerrazzanoPasswordData = "password"
 
 // RestartVersionAnnotation - the annotation used by user to tell Verrazzano applicaton to restart its components
 const RestartVersionAnnotation = "verrazzano.io/restart-version"
+
+// LifecycleAnnotation - the annotation perform lifecycle actions on a workload
+const LifecycleActionAnnotation = "verrazzano.io/lifecycle-action"
```
michael-cico, line 107: Not sure I want to do it for this PR, but it seems to overlap with `restart-version`.  Maybe in 1.2 we can consolidate the annotations/behavior.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -810,10 +843,87 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+// will not result in additional restarts.
+func setDomainLifecycleFields(log logr.Logger, wl *vzapi.VerrazzanoWebLogicWorkload, domain *unstructured.Unstructured) error {
+	if len(wl.Annotations[constants.LifecycleActionAnnotation]) > 0 && wl.Annotations[constants.LifecycleActionAnnotation] != wl.Status.LastLifecycleAction {
+		action := wl.Annotations[constants.LifecycleActionAnnotation]
+		if strings.EqualFold(action, constants.LifecycleActionStart) {
+			return startWebLogicDomain(log, domain)
+		}
+		if strings.EqualFold(action, constants.LifecycleActionStop) {
+			return stopWebLogicDomain(log, domain)
+		}
```
mbenmclean, line 857: Ok so if they enter a value that's not start or stop it will accept it but not do anything? Will the LastLifecycleAction be set?

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -810,10 +843,87 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+	}
+	if wl.Annotations != nil && wl.Annotations[constants.RestartVersionAnnotation] != wl.Status.LastRestartVersion {
```
mbenmclean, line 859: So if the user sets both the restartVersion and the lifecycle action, the first time through the lifecycle action will get hit and it will either stop or start based on that. But the restart version will still not match the LastRestartVersion so the next time that reconcile is triggered it will still try to restart. Also the domain could be either stopped or started at this point. How are you handling this edge case?

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -810,10 +843,87 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+		return restartWebLogic(log, domain, wl.Annotations[constants.RestartVersionAnnotation])
+	}
+	return nil
+}
+
+// Set domain restart version.  If it is changed from the previous value, then the WebLogic Operator will restart the domain
+func restartWebLogic(log logr.Logger, domain *unstructured.Unstructured, version string) error {
```
mbenmclean, line 866: Do we still need the restartVersion for weblogic. It might be better to just tell the user to use the stop lifecycle action and wait for the pods to spin down then use the start annotation. It might be less confusing

application-operator/constants/constants.go
pfmackin: it is different.  The restart version is a trigger to restart and it has to have some random number different from prev value.  The lifecycle action values have meaning and is a permanent life cycle change until the value changes (e.g. from stop to start)/

michael-cico: ok

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -810,10 +843,87 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+		log.Error(err, "Error setting restartVersion in domain")
+		return err
+	}
+	return nil
+}
+
+// Set the serverStartPolicy to stop WebLogic domain, return the current serverStartPolicy
+func stopWebLogicDomain(log logr.Logger, domain *unstructured.Unstructured) error {
+	// Return if serverStartPolicy is already never
+	currentServerStartPolicy, _, _ := unstructured.NestedString(domain.Object, specServerStartPolicyFields...)
```
mbenmclean, line 878: Should this check for an error?

michael-cico (approved):

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -810,10 +843,87 @@ func (r *Reconciler) addLoggingTrait(ctx context.Context, log logr.Logger, workl
...
+	if currentServerStartPolicy == Never {
+		return nil
+	}
+
+	// Save the last policy so that it can be used when starting the domain
+	if len(currentServerStartPolicy) == 0 {
+		currentServerStartPolicy = IfNeeded
+	}
+	annos, found, err := unstructured.NestedStringMap(domain.Object, metaAnnotationFields...)
```
mbenmclean, line 887: I don't love the annos var name but that's very nitpicky. I haven't seen that used anywhere I'd probably just use annotations.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
pfmackin: Yes it will be ignored.  It is set, i can fix this in next subtask.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
pfmackin: It will basically be a noop. For start, both the serverStartPolicy and resetVersion will get set.  For stop, the same thing.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
pfmackin: yes we need it since it is a general way to do a rolling restart.  The start/stop is not going to be documented, it is the interface used by the platform operator to stop/start WebLogic.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
mbenmclean: Oh really, so the platform operator goes in and sets both?

mbenmclean: Ok just want to make sure that after we stop weblogic, when the new app operator starts up and reconciles, that it doesn't detect the restartVersion changed and try to do a rolling restart and have things go haywire. Have you checked the app operator logs after the new one comes up to make sure that it is not entering the reconcile because of the restartVersion. It sounds like things are working in testing but it's possible this could be happening and you are not noticing it.

PR #1986, opened 2021-11-17
Add whitelist variable VZ_API_ALLOWED_ORIGINS to limit trusted origins for console

lake-of-dreams:
# Description

Added an environment variable VZ_API_ALLOWED_ORIGINS where we can specify a comma separated list of valid origins and limit the clients to which we send back the `Access-Control-Allow-Origin` headers.

Fixes VZ-4065

# Checklist 


platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -55,9 +55,11 @@ data:
         --      null in allowed origins.
 
         if h ~= null then
-            ngx.header["Access-Control-Allow-Origin"] = h
-            ngx.header["Access-Control-Allow-Credentials"] = "true"
-            ngx.header["Vary"] = "Origin"
+            if h == ingressUri or auth.isOriginAllowed(h) then
+                ngx.header["Access-Control-Allow-Origin"] = h
+                ngx.header["Access-Control-Allow-Credentials"] = "true"
+                ngx.header["Vary"] = "Origin"
```
wmhopkins, line 61: The "Vary" header should maybe be set even if we don't trust the given Origin.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -55,9 +55,11 @@ data:
         --      null in allowed origins.
 
         if h ~= null then
-            ngx.header["Access-Control-Allow-Origin"] = h
-            ngx.header["Access-Control-Allow-Credentials"] = "true"
-            ngx.header["Vary"] = "Origin"
+            if h == ingressUri or auth.isOriginAllowed(h) then
+                ngx.header["Access-Control-Allow-Origin"] = h
+                ngx.header["Access-Control-Allow-Credentials"] = "true"
```
wmhopkins, line 60: The recommendation in the bug report is to set Allow-Credentials to true even if the Origin isn't whitelisted. I'm a little confused by the advice to set the Allow-Origin to "null or the malicious page's domain", since "null" is not allowed as an Allow-Origin and the malicious page's domain is what we're trying to eliminate.

In the research I've done so far, I haven't seen it explicitly spelled out how a server should respond if it doesn't like the Origin. It seems you shouldn't return Allow-Origin, but is that enough? Do you still continue to process the request normally, and simply trust the browser not to share the data?

With respect to a "null" Origin, the approach I took was simply to not set any of the Access-Control headers.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -55,9 +55,11 @@ data:
         --      null in allowed origins.
 
         if h ~= null then
-            ngx.header["Access-Control-Allow-Origin"] = h
-            ngx.header["Access-Control-Allow-Credentials"] = "true"
-            ngx.header["Vary"] = "Origin"
+            if h == ingressUri or auth.isOriginAllowed(h) then
```
wmhopkins, line 58: We need to be mindful that this isn't just console processing; the same proxy is used for all console, VMI, and kubernetes API traffic, so we have to handle all cases. The list of whitelisted Origins will be different for each backend; in particular, some of the VMI components may reference each other; e.g., kibana making calls to elasticsearch. May need to make this processing backend-specific.

wmhopkins (commented):
Had a look at this and added some comments -- let me know what you think.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: Not sure why we should `Vary` header, looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary it seems to be related to caching related. Can you please explain why should we send `Vary` header at all?

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: The behvaiour still remains the same i.e. we process the header only when it is non-nil

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: made changes to allow whitelisting backend specific with separate env variable for each

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
wmhopkins: It was recommended by some of the CORS-handling descriptions I read. The reasoning is that the response returned by the server may be different depending on the Origin value specified in the request, so that the result shouldn't be cached, or if it is, should be cached on a per-Origin basis.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
wmhopkins: No, I understand that, but "null" is a special case. It's not a null or nil value, it's the string value "null". In that case, we act as if the header is not set at all; i.e., we ignore it. But in all other cases, we were processing the CORS headers and returning a valid response. (Although I see that null is not quoted above, so it's possible this code doesn't do what I thought it did -- null isn't a legitimate value in Lua, AFAIK.)

Now, with this change, for any non-whitelisted value of Origin, we do only partial processing. We don't set Allow-Origin or Allow-Credentials (although Allow-Credentials should still be set, according to some sources), but we do continue to process the request, and, in particular, we also add the Allowed-Methods and Allowed-Headers if it's an OPTIONS request. This is different from what we did for the null case.

I'm actually not sure what the proper/expected behavior is for the case that you want to disallow the Origin supplied in the request. None of the sources I looked at were very clear. Do you just not set Allowed-Origins? Should you set other Access-Control headers, or not? Should the request succeed or fail? (It doesn't seem very secure to me to return sensitive data even though you don't trust the Origin, but maybe that's the correct behavior since CORS is supposed to be enforced by the browser.)

I feel like I can't really evaluate this change without knowing for sure what the response should be to an invalid/denied Origin header. If we can't find out for sure, we should at least  clearly specify what *we* think the proper behavior should be, so that we can determine if the code is correct. Should we not set any of the CORS headers? Should we set some of them, selectively? Should the request overall succeed or fail? Should we process an OPTIONS request if the Origin is invalid? Etc.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
wmhopkins: Looks good. This will have to be tested carefully with all the VMI backends, to make sure we haven't broken anything (since we aren't whitelisting any origins other than the ingress). I suspect we'll need to provide some defaults for at least a couple of the backends -- which might want to be hard-wired -- one easy-ish fix might be to allow any VMI origin to access any VMI backend.

Note that I think most of the VMI tests are still disabled in the Verrazzano master pipelines. They run, but they're basically a no-op if the "flaky tests" aren't enabled, so a green run of the VZ pipelines may not find problems resulting from this change.

wmhopkins (commented):
Had a look at your comments and the updated code, and responded by replying to the original comments.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: ok, implemented

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: 1. As per https://datatracker.ietf.org/doc/rfc6454, "User Agent Requirements" section a "null" value for 
    Origin header is set by user agents for privacy-sensitive contexts. However it does not defined what 
    a privacy-sensitive context means. The "Privacy-Sensitive Contexts" section https://wiki.mozilla.org/Security/Origin
    defines certain contexts as privacy sensitive but there also it does not explain behaviour of server for the
    "Access-Control-Allow-Origin" header. Therefore we ignore the header processing.

lake-of-dreams: 1. As per https://datatracker.ietf.org/doc/rfc6454, "User Agent Requirements" section a "null" value for 
Origin header is set by user agents for privacy-sensitive contexts. However it does not defined what 
a privacy-sensitive context means. The "Privacy-Sensitive Contexts" section https://wiki.mozilla.org/Security/Origin
defines certain contexts as privacy sensitive but there also it does not explain behaviour of server for the
"Access-Control-Allow-Origin" header. Therefore we ignore the header processing.

2. From "Simple Cross-Origin Request, Actual Request, and Redirects" section of https://www.w3.org/TR/2020/SPSD-cors-20200602,
set the value of Access-Control-Allow-Origin and Access-Control-Allow-Credentials headers if there is a match.

3. From https://tools.ietf.org/id/draft-abarth-origin-03.html#server-behavior, if the request Origin is not in
whitelisted origins and the request method is not a safe non state changing (i.e. GET or HEAD), we should abort the request.

4. From https://fetch.spec.whatwg.org/#http-requests, we should allow headers and methods set in request Access-Control-Request-Method 
and Access-Control-Request-Headers headers.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: Currently no vmi frontend is directly accessing a different vmi backend and hence there is no use case where we have to default whitelist

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -39,40 +39,52 @@ data:
...
             ngx.header["Access-Control-Allow-Credentials"] = "true"
-            ngx.header["Vary"] = "Origin"
-
-            if ngx.req.get_method() == "OPTIONS" then
-                -- auth.info("Processing OPTIONS request, URI is '"..ngx.var.request_uri.."'")
-                ngx.header["Access-Control-Allow-Headers"] = "authorization, content-type"
-                ngx.header["Access-Control-Allow-Methods"] = "GET, HEAD, POST, PUT, DELETE, OPTIONS, PATCH"
-                ngx.header["Content-Length"] = 0
+        else
+            -- From https://tools.ietf.org/id/draft-abarth-origin-03.html#server-behavior, if the request Origin is not in
```
wmhopkins, line 61: But we should continue to process OPTIONS requests, presumably?

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -39,40 +39,52 @@ data:
...
+            local requestMethod = ngx.req.get_method()
+            if requestMethod ~= "GET" and requestMethod ~= "HEAD" then
+                auth.bad_request("Invalid Origin header: " .. h)
             end
         end
 
         if ngx.req.get_method() == "OPTIONS" then
+            -- From https://fetch.spec.whatwg.org/#http-requests, we should allow headers and methods set in 
+            -- request Access-Control-Request-Method and Access-Control-Request-Headers headers.
+            local allowedMethods, _ = ngx.req.get_headers()["access-aontrol-request-method"]
```
wmhopkins, line 73: We shouldn't just allow any header or method, though -- that's just as promiscuous as allowing any Origin. We should have a whitelisted set of headers and methods we allow. Not really sure how to generate the header list (don't remember how I came up with the current list), but we shouldn't allow the CONNECT or TRACE methods. Seems harmless to simply advertise the headers/methods we allow, regardless of what's requested. The browser should be reconciling the two lists.

wmhopkins (commented):
I think the logic here should really stay the same as it used to be, with one exception. Where it used to say:

    `if h ~= null then`

It should be changed to say:

    `if h ~= "null" and auth.isOriginAllowed(h, backend, ingressUri) then`

But all the other logic should stay the same. Basically, we were treating "null" as an invalid Origin before, so the treatment for other invalid Origins should be the same.

I think it's reasonable to add the logic that fails the request if there's a bad Origin and and request is not GET or HEAD, as long as OPTIONS is also handled without an error. And if we want to be more precise about allowed headers/methods, that's fine, too (as long as we have whitelists).

But the basic flow of the logic shouldn't be any different -- we're just increasing the set of Origins we won't accept from "null" to "null" + ingressUri + whitelisted URLs. We shouldn't need to change the overall flow of the request handling logic.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
wmhopkins: OK. That surprises me a little, but if so we should be fine.

wmhopkins:
> I think it's reasonable to add the logic that fails the request if there's a bad Origin and the request is not GET or HEAD, as long as OPTIONS is also handled without an error. If we want to be more precise about allowed headers/methods, that's fine, too, as long as we have whitelists, though not sure it's necessary.

Actually, just had a look at the referenced doc, and I don't see where it says to use Access-Control-Request-{Headers,Methods} to determine what goes in Access-Control-Allowed-{Headers,Methods}. It just says you should include the headers/methods that are acceptable to the server. Since this is only processed during the OPTIONS request, returning all allowed headers/methods, rather than just the one that may or may not be present as a Request-Header or Request-Method header would make sense -- the client may be able to cache the allowed set for future reference.

Given that, I think it makes sense to go back to the hard-wired whitelist we had before.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: there can not be an OPTIONS request without a Origin header

lake-of-dreams:
> I think the logic here should really stay the same as it used to be, with one exception. Where it used to say:
> 
> ```
> `if h ~= null then`
> ```
> 
> It should be changed to say:
> 
> ```
> `if h ~= "null" and auth.isOriginAllowed(h, backend, ingressUri) then`
> ```
> 
> But all the other logic should stay the same. Basically, we were treating "null" as an invalid Origin before, so the treatment for other invalid Origins should be the same.
> 
> I think it's reasonable to add the logic that fails the request if there's a bad Origin and the request is not GET or HEAD, as long as OPTIONS is also handled without an error. If we want to be more precise about allowed headers/methods, that's fine, too, as long as we have whitelists, though not sure it's necessary.
> 
> But the basic flow of the logic shouldn't be any different -- we're just increasing the set of Origins we won't accept from "null" to "null + ingressUri + whitelisted URLs". We shouldn't need to change the overall flow of the request handling logic.
The logic is still the same, only that we now check if header is not nil or empty or "null" for further processing and with the addition that if header is not an allowed one and the request is not GET or HEAD, we abort the request as per the recommendation

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: Added back the header list. Although I am not sure what the behaviour should be if the preflight request contains access-aontrol-request-method to be connect and trace but I can not think of scenarios that it will be applicable for our case.

wmhopkins:
> The logic is still the same, only that we now check if header is not nil or empty or "null" for further processing and with the addition that if header is not an allowed one and the request is not GET or HEAD, we abort the request as per the recommendation
>
> I think for the scope of this task these discussions can be taken to a separate jira because the main objective of this PR was to just add the whitelisting mechanism for Origin headers. As can be seen some improvements have been made in this PR itself but if we want to look again at the CORS implementation of the auth proxy, that should be a separate story and not block this.

I think there are changes to the logic that we don't really want. Shouldn't be a big deal to fix, but let's discuss offline.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -39,40 +39,41 @@ data:
...
             if ngx.req.get_method() == "OPTIONS" then
-                -- auth.info("Processing OPTIONS request, URI is '"..ngx.var.request_uri.."'")
-                ngx.header["Access-Control-Allow-Headers"] = "authorization, content-type"
-                ngx.header["Access-Control-Allow-Methods"] = "GET, HEAD, POST, PUT, DELETE, OPTIONS, PATCH"
-                ngx.header["Content-Length"] = 0
+              ngx.header["Access-Control-Allow-Headers"] = "authorization, content-type"
+              ngx.header["Access-Control-Allow-Methods"] = "GET, HEAD, POST, PUT, DELETE, OPTIONS, PATCH"
+              ngx.header["Content-Length"] = 0
+              ngx.status = 200
+              ngx.exit(ngx.HTTP_OK)
```
wmhopkins, line 72: This shouldn't be moved to the `if originAllowed` block; we want to handle OPTIONS even for the case that we aren't allowing the Origin. If it's a legitimate Origin, regardless of whether we want to authorize it or not, we should respond to an OPTIONS request with by not setting the *-Allow-* headers but returning HTTP_OK.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -1111,7 +1112,7 @@ data:
 
         return serverUrl
     end
-
+
```
wmhopkins, line 1115: Spurious whitespace change.

wmhopkins (changes requested):
Close, but not quite right. I really feel like the easiest/safest way to add the required changes without changing the original logic flow is to start with what's currently on tip of master, and make the following 4 changes:

- Change [line 57](https://github.com/verrazzano/verrazzano/blob/b099116306a545e02700512bba3059614924341b/platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml#L57) to read `if auth.isOriginAllowed(h, backend, ingressUri) then`.
- Move the `Vary` header setting above that line (as you currently have it).
- Add an `else if requestMethod ~= "GET" and requestMethod ~= "HEAD" and requestMethod ~= "OPTIONS" then` block to fail write requests if the Origin is forbidden.
- Move the `Content-Length` header into the separate block that returns HTTP_OK for ORIGIN requests.

I believe that will result in the smallest number of changed lines and ensure that only the changes we want are introduced.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: a space is added b/w this and next function so its actually ok

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: fixed

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-authproxy-configmap.yaml
lake-of-dreams: fixed

wmhopkins (approved):
Looks good to me.

PR #1995, opened 2021-11-17
Component retryable errors for install/upgrade

michael-cico:
# Description

Creates a custom error type to be returned from `spi.Component` to indicate to the controller that it should requeue, and is a non-fatal event. 

Also, adds support to `bom.KeyValue` and `HelmComponent` to allow a component to specify a custom helm file override.

# Checklist 

### DIFF
--- a/pkg/bom/bom.go
+++ b/pkg/bom/bom.go
@@ -110,6 +110,7 @@ type KeyValue struct {
 	Key       string
 	Value     string
 	SetString bool
+	IsFile    bool
 }
 
 // Create a new bom from a JSON file
--- a/pkg/controller/errors/errors.go
+++ b/pkg/controller/errors/errors.go
@@ -1,0 +1,41 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+package spi
+
+import (
+	"fmt"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"strings"
+)
+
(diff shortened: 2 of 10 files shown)


platform-operator/controllers/verrazzano/component/spi/errors.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: I think this is useful for all the controller in both operators.  Maybe put it in top level pkg.

platform-operator/controllers/verrazzano/component/spi/errors.go
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+package spi
+
+import "fmt"
+
+type RetryableError struct {
+	Component string
```
pfmackin, line 8: I thought this  struct would just have the original error plus a ctrl.Result{}.

michael-cico: OK, works for me.  I wasn't sure if folks wanted to record their errors locally, and if so if it made sense to propagate those up.

michael-cico: Pushed a second pass, where an optional error and/or result can be returned.

mbenmclean:
So does this change pass in the file overrides with -f instead of --set-file? That's what it looks like to me. Do you know if these do the same thing?

mbenmclean (dismissed):

mbenmclean (approved):

PR #2003, opened 2021-11-19
VZ-3985 upgrade test

pfmackin:
Add an e2e test to ensure that the application pods contain the Envoy 1.104 sidecar post-upgrade.  This proves that the applications were restarted and contain the new version of Envoy.

# Checklist 

### DIFF
--- a/tests/e2e/pkg/common.go
+++ b/tests/e2e/pkg/common.go
@@ -303,11 +303,11 @@ func CheckPodsForEnvoySidecar(namespace string, imageName string) bool {
 	// Every pod with istio enabled must containe the Envoy sidecar
 	for _, pod := range pods.Items {
 		// skip if istio sidecar disabled
-		v, ok := pod.Labels["sidecar.istio.io/inject"]
+		v := pod.Labels["sidecar.istio.io/inject"]
 		if v == "false" {
 			continue
 		}
-		_, ok = pod.Labels["istio.io/rev"]
+		_, ok := pod.Labels["istio.io/rev"]
 		if ok {
 			containers := pod.Spec.Containers
 			found := false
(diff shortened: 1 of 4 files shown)


tests/e2e/pkg/common.go
```diff
@@ -295,7 +296,17 @@ func CheckPodsForIstioImage(namespace string) bool {
...
 		return false
 	}
+	if len(pods.Items) == 0 {
+		Log(Info, fmt.Sprintf("No pods in namespace: %s, error: %v", namespace, err))
+		return false
+	}
+	// Every pod with istio enabled must containe the Envoy sidecar
 	for _, pod := range pods.Items {
+		// skip if istio sidecar disabled
+		v := pod.Labels["sidecar.istio.io/inject"]
```
mbenmclean, line 306: What pods will have this label?

tests/e2e/pkg/common.go
```diff
@@ -295,7 +296,17 @@ func CheckPodsForIstioImage(namespace string) bool {
...
+		if v == "false" {
+			continue
+		}
 		_, ok := pod.Labels["istio.io/rev"]
```
mbenmclean, line 310: Do we need both this and the check for sidecar.istio.io/inject?

tests/e2e/pkg/common.go
mbenmclean: I don't see this label on any pod. Do pods get deployed with this label by default or are we just checking to make sure it wasn't added on later?

anders-swanson (approved):

tests/e2e/pkg/common.go
anders-swanson: This label is new as of Istio 1.10, so it wouldn't have been applicable until now. It is a way of differentiating between 1.7.3 istio and 1.10.x+ Istio labels.

pfmackin: we actually should have a exclude list for pods that we know are not in the mesh; all other pods must have the annotation.  I am going to improve this after the break.

tests/e2e/pkg/common.go
pfmackin: coherence

PR #2014, opened 2021-11-22
fixed wls parse pattern to match more

shihchang:
# Description

Fix a case of 'pattern not matched' errors that the format13 pattern failed if the message field contains a ">", greater-than sign, in the message body.

Fixes VZ-3631

# Checklist 

### DIFF
--- a/application-operator/controllers/wlsworkload/weblogic_logging_test.go
+++ b/application-operator/controllers/wlsworkload/weblogic_logging_test.go
@@ -22,8 +22,7 @@ const log3 = `####<Nov 17, 2021 10:36:32,616 PM GMT> <Info> <RJVM> <bobs-booksto
  Http Enabled		 true
  Tunneling Enabled	 false
  Outbound Enabled	 false
- Admin Traffic Enabled	 true ResolveDNSName Enabled	 false> 
-
+ Admin Traffic Enabled	 true ResolveDNSName Enabled	 false>
 `
 const log4 = `####<Nov 17, 2021 10:37:35,553 PM GMT> <Info> <JDBC> <bobs-bookstore-managed-server1> <managed-server1> <[ACTIVE] ExecuteThread: '8' for queue: 'weblogic.kernel.Default (self-tuning)'> <<anonymous>> <> <c10cf642-4fbe-4753-9b2c-3c99b96df564-00000013> <1637188655553> <[severity-value: 64] [rid: 0] [partition-id: 0] [partition-name: DOMAIN] > <BEA-001156> <Stack trace associated with message 001129 follows: 
 
--- a/application-operator/controllers/wlsworkload/weblogic_logging_test.go
+++ b/application-operator/controllers/wlsworkload/weblogic_logging_test.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package wlsworkload
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	asserts "github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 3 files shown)


fredtibbitts:
I understand that this doesn't fix the original log interleaving issue, but does it change the behavior at all? Do we still report the log interleaving issue the same way and fail the test? I just want to make sure this isn't masking the original problem somehow.

tests/e2e/pkg/elasticsearch.go
```diff
@@ -319,10 +319,9 @@ func NoLog(index string, match []Match, mustNot []Match) bool {
 		return true
 	}
 	Log(Error, fmt.Sprintf("Found unexpected %d records", len(hits.([]interface{}))))
-	for i, hit := range hits.([]interface{}) {
-		if i < 10 {
-			Log(Error, fmt.Sprintf("Found unexpected log record: %v", hit))
-		}
+	for _, hit := range hits.([]interface{}) {
```
desagar, line 322: Won't this potentially log a LOT of error messages? Is it better to just count the length of the hits array and print one error msg saying you found n number of of unexpected records?

shihchang: The default is 50 or 100; we should not see that many but we may get more than 10.

shihchang:
@fredtibbitts The  interleaved messages should still fail in matching one of format1 ~ format13, as the message interleaved, and report one or two 'pattern not matched' errors

tests/e2e/pkg/elasticsearch.go
desagar: our tests already print too much cruft. I think it's better to count the number and just print one error msg that says how many we found.

shihchang: It did already print the count at line 322. I reduced the unexpected records from 10 to 5 as we will need these records to find out what caused the 'pattern not matched error' in the wls-server.log which could be huge.

tests/e2e/pkg/elasticsearch.go
```diff
@@ -320,7 +320,7 @@ func NoLog(index string, match []Match, mustNot []Match) bool {
 	}
 	Log(Error, fmt.Sprintf("Found unexpected %d records", len(hits.([]interface{}))))
 	for i, hit := range hits.([]interface{}) {
-		if i < 10 {
+		if i < 5 {
```
kminder, line 323: Should this also break instead of spinning through the reminder of the hits?

shihchang: yes, fixed

markxnelson (changes requested):
magic numbers

tests/e2e/pkg/elasticsearch.go
shihchang: It is the number of unexpected logs, or the "pattern not matched" logs, it would print out. So that we can locate the corresponding log entries in the huge adminserver.log. Some reviewer suggested we have already print/log too much so reduced from 10 to 5. We usually got 1 or 2; if we got more than 5, then the pattern should have obvious problem.

application-operator/controllers/wlsworkload/weblogic_logging_test.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	asserts "github.com/stretchr/testify/assert"
+)
+
+const log1 = `####<Nov 17, 2021 10:36:24,316 PM GMT> <Info> <Security> <bobs-bookstore-managed-server1> <> <main> <> <> <> <1637188584316> <[severity-value: 64] [partition-id: 0] [partition-name: DOMAIN] > <BEA-090905> <Disabling the CryptoJ JCE Provider self-integrity check for better startup performance. To enable this check, specify -Dweblogic.security.allowCryptoJDefaultJCEVerification=true.> 
+
```
markxnelson, line 15: are these blank lines significant?

markxnelson (dismissed):

application-operator/controllers/wlsworkload/weblogic_logging_test.go
shihchang: no, removed.

markxnelson (approved):

PR #2028, opened 2021-11-24
Validate request body for console and api

lake-of-dreams:
# Description

This change to validate that the body of a request targeted to console or verrazzano api is a well formed json.

Fixes VZ-4063

# Checklist 


wmhopkins (approved):
I think this might slow processing down significantly, but if it's only console or api requests they usually aren't too large, so hopefully not a noticeable difference.

lake-of-dreams:
>I think this might slow processing down significantly, but if it's only console or api requests they usually aren't too large, so hopefully not a noticeable difference.
Currently there is no request with body at all to console or api apart from tests so effect will be minimal

PR #2030, opened 2021-11-24
VZ-4124: Remove remaining Rancher scripts

anders-swanson:
Fixes VZ-4124


platform-operator/controllers/clusters/rancher.go
```diff
@@ -94,7 +95,7 @@ func registerManagedClusterWithRancher(rdr client.Reader, clusterName string, lo
 	rc.host = hostname
 
 	log.Debug("Getting Rancher TLS root CA")
-	caCert, err := getRancherTLSRootCA(rdr)
+	caCert, err := rancher.GetRootCA(rdr)
```
michael-cico, line 98: It looks like you are trying to share this, which is good, but probably create a common/rancher package or something under `platform-operator/controllers`, instead of having the `cluster` controller code point at the `verrazzano` controller component code.

platform-operator/controllers/verrazzano/component/common/pod.go
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
michael-cico, line 1: Is this general enough to go under the top-level `pkg` dir?  maybe under `pkg/k8sutil` ?

anders-swanson: yeah it is. will reorganize

platform-operator/controllers/clusters/rancher.go
anders-swanson: there's a decent amount of duplicated code in these two packages. I was having some trouble with import cycles by using a third package but it would work with some refactoring

platform-operator/controllers/verrazzano/component/rancher/rancher_client.go
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package rancher
```
michael-cico, line 4: Per earlier comment, probably move this up under `platform-operator/controllers` since it's used across multiple VPO controllers.

platform-operator/controllers/verrazzano/component/rancher/rancher_letsencrypt.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	bytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	c.cert = append(c.cert, bytes...)
+	return nil
+}
+
+func (c *certBuilder) buildLetsEncryptChain() error {
```
michael-cico, line 45: Maybe some comments around this stuff, explaining what the certs are and why we need them.

platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
```diff
@@ -43,69 +41,63 @@ func createAdminSecretIfNotExists(log *zap.SugaredLogger, c client.Client) error
...
-		return "", errors.New("failed to generate Rancher admin password, password is empty")
+		return "", fmt.Errorf("failed to reset Rancher admin password: %s", stderr)
 	}
 	return password, nil
 }
 
-// newAdminSecret generates the admin secret for rancher
+// hack to parse the stdout of Rancher reset password
+// we need to remove carriage returns and newlines from the stdout, since it is coming over from the pod's shell
+func parsePasswordStdout(stdout string) string {
```
michael-cico, line 74: this looks like it was a lot of fun... so there's no way to do this command from the rancher API, using the client code?

platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
```diff
@@ -90,13 +89,29 @@ func isUsingDefaultCACertificate(cm *vzapi.CertManagerComponent) bool {
 		cm.Certificate.CA.ClusterResourceNamespace == defaultSecretNamespace
 }
 
-func createAdditionalCertificates(log *zap.SugaredLogger, vz *vzapi.Verrazzano) error {
+func createAdditionalCertificates(log *zap.SugaredLogger, c client.Client, vz *vzapi.Verrazzano) error {
 	cm := vz.Spec.Components.CertManager
 	if (cm != nil && cm.Certificate.Acme != vzapi.Acme{} && useAdditionalCAs(cm.Certificate.Acme)) {
 		log.Infof("Creating additional Rancher certificates for non-production environment")
```
michael-cico, line 95: Wonder if this should be `Debugf`?  Trying to cut down on logging output generally I think now.

anders-swanson: agree

platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
anders-swanson: Rancher recommends to do this via the shell - i had done a cursory search for doing this via the HTTP API, but couldn't find anything. Maybe i missed it.

michael-cico (dismissed):

michael-cico (dismissed):

platform-operator/controllers/common/rancher.go
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
anders-swanson: it is used across multiple packages that interact with rancher. There's rancher component, and rancher multicluster that use this functionality

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -195,8 +196,8 @@ func (r rancherComponent) IsReady(ctx spi.ComponentContext) bool {
 		c := ctx.Client()
 		rancherDeploy := []types.NamespacedName{
 			{
-				Name:      ComponentName,
-				Namespace: ComponentNamespace,
+				Name:      common.RancherName,
```
pfmackin, line 199: same as before, common should not have rancher specific code

pfmackin (approved):

PR #2032, opened 2021-11-24
Package all overrides into a struct

mbenmclean:
This change adds support for --set-file helm overrides. It also packages all of the helm overrides into a struct. This struct has fields for the --set args, --set-string args, --set-file args, and -f args. This struct gets populated in the buildCustomHelmOverrides method. It defines a HelmComponent struct in the helm package to hold these overrides.

# Checklist 


pkg/bom/bom.go
```diff
@@ -110,6 +110,7 @@ type KeyValue struct {
 	Key       string
 	Value     string
 	SetString bool
+	SetFile   bool
```
michael-cico, line 113: Maybe we should add some comments here as to what these mean...

michael-cico: i.e., the difference between SetFile and IsFile... we can do that as a separate pass though.

platform-operator/internal/helm/helmcli.go
```diff
@@ -26,6 +26,14 @@ const ChartStatusFailed = "failed"
 // Package-level var and functions to allow overriding GetChartStatus for unit test purposes
 type ChartStatusFnType func(releaseName string, namespace string) (string, error)
 
+// HelmOverrides contains all of the overrides that gets passed to the helm cli runner
+type HelmOverrides struct {
+	SetOverrides       string   // for --set
```
michael-cico, line 31: Yeah, I think comments like these would be good in `bom.KeyValue`, but perhaps next pass.

michael-cico (approved):

platform-operator/internal/helm/helmcli.go
mbenmclean: I can throw them in now it's not a big deal

michael-cico (approved):

PR #2036, opened 2021-11-24
VZ-2528: Elasticsearch number of replicas wrong after upgrade

kminder:
# Description

Fixes issues that caused Elasticsearch to report a Yellow status.  This occurs when Elasticsearch has a single data node but is also configured to have index replicas.  This change allows Elasticsearch to manage the replica count so avoid this issue in both the Dev and Prod profiles.

Fixes VZ-2528: Elasticsearch number of replicas wrong after upgrade

Note that I decided to leave the code in the Verrazzano post upgrade hook for two reasons.
1. It didn’t seem to add significant time to the upgrade.
2. Once parallel upgrade is worked out this code won’t need to get found and move from the global post upgrade hook.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
+++ b/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
@@ -32,7 +32,6 @@ const containerName = "es-master"
 const portName = "http"
 const indexPattern = "verrazzano-*"
 
-
 var execCommand = exec.Command
 
 // ResolveVerrazzanoNamespace will return the default Verrazzano system namespace unless the namespace
@@ -150,7 +149,7 @@ func AppendOverrides(_ spi.ComponentContext, _ string, _ string, _ string, kvs [
 // fixupElasticSearchReplicaCount fixes the replica count set for single node Elasticsearch cluster
 func fixupElasticSearchReplicaCount(ctx spi.ComponentContext, namespace string) error {
 	// Only apply this fix to clusters with Elasticsearch enabled.
-	if *ctx.EffectiveCR().Spec.Components.Elasticsearch.Enabled {
+	if !*ctx.EffectiveCR().Spec.Components.Elasticsearch.Enabled {
 		ctx.Log().Info("Elasticsearch Post Upgrade: Replica count update unnecessary on managed cluster.")
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -136,3 +145,122 @@ func AppendOverrides(_ spi.ComponentContext, _ string, _ string, _ string, kvs [
...
+	ver1_1_0, err := semver.NewSemVersion("v1.1.0")
+	if err != nil {
+		return err
+	}
+	sourceVer, err := semver.NewSemVersion(ctx.ActualCR().Status.Version)
+	if err != nil {
+		ctx.Log().Errorf("Elasticsearch Post Upgrade: Invalid source Verrazzano version: %s", err)
+		return err
+	}
+	if sourceVer.IsGreatherThan(ver1_1_0) || sourceVer.IsEqualTo(ver1_1_0) {
```
michael-cico, line 167: nit pick, maybe do `!sourceVer.IsLessThan(ver1_1_0)` ?

michael-cico (approved):

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
kminder: I wrote it like this intentionally because I think this better matches the intended >= semantic.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -27,3 +27,20 @@ func NewComponent() spi.Component {
...
+	if err := c.HelmComponent.PostUpgrade(ctx); err != nil {
+		return err
+	}
+	return c.updateElasticsearchResources(ctx)
+}
+
+// updateElasticsearchResources updates elasticsearch resources
+func (c verrazzanoComponent) updateElasticsearchResources(ctx spi.ComponentContext) error {
+	if err := fixupElasticSearchReplicaCount(ctx, c.ChartNamespace); err != nil {
+		return err
```
pfmackin, line 43: i don’t think upgrade should fail if we cannot fix ES, should just be a warning.  Also, this would be better in elasticsearch.go

PR #2043, opened 2021-11-29
VZ-4112 Update log message for parsing components

davisgomes:
# Description

This fix includes an update to the logs that will make parsing for components easier. Additions to the logs look as follows:
`{"operation": "<install or upgrade>", "component": "<component-name>"}`

Fixes VZ-4112

# Checklist 


platform-operator/controllers/verrazzano/component/appoper/app_operator.go
```diff
@@ -159,3 +159,8 @@ func ApplyCRDYaml(log *zap.SugaredLogger, c client.Client, _ string, _ string, _
 
 	return nil
 }
+
+// Logger setup for the application operator install
+func appInstallLog(compContext spi.ComponentContext) *zap.SugaredLogger {
```
anders-swanson, line 164: Recommend adding a `*SugaredLogger` as a field to the component struct, and initialize in NewComponent(). This way we are not always making a new SugaredLogger, even for log levels which may be disabled.

anders-swanson: Also since these loggers are package private, it may be good to simply call them `log` or `logger`

davisgomes: Yea, I was struggling with how to ride the line of efficient and usable. I don't think this is possible because the logger requires the context of the current function which isn't inherent to the component object itself. At a minimum, we'd have to create a new logger for each function.

davisgomes: As for naming, I prefer the longer names for clarity, but I'm interested to hear other's opinions on usability.

michael-cico: An alternative is to create the logger as a member of the component object; for HelmComponent objects it should be part of the base structure and inherited.

I'm not sure why we started passing loggers around; it's no great shakes in my mind if we end up not using the logger passed in in the ComponentContext that much.  @pfmackin what do you think?

michael-cico:
Generally speaking, we should also use this logging pattern in `HelmComponent` methods also; so it'd be like

`ctx.Log().with("component", h.Name(), "operation", ...)`

platform-operator/controllers/verrazzano/component/appoper/app_operator.go
michael-cico: although, I'm not sure how much the thing I suggested would help; we won't know the operation until later during those calls.  

I think in general, what we can do is just use a local for the logger and pass that around within a component at the operation level, to avoid initializing a new logger every time as Anders points out.

It doesn't look like the logger creation is too expensive, but if we can optimize here and there that will help.

michael-cico: Another idea is, we could build something into Component context to make a copy with a refined logger, either at the component level or inside the component calls:

```
(c ComponentContext) func LogWith(args ...string) spi.Component {
   return componentContext {
       actualCR: c.actualCR,
       effectiveCR: c.effectiveCR,
       log: c.log.With(args...)
       ...
     }
}
```

or 

```
(c ComponentContext) func For(comp string) spi.Component {
   return componentContext {
       component: comp,
       actualCR: c.actualCR,
       effectiveCR: c.effectiveCR,
       log: c.log.With("component", comp)
       ...
     }
}

(c ComponentContext) func Operation(op string) spi.Component {
   return componentContext {
       component: c.component,
       operation: op,
       actualCR: c.actualCR,
       effectiveCR: c.effectiveCR,
       log: c.log.With("operation", op)
       ...
     }
}
```

and usage is something like

```
ctx ComponentContext = spi.NewContext(...)

ctx = ctx.For(component).Operation("install") // creates a copy of the ComponentContext, without re-merging the profiles
ctx.Log().Infof(...) // components just log as normal
```

davisgomes: Would this still run into the same issue? Each time we call `.For(...)` we would be creating a new logger with the `.With(...)` function?

davisgomes: Looking at the sugared logger documentation, they have functions Infow, Debugw, etc. That can include "with"commands. We could wrap the logger with our own definition those functions that include the `With` functionality.

platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -23,6 +23,10 @@ type ComponentContext interface {
 	IsDryRun() bool
 	// Copy returns a copy of the current context
 	Copy() ComponentContext
+	// For returns a copy of the current context with an updated logging component field
+	For(comp string) ComponentContext
+	// Operation specifies the logging operation field
+	Operation(op string) ComponentContext
```
michael-cico, line 29: We should store the component name and operation as fields in `componentContext`, and add `GetComponent()` and `GetOperation()` funcs to return the values.  If not set they can default to "undefined" or "unknown" or something.

We may want to know those at some point during processing with in a component, irrespective of logging.

platform-operator/tools.go
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
+//go:build tools
```
michael-cico, line 4: what do these do?

image-patch-operator/api/images/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
michael-cico, line 1: what's this for?

davisgomes: Same as previous comment

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -826,6 +826,7 @@ func (r *Reconciler) initializeComponentStatus(log *zap.SugaredLogger, cr *insta
 			if err != nil {
```
michael-cico, line 826: You should move the creation of the initial context outside of the loop

platform-operator/controllers/verrazzano/install.go
```diff
@@ -28,14 +28,14 @@ func (r *Reconciler) reconcileComponents(_ context.Context, log *zap.SugaredLogg
...
 
-	compContext, err := spi.NewContext(log, r, cr, r.DryRun)
-	if err != nil {
-		return newRequeueWithDelay(), err
-	}
-
 	// Loop through all of the Verrazzano components and upgrade each one sequentially for now; will parallelize later
 	for _, comp := range registry.GetComponents() {
 		compName := comp.Name()
+		compContext, err := spi.NewContext(log, r, cr, r.DryRun)
```
michael-cico, line 34: same comment as above, move initial creation outside the loop

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -49,6 +49,7 @@ func (r *Reconciler) reconcileUpgrade(log *zap.SugaredLogger, cr *installv1alpha
 		if err != nil {
```
michael-cico, line 49: same comment, move initial context creation outside the loop.

michael-cico: So, create `upgradeContext` outside the loop, and use a different variable name for the nested context used inside the loop.

anders-swanson (dismissed):

platform-operator/controllers/verrazzano/controller.go
davisgomes: Done!

michael-cico (approved):

PR #2044, opened 2021-11-30
VZ-4122 move EnvoyFilter creation to go code

pfmackin:
Remove the script that creates the Istio EnvoyFilter and create it using go code.  Also, rename istio_utils.go to reflect the code in that file.

# Checklist 


platform-operator/controllers/verrazzano/component/istio/envoy_filter.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/yaml"
+)
+
+// Envoy filter yaml
+const filterYaml = `
```
anders-swanson, line 17: matter of style, but there is now a general function for applying YAML from a directory: https://github.com/verrazzano/verrazzano/blob/master/pkg/k8sutil/k8sutil.go#L167
Snippets like this can be saved in YAML files instead of as in-line strings.

pfmackin: ok, is there some go best practice not to do this?  I would actually prefer to see the string in the same file as the code unless it is huge or there are many long strings.

anders-swanson (approved):

PR #2045, opened 2021-11-30
mbenmclean/vz-3933-mysql-install

mbenmclean:
Fixes VZ-3933

This change moves the mysql install into platform operator and out of the scripts. It builds the overrides needed to call the helm upgrade command. This includes create the temp file necessary to initiatlize the mysql db and also handles the pv claims in the effective CR.

# Checklist 


platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,245 @@ package mysql
...
+	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/status"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"os"
+	controllerruntime "sigs.k8s.io/controller-runtime"
 )
 
 // ComponentName is the name of the component
```
michael-cico, line 25: nit, orphaned comment?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,245 @@ package mysql
...
-	kvs = append(kvs, bom.KeyValue{
-		Key:   helmPwd,
-		Value: string(secret.Data[mysqlKey]),
-	})
-	kvs = append(kvs, bom.KeyValue{
-		Key:   helmRootPwd,
-		Value: string(secret.Data[mysqlRootKey]),
-	})
+	cr := compContext.EffectiveCR()
+	deployed, err := helmutil.IsReleaseDeployed(ComponentName, vzconst.KeycloakNamespace)
```
michael-cico, line 57: I'm not sure this is how we should be handling this here...I think before this move effort, the AppendOverrides was only being called at upgrade time, when the secret should exist.  So, waiting for the MySQL deployment here doesn't seem to make sense, since this is called before we install MySQL.

I'm not 100% sure on this, we may need to check with Bhat on this.  But likely we need to create the secret in the `PreInstall` hook for install-time, and handle things differently for upgrade.

I might not be fully understanding what needs to happen here, but this looks a bit suspicious as it is now.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,245 @@ package mysql
...
+	}
+	return nil
+}
+
+func generateVolumeSourceOverrides(compContext spi.ComponentContext, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	effectiveCR := compContext.EffectiveCR()
+	defaultVolumeSpec := effectiveCR.Spec.DefaultVolumeSource
+
+	// keycloak was not specified in CR so return defaults
+	if effectiveCR.Spec.Components.Keycloak == nil {
```
michael-cico, line 164: Is this check necessary?  I think this check for emptyDir is redundant with the check later on in this function.

Basically, the volume source override for MySQL should be determined by
1. the volume source defined in the MySQL component
2. if not defined in the MySQL component, use the default volume source
3. If the result of those checks is nil, then there's no override.

Once the correct volume source is determined, you can set things based on `EmptyDir` or a PVC-template as you're doing below.  So I think you can drop the code from lines 162-173.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: So I believe the reason that we save the secrets and append them as overrides is that so they don't get reset during upgrade. All this check is doing is skipping these overrides if mysql isn't installed yet. I think that the creation of the secret gets taken care of during the install.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: Well I think we need a check for keycloak != nil before `mysqlVolumeSource := effectiveCR.Spec.Components.Keycloak.MySQL.VolumeSource` because if keycloak is nil then the line checking for MySQL VolumeSource will throw an  error.
Also we need to see if the defaultVolumeSpec is emptyDir because we need to explicitly set persistence.enabled=false because it's true by default. Should the order of these things get moved around?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+		ns.Labels["istio-injection"] = "enabled"
+		return nil
+	}); err != nil {
+		return ctrlerrrors.RetryableError{Source: ComponentName, Cause: err}
+	}
+	return nil
+}
+
+// PostInstall Patch the controller service ports based on any user-supplied overrides
+func PostInstall(ctx spi.ComponentContext) error {
```
pfmackin, line 121: not capitalized -  postInstall

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+		return []bom.KeyValue{}, ctrlerrrors.RetryableError{Source: ComponentName, Cause: err}
+	}
+	// Convert NGINX install-args to helm overrides
+	kvs = append(kvs, helm.GetInstallArgs(getInstallArgs(cr))...)
+
+	return kvs, nil
+}
+
+// PreInstall Create and label the MySQL namespace, and create any override helm args needed
+func PreInstall(compContext spi.ComponentContext, namespace string) error {
```
pfmackin, line 100: not capitalized - preInstall

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+func IsEnabled(context spi.ComponentContext) bool {
+	keycloak := context.EffectiveCR().Spec.Components.Keycloak
+	if keycloak != nil && keycloak.Enabled != nil {
+		return *keycloak.Enabled
+	}
+	return false
+}
+
 // AppendMySQLOverrides appends the the password for database user and root user.
 func AppendMySQLOverrides(compContext spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
```
pfmackin, line 54: not capitalized - append...

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+var pvc100Gi, _ = resource.ParseQuantity("100Gi")
+
+func IsReady(context spi.ComponentContext, name string, namespace string) bool {
+	deployments := []types.NamespacedName{
+		{Name: name, Namespace: namespace},
+	}
+	return status.DeploymentsReady(context.Log(), context.Client(), deployments, 1)
+}
+
+func IsEnabled(context spi.ComponentContext) bool {
```
pfmackin, line 45: not capitalized - isEnabled

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+	helmPwd          = "mysqlPassword"
+	helmRootPwd      = "mysqlRootPassword"
+	mysqlKey         = "mysql-password"
+	mysqlRootKey     = "mysql-root-password"
+	mysqlDBFile      = "create-mysql-db.sql"
 )
 
+var pvc100Gi, _ = resource.ParseQuantity("100Gi")
+
+func IsReady(context spi.ComponentContext, name string, namespace string) bool {
```
pfmackin, line 38: not capitalized - isReady

pfmackin: why does name, namespace have to be passed in.  Isn't this always in keycloak ns?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: It looks like many of the other components. Should the Pre and Post install functions all be local?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: Same question

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: Ok np

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+	ns := v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	if _, err := controllerruntime.CreateOrUpdate(context.TODO(), compContext.Client(), &ns, func() error {
+		if ns.Labels == nil {
+			ns.Labels = make(map[string]string)
+		}
+		ns.Labels["verrazzano.io/namespace"] = namespace
+		ns.Labels["istio-injection"] = "enabled"
+		return nil
+	}); err != nil {
+		return ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
```
mbenmclean, line 115: @michael-cico am I doing this the right way? I am returning retryable errors for all of these functions. Is there anything you see here that shouldn't have a retryable error?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: OK, so it's not blocking or trying to requeue here then right?  My bad.  I knew we needed to preserve the secrets, just wasn't sure if we were trying to wait for the deployment.

thinking this through, I think if there's an error returned from this during upgrade it will fail the upgrade...but, I think this is something where we can't continue otherwise, preserving the credentials is kind of critical.  Really more of an issue around fixing upgrade, and we can have users force a retry of upgrade right?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: Anything that is not intended to be called from outside the package should not be exported from it.  So the `Component` interface overrides all have to be exported, but anything called from those, or from inherited `HelmComponent` interface impls via the component function hooks can be package-private.

michael-cico: This also includes, constants, etc.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: Maybe... so effectively I think it should be something like

```
var mySQLVolumeSource *corev1.VolumeSource
if effectiveCR.Spec.Components.Keycloak != nil {
   mySQLVolumeSource = effectiveCR.Spec.Components.Keycloak.MySQL.VolumeSource 
}
if mySQLVolumeSource == nil {
  mySQLVolumeSource = effectiveCR.Spec.DefaultVolumeSource
}

....

if mySQLVolumeSource == nil {
...
```

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: I think this is fine for now.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+		nsName := types.NamespacedName{
+			Namespace: vzconst.KeycloakNamespace,
+			Name:      secretName}
+
+		err = compContext.Client().Get(context.TODO(), nsName, secret)
+		if err != nil {
+			return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
+		}
+
+		// Force mysql to use the initial password and root password during the upgrade, by specifying as helm overrides
```
pfmackin, line 71: is there any harm in doing this for install also?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+func postInstall(ctx spi.ComponentContext) error {
+	if ctx.IsDryRun() {
+		ctx.Log().Infof("MySQL postInstall dry run")
+		return nil
+	}
+	// Delete create-mysql-db.sql after install
+	return os.Remove(os.TempDir() + "/" + mysqlDBFile)
+}
+
+func createDBFile(ctx spi.ComponentContext) error {
```
pfmackin, line 130: Can you add a comment for this function.  I don't know why it is creating a temp file.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+	fmt.Println(os.Getwd())
+	tmpDBFile, err := os.Create(os.TempDir() + "/" + mysqlDBFile)
```
pfmackin, line 132: When does this file get deleted?

mbenmclean: In the postinstall

mbenmclean: I am not able to use a defer to clean it up because the file needs to exist after the appendOverrides function exits

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+			Key:   helmRootPwd,
+			Value: string(secret.Data[mysqlRootKey]),
+		})
+		kvs = append(kvs, bom.KeyValue{
+			Key:   helmPwd,
+			Value: string(secret.Data[mysqlKey]),
+		})
+	}
+	kvs = append(kvs, bom.KeyValue{Key: mysqlUsernameKey, Value: mysqlUsername})
+	if !deployed {
```
mbenmclean, line 82: Maybe go to check for upgrade when that goes in

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,244 @@ package mysql
...
+		kvs = append(kvs, bom.KeyValue{
+			Key:   "persistence.enabled",
+			Value: "true",
+		})
+	}
 	return kvs, nil
 }
+
+// findVolumeTemplate Find a named VolumeClaimTemplate in the list
+func findVolumeTemplate(templateName string, templates []vzapi.VolumeClaimSpecTemplate) (*v1.PersistentVolumeClaimSpec, bool) {
```
mbenmclean, line 233: move to common function in vzconfig

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,263 @@ package mysql
...
+	helmRootPwd         = "mysqlRootPassword"
+	mySQLKey            = "mysql-password"
+	mySQLRootKey        = "mysql-root-password"
+	mySQLInitFilePrefix = "init-mysql-"
 )
 
-// AppendMySQLOverrides appends the the password for database user and root user.
-func AppendMySQLOverrides(compContext spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
-	secret := &corev1.Secret{}
+var pvc100Gi, _ = resource.ParseQuantity("100Gi")
```
michael-cico, line 40: This looks like a unit test thing?  Probably should be in that file.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: yeah, seems like a viable alternative.  What happens if we just always create the override?  If MySQL is already deployed, will it ignore it if unchanged?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -5,42 +5,263 @@ package mysql
...
 	nsName := types.NamespacedName{
 		Namespace: vzconst.KeycloakNamespace,
 		Name:      secretName}
+	// Get the mysql secret
+	err := compContext.Client().Get(context.TODO(), nsName, secret)
+	if err != nil && !k8serror.IsNotFound(err) {
+		return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
+	}
+	// see if the mysql exists
+	if !k8serror.IsNotFound(err) {
```
michael-cico, line 73: Not a blocker, but what you could do above is

```
if err != nil {
  if !k8serror.IsNotFound(err) {
      return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
  }
  kvs = append(kvs, bom.KeyValue{
	Key:   helmRootPwd,
	Value: string(secret.Data[mySQLRootKey]),
  })
  kvs = append(kvs, bom.KeyValue{
	  Key:   helmPwd,
	  Value: string(secret.Data[mySQLKey]),
  })
}
```

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: I think I created one already, but next pass...

michael-cico: I think there's a few things we can look to share consolidate, need to take a pass at that next sprint

platform-operator/controllers/verrazzano/component/mysql/mysql_component.go
```diff
@@ -4,33 +4,60 @@
...
-			ChartNamespace:          "keycloak",
+			ChartNamespace:          vzconst.KeycloakNamespace,
 			IgnoreNamespaceOverride: true,
+			SupportsOperatorInstall: true,
+			ImagePullSecretKeyname:  secret.DefaultImagePullSecretKeyName,
 			ValuesFile:              filepath.Join(config.GetHelmOverridesDir(), "mysql-values.yaml"),
-			AppendOverridesFunc:     AppendMySQLOverrides,
+			AppendOverridesFunc:     appendMySQLOverrides,
+			Dependencies:            []string{istio.ComponentName},
+			ReadyStatusFunc:         isReady,
```
michael-cico, line 40: you don't need this if you're overriding `Component.IsReady` like you are below.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/mysql/mysql_test.go
```diff
@@ -0,0 +1,329 @@
...
+			UnavailableReplicas: 1,
+		},
+	})
+	assert.False(t, isReady(spi.NewFakeContext(fakeClient, nil, false), "", vzconst.KeycloakNamespace))
+}
+
+// TestSQLFileCreatedAndDRemoved tests the creation and deletion of the mysql db init file
+// WHEN the appendMySQLOverrides and then postInstall functions are called
+// THEN ensure that the mysql db init file is created successfully and then deleted successfully
+func TestSQLFileCreatedAndDRemoved(t *testing.T) {
```
pfmackin, line 320: looks like extra D in DRemoved. Can be fixed as follow up

pfmackin (approved):

PR #2058, opened 2021-12-01
VZ-4122 rewrite Istio cert creation in go code

pfmackin:
Remove the bash scripts and openssl calls to create the Istio cert and write the code in go.


platform-operator/controllers/verrazzano/component/istio/istio_cert.go
```diff
@@ -0,0 +1,83 @@
...
+	"context"
+	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/certificate"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"time"
+)
+
+// createCertSecret creates the secret with cert data used by Istio MTLS
```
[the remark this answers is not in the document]
michael-cico: Never mind, that's something separate we can try to do later.  I thought it might be easy, but I'm not sure if that'll be more work that what you've already done.

platform-operator/internal/k8s/certificate/selfsigned.go
```diff
@@ -0,0 +1,181 @@
...
+// CertResult contains the generated cert results
+type CertResult struct {
+	PrivateKey    *rsa.PrivateKey
+	PrivateKeyPEM []byte
+	Cert          *x509.Certificate
+	CertPEM       []byte
+}
+
+// CreateSelfSignedCert creates a self signed cert and returns the generated PEM data
+func CreateSelfSignedCert(rootConfig CertConfig, intermConfig CertConfig) (*CertPemData, error) {
```
wmhopkins, line 60: The name of this method is a little misleading -- it's actually creating a self-signed cert and then another, intermediate-CA cert.

platform-operator/internal/k8s/certificate/selfsigned.go
```diff
@@ -0,0 +1,181 @@
...
+
+	return createCert(rootCert, rootCert, nil)
+}
+
+// Create the intermediate certificate
+func createIntermediateCert(config CertConfig, rootResult *CertResult) (*CertResult, error) {
+	// create the intermediate certificate info needed to create the certificate
+	intermediateCert := &x509.Certificate{
+		DNSNames:     []string{config.CommonName},
+		SerialNumber: rootResult.Cert.SerialNumber,
```
wmhopkins, line 119: I think this is wrong -- the intermediate CA needs its own serial number, not the root cert's serial number. The serial number for the root cert can probably be random, but the serial number for the intermediate CA should be unique among all certs generated by the self-signed root. And given that the root *is* self-signed, maybe the root serial number should also be unique among all serials generated by that cert. That said, since we'll only ever generate those two certs, as long as the serial numbers are unique and different, that's probably OK. Only possible glitch would be if the intermediate CA expired, and we wanted to regenerate from the root, we'd need another unique serial number.

platform-operator/internal/k8s/certificate/selfsigned.go
```diff
@@ -0,0 +1,181 @@
...
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageDigitalSignature | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+	}
+	return createCert(intermediateCert, rootResult.Cert, rootResult.PrivateKey)
+}
+
+// Create the certificate. The parent certificate will be used to sign the new cert
+func createCert(partialCert *x509.Certificate, parentCert *x509.Certificate, parentPrivKey *rsa.PrivateKey) (*CertResult, error) {
+	// create private key for this cert
+	privKey, err := rsa.GenerateKey(cryptorand.Reader, 4096)
```
wmhopkins, line 139: We only want to do this if the parentPrivKey parameter is nil, correct?

platform-operator/internal/k8s/certificate/selfsigned.go
```diff
@@ -0,0 +1,181 @@
...
+		IsCA:                  true,
+		AuthorityKeyId:        rootResult.Cert.SubjectKeyId,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageDigitalSignature | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+	}
+	return createCert(intermediateCert, rootResult.Cert, rootResult.PrivateKey)
+}
+
+// Create the certificate. The parent certificate will be used to sign the new cert
+func createCert(partialCert *x509.Certificate, parentCert *x509.Certificate, parentPrivKey *rsa.PrivateKey) (*CertResult, error) {
```
wmhopkins, line 137: The partialCert param would be better named "csr", or maybe csrStruct, or tbsCert (To Be Signed) right? Is it not effectively a CSR?

platform-operator/internal/k8s/certificate/selfsigned.go
```diff
@@ -0,0 +1,181 @@
...
+	Cert          *x509.Certificate
+	CertPEM       []byte
+}
+
+// CreateSelfSignedCert creates a self signed cert and returns the generated PEM data
+func CreateSelfSignedCert(rootConfig CertConfig, intermConfig CertConfig) (*CertPemData, error) {
+	// Create the object that will be loaded with the PEM data
+	pem := CertPemData{}
+
+	serialNumber, err := newSerialNumber()
```
wmhopkins, line 64: Where is `newSerialNumber()` declared/defined?

wmhopkins (changes requested):
Overall, looks good. I think the serial number is being set incorrectly for the intermediate CA, and I'm wondering where newSerialNumber() is defined and what it does. Other than that, had some comments on cosmetic issues, but nothing else substantive.

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: I changed it to CreateCertUsingSelfSignedRoot

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: no we still need the private key for the intermediate cert so that it can be stored in the secret.  Also, it is needed to get the public key from it.

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: The type is x509.Certificate so it is not exactly a CSR but sort of acting like one.  I can call it CSR if that is clearer

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: it is in with certificate.go, which already existed

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: fixed

platform-operator/internal/k8s/certificate/selfsigned.go
pfmackin: I am calling it certRequest

wmhopkins (approved):
lgtm.

PR #2059, opened 2021-12-01
MySQL second pass cleanup

mbenmclean:
This is a second pass on https://github.com/verrazzano/verrazzano/commit/34bf6e23945a2cd00aa275c3de462206f073aac3
and resolves comments from https://github.com/verrazzano/verrazzano/pull/2045
Contents:
- Leverages the operation field in the component context
- Moves findVolumeTemplate to common code and removes all duplicate non-exported functions
- Cleans up misplaced constant
- Cleans up typo

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/mysql/mysql.go
+++ b/platform-operator/controllers/verrazzano/component/mysql/mysql.go
@@ -6,6 +6,7 @@ package mysql
 import (
 	"context"
 	"fmt"
+
 	"github.com/verrazzano/verrazzano/pkg/bom"
 	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
@@ -14,6 +15,7 @@ import (
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/status"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/vzconfig"
+
 	"io/ioutil"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/platform-operator/controllers/verrazzano/component/mysql/mysql.go
+++ b/platform-operator/controllers/verrazzano/component/mysql/mysql.go
@@ -12,12 +12,10 @@ import (
 	vzconst "github.com/verrazzano/verrazzano/platform-operator/constants"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/helm"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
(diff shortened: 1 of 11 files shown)


platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -60,17 +56,17 @@ func isEnabled(context spi.ComponentContext) bool {
...
-		Namespace: vzconst.KeycloakNamespace,
-		Name:      secretName}
-	// Get the mysql secret
-	err := compContext.Client().Get(context.TODO(), nsName, secret)
-	if err != nil && !k8serror.IsNotFound(err) {
-		return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
-	}
-	// see if the mysql exists
-	if !k8serror.IsNotFound(err) {
+	if compContext.For(ComponentName).GetOperation() == "upgrade" {
```
michael-cico, line 59: We should put constants for these in the platform-operator/constants/constants.go

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -60,17 +56,17 @@ func isEnabled(context spi.ComponentContext) bool {
...
+		secret := &v1.Secret{}
+		nsName := types.NamespacedName{
+			Namespace: vzconst.KeycloakNamespace,
+			Name:      secretName}
+		// Get the mysql secret
+		err := compContext.Client().Get(context.TODO(), nsName, secret)
+		if err != nil {
+			compContext.Log().Errorf("Error getting mysql secret: %v", err)
+			return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
```
michael-cico, line 68: This is the right thing to do, but for now will still cause us to fail upgrade until we fix that.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -84,12 +80,7 @@ func appendMySQLOverrides(compContext spi.ComponentContext, _ string, _ string,
 
 	kvs = append(kvs, bom.KeyValue{Key: mySQLUsernameKey, Value: mySQLUsername})
 
-	// See if MySQL is deployed and create the MySQL init file if not
-	deployed, err := helmutil.IsReleaseDeployed(ComponentName, vzconst.KeycloakNamespace)
-	if err != nil {
-		return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
-	}
-	if !deployed {
+	if compContext.For(ComponentName).GetOperation() == "install" {
```
michael-cico, line 83: Same comment, we should create constants for the op names

michael-cico (approved):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mbenmclean: Well what the desired behavior. We could either fail upgrade or log that the mysql secret was not found so the password will be reset. What do you think

michael-cico: My comment was I htink this is the right thing to do, lol.  I probably should have stayed silent :-)

PR #2060, opened 2021-12-01
VZ-4111.  Make istioctl install non-blocking

michael-cico:
# Description

Use a simple goroutine to background the `istioctl install` command, which is blocking by default.

Fixes VZ-4111

# Checklist 


platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -115,19 +207,32 @@ func (i IstioComponent) Install(compContext spi.ComponentContext) error {
...
+		result := true
+		log.Debugf("Starting istioctl install...")
+		stdout, stderr, err := installFunc(log, overrideStrings, files...)
+		log.Debugf("istioctl stdout: %s", string(stdout))
+		if err != nil {
+			result = false
+			log.Errorf("Unexpected error %s during install, stderr: %s", err.Error(), string(stderr))
+		}
+		log.Debugf("Completed istioctl install successfully")
+		monitor.sendResult(result)
```
anders-swanson, line 233: Does the caller ever care about the error here? It is squashed by the goroutine.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -115,19 +207,32 @@ func (i IstioComponent) Install(compContext spi.ComponentContext) error {
...
+	}()
+	return ctrlerrors.RetryableError{Source: ComponentName}
```
anders-swanson, line 235: Why do we return an error here?

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -46,6 +53,62 @@ func setBashFunc(f bashFuncSig) {
...
+	case result := <-m.resultCh:
+		return result, nil
+	default:
+		return false, ctrlerrors.RetryableError{Source: ComponentName}
+	}
+}
+
+//reset - reset the monitor and close the channel
+func (m *installMonitorType) reset() {
+	m.running = false
```
anders-swanson, line 94: do we need m.running, can we not check the status of the chan?

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -46,6 +53,62 @@ func setBashFunc(f bashFuncSig) {
...
+	m.resultCh = make(chan bool)
+}
+
+//isRunning - returns true of the monitor/goroutine are active
+func (m *installMonitorType) isRunning() bool {
+	return m.running
+}
+
+// monitor - Install monitor for the component
+var monitor installMonitor = &installMonitorType{}
```
anders-swanson, line 110: can monitor be a struct member of istio component rather than package level variable?

platform-operator/controllers/verrazzano/component/istio/istio_install.go
anders-swanson: oh I get it, it's because install is running.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -115,19 +207,32 @@ func (i IstioComponent) Install(compContext spi.ComponentContext) error {
...
+	return forkInstallFunc(compContext, overrideStrings, files)
+}
 
-	return nil
+//forkInstall - istioctl install blocks, fork it into the background
+func forkInstall(compContext spi.ComponentContext, overrideStrings string, files []string) error {
+	log := compContext.Log()
+	log.Debugf("Creating background install goroutine for Istio")
+	monitor.init()
+	go func() {
```
anders-swanson, line 222: What if instead of using a goroutine for just the install, the entire lifecycle check runs in a goroutine? e.g., the full install or the full upgrade?

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: I thought about that, we want to retry on error, so the goroutine records it, and the component only cares about whether or not it finished and what the result was.  In the case that the install command fails, we want it to try again.

If I bubble the error up and return it when the component detects it, it is an extra requeue just to get back and try again.  So in this case when the component detects it failed, it just drops through and runs the install operation again.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: right

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: From what I read, it's not straightforward to detect if a channel is open or not without trying to read from it.  I'll take another look and see if it can be avoided.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: I thought about that, but when we call registry.GetComponents() we return new instances of the components as an array each time.  I can review changing that later.  Ideally we should only instantiate the registry once.

I was also thinking about adding pass-back objects via the ComponentContext, but then that has to be tracked somewhere.

I think if we can store the monitor as a member of the struct that's better.  Maybe if we just store the component array as a package-private list in `registry` that solves the problem, I'll give that a shot.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: Yeah, we want to do that for all components...so this might end up getting thrown away.  We wanted to do something simple for the Istio component right now, as that's a bigger change.

michael-cico: Since this might be short-term, I was trying to find the cheapest way to do it without spending too much effort on it.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -26,14 +27,20 @@ import (
...
 
 var installFunc installFuncSig = istio.Install
 
+type forkInstallFuncSig func(compContext spi.ComponentContext, overrideStrings string, files []string) error
+
+var forkInstallFunc forkInstallFuncSig = forkInstall
+
 type bashFuncSig func(inArgs ...string) (string, string, error)
 
 var bashFunc bashFuncSig = vzos.RunBash
```
mbenmclean, line 46: In the past I've used fake runners for this. Do you prefer this?

mbenmclean (approved):

platform-operator/controllers/verrazzano/component/registry/registry.go
```diff
@@ -27,6 +27,22 @@ type GetCompoentsFnType func() []spi.Component
 
 var getComponentsFn = getComponents
 
+var componentsRegistry = []spi.Component{
+	nginx.NewComponent(),
```
anders-swanson, line 31: I don't this will work in a var declaration, since it is using Config which is dynamically populated by program arguments. You will lose any program arguments to Config this way and only use the defaults.

anders-swanson: Initialization must be delayed until the first call of getComponents

michael-cico: good catch, I'll make that change.  It seemed to work, but probably by accident.

michael-cico: pushed

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: In some cases here I wanted to avoid invoking the goroutine in parallel and simulate the response.  The actual `forkInstall` function is tested separately.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -46,12 +53,65 @@ func setBashFunc(f bashFuncSig) {
...
+//reset - reset the monitor and close the channel
+func (m *installMonitorType) reset() {
+	m.running = false
+	close(m.resultCh)
+}
+
+//init - init the channel and running indicator
+func (m *installMonitorType) init() {
+	m.running = true
+	m.resultCh = make(chan bool)
```
pfmackin, line 101: I think this needs to be a buffered channel per https://gobyexample.com/channel-buffering

By default channels are unbuffered, meaning that they will only accept sends (chan <-) if there is a corresponding receive (<- chan) ready to receive the sent value.
--

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -115,22 +204,35 @@ func (i IstioComponent) Install(compContext spi.ComponentContext) error {
...
+	return forkInstallFunc(compContext, i.monitor, overrideStrings, files)
+}
 
-	return nil
+//forkInstall - istioctl install blocks, fork it into the background
+func forkInstall(compContext spi.ComponentContext, monitor installMonitor, overrideStrings string, files []string) error {
+	log := compContext.Log()
+	log.Debugf("Creating background install goroutine for Istio")
+	monitor.init()
+	go func() {
```
pfmackin, line 219: why aren't we using a channel to send the data needed by the background thread?

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: Fixed next push.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
michael-cico: fixed and pushed.

pfmackin (approved):

PR #2061, opened 2021-12-01
VZ 3503: Tests 

anders-swanson:


anders-swanson:
there are some extra diffs because I merged master. You can ignore anything that isn't Keycloak package related.

PR #2076, opened 2021-12-03
VZ 4218

anders-swanson:
Fixes VZ 4218 / Typo

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
@@ -5,6 +5,7 @@ package rancher
 
 import (
 	"context"
+	"fmt"
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common"
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
```diff
@@ -63,6 +64,9 @@ func copyDefaultCACertificate(log *zap.SugaredLogger, c client.Client, vz *vzapi
 		if err := c.Get(context.TODO(), namespacedName, defaultSecret); err != nil {
 			return err
 		}
+		if len(defaultSecret.Data[caCert]) < 1 {
```
mgianatagh, line 67: It might be valid for this to have a length of zero when using well known certs. @guoyongzhang can you take a look?

platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
```diff
@@ -63,6 +64,9 @@ func copyDefaultCACertificate(log *zap.SugaredLogger, c client.Client, vz *vzapi
 		if err := c.Get(context.TODO(), namespacedName, defaultSecret); err != nil {
```
bobdonat, line 64: do we need to retry if this secret is not found.  the scripts did this because there was a window where the secret had not been created yet and would be created.

anders-swanson: Retry is implicitly handled by returning an error. It's bubbled up to the component controller which will retry.

bobdonat: sounds good

platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
guoyongzhang: The secret `verrazzano-ca-certificate-secret` is for cert `verrazzano-ca-certificate` which is a CA.  So the ca.crt field should always be populated with self-signed CA cert.
In case of Letsencrypt or customer provided CA, we wouldn't have this secret `verrazzano-ca-certificate-secret`.

guoyongzhang (approved):
So it didn't work because we had field name ca.cert instead of ca.crt?  
If so, the commit is good.

PR #2079, opened 2021-12-06
Stshapir/vz 4021

stevshap:
# Description

This pull request adds the framework to push metrics from test suites using the VzDescribe wrappers. Only ci/kind tests currently have Prometheus credentials enabled; will be resolved in future changes. Also resolves error of invalid metrics name failure

Fixes VZ-9999

# Checklist 

### DIFF
--- a/pkg/test/framework/ginkgo_wrapper.go
+++ b/pkg/test/framework/ginkgo_wrapper.go
@@ -8,7 +8,7 @@ import (
 	"reflect"
 	"time"
 
-	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2"
 	"github.com/verrazzano/verrazzano/tests/e2e/pkg"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/test/framework/ginkgo_wrapper.go
```diff
@@ -69,8 +70,23 @@ func VzAfterEach(body interface{}) bool {
...
+		startTime := time.Now()
+
+		pkg.Log(pkg.Info, fmt.Sprintf("Describe block %q started - placeholder for making API call to emit test related metric(s)", ginkgo.CurrentSpecReport().LeafNodeText))
+		reflect.ValueOf(body).Call([]reflect.Value{})
+		pkg.Log(pkg.Info, fmt.Sprintf("Describe block %q ended - placeholder for making API call to emit test related metric(s)", ginkgo.CurrentSpecReport().LeafNodeText))
+
+		endTime := time.Now()
+		durationMillis := float64(endTime.Sub(startTime) / time.Millisecond)
+
+		if EmitGauge(text, "duration", durationMillis) != nil {
```
markxnelson, line 83: `text` might not be the best choice here?  are there any implication for it having spaces in it - how does prometheus handle that?  
i wonder if there is a better way of coming up with a good metric name - maybe we want the hierarchy, or the file at least?

markxnelson: i see you have a func makeMetricName, but you don't seem to use it here? - oh i see you call it down further  - ok

markxnelson (approved):

pkg/test/framework/ginkgo_wrapper.go
```diff
@@ -6,8 +6,9 @@ package framework
 import (
 	"fmt"
 	"reflect"
+	"time"
 
-	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2"
```
bobdonat, line 11: looks like the PR was already merged.
my understanding is we decided to use a dot import for ginkgo
should be 
. "github.com/onsi/ginkgo/v2"

stevshap: ack

PR #2089, opened 2021-12-07
Fix chart override keys for fluentd extra volume mounts

fredtibbitts:
# Description

The helm chart override keys in `verrazzano-logging.yaml` do match what the code is setting in the platform operator. As a result the `extraVolumeMounts` settings in the Verrazzano CR were being ignored. This fixes the key names.

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
@@ -566,8 +566,8 @@ spec:
             - mountPath: /run/log/journal
               name: run-log-journal
               readOnly: true
-{{- if .Values.logging.extraVolumeMounts }}
-{{- range $i, $e := .Values.logging.extraVolumeMounts }}
+{{- if .Values.fluentd.extraVolumeMounts }}
+{{- range $i, $e := .Values.fluentd.extraVolumeMounts }}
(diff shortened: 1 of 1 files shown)


platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
```diff
@@ -566,8 +566,8 @@ spec:
             - mountPath: /run/log/journal
               name: run-log-journal
               readOnly: true
-{{- if .Values.logging.extraVolumeMounts }}
-{{- range $i, $e := .Values.logging.extraVolumeMounts }}
+{{- if .Values.fluentd.extraVolumeMounts }}
```
michael-cico, line 569: is this the only place these are used?

michael-cico (approved):

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
fredtibbitts: As far as I can tell, yes.

michael-cico: OK.  Once this is in, I can review, maybe I'll flip it around to be consistent with the `logging` settings after, if needed.

PR #2093, opened 2021-12-09
VZ-3827 Add ability to set a DestinationRule with HTTP Cookie for session affinity

bryantidd:
# Description

Fixes VZ-9999

# Checklist 


application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -534,6 +539,56 @@ func (r *Reconciler) mutateVirtualService(virtualService *istioclient.VirtualSer
...
+										Ttl:  ptypes.DurationProto(rule.Destination.HTTPCookie.TTL)},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+
+		res, err := controllerutil.CreateOrUpdate(ctx, r.Client, destinationRule, func() error {
```
desagar, line 571: controllerutil.CreateOrUpdate first does a GET of the resource _into_ the object corresponding to the pointer you pass in, and _then_ calls the mutate function to give you a chance to edit it. 

The mutate function is where you're supposed to take the resource (if any) that already exists, and modify it.

If the destination rule already exists and is not in sync with the ingress trait, the way you're doing might not correct it to be in sync. The typical way to achieve this is:

- Create a destinationRule object that only has namespace and name populated
- Call controllerutil.CreateOrUpdate
- Do _all_ the manipulation inside the mutate function

bryantidd: Will do, Thanks

desagar: and given how complicated this is, probably would be good to add some coverage in the unit test

wmhopkins (approved):
lgtm.

PR #2094, opened 2021-12-09
Anders/k8sutil fakes

anders-swanson:
- Refactor ExecCmd code to use k8sutil.ExecPod
- Add `fake` package to pkg/k8sutil for faking k8s client stuff, especially w.r.t. ExecPod and SPDYExecutor
- Add ensure-default-admin command to Rancher PostInstall
- Add another UT to Keycloak as the test coverage dropped below 75% following the ExecPod refactor


platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
```diff
@@ -58,7 +62,12 @@ func resetAdminPassword(c client.Client) (string, error) {
 		return "", errors.New("no Rancher pods found")
 	}
 	pod := podList.Items[0]
-	stdout, stderr, err := k8sutil.ExecPod(cfg, restClient, &pod, common.RancherName, []string{resetPasswordCommand})
+	// Ensure the default rancer admin user is present
+	_, _, err = k8sutil.ExecPod(cli, cfg, &pod, common.RancherName, []string{ensureAdminCommand})
```
michael-cico, line 66: Might be good to log the `stderr` here?  We don't care about the stdout if it's successful of course.

platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
```diff
@@ -58,7 +62,12 @@ func resetAdminPassword(c client.Client) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	stdout, stderr, err := k8sutil.ExecPod(cli, cfg, &pod, common.RancherName, []string{resetPasswordCommand})
```
michael-cico, line 70: There was also logic to make sure the user was present before continuing, but perhaps that may not be necessary right?  I'm guessing if we call the `ensure` command, and then this fails, we should cycle back through and try again.

michael-cico: I think what was missing before was just forcing the admin user to exist if necessary

platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
anders-swanson: wrapped the error with stderr

michael-cico (approved):

PR #2101, opened 2021-12-10
VZ-4255 - skeleton webhook for metrics workloads

mgianatagh:
# Description

Create skeleton webhook for metrics workloads

Fixes VZ-4255

# Checklist 

### DIFF
--- a/application-operator/controllers/webhooks/scrape_generator_webhook.go
+++ b/application-operator/controllers/webhooks/scrape_generator_webhook.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/verrazzano/verrazzano/application-operator/internal/certificates"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -19,8 +20,6 @@ const ScrapeGeneratorloadPath = "/scrape-generator"
 
 var workloadLogger = ctrl.Log.WithName("webhooks.scrape-generator")
 
-var count = 0
-
 // WorkloadWebhook type for workload mutating webhook
 type WorkloadWebhook struct {
 	client.Client
@@ -31,8 +30,7 @@ type WorkloadWebhook struct {
 
 // Handle - handler for the mutating webhook
 func (a *WorkloadWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
-	count++
-	workloadLogger.Info(fmt.Sprintf("entered mutating webhook for group: %s, version: %s, kind: %s, namespace: %s, count: %d", req.Kind.Group, req.Kind.Version, req.Kind.Kind, req.Namespace, count))
+	workloadLogger.Info(fmt.Sprintf("entered %s webhook for group: %s, version: %s, kind: %s, namespace: %s, count: %d", certificates.ScrapeGeneratorWebhookName, req.Kind.Group, req.Kind.Version, req.Kind.Kind, req.Namespace))
 
-	return admission.Allowed("work in progress")
+	return admission.Allowed("not implemented yet")
 }
--- a/application-operator/controllers/webhooks/scrape_generator_webhook.go
+++ b/application-operator/controllers/webhooks/scrape_generator_webhook.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -38,18 +39,18 @@ func (a *ScrapeGeneratorWebhook) Handle(ctx context.Context, req admission.Reque
 	scrapeGeneratorLogger.Info(fmt.Sprintf("group: %s, version: %s, kind: %s, namespace: %s", req.Kind.Group, req.Kind.Version, req.Kind.Kind, req.Namespace))
 
 	// Determine what type of resource to handle
-	switch req.Kind.Kind {
-	case "Pod":
+	switch strings.ToLower(req.Kind.Kind) {
+	case "pod":
 		return a.handlePod(ctx, req)
-	case "Deployment":
+	case "deployment":
 		return a.handleDeployment(ctx, req)
-	case "ReplicaSet":
+	case "replicaset":
 		return a.handleReplicaSet(ctx, req)
-	case "StatefulSet":
+	case "statefulset":
 		return a.handleStatefulSet(ctx, req)
-	case "Domain":
+	case "domain":
 		return a.handleDomain(ctx, req)
-	case "Coherence":
+	case "coherence":
 		return a.handleCoherence(ctx, req)
 	default:
 		scrapeGeneratorLogger.Info(fmt.Sprintf("unsupported kind %s", req.Kind.Kind))
--- a/application-operator/controllers/webhooks/scrape_generator_webhook.go
+++ b/application-operator/controllers/webhooks/scrape_generator_webhook.go
@@ -20,6 +20,9 @@ import (
 // ScrapeGeneratorLoadPath specifies the path of scrape-generator webhook
 const ScrapeGeneratorLoadPath = "/scrape-generator"
 
+// StatusReasonSuccess
+const StatusReasonSuccess = "success"
+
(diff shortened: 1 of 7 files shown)


application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+	KubeClient    kubernetes.Interface
+	DynamicClient dynamic.Interface
+}
+
+// Handle - handler for the mutating webhook
+func (a *ScrapeGeneratorWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
+	scrapeGeneratorLogger.Info(fmt.Sprintf("group: %s, version: %s, kind: %s, namespace: %s", req.Kind.Group, req.Kind.Version, req.Kind.Kind, req.Namespace))
+
+	// Determine what type of resource to handle
+	switch strings.ToLower(req.Kind.Kind) {
```
kminder, line 42: Can leave this for now but I think this webhook should be agnostic to the actual kind.  It should be treated as an Unknown and either a template will match or it won't.  The expressions in the template need to be correct for the matched kind(s).

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -430,6 +430,39 @@ webhooks:
...
+        name: verrazzano-application-operator
+        namespace: verrazzano-system
+        path: "/scrape-generator"
+    rules:
+      - operations: ["CREATE"]
+        apiGroups: ["*"]
+        apiVersions: ["*"]
+        resources: ["deployments","coherences","domains","pods","replicasets","statefulsets"]
+        scope: "Namespaced"
+    sideEffects: None
```
kminder, line 457: I think we'll need to change sideEffects and possibly failurePolicy before we are finished but I don't know the correct answer right now.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -430,6 +430,39 @@ webhooks:
...
+    failurePolicy: Fail
+    matchPolicy: Exact
```
kminder, line 459: I think matchPolicy: Equivalent is generally recommended.
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -430,6 +430,39 @@ webhooks:
...
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values: ["kube-system", "verrazzano-system", "verrazzano-mc"]
+    clientConfig:
+      service:
+        name: verrazzano-application-operator
+        namespace: verrazzano-system
+        path: "/scrape-generator"
+    rules:
+      - operations: ["CREATE"]
```
bobdonat, line 452: Does update need to be supported too

mgianatagh: fixed in next push

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
mgianatagh: fixed in next push

application-operator/controllers/webhooks/scrape_generator_webhook.go
mgianatagh: Leaving for now, but in order for the webhook to mutate an object, isn't it going to need to know the kind and then decode it into the appropriate object type?

application-operator/main.go
```diff
@@ -212,6 +214,26 @@ func main() {
 			},
 		)
 
+		// Register the mutating webhook for plain old kubernetes objects workloads when the object exists
```
mbenmclean, line 217: I'm not sure if we should be checking the phrase plain old kubernetes objects

application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+	case "pod":
+		return a.handlePod(ctx, req)
+	case "deployment":
+		return a.handleDeployment(ctx, req)
+	case "replicaset":
+		return a.handleReplicaSet(ctx, req)
```
mbenmclean, line 48: Do you need a handler for Deployments and ReplicaSets? I don't entirely understand the difference but I thought ReplicaSets get deployed as deployments.
Also do you need a handler for Daemonsets. You probably don't because I don't know if user apps can even deploy those but just asking.

application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+	case "statefulset":
+		return a.handleStatefulSet(ctx, req)
+	case "domain":
+		return a.handleDomain(ctx, req)
+	case "coherence":
+		return a.handleCoherence(ctx, req)
+	default:
+		scrapeGeneratorLogger.Info(fmt.Sprintf("unsupported kind %s", req.Kind.Kind))
+		return admission.Allowed("not implemented yet")
```
mbenmclean, line 57: Will this get called on non-workload types of resources too?

application-operator/controllers/webhooks/scrape_generator_webhook.go
davisgomes: I think you can create ReplicaSet outside of a Deployment, but a Deployment creates ReplicaSet. So there might be a case where a ReplicaSet exists, but a Deployment does not.

application-operator/controllers/webhooks/scrape_generator_webhook.go
davisgomes: It depends on what you are mutating in the object. If you are just updating ObjectMeta fields. You can get away with using an unstructured.Unstructured: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured

application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// InjectDecoder injects the decoder.
+func (a *ScrapeGeneratorWebhook) InjectDecoder(d *admission.Decoder) error {
+	a.Decoder = d
+	return nil
+}
+
+// handlePod - handle Kind type of Pod
+func (a *ScrapeGeneratorWebhook) handlePod(ctx context.Context, req admission.Request) admission.Response {
```
davisgomes, line 68: If we find a way to be type agnostic, we should be able to combine these "handle" functions into one

application-operator/main.go
davisgomes: Maybe "standard" might be a better term?

application-operator/controllers/webhooks/scrape_generator_webhook.go
mgianatagh: It should not, but I wanted to log if the webhook got called for something we didn't expect.  The MutatingWebhookConfiguration defines "resources: ["deployments","coherences","domains","pods","replicasets","statefulsets"]".  If anyone added to that definition, the webhook would get called and log the unsupported kind message.

application-operator/controllers/webhooks/scrape_generator_webhook.go
mgianatagh: Yes, but I wasn't seeing an obvious way yet.  So I created some infrastructure for the different kinds.  But maybe it goes away later.

application-operator/main.go
mgianatagh: @kminder do you feel this is an issue as written?

application-operator/controllers/webhooks/scrape_generator_webhook.go
mbenmclean: Ok sounds good

application-operator/main.go
kminder: I think it is fine spelled out like this.

kminder (approved):

PR #2120, opened 2021-12-13
No JIRA.  Update BOM validator to ignore Rancher additional images

michael-cico:
# Description

Updates the BOM validator to ignore Rancher images that get installed in the background.  

At present we need to reference these images for private registry installs, however, Rancher does not strictly install what we declare.  Rancher requires a "min version" for these components, so in some cases (like upgrade) it may meet that requirement, but not match what's in our BOM.

# Checklist 


tools/bom-validator/bom-validator.go
```diff
@@ -40,6 +40,10 @@ type imageError struct {
 	clusterImageTags [tagLen]string
 }
 
+var ignoreSubComponents = []string{
```
pfmackin, line 43: Wondering if it make sense to have a  BOM ignore field for a subcomponent so that the BOM itself specifies ignore validation.  Anyway, this looks ok.

pfmackin (approved):

PR #2139, opened 2021-12-14
Basic Emit Metrics

anders-swanson:
Metric would look like:
```
{
   "timestamp":1639520348298250000,
   "caller":"metrics/ginkgo_logger.go:56",
   "suite_id":"ba2b8cff-40fa-490a-97ed-ce43c2a88038",
   "package":"metrics_test",
   "status":"passed",
   "attempts":1,
   "test":"Should do another thing"
}
```

Client has full control to emit additional metrics by operating on the SugaredLogger.

### DIFF
--- a/pkg/test/framework/ginkgo_wrapper.go
+++ b/pkg/test/framework/ginkgo_wrapper.go
@@ -5,11 +5,11 @@ package framework
 
 import (
 	"fmt"
-	"reflect"
-	"time"
-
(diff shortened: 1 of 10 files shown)


anders-swanson:
There's more metadata that can be gathered from ginkgo if we want to add more fields. I think this is a fine starting point though.

bryantidd (approved):

bryantidd (approved):

PR #2172, opened 2021-12-17
Remove special characters from returned password.  Make sure pw is sp…

ddrechse:
…ecified length

# Description

This change removes special characters being returned in a password generated by generatePassword
This caused a problem with a kcadm set-password command where the password had an "-" in the first character
Now, all passwords will just be comprised of alphanumerics

Fixes VZ-9999

# Checklist 

### DIFF
--- a/pkg/security/password/password.go
+++ b/pkg/security/password/password.go
@@ -24,7 +24,24 @@ func GeneratePassword(length int) (string, error) {
 		return "", err
 	}
 	pw := b64.URLEncoding.EncodeToString(b)
+	pw, err = makeAlphaNumeric(pw)
+	if err != nil {
+		return "", err
+	}
+	if len(pw) < length {
+		return GeneratePassword(length)
+	}
 	return pw[:length], nil
+}
+
+// makeAlphaNumeric removes all special characters from a password string
+func makeAlphaNumeric(input string) (string, error) {
+	// Make a Regex to say we only want letters and numbers
+	reg, err := regexp.Compile("[^a-zA-Z0-9]+")
+	if err != nil {
+		return "", err
+	}
+	return reg.ReplaceAllString(input, ""), nil
 }
 
 //MaskFunction creates a function intended to mask passwords which are substrings in other strings
--- a/pkg/security/password/password.go
+++ b/pkg/security/password/password.go
@@ -18,18 +18,16 @@ func GeneratePassword(length int) (string, error) {
 	if length < 1 {
 		return "", fmt.Errorf("cannot create password of length %d", length)
 	}
-	b := make([]byte, length)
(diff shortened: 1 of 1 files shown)


pkg/security/password/password.go
```diff
@@ -24,9 +24,26 @@ func GeneratePassword(length int) (string, error) {
 		return "", err
 	}
 	pw := b64.URLEncoding.EncodeToString(b)
+	pw, err = makeAlphaNumeric(pw)
```
pfmackin, line 27: Why wouldn't we just base64 encode?

ddrechse: wait, what happened to base58 which would require a new package?  base64.Stdencoding includes characters that break http calls. It broke our tests that use the keycloak rest API when I was using that one

pkg/security/password/password.go
```diff
@@ -24,9 +24,26 @@ func GeneratePassword(length int) (string, error) {
 		return "", err
 	}
 	pw := b64.URLEncoding.EncodeToString(b)
```
wmhopkins, line 26: Again, I would recommend using standard base64 encoding, not the URLEncoding. URL encoding generates a lot of characters we can't use. Regular base64 encoding generates just three, and likely not very many of them, at least compared to URLEncoding, so less biasing of the raw random output.

Pauls suggestion for base58 seems promising, but based on a quick read, it seems to be more concerned with not using 0 and O, which can be confused by humans, or i, I, and l. Not clear to me if it eliminates +, /, or =.

If base58 generates **only** alphanumeric chars, then it's probably the best choice. If not, then
I think we should just use standard base64 here, and strip out +, / and =.

Note that in the shell based version I did, I used 3*password-length as the size of the initial random data, just to be 100% sure I wouldn't go below the requested password length when stripping out +, /, and =. Base64 expands the input by about a third, IIRC, but making sure the original input is oversize makes sure we can't be too small even if the input is perfectly tuned to yield the most possible +, /, and = in the base64 output.

wmhopkins (changes requested):

pkg/security/password/password.go
ddrechse: std base64 broke the Keycloak ATs. Will, you were the one that suggested UrlEncoding when we discussed that

wmhopkins: Yes, that was my mistake, I was thinking of a different alternative encoding.

We'll still need to strip characters with the standard encoding, but we'll have to strip far fewer characters than with URLEncoding.

ddrechse: Done

pkg/security/password/password.go
ddrechse: done

PR #2194, opened 2021-12-21
JIRA:VZ-4313 generate cert per ingress

jmaron99:
# Description

This PR is attempting to address the issues seen with the todo-list app tests during application deletion.  The code now creates a certificate per ingress trait that is named based on namespace, app name, and trait name.  Previously certificates were named based on namespace and app name which could lead to naming conflicts and could possibly explain issues during application deletion(?).  The code is further enhanced with clearer logging.

The primary issue observed is that the ingress trait is removed during finalization prior to the removal of the certificate/secret associated with the ingress trait.  Once the trait is removed the code to remove the certificate is no longer executed (there is no trait to fetch and therefore no way of reconstructing the cert/secret name).  I'd ask the reviewers to focus on identifying any holes in the logic that would allow the finalizer to be removed and allow trait deletion prior to the removal of the cert/secret.  I see no such current hole in the logic, but maybe I've been looking at it too long....

Fixes VZ-4313

# Checklist 


examples/todo-list/todo-list-application.yaml
```diff
@@ -21,6 +21,8 @@ spec:
         - trait:
             apiVersion: oam.verrazzano.io/v1alpha1
             kind: IngressTrait
+            metadata:
```
pfmackin, line 24: does this force the app to name the ingress trait?  We currently don't do that for any of the examples.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -467,8 +480,13 @@ func (r *Reconciler) createGatewayCertificate(ctx context.Context, trait *vzapi.
...
 			return err
 		}
+		appName, ok := trait.Labels[oam.LabelAppName]
+		if !ok {
+			r.Log.Info("OAM app name label missing from metadata, will use wildcard for certificate host")
+			appName = "*"
+		}
 		certificate.Spec = certapiv1alpha2.CertificateSpec{
-			DNSNames:   []string{fmt.Sprintf("*.%s", appDomainName)},
+			DNSNames:   []string{fmt.Sprintf("%s.%s", appName, appDomainName)},
```
pfmackin, line 489: why is the willdcard being removed?

jmaron99: since we are creating a cert per ingress I thought it appropriate to associate to the specific ingress host.  If no app name label is found it defaults to wildcard.  I can go back to wildcard if you feel that's more appropriate, but this seemed like a valid approach.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -519,7 +538,7 @@ func buildCertificateNameFromAppName(trait *vzapi.IngressTrait) (string, error)
 	if !ok {
 		return "", errors.New("OAM app name label missing from metadata, unable to generate certificate name")
 	}
-	return fmt.Sprintf("%s-%s-cert", trait.Namespace, appName), nil
+	return fmt.Sprintf("%s-%s-%s-cert", trait.Namespace, appName, trait.Name), nil
```
pfmackin, line 541: what if the trait doesn't have a name?  Might be better to use the component name.

examples/todo-list/todo-list-application.yaml
jmaron99: No, the name isn't required - it just made it easier to test with a predictable name.  Otherwise the test would have to be modified to do wildcard searches, find the correct ingress/cert/secret, etc

application-operator/controllers/ingresstrait/ingresstrait_controller.go
jmaron99: If no name is specified the runtime creates a name that is composed of the component name and a uid

pfmackin (approved):

examples/todo-list/todo-list-application.yaml
mbenmclean: Does bobs books require a change too?

PR #2196, opened 2021-12-21
WKO 3.3.7

anders-swanson:
Uptake WKO 3.3.7


pfmackin (approved):

anders-swanson:
The build failures are not due to auxiliary images, rather they are related to other issues. Merging anyway, we can revert if there is strong sentiment against it.

PR #2197, opened 2021-12-21
VZ 4355: Ensure namespace deletion during tests

anders-swanson:


tests/e2e/pkg/kubernetes.go
```diff
@@ -521,6 +521,16 @@ func CreateNamespace(name string, labels map[string]string) (*corev1.Namespace,
 	return ns, nil
 }
 
+func RemoveNamespaceFinalizers(namespace *corev1.Namespace) error {
+	clientset, err := k8sutil.GetKubernetesClientset()
+	if err != nil {
+		return err
+	}
+	namespace.ObjectMeta.Finalizers = nil
```
pfmackin, line 529: won't this remove annotations?

anders-swanson: let me check. although it may not matter since we are using this in the context of cleaning up a namespace

pfmackin: yeah if you are deleting the NS then it doesn't matter

anders-swanson: annotations are not removed

pfmackin (approved):

PR #2200, opened 2021-12-21
VZ-4334: Fix Elasticsearch parsing errors

fredtibbitts:
# Description

The Fluentd logs have thousands of error messages stating that Elasticsearch has rejected log messages due to parsing errors when running acceptance tests. The errors are due to the same log field being logged as both a string and a complex type.

There are two fields that were being rejected: "trait" and "name". The "trait" field was being logged as a string in some cases and a `NamespacedName` in other cases. One bit of unexpected behavior I found is that when you call `Log.WithValues(...)` and pass it a `NamespacedName` type, it automatically converts the field to a string because the struct implements a `String()` function. I changed all of the places where the `NamespacedName` was being logged as a string so that "trait" is always logged as a complex type.

The "name" field was being logged as a string type in `ingresstrait_webhook.go` and logged as a `NamespacedName` in controllers. I changed the name of the field from "name" to "workload" in the controllers so there is no longer a collision on that field. 

I captured a cluster dump on successful run and verified that the Fluentd logs no longer report any rejected messages.

Fixes VZ-4334

# Checklist 

### DIFF
--- a/application-operator/controllers/helidonworkload/helidonworkload_controller.go
+++ b/application-operator/controllers/helidonworkload/helidonworkload_controller.go
@@ -74,9 +74,9 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	var workload vzapi.VerrazzanoHelidonWorkload
 	if err := r.Get(ctx, req.NamespacedName, &workload); err != nil {
 		if k8serrors.IsNotFound(err) {
-			log.Info("VerrazzanoHelidonWorkload has been deleted", "name", req.NamespacedName)
+			log.Info("VerrazzanoHelidonWorkload has been deleted")
 		} else {
-			log.Error(err, "Failed to fetch VerrazzanoHelidonWorkload", "name", req.NamespacedName)
+			log.Error(err, "Failed to fetch VerrazzanoHelidonWorkload")
 		}
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller.go
@@ -341,7 +341,7 @@ func (r *Reconciler) fetchWorkloadDefinition(ctx context.Context, workload *unst
 	workloadName := convertAPIVersionAndKindToNamespacedName(workloadAPIVer, workloadKind)
 	workloadDef := v1alpha2.WorkloadDefinition{}
 	if err := r.Get(ctx, workloadName, &workloadDef); err != nil {
-		r.Log.Error(err, "Failed to fetch workload definition", "name", workloadName)
+		r.Log.Error(err, "Failed to fetch workload definition", "workload", workloadName)
 		return nil, err
 	}
 	return &workloadDef, nil
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller.go
@@ -7,11 +7,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+	"strings"
+
 	"github.com/verrazzano/verrazzano/application-operator/controllers/clusters"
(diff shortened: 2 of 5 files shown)


application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -293,7 +293,7 @@ func (r *Reconciler) reconcileTraitCreateOrUpdate(ctx context.Context, trait *vz
 // The finalizer is only added if the trait is not being deleted and the finalizer has not previously been added
 func (r *Reconciler) addFinalizerIfRequired(ctx context.Context, trait *vzapi.MetricsTrait) error {
 	if trait.GetDeletionTimestamp().IsZero() && !vzstring.SliceContainsString(trait.Finalizers, finalizerName) {
-		traitName := vznav.GetNamespacedNameFromObjectMeta(trait.ObjectMeta).String()
+		traitName := vznav.GetNamespacedNameFromObjectMeta(trait.ObjectMeta)
 		r.Log.V(1).Info("Adding finalizer from trait", "trait", traitName)
```
pfmackin, line 297: I thought the description says we should not use NSN in the log message.

fredtibbitts: Sorry if the description was confusing, no, I think we should log NSN as a complex type, not as a string. My changes result in `trait` always being logged as a NSN.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -176,7 +176,7 @@ func (r *Reconciler) createOrUpdateChildResources(ctx context.Context, trait *vz
 // The finalizer is only added if the trait is not being deleted and the finalizer has not previously been added
 func (r *Reconciler) addFinalizerIfRequired(ctx context.Context, trait *vzapi.IngressTrait) error {
 	if trait.GetDeletionTimestamp().IsZero() && !vzstring.SliceContainsString(trait.Finalizers, finalizerName) {
-		traitName := vznav.GetNamespacedNameFromObjectMeta(trait.ObjectMeta).String()
+		traitName := vznav.GetNamespacedNameFromObjectMeta(trait.ObjectMeta)
```
pfmackin, line 179: it would be better if we used NSN as in traitNSN, but seems like name is being used in other places as well.

application-operator/controllers/helidonworkload/helidonworkload_controller.go
```diff
@@ -74,9 +74,9 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	var workload vzapi.VerrazzanoHelidonWorkload
 	if err := r.Get(ctx, req.NamespacedName, &workload); err != nil {
 		if k8serrors.IsNotFound(err) {
-			log.Info("VerrazzanoHelidonWorkload has been deleted", "name", req.NamespacedName)
+			log.Info("VerrazzanoHelidonWorkload has been deleted")
 		} else {
-			log.Error(err, "Failed to fetch VerrazzanoHelidonWorkload", "name", req.NamespacedName)
+			log.Error(err, "Failed to fetch VerrazzanoHelidonWorkload")
```
pfmackin, line 79: why is the logging of NSN removed?

application-operator/controllers/metricstrait/metricstrait_controller.go
pfmackin: never mind, i read it wrong

application-operator/controllers/helidonworkload/helidonworkload_controller.go
fredtibbitts: Line 70 already adds the NSN to all log messages in this function so adding the NSN again here is redundant.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
fredtibbitts: Yes, there is inconsistency in the naming of similar fields, but I'm not going to refactor all of that here. We can create a separate ticket.

pfmackin (approved):

PR #2205, opened 2021-12-21
Image registry overrides

anders-swanson:
Developer use case to override individual image repositories and registries.

### DIFF
--- a/pkg/bom/bom.go
+++ b/pkg/bom/bom.go
@@ -81,6 +81,12 @@ type BomImage struct {
 	// ImageTag specifies the name of the image tag, such as `0.46.0-20210510134749-abc2d2088`
 	ImageTag string `json:"tag"`
 
+	// Registry is the image registry. It can be used to override the subcomponent registry
+	Registry string `json:"registry,omitempty"`
+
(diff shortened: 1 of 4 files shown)


pkg/bom/bom.go
```diff
@@ -277,23 +282,30 @@ func (b *Bom) BuildImageOverrides(subComponentName string) ([]KeyValue, error) {
...
-func (b *Bom) ResolveRegistry(sc *BomSubComponent) string {
+func (b *Bom) ResolveRegistry(sc *BomSubComponent, img *BomImage) string {
 	// Get the registry ENV override, if it doesn't exist use the default
 	registry := os.Getenv(constants.RegistryOverrideEnvVar)
 	if registry == "" {
 		registry = b.bomDoc.Registry
 		if len(sc.Registry) > 0 {
 			registry = sc.Registry
 		}
+		if len(img.Registry) > 0 {
```
michael-cico, line 293: i think this should be reversed... if it's set at the image level, it has priority over the subcomponent level, etc.

So precedence is

`image -> subcomponent -> component -> bom-document`

michael-cico (dismissed):

pfmackin (approved):

PR #2226, opened 2021-12-23
ToDo cluster dump

anders-swanson:


tests/e2e/pkg/clusterdump.go
```diff
@@ -5,10 +5,53 @@ package pkg
 
 import (
 	"fmt"
+	"github.com/onsi/ginkgo/v2"
```
pfmackin, line 8: update copyright to 2021, 2022, Oracle...

tests/e2e/pkg/clusterdump.go
```diff
@@ -5,10 +5,53 @@ package pkg
...
+	return ginkgo.AfterSuite(func() {
+		if c.failed {
+			ExecuteClusterDumpWithEnvVarSuffix(fmt.Sprintf("fail-%d", time.Now().Unix()))
+		}
+
+		// ginkgo.Fail and gomega matchers panic if they fail. Recover is used to capture the panic and
+		// generate the cluster dump
+		defer func() {
+			if r := recover(); r != nil {
+				ExecuteClusterDumpWithEnvVarSuffix(fmt.Sprintf("aftersuite-%d", time.Now().Unix()))
```
pfmackin, line 48: Does this ExecuteClusterDumpWithEnvVarSuffix execute if there is not a panic?

anders-swanson: no it doesn't

anders-swanson: This dump should only happen if the aftersuite function panics

pfmackin: We want to dump if the aftersuite fails, for example deleting the namespace.

desagar: Am I reading correctly that this will generate 2 cluster dumps - one if there's a test failure and another if there's a panic? Is that needed? 
Unless there's a significant change in cluster state between the 2, isn't a single dump after the AfterSuite panic enough? Something like `if c.failed || r != nil`

anders-swanson: @pfmackin right, aftersuite panics when something goes wrong. Ginkgo eventually exits with status code 1 after this. 
@desagar you are reading this correctly, it will dump twice. That was intentional in case of needing to check any state changes.

tests/e2e/pkg/clusterdump.go
anders-swanson: updated copyrights

pfmackin (approved):

PR #2249, opened 2022-01-05
backport console test flag

anders-swanson:
backport console flag in ATs to 1.1

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
                 // 1st choice is the default value
                 choices: [ "nip.io", "sslip.io"])
         string (name: 'CONSOLE_REPO_BRANCH',
-                defaultValue: 'master',
+                defaultValue: 'release-1.1',
                 description: 'The branch to check out after cloning the console repository.',
                 trim: true)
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -506,7 +506,8 @@ pipeline {
                         parameters: [
                             string(name: 'GIT_COMMIT_TO_USE', value: env.GIT_COMMIT),
                             string(name: 'WILDCARD_DNS_DOMAIN', value: params.WILDCARD_DNS_DOMAIN),
-                            booleanParam(name: 'EMIT_METRICS', value: params.EMIT_METRICS)
+                            booleanParam(name: 'EMIT_METRICS', value: params.EMIT_METRICS),
+                            string(name: 'CONSOLE_REPO_BRANCH', value: params.CONSOLE_REPO_BRANCH)
```
michael-cico, line 510: So, I think when this gets backported we have to change the default for the parameter there right?  So each line defaults to a corresponding/matching console branch?

agv426: had the same question

anders-swanson: updated default values

michael-cico (approved):

PR #2264, opened 2022-01-07
vz-4315 -  private dns stages with jenkins 

theabmitra:
# Description

Today OKE tests using OCI DNS tests using global DNS This task will add another test to either triggered / long running pipeline to add a private DNS test 

Fixes VZ-4315 

# Checklist 


ci/JenkinsfileTestTrigger
```diff
@@ -198,6 +198,23 @@ pipeline {
                         }
                     }
                 }
+                stage('OCI Private DNS tests') {
```
agv426, line 201: This adds another OKE cluster to every triggered test run (ie: every merge, user branches doing triggered runs). 

Is this test case something that has to be tested that frequently, or can it be run in the periodic test runs instead (ie: every few hours on master)

ci/oke-ocidns/Jenkinsfile
```diff
@@ -111,11 +115,12 @@ pipeline {
         TF_VAR_ssh_public_key_path = credentials('oci-tf-pub-ssh-key')
         TF_VAR_compartment_id = credentials('oci-tiburon-dev-compartment-ocid')
 
+
         OCI_CLI_TENANCY = credentials('oci-tenancy')
         OCI_CLI_USER = credentials('oci-user-ocid')
         OCI_CLI_FINGERPRINT = credentials('oci-api-key-fingerprint')
         OCI_CLI_KEY_FILE = credentials('oci-api-key')
-        OCI_CLI_REGION = "${params.OKE_CLUSTER_REGION}"
+        //OCI_CLI_REGION = "${params.OKE_CLUSTER_REGION}"
```
agv426, line 123: nit: any reason to keep this commented out line?

ci/JenkinsfileTestTrigger
theabmitra: This is a feature for OCI DNS. my understanding is we do need to run this frequently to check whether things are fine .

theabmitra: but i do understand the resource point of view. in fact i was also in the process od adding instance-principal tests 
So basically for oci DNS we need 3 types of tests 
1. standard auth withh public dns 
2. standard auth with private dns (this one)
3. instance_principal with public dns 

All will require a separate OKE cluster . Whether we need 2 and 3 every run is a question that needs to be discussed on a wider forum

ci/oke-ocidns/Jenkinsfile
theabmitra: Thats because we need to dynamically determine the region now . If you go further down we have made that change 
def RUNNER_REGION = sh(returnStdout: true, script: "curl -s -H \"Authorization: Bearer Oracle\" http://169.254.169.254/opc/v2/instance/canonicalRegionName").trim()
                    env.TF_VAR_region = "${params.DNS_SCOPE == 'PRIVATE' ? RUNNER_REGION : params.OKE_CLUSTER_REGION}"
                    env.OCI_CLI_REGION = "${params.DNS_SCOPE == 'PRIVATE' ? RUNNER_REGION : params.OKE_CLUSTER_REGION}"
                    println("runner region is ${RUNNER_REGION}, tf var region is ${env.TF_VAR_REGION}")
                    
                    
                    We need this so that runners and oke clusters are in same region for every run . For public DNS , this is set accordingly

michael-cico: I think he means can we remove the commented out line?  Since it's not being used, and is being calculated differently.

ci/oke-ocidns/Jenkinsfile
```diff
@@ -304,9 +314,27 @@ pipeline {
...
         stage("create-dns-zone") {
             when { expression { return params.TEST_ENV == 'ocidns_oke' } }
+            environment {
+                            TF_VAR_label_prefix="${OKE_CLUSTER_PREFIX}"
+
+                        }
             steps {
                 script {
-                    dns_zone_ocid = sh(script: "${WORKSPACE}/tests/e2e/config/scripts/oci_dns_ops.sh -o create -c ${TF_VAR_compartment_id} -s z${zoneId}", returnStdout: true)
+                    //println("inside create-dns-zone , tf var region is ${env.TF_VAR_REGION}")
```
michael-cico, line 323: similar comment, can we remove the commented-out line?

tests/e2e/config/scripts/oci_dns_ops.sh
```diff
@@ -36,44 +44,94 @@ if [ -z "${SUBDOMAIN_NAME}" ] ; then
...
+  fi 
+  
 elif [ $OPERATION == "delete" ]; then
-  oci dns zone delete --zone-name-or-id ${ZONE_NAME} --force
-  status_code=$?
-  if [ ${status_code} -ne 0 ]; then
-    echo "DNS zone deletion failed on first try. Retrying once."
-    oci dns zone delete --zone-name-or-id ${ZONE_NAME} --force
+  if [ ${DNS_SCOPE} == "PRIVATE" ];then
+    DNS_ZONE_OCID=`(oci dns zone list --compartment-id ${COMPARTMENT_OCID} --scope PRIVATE --name ${ZONE_NAME} | jq -r '.data[].id')`
```
michael-cico, line 101: If you provide `--scope ${DNS_SCOPE}`, then you don't need to do if/else here right?  i.e. delete will operate the same in both cases?  And you can get rid of the `else` below.

ci/JenkinsfilePeriodicTests
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, Oracle and/or its affiliates.
+// Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
desagar, line 1: Not sure why this copyright's first year was updated to 2020. Previously it was 2021, which usually means the new one should be 2021, 2022.

ci/JenkinsfileTestTrigger
```diff
@@ -198,6 +198,25 @@ pipeline {
                         }
                     }
                 }
+                /*
+                stage('OCI Private DNS tests') {
```
desagar, line 202: If we are not going to run these in the triggered tests, please remove the new stage from this file instead of commenting out.

ci/oke-ocidns/Jenkinsfile
```diff
@@ -49,6 +49,9 @@ pipeline {
         choice (description: 'OKE node pool configuration', name: 'OKE_NODE_POOL',
```
desagar, line 49: As we discussed when coming up with this approach, the description of the OKE_CLUSTER_REGION parameter above should be modified to add a sentence something like "This value is ignored when the DNS scope parameter is set to PRIVATE".

ci/JenkinsfilePeriodicTests
theabmitra: ok, changes to 2021,2022

ci/oke-ocidns/Jenkinsfile
theabmitra: done

ci/oke-ocidns/Jenkinsfile
```diff
@@ -43,6 +43,7 @@ pipeline {
     }
 
     parameters {
+        // OKE_CLUSTER_REGION parameter will be ignored for private DNS tests. They get overwritten with runner region 
         choice (description: 'OCI region to launch OKE clusters in', name: 'OKE_CLUSTER_REGION',
```
desagar, line 47: Sorry if I was not clear earlier - I meant the "description" field, not the comment. The "description" field is what is shown to a user when trying to manually run a build from Jenkins.

ci/oke-ocidns/Jenkinsfile
desagar: Sorry if I was not clear earlier - I meant the "description" field, not the comment. The "description" field is what is shown to a user when trying to manually run a build from Jenkins.

theabmitra: Done, not sure why this string is not reflected in UI though

desagar (dismissed):

desagar (approved):

PR #2266, opened 2022-01-07
VZ-4468 specify port for Bobs Books

pfmackin:
Specify IngressTrait port for Bobs Books since we cannot automatically determine the correct port of a multi-service component.

# Checklist 

### DIFF
--- a/examples/bobs-books/bobs-books-app.yaml
+++ b/examples/bobs-books/bobs-books-app.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 apiVersion: core.oam.dev/v1alpha2
--- a/examples/bobs-books/bobs-books-app.yaml
+++ b/examples/bobs-books/bobs-books-app.yaml
@@ -39,6 +39,8 @@ spec:
                 - paths:
                     - path: "/bobbys-front-end"
                       pathType: Prefix
+                  destination:
+                    port: 8001
     - componentName: bobs-orders-wls
       traits:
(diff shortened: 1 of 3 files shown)


examples/bobs-books/bobs-books-app.yaml
```diff
@@ -39,6 +39,8 @@ spec:
                 - paths:
                     - path: "/bobbys-front-end"
                       pathType: Prefix
+                  destination:
+                    port: 8001
```
markxnelson, line 43: i thought we decided to put the host (service) too, to prevent any issues in the ingress controller logic?

pfmackin: I tested this and it is not needed.  The code waits for a service with the matching port then the normal hostname logic is used.

pfmackin: Of course the user is free to override the hostname but it is not required.

jmaron99: both the managed service and the cluster service have port 8001 configured.  Do they both appear simultaneously or does the cluster service appear first?  I'm wondering if we're still seeing the correct host selected due to finalizer issue.  I guess we'll find out when I test my branch and we can add the host if need be.

pfmackin: I can add the host if you want.  For BB it doesn't matter since both services resolve to the same managed server

pfmackin: I think there is a little confusion here on the host.  There are 2 hosts involved, the first is the one that we generate that is used in the VS hostname.  That has nothing to do with the services, for example:

bobs-books.bobs-books.172.18.0.230.nip.io

The second host is the destination host in the route.  This is the one that is the service name and would need to be specified by the user if there was ambiguity.

    route:
    - destination:
        host: bobbys-front-end-cluster-cluster-1

markxnelson (approved):

PR #2268, opened 2022-01-07
VZ-4147.  VPO namespace controller for OCI logging management

michael-cico:
# Description

Adds a namespace controller to the VPO to monitor namespace activity with respect to the new OCI logging feature
- Monitors namespace lifecycle (CRUD), looks for the OCI Logging annotation on the namespace
- If the annotation is detected during a namespace create/update, add a finalizer and call out to (currently placeholder) functions to update the fluentd configuration; it is expected that this will be idempotent
- if the annotation is not present during a namespace update, call the function to remove the OCI logging configuration for that namespace; it is expected that this will be idempotent
- During a namespace deletion, remove any OCI logging configuration for the namespace; it is expected that this will be idempotent
- add unit tests
- refactor common string and controller utils into a common package with unit tests

**NOTE:** This is currently using stubbed calls for the work being done in another PR to update the fluentd configmap.

Fixes VZ-4741.

# Checklist 

### DIFF
--- a/application-operator/controllers/namespace/namespace_controller_test.go
+++ b/application-operator/controllers/namespace/namespace_controller_test.go
@@ -234,8 +234,7 @@ func TestReconcileNamespaceDeletedErrorOnUpdate(t *testing.T) {
 
 	mocker.Finish()
 	asserts.Equal(expectedErr, err)
-	asserts.True(result.Requeue)
-	asserts.True(result.RequeueAfter > 0)
+	asserts.True(result.IsZero())
 }
 
 // Test_removeFinalizer tests the removeFinalizer method for the following use case
@@ -380,13 +379,11 @@ func Test_reconcileNamespaceErrorOnUpdate(t *testing.T) {
 			return expectedErr
 		})
 
-	result, err := nc.reconcileNamespace(context.TODO(), ns)
+	err = nc.reconcileNamespace(context.TODO(), ns)
 
 	mocker.Finish()
 	asserts.Error(err)
 	asserts.Equalf(expectedErr, err, "Did not get expected error: %v", err)
-	asserts.True(result.Requeue)
-	asserts.True(result.RequeueAfter > 0)
 }
 
 // Test_reconcileNamespace tests the reconcileNamespace method for the following use case
@@ -413,11 +410,10 @@ func Test_reconcileNamespace(t *testing.T) {
 		},
 	}
 
-	result, err := nc.reconcileNamespace(context.TODO(), ns)
+	err = nc.reconcileNamespace(context.TODO(), ns)
 
 	mocker.Finish()
 	asserts.NoError(err)
-	asserts.Equal(ctrl.Result{}, result)
 }
 
 // Test_reconcileNamespaceDelete tests the reconcileNamespaceDelete method for the following use case
--- a/application-operator/controllers/namespace/namespace_controller_test.go
+++ b/application-operator/controllers/namespace/namespace_controller_test.go
@@ -1,0 +1,736 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+package namespace
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/verrazzano/verrazzano/application-operator/constants"
+	"github.com/verrazzano/verrazzano/platform-operator/mocks"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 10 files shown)


platform-operator/controllers/common/collections.go
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+package common
+
+// ContainsString checks for a string in a slice of strings
+func ContainsString(slice []string, s string) bool {
```
pfmackin, line 6: These string functions should be in pkg/string

platform-operator/controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: Seems like this controller should be in application operator.  The VPO is responsible for lifecycle of the stack, not really configuration or wiring loggers, etc.

platform-operator/controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// removeFinalizer - Remove the finalizer and update the namespace resource if the post-delete processing is successful
+func (nc *NamespaceController) removeFinalizer(ctx context.Context, ns *corev1.Namespace) (reconcile.Result, error) {
+	nc.log.Infof("Removing finalizer %s", namespaceControllerFinalizer)
+	var removed bool
+	ns.Finalizers, removed = vpocommon.RemoveString(ns.Finalizers, namespaceControllerFinalizer)
+	if removed {
+		err := nc.Update(ctx, ns)
+		if err != nil && !k8serrors.IsConflict(err) {
```
pfmackin, line 86: retry needs to be done if conflict

platform-operator/controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+// reconcileNamespace - Reconcile any namespace changes
+func (nc *NamespaceController) reconcileNamespace(ctx context.Context, ns *corev1.Namespace) (reconcile.Result, error) {
+	if err := nc.reconcileOCILogging(ctx, ns); err != nil {
+		nc.log.Errorf("Error occurred during OCI Logging reconciliation: %s", err.Error())
+		return vpocommon.NewRequeueWithDelay(10, 30, time.Second), err
```
pfmackin, line 97: we should just requeue with error and set the backoff limit like we do in VAO ingress trait (max 60 sec)

platform-operator/controllers/namespace/namespace_controller.go
michael-cico: We debated that, because the logging configuration here is global and not scoped to an application we thought it should be here.  we also need to be able to recreate the namespace configmap updates post-upgrade.

platform-operator/controllers/common/collections.go
michael-cico: OK

platform-operator/controllers/namespace/namespace_controller.go
pfmackin: see...
IngressTrait.go

func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		WithOptions(controller.Options{
			RateLimiter: controllers.NewDefaultRateLimiter(),
		}).

platform-operator/controllers/common/collections.go
michael-cico: Fixed

platform-operator/controllers/namespace/namespace_controller.go
michael-cico: Moved to the VAO; will address the edge/upgrade cases in another PR.

platform-operator/controllers/namespace/namespace_controller.go
michael-cico: Fixed

application-operator/controllers/namespace/namespace_controller_test.go
```diff
@@ -0,0 +1,734 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
```
fredtibbitts, line 1: General comment about the tests, I've been using `sigs.k8s.io/controller-runtime/pkg/client/fake` for unit testing k8s interactions and I find the tests much easier to write (and read) over using mocks. Don't change anything, just something to consider for the future.

michael-cico: I typically use fakes in most cases, except here where I want to verify that some things were/weren't called, or to return an error from a Client interface... I'm not sure that's possible with fakes?

fredtibbitts: You can return errors with the controller-runtime fake, you just have to wrap the client and implement the CRUDL function to return an error.

fredtibbitts (approved):

application-operator/controllers/namespace/namespace_controller_test.go
michael-cico: ok, thanks...  I tried to limit my mocks usage as much as possible, maybe we can remove some of them next pass.

I do sometimes still like it for verifying things like calls to Update, etc, as long as it can be done sparingly.

pfmackin (approved):

PR #2271, opened 2022-01-07
VZ-4466: Change scrape-generator webhook to create a metrics binding resource

bobdonat:
# Description

This pull request changes the scrape-generator webhook to create a metrics binding resource instead of adding labels to a workload resource.

In addition, the mutuatingwebhookconfiguration for the scrape-generator webhook was changed to have the namespace selector match labels "verrazzano-managed: true" instead of the webhook code checking.

# Checklist 


application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -179,30 +169,64 @@ func (a *ScrapeGeneratorWebhook) processMetricsAnnotation(unst *unstructured.Uns
...
 
-	unst.SetLabels(labels)
+// function called by controllerutil.createOrUpdate to mutate a metricsBinding resource
+func (a *ScrapeGeneratorWebhook) mutateMetricsBinding(metricsBinding *vzapp.MetricsBinding, template *vzapp.MetricsTemplate, unst *unstructured.Unstructured) error {
+	metricsBinding.Spec.MetricsTemplate.Namespace = template.Namespace
+	metricsBinding.Spec.MetricsTemplate.Name = template.Name
+	metricsBinding.Spec.PrometheusConfigMap.Namespace = template.Spec.PrometheusConfig.TargetConfigMap.Namespace
+	metricsBinding.Spec.PrometheusConfigMap.Name = template.Spec.PrometheusConfig.TargetConfigMap.Name
+	trueValue := true
+	metricsBinding.OwnerReferences = []metav1.OwnerReference{
```
kminder, line 220: Might want to consider using one of these instead of doing this manually.
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.0/pkg/controller/controllerutil/controllerutil.go#L59
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.0/pkg/controller/controllerutil/controllerutil.go#L96

application-operator/controllers/webhooks/scrape_generator_webhook.go
```diff
@@ -179,30 +169,64 @@ func (a *ScrapeGeneratorWebhook) processMetricsAnnotation(unst *unstructured.Uns
...
-		scrapeGeneratorLogger.Error(err, "error getting Prometheus target config map")
+		scrapeGeneratorLogger.Error(err, "error getting Prometheus target config map", "Namespace", template.Namespace, "Name", template.Name)
 		return err
 	}
-	labels := unst.GetLabels()
-	if labels == nil {
-		labels = make(map[string]string)
+
+	// For at least deployments, the webhook is called multiple times.  The first time the UID is empty.
+	// A UID is needed for setting up the owner reference.  Return and do nothing if thd UID is empty.
```
kminder, line 187: typo in the comment

bobdonat: fixed

application-operator/controllers/webhooks/scrape_generator_webhook.go
bobdonat: I original was going to use that.  That function takes a runtime.Scheme.  As I understand the scheme would need all workload resource objects that we support in the webhook.  That being the case, I don't think we could guarantee things would always work.

kminder: Good point.

kminder (approved):
Approved.

application-operator/internal/app/scripts/install-app-resources.sh
```diff
@@ -26,13 +26,8 @@ metadata:
 webhooks:
   - name: verrazzano-application-scrape-generator.verrazzano.io
     namespaceSelector:
-      matchExpressions:
-        - key: kubernetes.io/metadata.name
-          operator: NotIn
-          values: ["kube-system", "verrazzano-mc"]
-        - key: verrazzano.io/namespace
-          operator: NotIn
-          values: ["verrazzano-system"]
```
davisgomes, line 35: Should we keep these, just in case they get labeled by the user?

bobdonat: I don't think they are needed. We require the namespace to be labeled with verrazzano-managed: true

PR #2272, opened 2022-01-07
VZ-4453: Update Fluentd config map with namespace-specific logging config for OCI Logging

fredtibbitts:
# Description

This PR adds functions that update the verrazzano-system Fluentd config map to add and remove namespace-specific logging configuration for OCI Logging. These functions will be called by a new platform operator controller that watches namespaces and looks for a verrazzano annotation the specifies an OCI Log object id. The controller work is in a separate PR.

Implements VZ-4453

# Checklist 

### DIFF
--- a/platform-operator/controllers/common/fluentd_configmap.go
+++ b/platform-operator/controllers/common/fluentd_configmap.go
@@ -1,7 +1,7 @@
 // Copyright (c) 2022, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-package configmap
+package common
 
 import (
 	"bytes"
--- a/platform-operator/internal/k8s/configmap/fluentd_configmap.go
+++ b/platform-operator/internal/k8s/configmap/fluentd_configmap.go
@@ -1,0 +1,138 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package configmap
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+	"text/template"
+
+	"github.com/verrazzano/verrazzano/pkg/constants"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	fluentdConfigMapName = "fluentd-config"
+
+	nsConfigKeyTemplate         = "oci-logging-ns-%s.conf"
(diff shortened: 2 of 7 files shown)


platform-operator/internal/k8s/configmap/fluentd_configmap.go
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
```
pfmackin, line 2: internal/k8s is supposed to be for generic k8s code, not sure this fits here.   Maybe it should go with the controller, seems to be specific to the oci logging work (or maybe not).

fredtibbitts: OK, I'll move it to be collocated with the controller.

michael-cico: So we're adding a new controller for this.  I suggested it be in a common area.  I'm creating a `common` package under the `controllers` package, maybe it should go there.

michael-cico: It will have to be shared likely between the `verrazzano` controller and the new `namespace` controller.

platform-operator/controllers/common/fluentd_configmap.go
```diff
@@ -0,0 +1,138 @@
...
+	loggingTemplate, _ = template.New("loggingConfig").Parse(loggingTemplateBody)
+}
+
+// AddNamespaceLogging updates the system Fluentd config map to include a match section that directs all logs for the given
+// namespace to the given OCI Log object id. It returns true if the config map was updated.
+func AddNamespaceLogging(ctx context.Context, cli client.Client, namespace string, ociLogID string) (bool, error) {
+	cm := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: fluentdConfigMapName, Namespace: constants.VerrazzanoSystemNamespace}}
+
+	opResult, err := controllerutil.CreateOrUpdate(ctx, cli, cm, func() error {
+		if cm.ObjectMeta.CreationTimestamp.IsZero() {
```
michael-cico, line 58: Is this just because we expect the configmap to be created on install?  Is there a problem if we just go ahead and create it even if it wasn't?

michael-cico (approved):

platform-operator/controllers/common/fluentd_configmap.go
fredtibbitts: Yes, that config map needs to exist. If it doesn't something is seriously wrong. I'd rather fail and log an error than create a config map that only contains a portion of what's needed for a functioning system. Creating it and inserting just the config that this func is concerned with won't help and might cause more confusion, maybe.

fredtibbitts: Also if we fail and attempt to reconcile again, there's a chance the config map will exist.

michael-cico (approved):

PR #2280, opened 2022-01-11
VZ-4528 Move the owner reference creation to the controller

davisgomes:
# Description
Pods were failing to be properly registered by the webhook because the UID was not present. Thus, the owner reference creation has been moved to the controller.

Fixes VZ-4528

# Checklist 


application-operator/apis/app/v1alpha1/metricsbinding_types.go
```diff
@@ -46,3 +49,14 @@ type NamespaceName struct {
 	// Name of a resource
 	Name string `json:"name"`
 }
+
+type Workload struct {
+	// Namespace of a resource
+	Namespace string `json:"namespace"`
```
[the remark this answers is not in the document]
davisgomes: NamespaceName is used for the other fields. I had to add the TypeMeta so I decided to create a new resource with both. I'm going to get rid of the namespace since it seems like I don't use it in the controller. However, I think I'll need Bob's help regenerating the CRDs.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -207,19 +207,38 @@ func (r *Reconciler) createOrUpdateScrapeConfig(metricsBinding *vzapi.MetricsBin
...
+		{
+			Name:       owner.Name,
+			APIVersion: owner.TypeMeta.APIVersion,
+			Kind:       owner.TypeMeta.Kind,
+			UID:        workloadObject.GetUID(),
+		},
+	})
+
+	// Apply the updated MetricsBinding
+	err = r.Client.Update(context.TODO(), metricsBinding)
```
kminder, line 239: This is starting to break normal controller patterns.  I believe there are now two updates to the metrics binding both here and in add/removeFinalizerIfRequired.  There really shouldn't be two updates in a single reconcile.  These updates need to be merged and ideally controllerutil.createOrUpdate used.

davisgomes: I'm working on consolidating this now for both the Binding and the ConfigMap

application-operator/apis/app/v1alpha1/metricsbinding_types.go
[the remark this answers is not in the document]
bobdonat: How does the TypeMeta work if workload is a pod where there is no group

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -207,19 +207,38 @@ func (r *Reconciler) createOrUpdateScrapeConfig(metricsBinding *vzapi.MetricsBin
...
+	if len(workloadObject.GetUID()) == 0 {
+		err = fmt.Errorf("Could not get UID from workload resource: %s, %s", workloadObject.GetKind(), workloadObject.GetName())
+		r.Log.Error(err, "UID for workload not found", "resource", workloadObject.GetName())
+		return nil, err
+	}
+
+	// Set the owner reference for the MetricsBinding so that it gets deleted with the workload
+	metricsBinding.SetOwnerReferences([]k8smetav1.OwnerReference{
+		{
+			Name:       owner.Name,
```
bobdonat, line 231: Do we need to set these field too
			Controller:         &trueValue,
			BlockOwnerDeletion: &trueValue,
I thought @kminder suggested they be set when in the webhook

davisgomes: Yes I can add those.

application-operator/apis/app/v1alpha1/metricsbinding_types.go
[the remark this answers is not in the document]
davisgomes: That's correct. I've built the controller to handle both cases.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: The implementation has been simplified to handle the controls of the binding and the configmap separately.

application-operator/apis/app/v1alpha1/metricsbinding_types.go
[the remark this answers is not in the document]
davisgomes: And I'm using the Kubernetes-defined TypeMeta that handles this case.

application-operator/apis/app/v1alpha1/metricsbinding_types.go
```diff
@@ -36,6 +36,9 @@ type MetricsBindingSpec struct {
 
 	// Identifies a namespace and name for a Prometheus configMap resource
 	PrometheusConfigMap NamespaceName `json:"prometheusConfigMap"`
+
+	// Identifies the name, namespace, and gvk for a workload
```
bobdonat, line 40: comment is wrong - no namespace

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -6,12 +6,12 @@ package metricsbinding
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```
bobdonat, line 9: imports out of order

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -63,8 +63,11 @@ func (r *Reconciler) reconcileBindingDelete(ctx context.Context, metricsBinding
 		return k8scontroller.Result{Requeue: true}, err
 	}
 
-	// For deletion, we have to remove the finalizer if it exists
-	err := r.removeFinalizerIfRequired(ctx, metricsBinding)
+	// Remove the finalizer if deletion was successful
+	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, metricsBinding, func() error {
```
bobdonat, line 67: this seems odd.  why is controllerutil.CreateOrUpdate being called to remove the finalizer

davisgomes: The remove finalizer function only removes it from the local copy. The binding has to be updated to actually remove the finalizer and delete. I thought this was more fitting to the controller pattern to use CreateOrUpdate rather than manually updating.

application-operator/apis/app/v1alpha1/metricsbinding_types.go
davisgomes: Thanks. will update

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: You're right. Intellij loves to switch up the imports. Will fix.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -63,8 +63,11 @@ func (r *Reconciler) reconcileBindingDelete(ctx context.Context, metricsBinding
...
+		controllerutil.RemoveFinalizer(metricsBinding, finalizerName)
+		return nil
+	})
 	if err != nil {
 		return k8scontroller.Result{Requeue: true}, err
```
bobdonat, line 72: my understanding is that when an error is returned then a requeue will happen.  nit: sometimes when there is an error Requeue:true is used and sometimes not.

bobdonat (approved):

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: I'm not sure if errors always requeue, but I thought it best to be safe and explicitly requeue.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -89,76 +94,68 @@ func (r *Reconciler) reconcileBindingCreateOrUpdate(ctx context.Context, metrics
...
+	// Return error if UID is not found
+	if len(workloadObject.GetUID()) == 0 {
+		err = fmt.Errorf("Could not get UID from workload resource: %s, %s", workloadObject.GetKind(), workloadObject.GetName())
+		r.Log.Error(err, "UID for workload not found", "resource", workloadObject.GetName())
+		return err
 	}
+
+	// Set the owner reference for the MetricsBinding so that it gets deleted with the workload
+	trueValue := true
+	metricsBinding.SetOwnerReferences([]k8smetav1.OwnerReference{
```
kminder, line 117: Can't this possibly return an error?

application-operator/controllers/metricsbinding/metricsbinding_controller.go
```diff
@@ -323,21 +304,31 @@ func (r *Reconciler) getMetricsTemplate(metricsBinding *vzapi.MetricsBinding) (*
 }
 
 // getPromConfigMap returns the Prometheus ConfigMap given in the MetricsBinding
-func (r *Reconciler) getPromConfigMap(metricsBinding *vzapi.MetricsBinding) (*k8scorev1.ConfigMap, error) {
-	configMap := k8scorev1.ConfigMap{
+func (r *Reconciler) getPromConfigMap(metricsBinding *vzapi.MetricsBinding) *k8scorev1.ConfigMap {
```
kminder, line 307: This is a bit confusingly named since it doesn't actually get the CM like getWorkloadObject does below.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: No. This function has no return value.

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: What about getPromConfigMapTemplate?

application-operator/controllers/metricsbinding/metricsbinding_controller.go
kminder: Docs say it will be requeued if error != nil
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

application-operator/controllers/metricsbinding/metricsbinding_controller.go
davisgomes: I'm going to merge, but will update this name in the next PR.

PR #2293, opened 2022-01-12
VZ-4412 jenkins tests with oci dns and  instance principal

theabmitra:
# Description

Implement jenkins stages to test out instance principal support with OCI DNS

OCI secret
auth:
  authtype: instance_principal

# Checklist 


ci/oke-ocidns/Jenkinsfile
```diff
@@ -359,6 +362,10 @@ pipeline {
             when { expression { return params.TEST_ENV == 'ocidns_oke' } }
             steps {
                 script {
+                    if (params.OCI_DNS_AUTH == "instance_principal") {
+                            env.OCI_CLI_AUTH="instance_principal"
```
desagar, line 366: As I understand it, this will result in the auth for things like oci_dns_ops.sh also being set to instance principal. Do we need that to happen? Also, last I recall we didn't have permissions for dns zones for the jenkins runners - did you end up adding that?

platform-operator/helm_config/overrides/external-dns-values.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, Oracle and/or its affiliates.
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
desagar, line 1: Why was the copyright here changed? This file doesn't appear to have been modified.

tools/scripts/generate_operator_yaml.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright (c) 2021, Oracle and/or its affiliates.
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
desagar, line 3: same q. - why was copyright changed, since the file was not modified?

ci/oke-ocidns/Jenkinsfile
```diff
@@ -624,7 +631,7 @@ pipeline {
 
             sh """
                 if [ "${TEST_ENV}" == "ocidns_oke" ]; then
-                  ${WORKSPACE}/tests/e2e/config/scripts/oci_dns_ops.sh -o delete -s z${zoneId} -k ${params.DNS_SCOPE} || echo "Failed to delete DNS zone z${zoneId}"
+                  ${WORKSPACE}/tests/e2e/config/scripts/oci_dns_ops.sh -o delete -c ${TF_VAR_compartment_id} -s z${zoneId} -k ${params.DNS_SCOPE} || echo "Failed to delete DNS zone z${zoneId}"
```
desagar, line 634: I'm curious why this was needed - in the script, the default comes from TF_VAR_compartment_id

platform-operator/helm_config/overrides/external-dns-values.yaml
theabmitra: i had made a change earlier in this file (should be there in my commits),, then changed it back. After that it started erroring out at copyright

tools/scripts/generate_operator_yaml.sh
theabmitra: i had made a change earlier in this file (should be there in my commits),, then changed it back. After that it started erroring out at copyright

ci/oke-ocidns/Jenkinsfile
theabmitra: This was not there in master, but i felt its better, since while create we pass it.

ci/oke-ocidns/Jenkinsfile
theabmitra: i agree, i think this is not needed. Let me remove and retry a run

ci/oke-ocidns/Jenkinsfile
theabmitra: i checked in the script. the compartment is not needed for delete , it basically is a NOP. removed this parameter

PR #2313, opened 2022-01-14
Add reconcile to component interface

anders-swanson:
Most of the diff is generated files.


michael-cico (dismissed):

platform-operator/controllers/verrazzano/install.go
```diff
@@ -50,6 +50,9 @@ func (r *Reconciler) reconcileComponents(_ context.Context, spiCtx spi.Component
 		case vzapi.Ready:
 			// For delete, we should look at the VZ resource delete timestamp and shift into Quiescing/Uninstalling state
 			log.Debugf("component %s is ready", compName)
+			if err := comp.Reconcile(spiCtx); err != nil {
+				return ctrl.Result{Requeue: true}, err
```
pfmackin, line 54: Need a delay, use newRequeueWithDelay

michael-cico: I actually moved that to the top-level `pkg/controller` package, `NewRequeueWithDelay`

anders-swanson: updated

pfmackin (approved):

PR #2340, opened 2022-01-19
VZ-4140: Support user-supplied API creds for OCI Logging integration

fredtibbitts:
# Description

This PR allows a user to configure OCI Logging using a secret containing OCI API configuration. There are two parts to this:

1. Update the helm template to mount the secret data into the Fluentd pods on install
2. Install pre-configuration function update that copies the user-supplied secret from the `verrazzano-install` namespace to the `verrazzano-system` namespace. I also did some refactoring to move the secret copying into a function to reduce duplication.

Implements VZ-4140

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
+++ b/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
@@ -494,48 +494,66 @@ func LabelKubeSystemNamespace(client clipkg.Client) error {
 	return nil
 }
 
+// loggingPreInstall copies logging secrets from the verrazzano-install namespace to the verrazzano-system namespace
 func loggingPreInstall(ctx spi.ComponentContext) error {
 	if vzconfig.IsFluentdEnabled(ctx.EffectiveCR()) {
-		// If fluentd is enabled, copy any custom Elasticsearch secret if found
+		// If fluentd is enabled, copy any custom secrets
 		fluentdConfig := ctx.EffectiveCR().Spec.Components.Fluentd
-		if fluentdConfig != nil &&
-			len(fluentdConfig.ElasticsearchURL) > 0 &&
-			fluentdConfig.ElasticsearchSecret != globalconst.DefaultElasticsearchSecretName {
-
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -494,51 +494,69 @@ func LabelKubeSystemNamespace(client clipkg.Client) error {
...
-				ObjectMeta: metav1.ObjectMeta{Name: esSecret, Namespace: globalconst.VerrazzanoSystemNamespace},
-			}
-			opResult, err := controllerruntime.CreateOrUpdate(context.TODO(), ctx.Client(), &targetSecret, func() error {
-				sourceSecret := corev1.Secret{}
-				if err := ctx.Client().Get(context.TODO(),
-					types.NamespacedName{Name: esSecret, Namespace: constants.VerrazzanoInstallNamespace},
-					&sourceSecret); err != nil {
+		if fluentdConfig != nil {
+			// Copy the external Elasticsearch secret
+			if len(fluentdConfig.ElasticsearchURL) > 0 && fluentdConfig.ElasticsearchSecret != globalconst.DefaultElasticsearchSecretName {
```
michael-cico, line 504: So we can have both the ES and OCI secrets configured at the same time?

fredtibbitts: There's nothing that currently prevents the user from specifying both ES and OCI config. The ES plugin config won't get configured in the Fluentd configmap if OCI Logging is enabled. I didn't feel like this was the best place to enforce that logic so it just copies whatever secrets are specified.

I think maybe the validating webhook should have the rules for this sort of thing.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
michael-cico: Sounds good

PR #2359, opened 2022-01-24
VZ-4619 add progress logging package

pfmackin:
Add support for logging progress messages during reconcile.  This package is not used yet, subsequent commits will use this code.  See doc.go for details.


pkg/log/progress/progress.go
```diff
@@ -0,0 +1,201 @@
...
+import (
+	"fmt"
+	"go.uber.org/zap"
+	"time"
+)
+
+// RootLoggerMap contains a map of RootLogger objects
+var RootLoggerMap = make(map[string]*RootLogger)
+
+var fakeInfoLogger VerrazzanoLogger
```
kminder, line 15: Why "fake"

pfmackin: maybe, let me check.  I refactored the design and before i could not do that but maybe now i can

pkg/log/progress/doc.go
```diff
@@ -0,0 +1,40 @@
...
+//
+// This logger is initialized with the zap.SugaredLogger and then used instead of the zap logger directly.
+// The same SugaredLogger calls can be made: Debug, Debugf, Info, Infof, Error, and Errorf.  The
+// two new calls are Progress and Progressf. The S() method will return the underlying SugaredLogger.
+// The following psuedo-code shows how this should be used:
+//
+//   l := EnsureLogger("namespace/myresource", zap.S()).DefaultProgressLogger()
+//
+// Display info and errors as usual
+//   p.Errorf(...)
```
davisgomes, line 26: What's the justification behind supporting error messages in this format? I'd assume we want progress messages for Reconciliation Info. It seems odd to ignore error messages.

michael-cico: IIUC it will throttle repeated messages

* IFF the message is logged using `Progress` 
* the same exact message is being recorded within the throttling interval

Is that correct?  If so, it's up to the caller to decide what message to throttle and which ones not to, so the caller would have to be careful not to throttle things that shouldn't be.

pfmackin: yes that is correct.  I think most of the stuff our controllers log would be progress messages, since reconcile gets called repeatedly and we have no state.

pfmackin: Also, we don't ignore error message, the info, error, and debug messages are still logged directly by zap

pfmackin: I thought about changing the info sematics to be progress but thought it was better for the code to explicitly log progress messages.

PR #2366, opened 2022-01-25
VZ-4655 ginkgo validate upgrade

mbenmclean:
This change takes converts the groovy and shell tests from the upgrade pipeline to ginkgo code. These tests validate the success of the upgrade after it occurs. It also reworks and cleans up the upgrade Jenkinsfile. It also moves the istio package out of plaform-operator/internal and into the top level pkg directory to use it in test code. There was also a bug with the istioctl verify-upgrade function that this fixes as well.

# Checklist 


pkg/istio/istioctl.go
```diff
@@ -90,7 +90,7 @@ func IsInstalled(log *zap.SugaredLogger) (bool, error) {
 
 // VerifyInstall verifies the Istio installation
 func VerifyInstall(log *zap.SugaredLogger) (stdout []byte, stderr []byte, err error) {
-	args := []string{}
+	args := []string{"verify-install"}
```
markxnelson, line 93: not sure about this - isn't it added on 96?

tests/e2e/pkg/upgrade.go
```diff
@@ -0,0 +1,75 @@
...
+		return false
+	}
+	pods, err := ListPodsInCluster(namespace, clientset)
+	if err != nil {
+		Log(Error, fmt.Sprintf("Error listing pods in cluster for namespace: %s, error: %v", namespace, err))
+		return false
+	}
+	for _, pod := range pods.Items {
+		_, hasAnnotation := pod.Annotations[annotation]
+		if !hasAnnotation &&
```
markxnelson, line 26: i don't get what this logic is for here - what makes these special cases? maybe add a comment here?

tests/e2e/pkg/upgrade.go
```diff
@@ -0,0 +1,75 @@
...
+		v := pod.Labels["sidecar.istio.io/inject"]
+		if v == "false" {
+			continue
+		}
+		_, ok := pod.Labels["istio.io/rev"]
+		if ok {
+			containers := pod.Spec.Containers
+			found := false
+			for _, container := range containers {
+				if strings.Contains(container.Image, "proxyv2:1.10") {
```
markxnelson, line 63: this seems pretty specific - should this version be hardcoded here?

tests/e2e/pkg/upgrade.go
```diff
@@ -0,0 +1,75 @@
...
+			!strings.Contains(pod.Name, "vmi-system-kiali") &&
+			!strings.Contains(pod.Name, "vmi-system-es-data") {
+			return false
+		}
+	}
+	return true
+}
+
+// CheckPodsForEnvoySidecar checks if a pods which have Envoy sidecars, have the specified image
+func CheckPodsForEnvoySidecar(namespace string, imageName string) bool {
```
markxnelson, line 36: i think the name is a bit vague here - you are really checking for a container with a specific image right, it just happens to be envoy in this case?

pkg/istio/istioctl.go
mbenmclean: So it's actually confusing. That parameter actually is only used for logging the operation. It's because the Install function and the Upgrade function both use the install command. It should probably be changed because it's misleading but I'd like to do that as a separate change.

tests/e2e/pkg/upgrade.go
mbenmclean: This code is already in master I just reorganized it into a new location

mbenmclean: This tests to make sure the pods have been restarted. Kiali won't have been because it got installed when we upgraded. For vmi-system-es-data, I ran into an issue where the pods don't have the restart annotation, although they are getting restarted it. I filed a bug to look into it (VZ-4673).

tests/e2e/pkg/upgrade.go
mbenmclean: This code is already in master I just reorganized it into a new location. You're right though it isn't used. I can remove it.

tests/e2e/pkg/upgrade.go
mbenmclean: This code is already in master I just reorganized it into a new location. You're right it probably should be parameterized.

tests/e2e/pkg/upgrade.go
mbenmclean: I modified the function to pass through the image name instead of hardcoding it

tests/e2e/pkg/upgrade.go
mbenmclean: Now passed through as a parameter

markxnelson (approved):

PR #2368, opened 2022-01-25
VZ 4208: OpenSearch Support

anders-swanson:


markxnelson (changes requested):
please address offline comments

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
```diff
@@ -633,8 +628,8 @@ spec:
           emptyDir: {}
         - name: secret-volume
           secret:
-  {{- if .Values.logging.elasticsearchSecret }}
-            secretName: {{ .Values.logging.elasticsearchSecret }}
+  {{- if .Values.fluentd.elasticsearchSecret }}
```
fredtibbitts, line 631: This change doesn't look right to me. The helm value is being set here: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go#L337 and it's under `logging`, not `fluentd`.

anders-swanson: thanks, this was clobbered in a find/replace

markxnelson (dismissed):

PR #2375, opened 2022-01-26
VZ-4619 more VPO logging cleanup

pfmackin:
More component level VPO logging changes to use progress message.
Fix the new VerrazzanoLogger to log the valid filenames and line numbers
Make log messages for components more consistent

# Checklist 


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -5,7 +5,9 @@ package verrazzano
 
 import (
 	"context"
+	errors2 "errors"
```
michael-cico, line 8: Nit, and not a blocker, but this and the `vzlog2` package aliases could be more descriptive... and I don't see a collision with this package anywhere that requires the alias?  Unless I've gone blind, which is possible.

michael-cico (approved):

platform-operator/controllers/verrazzano/controller.go
pfmackin: the errors2 is because there is a controller runtime errors.  The vzlog2 is for the log package..  I will look at combining the loginit and vzlog into the same package.

PR #2378, opened 2022-01-27
VZ-4619 more VPO logging cleanup, mostly Keycloak

pfmackin:
More VPO logging cleanup.
Keycloak logging cleanup
Providing component info waiting for deployments, statefulsets, ingresses

# Checklist 


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -621,7 +629,7 @@ func configureKeycloakRealms(ctx spi.ComponentContext) error {
 		return err
 	}
 
-	ctx.Log().Debugf("configureKeycloakRealm: successfully configured realm %s", vzSysRealm)
+	ctx.Log().Oncef("Component Keycloak successfully configured realm %s", vzSysRealm)
```
kminder, line 632: What are the semantics here?
What happens if a VZ in uninstalled and reinstalled?  Will this message ever be seen again?
Will this message be stored in a map forever?

pfmackin: Once is in the context of a reconcile session for a resource change.  This means if you change resource foo, and the controller reconciler is called 100 times, this message will only be displayed once.  When the resource foo is finally reconciled (return ctrl.Result{}, nil to controller runtime), then we delete the logging context.  So later, if you changed the resource foo an hour later, and this code executed, then the message will be displayed once, for the new reconcile session.

kminder: Perfect, thanks.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -73,7 +74,8 @@ func (c verrazzanoComponent) IsReady(ctx spi.ComponentContext) bool {
 			{Name: "verrazzano-monitoring-operator", Namespace: globalconst.VerrazzanoSystemNamespace},
 		}...)
 	}
-	if !status.DeploymentsReady(ctx.Log(), ctx.Client(), deployments, 1) {
+	prefix := fmt.Sprintf("Component %s", ComponentName)
```
mgianatagh, line 77: I have seen this line repeated many times.  Maybe a function would be better for returning the prefix.

kminder (approved):

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -97,7 +98,8 @@ func isReady(compContext spi.ComponentContext) bool {
 	deployments := []types.NamespacedName{
 		{Name: externalDNSDeploymentName, Namespace: externalDNSNamespace},
 	}
-	return status.DeploymentsReady(compContext.Log(), compContext.Client(), deployments, 1)
+	prefix := fmt.Sprintf("Component %s", ComponentName)
```
michael-cico, line 101: Are we losing the "component" log fields, so we have to do these kinds of prefixes?

PR #2388, opened 2022-01-28
VZ-4747 - Add an affinity configuration for the AuthProxy deployment

mgianatagh:
# Description

Add an affinity configuration for the AuthProxy deployment.  The option to customize the affinity configuration via The Verrazzano custom resource will be handled in a separate task.

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/values.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/values.yaml
@@ -105,6 +105,18 @@ api:
     AuthnStateTTL: "300"
     MaxRequestSize: 65m
     ProxyBufferSize: 8k
+  affinity: |
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                    - {{ .Values.api.name }}
+            topologyKey: kubernetes.io/hostname
 
 # OCI-related values
 oci:
--- a/tests/e2e/pkg/kubernetes.go
+++ b/tests/e2e/pkg/kubernetes.go
@@ -8,11 +8,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
 	"net/http"
 	"os"
 	"strings"
 
+	"github.com/google/uuid"
 	vpClient "github.com/verrazzano/verrazzano/application-operator/clients/clusters/clientset/versioned"
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
 	"github.com/verrazzano/verrazzano/pkg/semver"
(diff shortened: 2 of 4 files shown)


platform-operator/helm_config/charts/verrazzano/values.yaml
```diff
@@ -105,6 +105,18 @@ api:
     AuthnStateTTL: "300"
     MaxRequestSize: 65m
     ProxyBufferSize: 8k
+  affinity: |
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
```
theabmitra, line 110: A scenario to consider here .
preferredDuringSchedulingIgnoredDuringExecution is a soft check by the scheduler, so in a scenario if number of nodes < number of pods, the pods will be assigned successfully on that node . 
However they will remain like that unless some external action is taken . 
Just like we are adding checks in go-code to check if 2 replicas are running with affinity configured, we may also have this check and just scale down -> scale up the deployment only when number of nodes > number of pods.

tests/e2e/pkg/kubernetes.go
```diff
@@ -128,6 +128,36 @@ func ListDeployments(namespace string) (*appsv1.DeploymentList, error) {
...
+		return nil, err
+	}
+	return deployment, nil
+}
+
+// DoesDeploymentExist returns whether a deployment with the given name and namespace exists for the cluster
+func DoesDeploymentExist(namespace string, name string) (bool, error) {
+	deployments, err := ListDeployments(namespace)
+	if err != nil {
+		Log(Error, fmt.Sprintf("Error listing deployments in cluster for namespace: %s, error: %v", namespace, err))
```
pfmackin, line 150: should be "Failed listing.  You don't need error: %v,  just : %v

tests/e2e/pkg/kubernetes.go
```diff
@@ -128,6 +128,36 @@ func ListDeployments(namespace string) (*appsv1.DeploymentList, error) {
...
+// GetDeployment returns a deployment with the given name and namespace
+func GetDeployment(namespace string, deploymentName string) (*appsv1.Deployment, error) {
+	// Get the Kubernetes clientset
+	clientSet, err := k8sutil.GetKubernetesClientset()
+	if err != nil {
+		return nil, err
+	}
+	deployment, err := clientSet.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
+	if err != nil {
+		Log(Error, fmt.Sprintf("Failed to get Deployment %v from namespace %v:  Error = %v ", deploymentName, namespace, err))
```
pfmackin, line 140: don't need : Error = %v,  just : %v

pfmackin (approved):

tests/e2e/pkg/kubernetes.go
mgianatagh: fixed in my next push.  I am reviewing all the messages in the file to fix them according to your feedback.

platform-operator/helm_config/charts/verrazzano/values.yaml
kminder: This goes back to the discussion of preferredDuringSchedulingIgnoredDuringExecution vs requiredDuringSchedulingIgnoredDuringExecution.  Ideally I think a prod profile should use requiredDuringSchedulingIgnoredDuringExecution but this won't work if someone tries to use prod on a single node cluster (even though they shouldn't).  In the end I think this is something that we should expect customer to tweak and make it easy for them.  Specifically I imagine most "real" customer will switch to requiredDuringSchedulingIgnoredDuringExecution and add zone anti-affinity (i.e. topology.kubernetes.io/zone).

pfmackin (dismissed):

kminder (approved):

tests/e2e/pkg/kubernetes.go
```diff
@@ -268,7 +298,7 @@ func GetVerrazzanoInstallResourceInCluster(kubeconfigPath string) (*v1alpha1.Ver
 func IsDevProfile() bool {
 	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
 	if err != nil {
-		Log(Error, fmt.Sprintf("Error getting kubeconfig, error: %v", err))
+		Log(Error, fmt.Sprintf("Error getting kubeconfig: %v", err))
```
kminder, line 301: Should this start with "Failed" too?

tests/e2e/pkg/kubernetes.go
```diff
@@ -286,7 +316,7 @@ func IsDevProfile() bool {
 func GetVerrazzanoVersion() (string, error) {
```
kminder, line 316: Nothing to do with your change bu

tests/e2e/pkg/kubernetes.go
```diff
@@ -324,7 +354,7 @@ func IsVerrazzanoMinVersion(minVersion string) (bool, error) {
 func IsProdProfile() bool {
```
kminder, line 354: Nothing to do with your change but these should probably return the error instead of defaulting to false.

kminder (approved):
Approved as fixing the logging is really out of scope of the story.

PR #2392, opened 2022-01-28
add CN in ingress certs

guoyongzhang:
# Description

As a part of https://jira.oraclecorp.com/jira/browse/VZ-4588, add CN to keycloak, rancher, kiali, vmi, and verrazzano ingress subject.

Fixes VZ-4588

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -431,19 +431,25 @@ func getEnvironmentName(envName string) string {
 	return envName
 }
 
-// updateKeycloakIngress updates the Ingress when using externalDNS
+// updateKeycloakIngress updates the Ingress
 func updateKeycloakIngress(ctx spi.ComponentContext) error {
 	ingress := networkv1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{Name: "keycloak", Namespace: "keycloak"},
 	}
 	_, err := controllerruntime.CreateOrUpdate(context.TODO(), ctx.Client(), &ingress, func() error {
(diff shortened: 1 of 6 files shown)


wmhopkins (commented):
Similar comment as for the VMO PR -- I think the Subject should be a DN that better represents the identity of the service; not just adding the SAN as a CN in the Subject.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -429,19 +429,25 @@ func getEnvironmentName(envName string) string {
...
+// updateKeycloakIngress updates the Ingress
 func updateKeycloakIngress(ctx spi.ComponentContext) error {
 	ingress := networkv1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{Name: "keycloak", Namespace: "keycloak"},
 	}
 	_, err := controllerruntime.CreateOrUpdate(context.TODO(), ctx.Client(), &ingress, func() error {
-		dnsSubDomain, err := vzconfig.BuildDNSDomain(ctx.Client(), ctx.EffectiveCR())
-		if err != nil {
-			return err
+		dnsSuffix, _ := vzconfig.GetDNSSuffix(ctx.Client(), ctx.EffectiveCR())
```
markxnelson, line 438: why would we do this in upgrade if we don't do it in install?

guoyongzhang: updateKeycloakIngress() is in the install code path.  We used to do it only for OCI DNS.  Now it is expanded to run in all cases to add subject name to the generated ingress cert.

PR #2415, opened 2022-02-01
VZ- 4774: add fluentd parsing for oam-kubernetes-runtime log files

bobdonat:
# Description

This pull request adds fluentd parsing/filter rules for oam-kubernetes runtime log files.  In addition, removes unneeded settings from previous fluentd PR.

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
+++ b/platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
@@ -248,7 +248,6 @@ data:
           format /^(?<logtime>\d{4}\/\d{2}\/\d{2} \d{2}\:\d{2}\:\d{2}) \[(?<level>.*?)\] ([\s\S]*?)$/
           time_key logtime
           time_format %Y/%m/%d %H:%M:%S
-          keep_time_key true
         </pattern>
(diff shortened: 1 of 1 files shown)


platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
```diff
@@ -259,7 +258,6 @@ data:
     <filter kubernetes.**verrazzano-authproxy**verrazzano-authproxy**>
       @type record_transformer
       enable_ruby true
-      remove_keys remaining,logtime
```
kminder, line 262: Was this line intentionally removed?  I imagine logtime is removed because keep_time_key defaults to false.  But what about remaining.  Is that now being included in the message sent to ES and if so, should it be?

kminder (approved):
Approved with question.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
bobdonat: Yes, this was intentionally removed.  logtime does not need to be removed because keep_time_key is false.  remaining was no longer being used so there is no need to remove it.

PR #2419, opened 2022-02-01
VZ-4767 multi-cluster logging

pfmackin:
Cleanup logging for Multicluster

1. Use VerrazzanoLogger
2. Don't return error to controller-runtime
3. Cleanup log message

### DIFF
--- a/platform-operator/controllers/clusters/vmc_controller.go
+++ b/platform-operator/controllers/clusters/vmc_controller.go
@@ -54,7 +54,7 @@ func (r *VerrazzanoManagedClusterReconciler) Reconcile(req ctrl.Request) (ctrl.R
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
-		zap.S().Errorf("Failed to fetch Verrazzano resource: %v", err)
+		zap.S().Errorf("Failed to fetch VerrazzanoManagedCluster resource: %v", err)
 		return newRequeueWithDelay(), nil
 	}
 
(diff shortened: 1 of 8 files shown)


platform-operator/controllers/clusters/vmc_controller.go
```diff
@@ -155,7 +192,7 @@ func (r *VerrazzanoManagedClusterReconciler) syncServiceAccount(vmc *clustersv1a
 	// Does the VerrazzanoManagedCluster object contain the service account name?
 	saName := generateManagedResourceName(vmc.Name)
 	if vmc.Spec.ServiceAccount != saName {
-		r.log.Infof("Updating ServiceAccount from %q to %q", vmc.Spec.ServiceAccount, saName)
+		r.log.Oncef("Updating ServiceAccount from %q to %q", vmc.Spec.ServiceAccount, saName)
```
mgianatagh, line 195: Do we have a guideline for using %s versus %q

pfmackin: i didn't notice %q, it would be in the go fmt doc, let me look.

pfmackin: this logs the string in a double quoted string

pfmackin: https://pkg.go.dev/fmt

mgianatagh: Right, my question was more aimed towards do we have guidelines of when to log strings enclosed in double quotes versus not quoted at all.  In almost all cases we do not double quote.  But if the content could include white space I can see the value of the double quotes.

mgianatagh (approved):

platform-operator/controllers/clusters/vmc_controller.go
pfmackin: that is a good point, i can add it to the guidelines.

PR #2502, opened 2022-02-10
VPO cleanup some state name strings

pfmackin:
There is some overlap between component states and VZ CR states and they used to use the same constants.  This makes it difficult to do any refactoring of the state code.  This PR just does some renaming to make component states, VZ CR states, and conditions clear.

# Checklist 


stevshap (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -213,39 +213,53 @@ type Condition struct {
...
+
+	// CompStatePreInstalling is the state when an install is about to be started
+	CompStatePreInstalling CompStateType = "PreInstalling"
+
+	// CompStateInstalling is the state when an install is in progress
+	CompStateInstalling CompStateType = "Installing"
 
-	// Updating is the state when a component configuration update is being applied
-	Updating StateType = "Updating"
+	// CompStateUninstalling is the state when an uninstall is in progress
```
michael-cico, line 249: I think generally, as a nitpick, I prefer more verbose naming, e.g., `ComponentStateType` on the type, and perhaps for the individual values things like `ComponentInstalling`, `ComponentUninstalling`, etc.  Not sure if we want to repeat `State` everywhere, but if we did, `ComponentInstallingState` if that's clearer.  Personally seems more readable.

I'd care more about keeping it brief if we didn't have IDE autocomplete 😄

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
pfmackin: I can do that in a follow up PR.  I do like having State in the name, though it is verbose.

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -26,9 +27,9 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 	targetVersion := cr.Spec.Version
 
 	// Only write the upgrade started message once
-	if !isLastCondition(cr.Status, installv1alpha1.UpgradeStarted) {
+	if !isLastCondition(cr.Status, installv1alpha1.CondUpgradeStarted) {
```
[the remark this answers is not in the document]
pfmackin: I can do that in a follow up PR, i do like Condition in the name, just makes is clearer IMO

PR #2530, opened 2022-02-14
fix  letsencryt production 

guoyongzhang:
# Description

When Verrazzano is using letsencryt production mode, root CA would be empty, and there should not be a secret tls-ca-additional, we can not throw an error in this case.  Instead, we use the default RootCAs.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/rancher.go
+++ b/platform-operator/controllers/verrazzano/component/common/rancher.go
@@ -116,17 +116,18 @@ func GetRootCA(c client.Reader) ([]byte, error) {
 }
 
 // GetAdditionalCA fetches the Rancher additional CA secret
-func GetAdditionalCA(c client.Reader) ([]byte, error) {
+// returns empty byte array of the secret tls-ca-additional is not found
+func GetAdditionalCA(c client.Reader) []byte {
 	secret := &corev1.Secret{}
 	nsName := types.NamespacedName{
 		Namespace: CattleSystem,
 		Name:      RancherAdditionalIngressCAName}
 
 	if err := c.Get(context.TODO(), nsName, secret); err != nil {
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -116,17 +116,18 @@ func GetRootCA(c client.Reader) ([]byte, error) {
...
+// returns empty byte array of the secret tls-ca-additional is not found
+func GetAdditionalCA(c client.Reader) []byte {
 	secret := &corev1.Secret{}
 	nsName := types.NamespacedName{
 		Namespace: CattleSystem,
 		Name:      RancherAdditionalIngressCAName}
 
 	if err := c.Get(context.TODO(), nsName, secret); err != nil {
-		return nil, client.IgnoreNotFound(err)
+		return []byte{}
```
mgianatagh, line 127: What type of error other than NotFound was coming back that we are now ignoring and treating as not found?

anders-swanson: probably should be something along the lines of 
```
if not found
 return []byte{}
else if some other type of error
 return nil, err
```

guoyongzhang: can not think of anything except Kubernetes system error, in which case, I think it is ok to assume "no additional CA".

mgianatagh: Before your change it appears this function was return err of nil when the error was not found.  So I was wasn't sure why that wasn't already working.  Just a little concerned if we are ignoring some other type of fatal error.

mgianatagh (approved):

PR #2537, opened 2022-02-15
fix letsencrypt production in 1.2

guoyongzhang:
# Description

When Verrazzano is using letsencryt production mode, root CA would be empty, and there should not be a secret tls-ca-additional, we can not throw an error in this case. Instead, we use the default RootCAs.

Fixes VZ-9999

# Checklist 

### DIFF
--- a/platform-operator/controllers/clusters/rancher.go
+++ b/platform-operator/controllers/clusters/rancher.go
@@ -102,11 +102,7 @@ func registerManagedClusterWithRancher(rdr client.Reader, clusterName string, lo
 	rc.certificateAuthorityData = caCert
 
 	log.Debugf("Checking for Rancher additional CA in secret %s", rancherTLSAdditional)
-	additionalCA, err := common.GetAdditionalCA(rdr)
-	if err != nil {
-		return "", log.ErrorfNewErr("Failed getting Rancher additional CA: %v", err)
-	}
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/clusters/rancher.go
```diff
@@ -367,10 +363,18 @@ func doRequest(req *http.Request, rc *rancherConfig, log vzlog.VerrazzanoLogger)
 
 	proxyURL := getProxyURL()
 
-	tlsConfig := &tls.Config{
-		RootCAs:    common.CertPool(rc.certificateAuthorityData, rc.additionalCA),
-		ServerName: rc.host,
-		MinVersion: tls.VersionTLS12,
+	var tlsConfig *tls.Config
```
markxnelson, line 366: is there a test for this logic?

mgianatagh (approved):

markxnelson (changes requested):

platform-operator/controllers/clusters/rancher.go
guoyongzhang: no.  I found this bug yesterday and want to squeeze it in quickly after tested with OCI DNS and letsencrypt production mode.  Now we are out of letsencrypt cert quota, I am very hesitant to change code.

markxnelson (approved):

PR #2546, opened 2022-02-15
Dev install version

anders-swanson:


platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -82,8 +82,12 @@ func GetCurrentBomVersion() (*semver.SemVersion, error) {
 	}
 	v := bom.GetVersion()
 	if v == constants.BomVerrazzanoVersion {
-		// This will only happen during development testing, the value doesn't matter
-		v = "1.0.1"
+		// This branch is only hit when using a development BOM
```
pfmackin, line 85: Does it make sense to have a test for this, or is it just something a developer would manually do

anders-swanson: Maybe, but it's really only for developer use

pfmackin (approved):

PR #2549, opened 2022-02-16
JIRA:VZ-3760 support multiple metric ports per container

jmaron99:
# Description

Support the scraping of multiple metric ports per container.  This is required to support the scraping of both helidon and coherence metrics in a deployment such as sock shop.

An component configured with multiple ports will trigger the creation of:
- a set of metrics related annotations per port
- a scrape job per port. 

If only one metric port is defined the resulting configuration will be identical to the configuration yielded by existing versions of Verrazzano.

NOTE:  I can think of one enhancement - adding a "name" attribute to the port spec for the metrics trait, and using that name to explicitly identify the resulting scrape jobs (rather than using an increment as is done currently).  Let me know if you feel this is preferable.

Fixes VZ-3760

# Checklist 


application-operator/apis/oam/v1alpha1/metricstrait_types.go
```diff
@@ -36,9 +36,12 @@ type MetricsTrait struct {
 
 // MetricsTraitSpec specifies the desired state of a metrics trait.
 type MetricsTraitSpec struct {
-	// The HTTP port for the related metrics endpoint. Defaults to 8080.
+	// The HTTP port for the related metrics trait. Defaults to 8080.
 	Port *int `json:"port,omitempty"`
 
+	// The HTTP ports for the related metrics trait. Defaults to 8080.
+	Ports []PortSpec `json:"ports,omitempty"`
```
pfmackin, line 43: Can the trait have Port and Ports both (or either) be used?

jmaron99: If a Port is specified it is converted to a port spec in code in processed.  I'd like to validate that only one or the other is used, which is why I was asking about validation in the metrics defaulter last week.  I figure that validation can be added as a follow up task.  I'd like to keep the existing Port for backwards compatibility.

application-operator/apis/oam/v1alpha1/metricstrait_types.go
```diff
@@ -53,6 +56,14 @@ type MetricsTraitSpec struct {
 	WorkloadReference oamrt.TypedReference `json:"workloadRef"`
 }
 
+type PortSpec struct {
+	// The HTTP port for the related metrics trait. Defaults to 8080.
+	Port *int `json:"port,omitempty"`
+
+	// The HTTP path for the related metrics endpoint. Defaults to /metrics.
+	Path *string `json:"path,omitempty"`
```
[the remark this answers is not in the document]
jmaron99: Because it isn't always "/metrics".

application-operator/apis/oam/v1alpha1/metricstrait_types.go
pfmackin: so the user has to do one of the other, not both?  Can we just let them do both and we can convert internally?

jmaron99: I guess we could do that

application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -845,79 +857,136 @@ func updateStatusIfRequired(status *vzapi.MetricsTraitStatus, results *reconcile
...
-			// This will occur in two situations.
-			// 1. The trait is being deleted.
-			// 2. The trait scraper has been changed and the old scrape config is being updated.
-			//    In this case the traitDefaults and newScrapeConfig will be nil.
-			if trait.DeletionTimestamp.IsZero() && traitDefaults != nil && newScrapeConfig != nil {
-				prometheusScrapeConfig.ArrayAppendP(newScrapeConfig.Data(), prometheusScrapeConfigsLabel)
+	ports := trait.Spec.Ports
+	if len(ports) == 0 {
+		// create a port spec from the existing port
+		ports = []vzapi.PortSpec{{Port: trait.Spec.Port, Path: trait.Spec.Path}}
```
pfmackin, line 863: what if the existing port doesn't exist?  You will have a PortSpec with nil in the fields.  Will that work?

jmaron99: that's a good point.  I can leverage the trait defaults to guard against that.

jmaron99: Actually, I take that back :)

The way it works is that the scrape jobs are actually dependent on the associated metrics annotations.  The code to create those is mutateAnnotations and will handle the nil case by using the trait defaults specified.  So the only need here is simply to have a matching size slice of PortSpec entries - their values don't really matter in this part of the code.

pfmackin (approved):

application-operator/apis/oam/v1alpha1/metricstrait_types.go
kminder, line 43: Wouldn't there need to be several port/path pairs?  This change seems to imply that all the ports will share the one path?

application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -62,9 +63,10 @@ const (
 	prometheusPathAnnotation = "prometheus.io/path"
 
 	// Annotation names for metrics set by the controller
-	verrazzanoMetricsPortAnnotation    = "verrazzano.io/metricsPort"
-	verrazzanoMetricsPathAnnotation    = "verrazzano.io/metricsPath"
-	verrazzanoMetricsEnabledAnnotation = "verrazzano.io/metricsEnabled"
+	verrazzanoMetricsAnnotationPrefix  = "verrazzano.io/metrics"
+	verrazzanoMetricsPortAnnotation    = "verrazzano.io/metricsPort%s"
```
kminder, line 67: I'm sure I'll understand later but this seems strange.

kminder (approved):
Approved.

PR #2552, opened 2022-02-16
VZ-4954: Fix upgrade bugs for WebLogic-based apps

fredtibbitts:
# Description

When upgrading from 1.0.4 to 1.2, I encountered the following bugs with the todo list example application:
1. In the WebLogic adminserver pod, the fluentd sidecar container fails to start. The error in the logs complains about the fluentd config being bad.
2. In the WebLogic adminserver pod, the WebLogic server container fails to reach ready state.

I fixed two bugs related to the Fluentd issue:
1. When reconciling the Fluentd config, the only check to see if changes needed to be made was to compare the Fluentd daemonset. In the case of upgrade to 1.2 for WebLogic workloads, there are environment variable changes needed on the domain spec and the Fluentd configmap needs to be updated. I removed the check altogether. The Fluentd reconciling should only create/update what is needed so the check should not be necessary.
2. The upgrade component order needed to change so that the Verrazzano application operator is higher in the list, otherwise the WebLogic workload was not being reconciled correctly.

For the bug related to the WebLogic container ready state:
1. Since we upgraded the version of Istio in 1.2, there is a WebLogic setting `localhostBindingsEnabled` (controlled by helm chart value `istioLocalhostBindingsEnabled`) that needed to change from true to false. Also the WebLogic operator needed to be moved up the component upgrade list otherwise the WebLogic domain was restarted before the new WKO operator was installed.

This PR also includes two related changes:
1. Removed a block of duplicate code from a controller
2. Added an `isReady` check on upgrade to ensure the component is ready before proceeding to the next component

I also tested these fixes by installing 1.1.1 and upgrading and verified that the todo list app comes up cleanly and functions properly.

Fixes VZ-4954

# Checklist 


platform-operator/controllers/verrazzano/component/weblogic/weblogic_operator.go
```diff
@@ -41,6 +41,10 @@ func AppendWeblogicOperatorOverrides(_ spi.ComponentContext, _ string, _ string,
 			Key:   "enableClusterRoleBinding",
 			Value: "true",
 		},
+		{
+			Key:   "istioLocalhostBindingsEnabled",
+			Value: "false",
```
pfmackin, line 46: what does this bindingEnabled to false do?  Why do we need it now?

platform-operator/controllers/verrazzano/component/registry/registry_test.go
```diff
@@ -49,17 +49,17 @@ func TestGetComponents(t *testing.T) {
...
-	assert.Equal(comps[3].Name(), externaldns.ComponentName)
-	assert.Equal(comps[4].Name(), rancher.ComponentName)
-	assert.Equal(comps[5].Name(), verrazzano.ComponentName)
-	assert.Equal(comps[6].Name(), authproxy.ComponentName)
-	assert.Equal(comps[7].Name(), coherence.ComponentName)
-	assert.Equal(comps[8].Name(), weblogic.ComponentName)
-	assert.Equal(comps[9].Name(), oam.ComponentName)
-	assert.Equal(comps[10].Name(), appoper.ComponentName)
+	assert.Equal(comps[0].Name(), oam.ComponentName)
+	assert.Equal(comps[1].Name(), weblogic.ComponentName)
```
pfmackin, line 53: Is there a reason we are upgrading weblogic before istio?  This means the pod will need to get (potentially) restarted twice, once during upgrade and once after istio upgrade.  I suppose it doesn't matter too much.

platform-operator/controllers/verrazzano/component/weblogic/weblogic_operator.go
fredtibbitts: See this doc: https://oracle.github.io/weblogic-kubernetes-operator/userguide/istio/istio/
When using Istio pre 1.10, that setting needs to be true. 1.10+ and that setting needs to be false. It's correct in the helm chart for 1.2, but the value gets overridden because it was true prior to the upgrade. This forces it to false.

Without this change, the WebLogic container readiness probe returns HTTP 500.

platform-operator/controllers/verrazzano/component/registry/registry.go
```diff
@@ -49,6 +49,9 @@ func GetComponents() []spi.Component {
 func getComponents() []spi.Component {
 	if len(componentsRegistry) == 0 {
 		componentsRegistry = []spi.Component{
+			oam.NewComponent(),
+			weblogic.NewComponent(),
```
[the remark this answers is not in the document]
fredtibbitts: Before moving the component towards the top, the WebLogic server container was in a bad state after the Istio upgrade (due to the binding property not being set correctly), until the WKO operator was upgraded. This makes sure the binding property is set correctly and removes the period of time that the container is in a bad state. If this is a risky change, I can move it back towards the bottom but the WebLogic server container will be in a bad state until WKO is upgraded.

I admit I don't understand all the interdependencies between the components and this is fragile, so if moving it introduces risk that outweighs having the container in a bad state, we can move it back.

platform-operator/controllers/verrazzano/component/weblogic/weblogic_operator.go
pfmackin: ok

platform-operator/controllers/verrazzano/component/registry/registry.go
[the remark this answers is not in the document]
pfmackin: Ok, leave it as is.  We just need to so several upgrade tests to make sure it is solid.

pfmackin (approved):

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -67,6 +67,10 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 			err := r.updateStatus(log, cr, msg, installv1alpha1.CondUpgradeFailed)
 			return ctrl.Result{}, err
 		}
+		if !comp.IsReady(compContext) {
```
michael-cico, line 70: This will force a re-run of the other upgrades, as upgrade is sequential and blocking, but if it's working then probably OK.

michael-cico (approved):

PR #2562, opened 2022-02-17
Fix for repeated reconcile of MetricsTrait for unsupported workloads

lake-of-dreams:
# Description

- The `MetricsTrait` `Defaulter` was adding the `MetricsTrait` to each component in an `ApplicationConfiguration`. This would result in a new `MetricsTrait` being created. The `MetricsTrait` controller would then verify if the trait is supported for the workload of corresponding component and and would remove it if it was not a supported workload. However since the MetricsTrait still existed on the component in `ApplicationConfiguration`, it kept on getting created as a resource and the controller repeatedly removing it.
- This fix changes the `Defaulter` behaviour to only add a default  `MetricsTrait` if the workload type of the concerned component can not be established (for example in cases when the workload has still not been created) or when the work load type is supported one.
- The controller, when removing a `MetricsTrait` will also remove the corresponding specification from the owning `ApplicationConfiguration` so that the resources are in consistent state and the Defaulter will not attempt to re-initiate a `MetricsTrait` for an unsupported workload

Fixes VZ-4548

# Checklist 


application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -407,6 +411,15 @@ func (r *Reconciler) deleteOrUpdateObsoleteResources(ctx context.Context, trait
 			update.RecordOutcome(status.Relations[i], status.Results[i], status.Errors[i])
 		}
 	}
+
+	if !trait.DeletionTimestamp.IsZero() && trait.OwnerReferences != nil {
+		for i := range trait.OwnerReferences {
+			if trait.OwnerReferences[i].Kind == "ApplicationConfiguration" {
+				update.RecordOutcome(r.removedTraitReferencesFromOwner(ctx, &trait.OwnerReferences[i], trait, log))
```
kminder, line 418: I'm concerned here that updating the application configuration resource will result in the redeployment of the application.  The OAM operator will detect the app conf and may regenerate the components (and traits).  Have you confirmed that this is not the case?

application-operator/controllers/webhooks/metrics_trait_defaulter.go
```diff
@@ -58,3 +66,49 @@ func (m *MetricsTraitDefaulter) Default(appConfig *oamv1.ApplicationConfiguratio
 func (m *MetricsTraitDefaulter) Cleanup(appConfig *oamv1.ApplicationConfiguration, dryRun bool, log *zap.SugaredLogger) error {
 	return nil
 }
+
+//defaultTraitToBeAdded method verifies whether a trait should be applied to the component
+func (m *MetricsTraitDefaulter) defaultTraitToBeAdded(appConfig *oamv1.ApplicationConfiguration, appConfigComponent *oamv1.ApplicationConfigurationComponent, log *zap.SugaredLogger) bool {
```
kminder, line 71: Found name confusing.  How about shouldDefaultTraitBeAdded or defaultTraitShouldBeAdded?

lake-of-dreams: fixed.

application-operator/controllers/metricstrait/metricstrait_controller.go
lake-of-dreams: I have confirmed that neither the `workloads` nor the `traits` are regenerated when app config is updated for removing the metricstrait from unsupported workloads. Also the resulting pods/deployments are also unaffected when this happens and there are no restarts.

kminder (approved):
Approved.

PR #2571, opened 2022-02-18
VZ 4993: Deprecate Verrazzano Operator (VO)

anders-swanson:
JSON dashboard files were copied over from VO.


michael-cico:
Still reviewing, but I'm wondering if we should move all the non-Go code under `platform-operator/manifests`, or `platform-operator/resources` (and maybe move manifests under `resources`).  So the dashboards, etc.

cc: @pfmackin

anders-swanson:
> Still reviewing, but I'm wondering if we should move all the non-Go code under `platform-operator/manifests`, or `platform-operator/resources` (and maybe move manifests under `resources`). So the dashboards, etc.
> 
> cc: @pfmackin

At platform-operator directory, the following directories have non-code files in them:
- build
- config
- dashboards
- hack
- helm_config
- manifests
- scripts
- thirdparty

michael-cico:
> > Still reviewing, but I'm wondering if we should move all the non-Go code under `platform-operator/manifests`, or `platform-operator/resources` (and maybe move manifests under `resources`). So the dashboards, etc.
> > cc: @pfmackin
> 
> At platform-operator directory, the following directories have non-gocode files in them:
> 
> * build
> * config
> * dashboards
> * hack
> * helm_config
> * manifests
> * scripts
> * thirdparty

Right, I Paul and I had discussed trying to consolidate things under the manifests dir a while back.  I'm not suggesting you do all of those now, but trying to avoid proliferation of more directories like those above.

But if we don't want to do it now that's fine.

go.mod
```diff
@@ -39,14 +37,13 @@ require (
 	istio.io/api v0.0.0-20200911191701-0dc35ad5c478
 	istio.io/client-go v0.0.0-20200807182027-d287a5abb594
 	k8s.io/api v0.21.1
-	k8s.io/apiextensions-apiserver v0.19.0
+	k8s.io/apiextensions-apiserver v0.19.2
```
pfmackin, line 40: I see VO already uses this

platform-operator/controllers/verrazzano/component/verrazzano/configmaps.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: can you put this in verrazzano/dashboards instead of verrazzano/configmaps

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -44,6 +44,9 @@ func NewComponent() spi.Component {
 // PreInstall Verrazzano component pre-install processing; create and label required namespaces, copy any
 // required secrets
 func (c verrazzanoComponent) PreInstall(ctx spi.ComponentContext) error {
+	if err := setupSharedVMIResources(ctx); err != nil {
```
pfmackin, line 47: This is good, creating Verrazzano secret here will speed up install since Keycloak (the slowest component tp install) preinstall waits for it.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -82,6 +84,9 @@ func (c verrazzanoComponent) IsReady(ctx spi.ComponentContext) bool {
 // PostInstall - post-install, clean up temp files
 func (c verrazzanoComponent) PostInstall(ctx spi.ComponentContext) error {
 	cleanTempFiles(ctx)
+	if err := createVMI(ctx); err != nil {
+		return err
```
pfmackin, line 88: Need to log error
return ctx.Log().ErrorfNewErr("Failed ...

I createVMI has multiple places where it is likely to fail then pass the log to it and log inside createVMI instead, up to you.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -94,6 +99,9 @@ func (c verrazzanoComponent) PostUpgrade(ctx spi.ComponentContext) error {
 		return err
 	}
 	cleanTempFiles(ctx)
+	if err := createVMI(ctx); err != nil {
```
pfmackin, line 102: why are we calling createVMI twice?

pfmackin:
> > > Still reviewing, but I'm wondering if we should move all the non-Go code under `platform-operator/manifests`, or `platform-operator/resources` (and maybe move manifests under `resources`). So the dashboards, etc.
> > > cc: @pfmackin
> > 
> > 
> > At platform-operator directory, the following directories have non-gocode files in them:
> > 
> > * build
> > * config
> > * dashboards
> > * hack
> > * helm_config
> > * manifests
> > * scripts
> > * thirdparty
> 
> Right, I Paul and I had discussed trying to consolidate things under the manifests dir a while back. I'm not suggesting you do all of those now, but trying to avoid proliferation of more directories like those above.
> 
> But if we don't want to do it now that's fine.

I think the dashboards should go in manifests/dashboards.  In general the manifests dir should have runtime files related to:
 
- verrazzano bom
- helm charts and anything related, helm value overrides, third party charts, etc.
- profiles
- dashboards

Can't think of anything else right now.   I would not put scripts and other non-go code there, though i do think that stuff should be consolidated somewhere. I would be happy if we could just move the 4 i listed to manifests.  We have proposed dir struct if we decide to do that.

Manifests are used at runtime, versus build.

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+	if !isVMOEnabled(ctx.EffectiveCR()) {
+		return nil
+	}
+
+	if err := createGrafanaConfigMaps(ctx); err != nil {
+		return err
+	}
+	values := &verrazzanoValues{}
+	if err := appendVerrazzanoValues(ctx, values); err != nil {
+		return err
```
pfmackin, line 38: return ctx.Log().ErrorfNewErr("Failed ...

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	storage, err := findStorageOverride(ctx.EffectiveCR())
+	if err != nil {
+		return err
```
pfmackin, line 42: return ctx.Log().ErrorfNewErr("Failed ...

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+		vmi.Spec.Prometheus = newPrometheus(cr, storage, existingVMI)
+		opensearch, err := newOpenSearch(cr, storage, existingVMI)
+		if err != nil {
+			return err
+		}
+		vmi.Spec.Elasticsearch = *opensearch
+		vmi.Spec.Kibana = newOpenSearchDashboards(cr)
+		return nil
+	})
+	return err
```
pfmackin, line 69: if err != nil {
  return ctx.Log().ErrorfNewErr("Failed ...
}
return nil

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+	values := &verrazzanoValues{}
+	if err := appendVerrazzanoValues(ctx, values); err != nil {
+		return err
+	}
+	storage, err := findStorageOverride(ctx.EffectiveCR())
+	if err != nil {
+		return err
+	}
+
+	existingVMI := getExistingVMI(ctx.Client())
```
pfmackin, line 45: You don't need to fetch the existing VMI, create or update does that.  You just need a VMI var with namespace/name filled in, for example:

	ingress := v1.Ingress{
		ObjectMeta: metav1.ObjectMeta{Name: kialiSystemName, Namespace: namespace},
	}
	_, err := controllerruntime.CreateOrUpdate(context.TODO(), ctx.Client(), &ingress, func() error {

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+			return err
+		}
+		vmi.Spec.Elasticsearch = *opensearch
+		vmi.Spec.Kibana = newOpenSearchDashboards(cr)
+		return nil
+	})
+	return err
+}
+
+func getExistingVMI(cli client.Client) *vmov1.VerrazzanoMonitoringInstance {
```
pfmackin, line 72: Don't think this is needed

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+		if err != nil {
+			return err
+		}
+		*val = intVal
+		return nil
+	}
+
+	// The install args were designed for helm chart, not controller code.
+	// The switch statement is a shim around this design.
+	for _, arg := range opensearchValues.ESInstallArgs {
```
pfmackin, line 183: We have similar logic for istio and maybe others.  We should have a file naming convention where we do this type of VZ CR -> Component config tranformation.  Just an FYI for going forward, not action now, will be part of our cleanup pass.

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -0,0 +1,251 @@
...
+	return ensureVMISecret(ctx.Client())
+}
+
+func ensureVMISecret(cli client.Client) error {
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ComponentName,
+			Namespace: globalconst.VerrazzanoSystemNamespace,
+		},
+		Data: map[string][]byte{},
```
pfmackin, line 235: this line is not needed

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -82,6 +84,9 @@ func (c verrazzanoComponent) IsReady(ctx spi.ComponentContext) bool {
 // PostInstall - post-install, clean up temp files
 func (c verrazzanoComponent) PostInstall(ctx spi.ComponentContext) error {
 	cleanTempFiles(ctx)
+	if err := createVMI(ctx); err != nil {
```
pfmackin, line 87: It would be nice if we could do this pre-install but i don't think we can unless we pull dependencies out of the Verrazzano helm chart like network policies and authorizationpolicies.  If we did that I am wondering if we could have a VMI helm chart at the same level as Verrazzano (like the new AuthProxy chart).  I know that is beyond the scope of this PR but just a thought.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -8,8 +8,10 @@ import (
 	"fmt"
```
pfmackin, line 8: This file is getting way to big, can you move the 5 or 6 override related functions to override.go

pfmackin:
As always, we need to test upgrade with data in ES and make sure this change doesn't cause issues.  I don't see any obvious problems but you never know.  Same for multicluster

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
michael-cico: Yeah, I was also wondering if maybe if some things could be organized into subdirectories, since VZ is an uber-component sort of.  At least for now, if we could separate things by related things within the VZ chart/comp.

But, files would be a good start.

michael-cico:
> As always, we need to test upgrade with data in ES and make sure this change doesn't cause issues. I don't see any obvious problems but you never know. Same for multicluster

We do, but that's orthogonal to this change right?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
michael-cico: I also don't think you can do it unless you have the VMI CRD installed.  so as you say, you need to break the VMI/VMO out into its own chart.  So I think this is a good start.

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
michael-cico: Are you strictly talking about `installArgs`?  Not sure I follow this completely.

platform-operator/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
```
michael-cico: This feels like it's ancient history.  Microoperators were removed a long time ago, is this being used?

If it is, seems like it should be located under the `manifests` subdirectory.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
anders-swanson: Can't be in preinstall for the reasons Mike stated, I had initally started with that. Having a VMI chart would be a good step 2 after this.

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
anders-swanson: It's for the PVC Names, I wasn't sure whether that was used or not. If not, we can remove.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
anders-swanson: postinstall vs. postupgrade?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
anders-swanson: Yeah totally possible. VMI should probably be it's own component.

michael-cico (approved):

pfmackin (approved):

PR #2576, opened 2022-02-21
VZ-5015 Check for Image Pulls before checking application pod deployment in tests

adiforluls:
# Description

PR to implement a mechanism to wait and check for image pull success for all the container images before checking pod deployment status. The process is to enumerate all the containers from the target pods (wait if all of them haven't been scheduled), then drill down the event data to check if all the images have been pulled successfully.

Note that for ImagePullBackOff and CrashLoopBackOff the function returns true, which is intentional as it will break the Eventually loop, and then the control will reach to PodsRunning func where these states will be caught and the test suite will be aborted.

Implements VZ-5015

# Checklist 


tests/e2e/pkg/common.go
```diff
@@ -347,3 +348,98 @@ func SliceContainsPolicyRule(ruleSlice []rbacv1.PolicyRule, rule rbacv1.PolicyRu
 	}
 	return false
 }
+
+func ContainerImagePullWait(namespace string, namePrefixes []string) bool {
+	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
```
bhatpmk, line 353: Multi-cluster tests make use of environment variables ADMIN_KUBECONFIG and MANAGED_KUBECONFIG for admin server and managed server respectively. The function GetKubeConfigLocation expects the kube config defined by either TEST_KUBECONFIG or KUBECONFIG, before using the default location. I think we can follow the approach used by other functions to add another function accepting additional argument for the kube config location.

adiforluls: I'll address it in following commit

bryantidd (dismissed):

tests/e2e/pkg/common.go
```diff
@@ -347,3 +348,101 @@ func SliceContainsPolicyRule(ruleSlice []rbacv1.PolicyRule, rule rbacv1.PolicyRu
...
+				if event.InvolvedObject.Kind == "Pod" && event.InvolvedObject.Name == podName && len(event.InvolvedObject.FieldPath) > 0 && strings.Contains(event.InvolvedObject.FieldPath, containerRegex) {
+
+					if event.Reason == "Pulled" {
+						imagesYetToBePulled--
+						Log(Info, fmt.Sprintf("Pod: %v container: %v status: %v ", podName, container, event.Reason))
+					}
+					// Stop waiting in case of ImagePullBackoff and CrashLoopBackOff
+					if event.Reason == "Failed" {
+						Log(Info, fmt.Sprintf("Pod: %v container: %v status: %v ", podName, container, event.Reason))
+						if strings.Contains(event.Message, "ImagePullBackOff") || strings.Contains(event.Message, "CrashLoopBackOff") {
```
bhatpmk, line 432: We have constants defined for ImagePullBackOff and CrashLoopBackOff. You can add more for "Pulled" and "Failed"

tests/e2e/pkg/common.go
```diff
@@ -347,3 +348,100 @@ func SliceContainsPolicyRule(ruleSlice []rbacv1.PolicyRule, rule rbacv1.PolicyRu
...
+						if strings.Contains(event.Message, ImagePullBackOff) || strings.Contains(event.Message, CrashLoopBackOff) {
+							return true
+						}
+					}
+				}
+			}
+		}
+	}
+
+	Log(Info, fmt.Sprintf("%d images yet to be pulled", imagesYetToBePulled))
```
bhatpmk, line 445: In your follow-up task, add a check here to display the message only when there is at least one image yet to be pulled.

tests/e2e/pkg/common.go
```diff
@@ -347,3 +348,100 @@ func SliceContainsPolicyRule(ruleSlice []rbacv1.PolicyRule, rule rbacv1.PolicyRu
...
+		for _, container := range containers {
+			for _, event := range events.Items {
+				// used to match exact container name in event data
+				containerRegex := "{" + container.(string) + "}"
+
+				if event.InvolvedObject.Kind == "Pod" && event.InvolvedObject.Name == podName && len(event.InvolvedObject.FieldPath) > 0 && strings.Contains(event.InvolvedObject.FieldPath, containerRegex) {
+
+					if event.Reason == "Pulled" {
+						imagesYetToBePulled--
+						Log(Info, fmt.Sprintf("Pod: %v container: %v status: %v ", podName, container, event.Reason))
```
bhatpmk, line 431: In your follow-up task, consider whether we really need this info message.

bhatpmk (approved):

tests/e2e/pkg/common.go
adiforluls: Yeah, we can check for "Pulling" instead of "Pulled", that will be more informative

PR #2581, opened 2022-02-22
VZ-5019: Fix bug, don't create Issuer and Certificate when user specifies CA cert

fredtibbitts:
# Description

While manually testing custom CA certs in release 1.2, I discovered that the CA cert I provided was not being used by the system. The problem is that Verrazzano was incorrectly creating a new CA cert and Issuer, and that was being used in place of the CA cert I specified. The fix is to only create the Certificate and Issuer resources when the secret containing the CA cert does not exist.

Fixes VZ-5019

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
@@ -357,7 +357,7 @@ func createAcmeResources(compContext spi.ComponentContext) error {
 	// Verify that the secret exists
 	secret := v1.Secret{}
 	if err := compContext.Client().Get(context.TODO(), client.ObjectKey{Name: ociDNSConfigSecret, Namespace: ComponentNamespace}, &secret); err != nil {
-		return compContext.Log().ErrorfNewErr("Failed to retireve the OCI DNS config secret: %v", err)
+		return compContext.Log().ErrorfNewErr("Failed to retrieve the OCI DNS config secret: %v", err)
 	}
 
 	// Verify the acme environment and set the server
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -404,52 +404,58 @@ func createAcmeResources(compContext spi.ComponentContext) error {
...
-			Namespace: vzCertCA.ClusterResourceNamespace,
-		},
-		Spec: certv1.IssuerSpec{
-			IssuerConfig: certv1.IssuerConfig{
-				SelfSigned: &certv1.SelfSignedIssuer{},
+
+	// if the CA cert secret does not exist, create the Issuer and Certificate resources
+	secret := v1.Secret{}
+	secretKey := client.ObjectKey{Name: vzCertCA.SecretName, Namespace: vzCertCA.ClusterResourceNamespace}
+	if err := compContext.Client().Get(context.TODO(), secretKey, &secret); err != nil {
```
michael-cico, line 412: Do we need to verify that the error is a `NotFound` error, and return an error otherwise?  Could be done in a separate pass also.

michael-cico (approved):

guoyongzhang (approved):
LGTM.  
Please file a JIRA to track 
1) webhook to validate the secret when customers provide CA secret in CR
2) we don't want to create issuer/CA secret when the provided secret is not default

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
fredtibbitts: OK, I think I'll do it in a separate pass since this is blocking the release. Thanks.

fredtibbitts:
> LGTM. Please file a JIRA to track
> 
> 1. webhook to validate the secret when customers provide CA secret in CR
> 2. we don't want to create issuer/CA secret when the provided secret is not default

Thanks, I'll enter a JIRA for the webhook check. We had a conversation about the 2nd point, in the documentation we don't tell users they can't use the default secret name, so checking if the secret exists (rather than the name of the secret) seems to cover us in all cases.

PR #2583, opened 2022-02-22
Obsolete - separate Verrazzano context from component context

pfmackin:
Obsolete

Verrazzano has a ComponentContext struct that is overloaded and used also as a Verrazzano CR context.  This is confusing and makes sections of the code hard to understand. This PR introduces a VerrazzanoComponent that applies to all components.  Also, refactor the ComponentContext initialization.

One thing to note is that we now create the effective CR at the controller level for the VerrazzanoContext vs the Component level in the ComponentContext.

# Checklist 


platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -20,46 +20,42 @@ type ComponentContext interface {
 	ActualCR() *vzapi.Verrazzano
 	// EffectiveCR returns the effective merged Verrazzano CR
 	EffectiveCR() *vzapi.Verrazzano
-	// IsDryRun indicates the component context is in DryRun mode
+	// IsDryRun indicates the componentName context is in DryRun mode
```
kminder, line 23: Is this change intentional?

platform-operator/controllers/verrazzano/component/spi/component_context.go
```diff
@@ -5,46 +5,64 @@ package spi
...
-		client:      c,
-		dryRun:      dryRun,
-		cr:          actualCR,
-		effectiveCR: effectiveCR,
-	}, nil
+		componentName: compName,
+		log:           log,
+		client:        vzContext.Client,
+		dryRun:        vzContext.DryRun,
+		cr:            vzContext.ActualCR,
```
kminder, line 54: Would actualCR be more clear?

platform-operator/controllers/verrazzano/component/spi/component_context.go
```diff
@@ -5,46 +5,64 @@ package spi
...
+	}
 }
 
 // NewFakeContext creates a fake ComponentContext for unit testing purposes
-// c Kubernetes client
-// actualCR The user-supplied Verrazzano CR
-// dryRun Dry-run indicator
-// profilesDir Optional override to the location of the profiles dir; if not provided, EffectiveCR == ActualCR
 func NewFakeContext(c clipkg.Client, actualCR *vzapi.Verrazzano, dryRun bool, profilesDir ...string) ComponentContext {
+	return NewFakeContextWithNameAndOperation(c, actualCR, dryRun, "", "", profilesDir...)
```
kminder, line 61: Should NewFakeContext be in this file?

platform-operator/controllers/verrazzano/component/spi/component.go
michael-cico: yeah, looks like a Goland refactoring victim.

platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -20,46 +20,42 @@ type ComponentContext interface {
...
 	// Copy returns a copy of the current context
 	Copy() ComponentContext
-	// Init returns a copy of the current context with an updated logging component field
-	Init(comp string) ComponentContext
-	// Operation specifies the logging operation field
-	Operation(op string) ComponentContext
 	// GetOperation returns the operation object in the context
 	GetOperation() string
-	// GetComponent returns the component object in the context
+	// GetComponent returns the componentName object in the context
```
michael-cico: similar refactoring sideffect here?

platform-operator/controllers/verrazzano/component/spi/component_context.go
pfmackin: I can change that, i was using what was there.

platform-operator/controllers/verrazzano/component/spi/component_context_test.go
```diff
@@ -152,8 +157,8 @@ func TestContextProfilesMerge(t *testing.T) {
 			assert.NotNil(context, "Context was nil")
 			assert.NotNil(context.ActualCR(), "Actual CR was nil")
 			assert.Equal(test.actualCR, *context.ActualCR(), "Actual CR unexpectedly modified")
+			context.EffectiveCR().Status = vzapi.VerrazzanoStatus{}
```
kminder, line 160: This should be after the next line.  Is this just to "fake out" the assert.Equal call below?

platform-operator/controllers/verrazzano/component/spi/component_context.go
pfmackin: I tried moving to a test file but doesn't compile.  It was already there

platform-operator/controllers/verrazzano/context/vzcontext.go
```diff
@@ -0,0 +1,84 @@
...
+	Client clipkg.Client
+	// dryRun If true, do a dry run of operations
+	DryRun bool
+	// ActualCR is the CR passed to top level Reconcile.  It epresents the desired Verrazzano state in the cluster
+	ActualCR *vzapi.Verrazzano
+	// effectiveCR Represents the configuration resulting from any named profiles used and any configured overrides in the CR
+	EffectiveCR *vzapi.Verrazzano
+}
+
+// NewContext creates a ComponentContext from a raw CR
```
kminder, line 34: Comment doesn't match function name.  Is New as a function name a good idea?

platform-operator/controllers/verrazzano/context/vzcontext.go
```diff
@@ -0,0 +1,84 @@
...
+func New(log vzlog.VerrazzanoLogger, c clipkg.Client, actualCR *vzapi.Verrazzano, dryRun bool) (VerrazzanoContext, error) {
+	// Generate the effective CR based ond the declared profile and any overrides in the user-supplied one
```
kminder, line 36: typo

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -138,7 +140,7 @@ func (r *Reconciler) doReconcile(log vzlog.VerrazzanoLogger, vz *installv1alpha1
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	spiContext, err := spi.NewContext(log, r, vz, r.DryRun)
+	vzContext, err := vzctx.New(log, r, vz, r.DryRun)
```
kminder, line 143: Should this return a pointer?

platform-operator/controllers/verrazzano/component/spi/component_context_test.go
pfmackin: It is needed since we now clone the status from the actual CR to the effective CR, since we are now creating the effective CR earlier.

platform-operator/controllers/verrazzano/context/vzcontext.go
```diff
@@ -0,0 +1,84 @@
...
+		profileFiles = append(profileFiles, config.GetProfile(profile))
+	}
+	// Merge the profile files into an effective profile YAML string
+	effectiveCR, err := transform.MergeProfiles(actualCR, profileFiles...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Copy actual status to effective CR
+	actualCR.Status.DeepCopyInto(&effectiveCR.Status)
```
michael-cico, line 73: Originally, I'd left the effective CR status empty, so it didn't duplicate what was in the actualCR, but maybe that was more confusing.

I think the status field should be populated via the merge above anyway, if we truly want to duplicate the status field in the Effective CR.  I think initially, we had thought we only wanted the Spec field, then we decided to keep the Metadata and other info.  But not 100% sure we want to have the status duplicated.

platform-operator/controllers/verrazzano/context/vzcontext.go
pfmackin: New is a go convention if you have single purpose package like this.

platform-operator/controllers/verrazzano/component/spi/component_context.go
kminder: ok

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -688,7 +689,8 @@ func (r *Reconciler) checkComponentReadyState(context spi.ComponentContext) (boo
 
 	// Return false if any enabled component is not ready
 	for _, comp := range registry.GetComponents() {
-		if comp.IsEnabled(context) && cr.Status.Components[comp.Name()].State != installv1alpha1.CompStateReady {
+		compContext := spi.NewComponentContext(vzContext, comp.Name(), "")
```
michael-cico, line 692: Wonder if we need create an operation name here instead of ""

platform-operator/controllers/verrazzano/component/spi/component_context.go
michael-cico: I had originally tried doing that, but then it couldn't be shared with other test code.  I thought it lined up with things like the fake controllerruntime client, although perhaps this kind of stuff should be in its own `fake` package?

kminder (commented):
Approved

kminder (dismissed):
Approved.

platform-operator/controllers/verrazzano/context/vzcontext.go
pfmackin: This is needed for upgrade.  We can look at removing the code that checks the status now that the failed state is not used.

platform-operator/controllers/verrazzano/controller.go
pfmackin: no really, this is logging at the global level

platform-operator/controllers/verrazzano/component/spi/component.go
pfmackin: fixed them all

platform-operator/controllers/verrazzano/context/vzcontext.go
pfmackin: I also changed this to NewVerrazzanoContext

kminder (approved):
Approved.

michael-cico (approved):

PR #2584, opened 2022-02-22
WIP VZ-5017: Get correct image overrides for mysql

pshirodkar-orcl:
Signed-off-by: Prasad Shirodkar <prasad.shirodkar@oracle.com>

# Description

Private registry functionality broken for mysql due to extraInitContainers

Fixes VZ-5017

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/mysql/mysql.go
+++ b/platform-operator/controllers/verrazzano/component/mysql/mysql.go
@@ -4,12 +4,14 @@
 package mysql
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"text/template"
 
 	"github.com/verrazzano/verrazzano/pkg/bom"
 	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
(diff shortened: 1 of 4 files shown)


platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -246,10 +283,46 @@ func appendCustomImageOverrides(kvs []bom.KeyValue) ([]bom.KeyValue, error) {
 		return kvs, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
 	}
 
+	if len(imageOverrides) < 2 {
+		return nil, fmt.Errorf("Component MySQL failed, expected atleast 2 image override keys for oraclelinux theme, found %v", len(imageOverrides))
```
desagar, line 287: minor nit - the word "theme" should not be in this error message

pshirodkar-orcl: Removed. Thanks.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -35,8 +37,43 @@ const (
 	mySQLKey            = "mysql-password"
 	mySQLRootKey        = "mysql-root-password"
 	mySQLInitFilePrefix = "init-mysql-"
+	busyboxImageNameKey = "busybox.image"
```
michael-cico, line 40: Since the template is rendered in-memory, are these helm keys needed?  The rendered string is all that's passed to the helm chart right?

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -35,8 +37,43 @@ const (
...
+// (uid=27 gid=27) so that when upgrading, mysql user can definitely access data dir
+// from previous volume. The Helm chart expects this to be a STRING, so using a multi-line string here
+// Note: if the Helm chart for MySQL changes, this should be reviewed for correctness
+const mySQLInitContainerValueTemplate = `
+    - command:
+      - chown
+      - -R
+      - 27:27
+      - /var/lib/mysql
+    image: {{.Image}}
```
michael-cico, line 58: I think it might also be possible to leave this whole snippet in the chart, and then reference the image as helm values like `image: {{ .Values.busybox.image:.Values.busybox.key }}`.  But if things are working and since this is blocking the release then keep it as-is.  I think there are other examples in other charts.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
michael-cico: Also, if the above is true, you can clean it up after merging this, it's not a big deal.

michael-cico: actually, I think I get it now, you need this to leverage the overrides logic right? We probably need to enhance the BOM package at some point for cases like this.

So, nm...

michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
pshirodkar-orcl: Yes, the two new keys are used in the image overrides logic and referred in the updated unit tests. They are not exposed in the helm charts.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
pshirodkar-orcl: I had tried this approach previously but it does not work since everything after `extraInitContainers: |` is a string literal and the string substitution does not work.

pshirodkar-orcl:
Closing this PR since no longer needed. https://github.com/verrazzano/verrazzano/pull/2594 makes the clean fix.

PR #2598, opened 2022-02-24
VZ-5055 Remove Integrtion Tests from Pipeline

bryantidd:
# Description

Removed the Stage Integration Tests from the pipeline
Fixes VZ-5055

# Checklist 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -389,24 +389,6 @@ pipeline {
             }
         }
 
-        stage('Integration Tests') {
-            when { not { buildingTag() } }
-            steps {
-                integrationTests("${DOCKER_IMAGE_TAG}")
-            }
-            post {
-                failure {
-                    script {
-                        SKIP_TRIGGERED_TESTS = true
-                    }
-                }
-                always {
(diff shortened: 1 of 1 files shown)


agv426 (approved):

stevshap (dismissed):

Jenkinsfile
```diff
@@ -389,24 +389,6 @@ pipeline {
             }
         }
 
-        stage('Integration Tests') {
-            when { not { buildingTag() } }
-            steps {
-                integrationTests("${DOCKER_IMAGE_TAG}")
```
bobdonat, line 395: shouldn't the integrationTests def be removed too since it not being used now

bobdonat: and the SIMULATE_FAILURE parameter can be removed since it is only referenced by integration tests

agv426: missed that, we should move that to fail from a different stage rather than remove the param

agv426 (approved):

Jenkinsfile
bryantidd: SIMULATE_FAILURE will move to a separate task.

PR #2646, opened 2022-03-01
VZ 5128: Copy existing storage options during VMI upgrade

anders-swanson:


bobdonat (changes requested):
Isn't there a more general problem where we are calling createVMI during upgrade and this could result in the vmi resource being changed.

anders-swanson:
> Isn't there a more general problem where we are calling createVMI during upgrade and this could result in the vmi resource being changed.

The user may have changes to their VMI, so we need to do this

michael-cico (approved):

bobdonat (approved):

PR #2658, opened 2022-03-02
VZ-4869 restore url

theabmitra:
# Description

Today at install vz instance is updated with conditions and install urls under .status.instance . On a restore of VZ instance, these dont get populated. Since the components are already installed, with this change VPO will update them . 
This is required since console fetches them and displays in UI.  

Fixes VZ-9999

# Checklist 


kminder (commented):
These changes look ok to me but need to reviewed by someone more familiar with VPO.  I suggest Paul Mackin and Mike Cico.

In addition there needs to be a test added that covers this change specifically.

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -84,6 +85,21 @@ func TestUpgradeNoVersion(t *testing.T) {
...
+		List(gomock.Any(), gomock.Not(gomock.Nil())).
+		DoAndReturn(func(ctx context.Context, ingressList *networkingv1.IngressList) error {
+			ingressList.Items = []networkingv1.Ingress{}
+			return nil
+		}).AnyTimes()
+	mock.EXPECT().Status().Return(mockStatus).AnyTimes()
+	mockStatus.EXPECT().
+		Update(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, verrazzano *vzapi.Verrazzano, opts ...client.UpdateOption) error {
+			asserts.NotZero(len(verrazzano.Status.Components), "Status.Components len should not be zero")
```
michael-cico, line 99: Do you need to validate that the URLs are present?  So it looks like this is verifying the `status.components` field is not empty, which is good.  

I'm wondering if it would be better to break out this specific test case into its own unit test, instead of side-effect testing here.

If you do, you could minimize or avoid using the mocks possibly by using the `fakeComponent` struct and overriding the `GetComponents` func.  There are other places where we do that, like this

```
	registry.OverrideGetComponentsFn(func() []spi.Component {
		return []spi.Component{
			fakeComponent{},
		}
	})
	defer registry.ResetGetComponentsFn()
```

theabmitra: Well we will still need the expect in the tests, as the List, Status and Update will be called always when Component status is Ready. 
I can add a non zero check for ingress list as well.

kminder: I was also expecting to see some test where the URLs were initially empty and after a call to reconcileComponents they were corrected.

theabmitra: @kminder and @michael-cico added new tests with the scenarios. please check

michael-cico (approved):

kminder (approved):
Approved.

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -1503,3 +1545,410 @@ func (r goodRunner) Run(_ *exec.Cmd) (stdout []byte, stderr []byte, err error) {
...
+	result, err := reconciler.Reconcile(request)
+
+	// Validate the results
+	mocker.Finish()
+	asserts.NoError(err)
+	asserts.Equal(false, result.Requeue)
+	asserts.Equal(time.Duration(0), result.RequeueAfter)
+
+	// validating instance urls are updated
+	fakeInstanceInfo := vzapi.InstanceInfo{}
```
pfmackin, line 1905: This should be done in the update status mock

theabmitra: This was added to cover the case under normal operation, where instance info is already populated.

PR #2688, opened 2022-03-07
WIP: VZ-5061 setup a scheduled run of the upgrade resiliency tests

jmaron99:
# Description

Setup a build trigger to initiate a complete build every 3 hours.

Fixes VZ-5061

# Checklist 

### DIFF
--- a/ci/chaos/JenkinsfileResiliencyTrigger
+++ b/ci/chaos/JenkinsfileResiliencyTrigger
@@ -388,7 +388,7 @@ def metricBuildDuration() {
 
 def getCronSchedule() {
     if (env.BRANCH_NAME.equals("jmaron/VZ-5061")) {
-        return "H */2 * * *"
+        return "H */3 * * *"
     } else if (env.BRANCH_NAME.startsWith("release-")) {
         return "@daily"
     }
--- a/ci/chaos/JenkinsfileResiliencyTrigger
+++ b/ci/chaos/JenkinsfileResiliencyTrigger
@@ -8,6 +8,11 @@ pipeline {
         skipDefaultCheckout true
     }
 
+    triggers {
+        // run this job on master every 3 hours between 6 AM and 6 PM
+        cron(env.BRANCH_NAME == 'master' ? 'H 06-18/3 * * *' : '')
+    }
+
     agent {
        docker {
             image "${RUNNER_DOCKER_IMAGE}"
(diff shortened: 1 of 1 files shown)


ci/chaos/JenkinsfileResiliencyTrigger
```diff
@@ -2,12 +2,18 @@
...
 def agentLabel = env.JOB_NAME.contains('master') ? "phxlarge" : "VM.Standard2.8"
+def branchSpecificSchedule = getCronSchedule()
 
 pipeline {
     options {
         skipDefaultCheckout true
     }
 
+    triggers {
+        // run this job on master every 2 hours
```
kminder, line 13: Fix or remove this.

ci/chaos/JenkinsfileResiliencyTrigger
```diff
@@ -379,3 +385,13 @@ def metricBuildDuration() {
         }
     }
 }
+
+def getCronSchedule() {
+    if (env.BRANCH_NAME.equals("jmaron/VZ-5061")) {
+        return "H */3 * * *"
+    } else if (env.BRANCH_NAME.startsWith("release-")) {
```
kminder, line 392: I think this needs discussion.
Does it really make sense to run this daily on release branches or per commit?
How frequently would this run for master?

ci/chaos/JenkinsfileResiliencyTrigger
jmaron99: OK

ci/chaos/JenkinsfileResiliencyTrigger
jmaron99: This follows the pattern used for the VZ periodic builds, except for the 3 hours cycle instead of 2 hours.  The cron just dictates that for the master branch it'll run every 3 hours, for any release branches it'll run once per day.  I suppose the frequency of commits to release branches is relatively infrequent so the argument could be made to scan for commits in that branch, but daily seems relatively safe as well.

jmaron99: As far as I can tell the Jenkins build for our periodic builds still have a configured periodic scan to run periodically if no other builds are noted otherwise.  So I'd imagine a daily cron job would be augmented by other runs if a commit is pushed?  @agv426 ?

agv426: The periodic tests are not driven by commits, only by the schedule, but you can fire them off from Jenkins (ie: build with params). We haven't added in a trigger when a commit is promoted to stable, but we could do that if we decided to.

If you modeled this jobs configuration after the Periodics config, then it would do similar on whatever cadence you define the test to run at. But you could configure the job itself to fire off based on commit changes.

jmaron99: Then this configuration in the periodic VZ build has no effect?
verrazzano-periodic-triggered-tests_Config__Jenkins_

jmaron99: I see that the periodic VZ build file disables concurrent builds, which is probably a good directive for periodic runs, but I'm not sure why it would disable to configured trigger.

agv426: that is a scan repository trigger, that only scans the repository to see if something changes. Other triggers/settings control whether a job is triggered based on that information. Ie: just because it sees a new branch show up in a repo scan doesn't mean it will trigger a build there, there needs to be something else defining how builds are triggered.

kminder (approved):
Approved.

agv426 (dismissed):

agv426 (approved):

PR #2698, opened 2022-03-08
VZ-5171 re-entry into the install flow

shihchang:
# Description
* allow re-entry of install flow by resetting the component state to disabled to restart the install flow 
* add LastReconciledGeneration to component state
	* at the end of install or upgrade, component.LastReconciledGeneration is set to VerrazzanoCR.Generation
	* update(re-install) is triggered when component.LastReconciledGeneration < VerrazzanoCR.Generation
* test with following CR, it moved an 'InstallComplete' verrazzano to 'InstallStarted' to update/scale the authProxy to 2 replicas:  
```
apiVersion: install.verrazzano.io/v1alpha1
kind: Verrazzano
metadata:
  name: my-verrazzano
spec:
  profile: dev
  components:
    authProxy:
      kubernetes:
        replicas: 2
```
Fixes VZ-5171

# Checklist 

### DIFF
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -162,6 +162,8 @@ type ComponentStatusDetails struct {
 	State CompStateType `json:"state,omitempty"`
 	// The version of Verrazzano that is installed
 	Version string `json:"version,omitempty"`
+	// The generation of the last VZ resource the Component was reconciled against
+	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
 }
 
 // ConditionType identifies the condition of the install/uninstall/upgrade which can be checked with kubectl wait
--- a/platform-operator/controllers/verrazzano/install.go
+++ b/platform-operator/controllers/verrazzano/install.go
@@ -51,6 +51,20 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
 			compLog.Debugf("Did not find status details in map for component %s", comp.Name())
 			continue
 		}
+		if checkConfigUpdated(spiCtx, componentStatus) {
+			if err := r.updateComponentStatus(compContext, "PreInstall started", vzapi.CondPreInstall); err != nil {
+				return ctrl.Result{Requeue: true}, err
+			}
+			compLog.Oncef("Reset component %s state to %v for generation %v", compName, componentStatus.State, spiCtx.ActualCR().Generation)
+			if spiCtx.ActualCR().Status.State == vzapi.VzStateReady {
+				err = r.setInstallingState(vzctx.Log, spiCtx.ActualCR())
+				compLog.Oncef("Reset Verrazzano state to %v for generation %v", spiCtx.ActualCR().Status.State, spiCtx.ActualCR().Generation)
+				if err != nil {
+					spiCtx.Log().Errorf("Failed to reset state: %v", err)
+					return newRequeueWithDelay(), err
+				}
+			}
+		}
 		switch componentStatus.State {
 		case vzapi.CompStateReady:
 			// For delete, we should look at the VZ resource delete timestamp and shift into Quiescing/Uninstalling state
@@ -128,6 +142,11 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
 		return newRequeueWithDelay(), nil
 	}
 	return ctrl.Result{}, nil
+}
+
+func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.ComponentStatusDetails) bool {
+	return (componentStatus.State == vzapi.CompStateReady) &&
+		(ctx.ActualCR().Generation > componentStatus.LastReconciledGeneration)
 }
 
 // Check if the component can be installed in this Verrazzano installation based on version
--- a/platform-operator/controllers/verrazzano/install_test.go
+++ b/platform-operator/controllers/verrazzano/install_test.go
@@ -1,0 +1,369 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package verrazzano
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	networkingv1 "k8s.io/api/networking/v1"
+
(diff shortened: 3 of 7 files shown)


platform-operator/controllers/verrazzano/install.go
```diff
@@ -51,6 +51,18 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
 			compLog.Debugf("Did not find status details in map for component %s", comp.Name())
 			continue
 		}
+		if checkReInstall(spiCtx, componentStatus) {
```
pfmackin, line 54: I am missing why you have to do this.  We already call reconcile in the ready state.  Is this supposed to handle the case where install is in progress and the user changes the CR?

shihchang: When the component is in Ready state, this is to check if the config(generation) is changed. If the config is changed, it move the state from Ready back to Disabled to re-enter the install flow. I am trying to introducing a new state of 'ConfigUpdated' for this purpose, but it technically is just a alias of Disabled.

platform-operator/controllers/verrazzano/install_test.go
```diff
@@ -0,0 +1,352 @@
...
+	asserts.NoError(err)
+	asserts.Equal(vzapi.VzStateReady, vz.Status.State)
+	asserts.Equal(false, result.Requeue)
+	asserts.Equal(time.Duration(0), result.RequeueAfter)
+}
+
+// TestUpdate tests the reconcile func with updated generation
+// GIVEN a request to reconcile an verrazzano resource after install is completed
+// WHEN all components have the smaller LastReconciledGeneration than verrazzano CR in the request
+// THEN ensure a condition with type InstallStarted
```
michael-cico, line 248: I think the comment is supposed to say that the expected condition is `UpgradeStarted`?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -162,6 +162,8 @@ type ComponentStatusDetails struct {
 	State CompStateType `json:"state,omitempty"`
 	// The version of Verrazzano that is installed
 	Version string `json:"version,omitempty"`
+	// The generation of the last VZ resource the Component was reconciled against
+	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
```
michael-cico, line 166: I think during the Zoom review, we should determine if we want/need to gate things, or just let any update cause us to go through the motions.  Some folks have said we should not worry about gating the updates like this.

platform-operator/controllers/verrazzano/install_test.go
shihchang: fixed

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
shihchang: Setting the LastReconciledGeneration at the end of upgrade fixed the upgrade test failures. I suspect some components have issue running re-install

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -109,13 +109,18 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 				compName := comp.Name()
 				compContext := spiCtx.Init(compName).Operation(vzconst.UpgradeOperation)
 				installed, err := comp.IsInstalled(compContext)
+				componentStatus := cr.Status.Components[compName]
```
pfmackin, line 112: move this line to right before 120

platform-operator/controllers/verrazzano/install.go
shihchang: changed to entering pre-install state

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
shihchang: next subtask will investigate to remove LastReconciledGeneration

platform-operator/controllers/verrazzano/upgrade.go
shihchang: moved setting componentStatus.LastReconciledGeneratio to the spot you suggested

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -117,7 +117,11 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 					return newRequeueWithDelay(), nil
 				}
 				log.Oncef("Component %s is ready after being upgraded", compName)
-
+				componentStatus := cr.Status.Components[compName]
```
pfmackin, line 120: This actually needs to be done in the vzStateUpgradeDone where we update the status.  Once we move out of this state, and we have to re-reconcile then the change will be lost.   Just move this code to before line 132 in vzStateUpgradeDone

shihchang: moved setting LastReconciledGeneration to vzStateUpgradeDone

mgianatagh (approved):

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -126,6 +127,14 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 			msg := fmt.Sprintf("Verrazzano successfully upgraded to version %s", cr.Spec.Version)
 			log.Once(msg)
 			cr.Status.Version = targetVersion
+			for _, comp := range registry.GetComponents() {
+				compName := comp.Name()
+				componentStatus := cr.Status.Components[compName]
+				if componentStatus != nil {
+					log.Oncef("Component %s has been upgraded from generation %v to %v %v", compName, componentStatus.LastReconciledGeneration, cr.Generation, componentStatus.State)
+					componentStatus.LastReconciledGeneration = cr.Generation
```
michael-cico, line 135: Not for this PR, but we might want to set this as each component finishes, if we can; it might also be used to prevent re-entry into an already upgraded component.  Although converging the code paths is probably the better option.

Probably a larger discussion.

michael-cico (approved):

platform-operator/controllers/verrazzano/upgrade.go
shihchang: i did try setting componentStatus.LastReconciledGeneration at upgradeComponents but that does not fix upgrade test failures; need to investigate why.

PR #2702, opened 2022-03-08
VZ 5212: Use updated VMO CR

anders-swanson:


michael-cico (approved):

platform-operator/helm_config/charts/verrazzano/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
```diff
@@ -1,274 +1,324 @@
...
 spec:
-  conversion:
-    strategy: None
   group: verrazzano.io
   names:
     kind: VerrazzanoMonitoringInstance
     listKind: VerrazzanoMonitoringInstanceList
     plural: verrazzanomonitoringinstances
-    shortNames:
-      - vmi
```
fredtibbitts, line 16: Today users can run `kubectl get vmi ...` to get the VMIs, without the short name this no longer works, which would be surprising to users. We probably need to add the short name on the VMO side?

platform-operator/helm_config/charts/verrazzano/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
```diff
@@ -288,56 +338,93 @@ spec:
...
                   properties:
                     currentVersion:
-                      pattern: '[A-Za-z0-9-_]+'
                       type: string
                     desiredVersion:
-                      pattern: '[A-Za-z0-9-_]+'
                       type: string
-                contactemail:
+                  type: object
+              required:
```
fredtibbitts, line 384: If these weren't previously required, wondering if there will be some subtle change in behavior on the VMO side?

anders-swanson: There's actually a lot of 'required' things that we should probably put `omitempty` on. It didn't seem to make any difference during testing, but it's better to be sure.

anders-swanson: Looking at the types, none of these are pointer values. I don't think it will matter from the VPO side, since we initialize the CR before applying it (all 'required' values will be populated with the defaults, which is the expected values anyway).

fredtibbitts (dismissed):
Approved. It's a little troubling that the CRDs were so out of sync. I'd feel better if we had better test coverage of the VMO functionality.

fredtibbitts (dismissed):

markxnelson (approved):

PR #2705, opened 2022-03-09
Vicky/del resilency

absBitm:
# Description

PR to add the deletion confirmation status while Verrazzano comes up with any issues while upgrading,making sure that the vz is deletable.   

Fixes VZ-4984

# Checklist 

### DIFF
--- a/ci/chaos/Jenkinsfile
+++ b/ci/chaos/Jenkinsfile
@@ -406,7 +406,7 @@ pipeline {
                         echo "-----------------------------------------------------"
 
 
-                        verify Uninstall
+                        # verify Uninstall
                         ${LOOPING_TEST_SCRIPTS_DIR}/dump_cluster.sh ${WORKSPACE}/verrazzano/build/resources/post-uninstall-resources false
                         ${LOOPING_TEST_SCRIPTS_DIR}/verify_uninstall.sh ${WORKSPACE}/verrazzano/build/resources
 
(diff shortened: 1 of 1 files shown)


ci/chaos/Jenkinsfile
```diff
@@ -398,6 +398,10 @@ pipeline {
                         if [[ \${vz_status} == "UpgradeStarted" ]]; then
                             # Uninstall VZ
                             kubectl delete verrazzano my-verrazzano
+                            len=\$(kubectl get vz -o json | jq '.items | length')
```
kminder, line 401: Now that we have the command sorted out it seems like this won't actually fail if the delete fails, right?

absBitm: yes correct

kminder: Then you need to fix it.  The test/job should fail if the uninstall (i.e. delete) fails.

absBitm: if 'kubectl delete ..' fails Jenkins reads the non zero exit status and it will fail the job automatically.

ci/chaos/Jenkinsfile
```diff
@@ -387,40 +386,30 @@ pipeline {
...
 
-                    if [[ "uninstall.failed.upgrade" != ${params.CHAOS_TEST_TYPE} ]]
-                    then
-                        # wait for the upgrade to complete
-                        # Turn it off for now, Upgrade Fails
-                        kubectl wait --timeout=20m --for=condition=UpgradeComplete verrazzano/my-verrazzano
+                        # Uninstall VZ
+                        kubectl delete verrazzano my-verrazzano
 
+                        ## dump out uninstall logs
```
kminder, line 400: Doesn't the cluster dump below make capturing these extra?

ddrechse: i don't think so. This was copied straight form the uninstall pipeline

kminder (approved):
Approved.

PR #2722, opened 2022-03-11
VZ-5140 Added assets make target as dependency to docker-build target

anajosep:
# Description

Problem

When updates are made to the system-dashboards, the "assets" target must be explicitly run to regenerate the gobin data.  
Adding "assets" as a dependency to the "docker-build" target would help in resolving this issue. Another benefit of adding this dependency is that if someone makes a change to the dashboards without running "assets" target and pushes their branch, the build will fail because running the target will create changed files.  The build pipeline checks that running our targets do not create uncommitted file changes. go-bindata always updates the timestamp, in order to avoid it being updated each time, hardcoded timestamp is being set in the assets target.

Fixes VZ-5140

# Checklist 


mgianatagh (approved):

anajosep:
Changes no longer required as the fix is addressed by another PR https://github.com/verrazzano/verrazzano/pull/2571

PR #2745, opened 2022-03-16
VZ-5236 Reject changes to persistence settings until supported

guoyongzhang:
# Description

Reject any persistence changes that are not supported:
- If the volumeSource or defaultVolumeSource is changed from emptyDir to persistent storage, or vice-versa
- If the requested volume size is different than the current size
- If the storage class is changed

Fixes VZ-5236

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/mysql/mysql.go
+++ b/platform-operator/controllers/verrazzano/component/mysql/mysql.go
@@ -96,6 +96,7 @@ func appendMySQLOverrides(compContext spi.ComponentContext, _ string, _ string, 
 	// generate the MySQl PV overrides
 	kvs, err = generateVolumeSourceOverrides(compContext, kvs)
 	if err != nil {
+		compContext.Log().Error(err)
 		return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
 	}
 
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -96,6 +96,7 @@ func appendMySQLOverrides(compContext spi.ComponentContext, _ string, _ string,
 	// generate the MySQl PV overrides
 	kvs, err = generateVolumeSourceOverrides(compContext, kvs)
 	if err != nil {
+		compContext.Log().Error(err)
 		return []bom.KeyValue{}, ctrlerrors.RetryableError{Source: ComponentName, Cause: err}
```
michael-cico, line 100: Shouldn't this return `kvs` here?  This seems like a bug (since it's not part of your change)

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -149,9 +149,42 @@ func (c verrazzanoComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Ve
...
+	}
+	if oldSetting == nil && newSetting != nil || oldSetting != nil && newSetting == nil {
+		return fmt.Errorf("can not change default volume")
+	}
+	if oldSetting != nil && newSetting != nil {
+		if oldSetting.Storage != newSetting.Storage {
+			return fmt.Errorf("can not change default volume size")
+		}
+	}
+	if getOpenSearchDataNodeStorageOverride(old) != getOpenSearchDataNodeStorageOverride(new) {
```
michael-cico, line 169: I hate to say it, but I think @anders-swanson has a PR out that will work for all persistence changes... but until that's in, we should probably keep this.

michael-cico: To be clear, it will only work for Opensearch.  Which not that I think about it, is a bit of a mess, since we don't allow overrides at the component level for the other VMI.  We may need to change that this release, maybe Anders' change is going to address at least part of that.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
guoyongzhang: I think it is ok.  When there is err, the kvs returned is ignored.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
anders-swanson: Once I merge the PVC changes into VMO, a companion PR needs to be made into the Verrazzano repo to allow opensearch storage changes (but only for the data nodes).
Right now there is some code in vmi.go that also rejects (overwrites) attempted changes to the storage spec of the VMI resource.

PR #2746, opened 2022-03-16
aamitra/VZ-4061 Support NodePort across all components 

theabmitra:
# Description

Support Nodeport configurations across all components and VMI


platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -129,5 +128,107 @@ func getInstallArgs(cr *vzapi.Verrazzano) []vzapi.InstallArgs {
...
+	if serviceType == vzapi.LoadBalancer || serviceType == vzapi.NodePort {
+		svc := v1.Service{}
+		if err := client.Get(context.TODO(), types.NamespacedName{Name: ControllerName, Namespace: ComponentNamespace}, &svc); err != nil {
+			return "", err
+		}
+		// Test for IP from status, if that is not present then assume an on premises installation and use the externalIPs hint
+		if len(svc.Status.LoadBalancer.Ingress) > 0 {
+			ingressIP = svc.Status.LoadBalancer.Ingress[0].IP
+		} else if len(svc.Spec.ExternalIPs) > 0 {
+			// In case of OLCNE, the Status.LoadBalancer.Ingress field will be empty, so use the external IP if present
```
pfmackin, line 168: I think External IPs can be used for OKE with Nodeport type.  I assume OLCNE is just an example use case.

platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -129,5 +128,107 @@ func getInstallArgs(cr *vzapi.Verrazzano) []vzapi.InstallArgs {
...
+			ingressIP = svc.Spec.ExternalIPs[0]
+		} else {
+			return "", fmt.Errorf("No IP found for LoadBalancer service type")
```
pfmackin, line 171: You should log an error like...

return "", log.ErrorNewErr("Failed, IP not found in LoadBalancer or Nodeport service")

platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -129,5 +128,107 @@ func getInstallArgs(cr *vzapi.Verrazzano) []vzapi.InstallArgs {
...
+			return "", err
+		}
+		dnsSuffix = fmt.Sprintf("%s.%s", ingressIP, getWildcardDomain(dnsConfig))
+	} else if dnsConfig.OCI != nil {
+		dnsSuffix = dnsConfig.OCI.DNSZoneName
+	} else if dnsConfig.External != nil {
+		dnsSuffix = dnsConfig.External.Suffix
+	}
+	if len(dnsSuffix) == 0 {
+		return "", fmt.Errorf("Invalid OCI DNS configuration, no zone name specified")
```
pfmackin, line 194: Same logging comment

platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -129,5 +128,107 @@ func getInstallArgs(cr *vzapi.Verrazzano) []vzapi.InstallArgs {
...
+	if vz.Spec.Components.DNS != nil && vz.Spec.Components.DNS.OCI != nil {
+		return true
+	}
+	return false
+}
+
+// getWildcardDomain Get the wildcard domain from the Verrazzano config
+func getWildcardDomain(dnsConfig *vzapi.DNSComponent) string {
+	wildcardDomain := "nip.io"
+	if dnsConfig != nil && dnsConfig.Wildcard != nil && len(dnsConfig.Wildcard.Domain) > 0 {
```
pfmackin, line 230: slip.io is also a wildcard domain

michael-cico: I think he's just coding in the default, and overriding it if it's explicitly set via the `wildcard` field.

platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -129,5 +128,107 @@ func getInstallArgs(cr *vzapi.Verrazzano) []vzapi.InstallArgs {
 	if cr.Spec.Components.Ingress == nil {
 		return []vzapi.InstallArgs{}
 	}
+
 	return cr.Spec.Components.Ingress.NGINXInstallArgs
 }
+
+// Identify the service type, LB vs NodePort
+func GetServiceType(cr *vzapi.Verrazzano) (vzapi.IngressType, error) {
```
michael-cico, line 136: I think this branch might be old?  I think we'd moved these helpers to `platform-operator/internal/vzconfig` as they're used by different components.

platform-operator/controllers/verrazzano/component/nginx/nginx.go
theabmitra: With NodePort , externalIPS are mandatory in input. We need to add a validation for that in webhook

components:
    ingress:
      nginxInstallArgs:
      - name: controller.service.externalIPs
          valueList:
            - 100.111.142.150
      ports:
        - name: http
          nodePort: 30080
          port: 80
          protocol: TCP
          targetPort: 0
        - name: https
          nodePort: 30443
          port: 443
          protocol: TCP
          targetPort: 0
      type: NodePort
    istio:
      istioInstallArgs:
      - name: gateways.istio-ingressgateway.externalIPs
        valueList:
        - 100.111.142.150

platform-operator/controllers/verrazzano/component/nginx/nginx.go
theabmitra: yes , this is just assuming default

platform-operator/controllers/verrazzano/vzinstance/verrazzano_instance.go
```diff
@@ -64,20 +64,35 @@ func getComponentIngressURL(ingresses []networkingv1.Ingress, compContext spi.Co
...
 	}
-	url := fmt.Sprintf("https://%s", ingress.Spec.Rules[0].Host)
+	cr := compContext.EffectiveCR()
+	ingressType, err := vzconfig.GetServiceType(cr)
+	if err != nil {
+		return nil
+	}
+	switch ingressType {
+	case v1alpha1.LoadBalancer:
+		url = fmt.Sprintf("https://%s", ingress.Spec.Rules[0].Host)
```
michael-cico, line 88: Not for this PR, but is there a way we can generalize the ports for the single-LB configuration?  We had a question a while back about using an external LB to support both application and system traffic, which in the current system would require different ports most likely since we require a TCP/L4 LB at the front end right now.

Although if we consolidate all traffic behind Istio as has been discussed then that becomes less of an issue.

platform-operator/scripts/install/config.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
michael-cico, line 3: I think this script is obsolete, isn't it @pfmackin ?  Looks like there's a bunch of obsolete stuff under `scripts/install` actually.

Some follow-up cleanup for the backlog I think if you agree.

platform-operator/scripts/install/install-overrides/effective.config.json
```diff
@@ -0,0 +1,9 @@
+{
```
michael-cico, line 1: This looks like it was accidentally checked in.  This was generated by the old install scripts at runtime in the install job I believe (which is now gone).

platform-operator/scripts/install/config.sh
theabmitra: yes, that can be taken up by a separate PR

platform-operator/scripts/install/install-overrides/effective.config.json
theabmitra: done. i removed the file. Noticed its not there in master

michael-cico (approved):

platform-operator/controllers/verrazzano/vzinstance/verrazzano_instance.go
theabmitra: yes, it can be . We will have to use different ports 
So essentially today 
LB1 - Nginx - 80/443 are allowed
LB2 - Istio Ingress - 80/443/15021
They can both be combined into one , but the LB needs to have selectors for both nginx and istio. 
However when you hit 443 with LB3(the combined one), how do you diifferentiate where traffic should hit nginx pods/istio pods? Thats where the port number comes in . Ull need to have separate port numbers for 443 for each app. 
Article here - https://confluence.oraclecorp.com/confluence/display/~abhishek.a.mitra@oracle.com/Setting+up+Load+Balancer+across+multiple+K8s+deployments

michael-cico: You can't differentiate on the LB side without a different port, unless you can do L7 path-based routing, which we can't.  In my experiments I used a different LB listener/port for application traffic, and used the default 443 for system traffic.  

Actually, I think with the ability to set the nodePort on the services for Istio we'd be covered.  I'm updating a JIRA for that now.

theabmitra: yes, ull need a different port

PR #2749, opened 2022-03-17
VZ-5326: ExternalIP validations for NodePort

theabmitra:
# Description
When type is Nodeport , we need to ensure the Verrazzano spec has the externalIp addresses set correctly for both nginx and istio components

The PR has a few scanrios validated for ip address parameter present for the specific keys and also whether ip address itself is valid


platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
 	return currentSemVer.IsEqualTo(requestedSemVer) || currentSemVer.IsGreatherThan(requestedSemVer)
 
 }
+
+//validateExternalIPSForNodePort checks that externalIPs are set when Type=NodePort
+func validateExternalIPSForNodePort(spec *VerrazzanoSpec) error {
```
michael-cico, line 432: We added a component validator framework to allow individual components to validate CRs (create and update), I would implement this in the validator methods for the Istio and NGINX components in the VPO.  

See

https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/istio/istio_component.go#L132

and

https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/nginx/nginx_component.go#L75

You may need to add a `ValidateCreate` impl to the NGINX component.

michael-cico: We have the ability to update installs, so we need to validate both.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
...
+							if net.ParseIP(installarg.ValueList[0]) == nil {
+								return fmt.Errorf("Gateway external service key \"%v\" with IP \"%v\" is of invalid format. Must be a proper IP address format", installarg.Name, installarg.ValueList[0])
+							}
+						}
+					}
+				}
+
+			}
+
+			if !nginxExternalIPInputs {
```
michael-cico, line 467: when you move these to the validators, you can return separate errors for both NGINX and Istio, from those components.  Those will get rolled up and returned to the user I believe.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
...
+							return fmt.Errorf("At least one nginx external ips need to be set as an array for the key \"%v\"", installarg.Name)
+						}
+						if net.ParseIP(installarg.ValueList[0]) == nil {
+							return fmt.Errorf("Controller external service key \"%v\" with IP \"%v\" is of invalid format. Must be a proper IP address format", installarg.Name, installarg.ValueList[0])
+						}
+					}
+				}
+			}
+
+			if spec.Components.Istio != nil {
```
michael-cico, line 450: It might be more concise to do something like

```
externalIP := ""
// try to find the arg in the args list

if len(externalIP) == 0 {
  // return not found
}
if net.ParseIP() == nil {
  // error
}
// OK
```

michael-cico: This would apply to both NGINX and Istio

platform-operator/apis/verrazzano/v1alpha1/validate.go
michael-cico: When you do break this out, I think you may also be able to create a common helper between the two to validate the existence and format of the IP, just passing in the installArg key to look for.

theabmitra: The intention of this PR is to prevent install when config is not valid. I checked the code you are pointing at, those are validations after install has been kicked off.  What am i missing here ?

pfmackin: There is a new validate that is intended for Webhooks, that is what Mike is talking.  Sounds like this code should go there

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
 	return currentSemVer.IsEqualTo(requestedSemVer) || currentSemVer.IsGreatherThan(requestedSemVer)
 
 }
+
+//validateExternalIPSForNodePort checks that externalIPs are set when Type=NodePort
+func validateExternalIPSForNodePort(spec *VerrazzanoSpec) error {
+	var nginxExternalIPInputs, istioExternalIPInputs bool
+	if spec.Components.Ingress != nil {
```
pfmackin, line 434: better to return if nil and remove a layer of nesting.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
...
+		if spec.Components.Ingress.Type == NodePort {
```
pfmackin, line 435: Better to return if not Nodeport and remove a level of nesting.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -424,3 +427,53 @@ func ValidateVersionHigherOrEqual(currentVersion string, requestedVersion string
...
+								return fmt.Errorf("Gateway external service key \"%v\" with IP \"%v\" is of invalid format. Must be a proper IP address format", installarg.Name, installarg.ValueList[0])
+							}
+						}
+					}
+				}
+
+			}
+
+			if !nginxExternalIPInputs {
+				return fmt.Errorf("External ips for nginx was not set")
```
pfmackin, line 468: You might want to mention that Nodeport service type requires ExternalIP.  Same for other messages

platform-operator/apis/verrazzano/v1alpha1/validate.go
michael-cico: As paul mentinoed, those validators are called from the validating webhook during Create/Update.

theabmitra: Refactored the changes.

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/validate.go
theabmitra: Done

PR #2751, opened 2022-03-17
VZ-5030: 400 error page

mbenmclean:
This change picks up the new nginx image that returns the correct 400 error page after a directory traversal. It also adds a new e2e test to confirm this works successfully. To do this, it also mounts the 400 error page in the controller pod. Finally, it adds support for json error formats. It also removes the custom-http-errors from the controller configmap so it will not intercept upstream errors unneccessarily.

Fixes VZ-9999

# Checklist 


platform-operator/helm_config/overrides/ingress-nginx-values.yaml
```diff
@@ -146,55 +155,55 @@ defaultBackend:
             path: 504.htm
           - key: 505.html
             path: 505.htm
-          - key: 301.html
```
wmhopkins, line 149: Not sure what's going on here -- we're switching these keys from html to json?

platform-operator/thirdparty/charts/ingress-nginx/templates/default-backend-configmap-errors.yaml
```diff
@@ -184,4 +187,56 @@ data:
     <body>
     <center><h1>505 HTTP Version Not Supported/h1></center>
     </body>
-    </html>
\ No newline at end of file
+    </html>
+  301.json: |
```
wmhopkins, line 191: And we're adding json error messages?

tests/e2e/verify-install/web/nginx_errors_test.go
```diff
@@ -21,8 +21,7 @@ import (
 
 const (
 	minimumVersion = "1.3.0"
-	expected404    = "<html>\n<head><title>404 Not Found</title></head>\n<body>\n<center><h1>404 Not Found</h1></center>\n</body>\n</html>"
```
wmhopkins, line 24: Is this testing 404 from an arbitrary upstream, or the default backend? If the latter, we should probably leave it in.

platform-operator/helm_config/overrides/ingress-nginx-values.yaml
wmhopkins: Nevermind, I see what's going on.

platform-operator/thirdparty/charts/ingress-nginx/templates/default-backend-configmap-errors.yaml
wmhopkins: Nevermind, see what you're doing.

platform-operator/helm_config/overrides/ingress-nginx-values.yaml
```diff
@@ -25,6 +24,16 @@ controller:
     traffic.sidecar.istio.io/excludeInboundPorts: "80,443"
     traffic.sidecar.istio.io/includeInboundPorts: ""
     sidecar.istio.io/rewriteAppHTTPProbers: "true"
+  extraVolumeMounts:
```
wmhopkins, line 27: I wonder if we should include a few other errors we think the ingress controller may generate on its own, specifically 502 and/or 503 errors. Wouldn't hurt to add those since we're already adding 400.

That said, while I think you should try it and add those if it works, if it doesn't work for some reason I'd say go ahead and get this PR in without them rather than spend a bunch of time trying to get it working.

wmhopkins (commented):
Couple more comments.

tests/e2e/verify-install/web/nginx_errors_test.go
mbenmclean: Well I took the errorpage directive out so it just returns the upstream

wmhopkins (approved):
lgtm

PR #2759, opened 2022-03-17
VZ-5254 only restart apps if they have the old Istio sidecar proxy

pfmackin:
During upgrade, if Verrazzano installs a new Istio version with a new proxyv2 image, then all apps with the old Istio sidecar must be restarted.  Previously, we always restarted the apps.
Also, make some log messages consistent.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/istio/app_restart.go
+++ b/platform-operator/controllers/verrazzano/component/istio/app_restart.go
@@ -143,7 +143,7 @@ func startDomainIfNeeded(log vzlog.VerrazzanoLogger, client clipkg.Client, wlNam
 
 // RestartAllApps restarts all the applications
 func RestartAllApps(log vzlog.VerrazzanoLogger, client clipkg.Client, restartVersion string) error {
-	log.Progressf("RestartApps: restarting all apps")
+	log.Progressf("Restarting all OAM applications that have an old Istio proxy sidecar")
 
 	// Get the latest Istio proxy image name from the bom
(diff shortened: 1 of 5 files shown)


platform-operator/controllers/verrazzano/component/istio/app_restart.go
```diff
@@ -40,37 +47,37 @@ func StopDomainsUsingOldEnvoy(log vzlog.VerrazzanoLogger, client clipkg.Client)
...
-				log.Progressf("StopWebLogicApps: stopping domain for workload %s ", wlName)
-				err := stopDomain(client, appConfig.Namespace, wlName)
-				if err != nil {
-					return log.ErrorfNewErr("Failed annotating VerrazzanoWebLogicWorkload %s to stop the domain: %v", wlName, err)
-				}
-				return err
-			}
-		}
+	// Check if any pods contain the old Istio proxy image
+	found, oldImage := doesPodContainOldIstioSidecar(podList, istioProxyImage)
```
kminder, line 74: Are there conditions other than an image change that should result in restarting the Domain?

pfmackin: This code is only automatically restarting only if the Istio sidecar changed.  There are other use cases where WebLogic domain gets updated with logging config, for example, but the user has to initiate those changes which will cause WebLogic controller to restart the domain.

mgianatagh (approved):

PR #2760, opened 2022-03-18
VZ-5117.  Fixes for DNS and Cert-Manager updates

michael-cico:
# Description

Fixes the re-entry into the CertManager and Keycloak components when DNS and Cert-Manager are updated
- Have Keycloak rebuild the PKCE client URIs in the realm config
- Have Cert-Manager update the ClusterIssuer in both `PostInstall` and `PostUpgrade`
- Add some Cert-Manager and Keycloak component unit tests to validate updates, and update during upgrade
- Also fix an issue in the Keycloak component to fix the certificate checks

Fixes VZ-5117.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
@@ -54,6 +54,13 @@ const (
 	clusterResourceNamespaceKey = "clusterResourceNamespace"
 )
 
+// Template for ClusterIssuer for looking up Acme certificates for controllerutil.CreateOrUpdate
+const clusterIssuerLookupTemplate = `
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: verrazzano-cluster-issuer`
+
 // Template for ClusterIssuer for Acme certificates
 const clusterIssuerTemplate = `
 apiVersion: cert-manager.io/v1
@@ -133,13 +140,13 @@ func (c certManagerComponent) createOrUpdateClusterIssuer(compContext spi.Compon
 	}
 	if !isCAValue {
 		// Create resources needed for Acme certificates
-		err := createAcmeResources(compContext)
+		err := createOrUpdateAcmeResources(compContext)
 		if err != nil {
 			return compContext.Log().ErrorfNewErr("Failed creating Acme resources: %v", err)
 		}
 	} else {
 		// Create resources needed for CA certificates
-		err := createCAResources(compContext)
+		err := createOrUpdateCAResources(compContext)
 		if err != nil {
 			return compContext.Log().ErrorfNewErr("Failed creating CA resources: %v", err)
 		}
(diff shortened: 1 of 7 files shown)


platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -361,55 +325,69 @@ func createAcmeResources(compContext spi.ComponentContext) error {
...
 		OCIZoneName: ociDNSZoneName,
 	}
 
+	ciObject, err := createAcmeClusterIssuer(compContext.Log(), clusterIssuerData)
+	if err != nil {
+		return err
+	}
+	// Create a copy of the object for the case where it's an update, as what we generate will get wiped out on the
+	// GET operation if the resource exists
+	getCIObj := ciObject.DeepCopy()
```
pfmackin, line 356: not sure why this is needed.   CreateOrUpdate handles this, we should generate the ciObject in the callback (line 362)

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -961,7 +961,13 @@ func createVerrazzanoPkceClient(ctx spi.ComponentContext, cfg *restclient.Config
 	data := templateData{}
 
 	keycloakClients, err := getKeycloakClients(ctx)
-	if err == nil && clientExists(keycloakClients, "verrazzano-pkce") {
+	if err != nil {
+		return err
+	}
+	if clientExists(keycloakClients, "verrazzano-pkce") {
+		if err := updateKeycloakUris(ctx); err != nil {
```
pfmackin, line 968: Why are we updating the URIs here?  I would not expect that in a function named createVerrazzanoPkceClient

platform-operator/controllers/verrazzano/component/keycloak/keycloak_component.go
```diff
@@ -140,11 +140,11 @@ func (c KeycloakComponent) PostInstall(ctx spi.ComponentContext) error {
 
 // PostUpgrade Keycloak-post-upgrade processing, create or update the Kiali ingress
 func (c KeycloakComponent) PostUpgrade(ctx spi.ComponentContext) error {
+	// populate the certificate names before calling PostInstall on Helm component because those will be needed there
+	c.HelmComponent.Certificates = c.GetCertificateNames(ctx)
```
pfmackin, line 144: Why would helm postUpgrade need certs?  Seems like that is component specific logic, not generic helm component logic.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
michael-cico: I generated it for the Get/lookup.  but I'll see about moving it around, makes sense.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
michael-cico: It's called during Install. I can rename the function, but because we're in the install flow during an update we need to update the PCKE client when the DNS domain/endpoints change.

platform-operator/controllers/verrazzano/component/keycloak/keycloak_component.go
michael-cico: It's because when `HelmComponent.PostUpgrade` calls `CheckCertificatesReady` it ends up calling the `HelmComponent` impl, not the `keycloakCompoment` impl, which returns an empty set.  We had to do something similar for the ingress checks.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
michael-cico: Fixed next push

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
michael-cico: Fixed next push

pfmackin (approved):

PR #2765, opened 2022-03-18
VZ-5319 Support weblogic-monitoring-exporter in private registry configurations

guoyongzhang:
# Description

- add weblogic-monitoring-exporter image to BOM
-  weblogic-monitoring-exporter image in BOM is passed to VPO as env WEBLOGIC_MONITORING_EXPORTER_IMAGE
- VAO set spec.monitoringExporter.image in domain to the specified WEBLOGIC_MONITORING_EXPORTER_IMAGE; unset spec.monitoringExporter.image for empty WEBLOGIC_MONITORING_EXPORTER_IMAGE, which means WebLogic Monitoring Exporter uses its default image.
- update private registry instruction to add more public images

Fixes VZ-5319

# Checklist 

### DIFF
--- a/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
+++ b/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
@@ -4,17 +4,20 @@
 package wlsworkload
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"fmt"
-	vzlogInit "github.com/verrazzano/verrazzano/pkg/log"
 	"math/big"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


michael-cico:
I kind of had something different in mind:

- Pass in the registry overrides at VAO install time (same env vars we give to the VPO)
- Have the VAO query the WME image directly from the BOM using the code in `pkg/bom`

That being said, this is a very focused solution on the WME problem.  I think the question is is there a problem with the VAO looking something like this up directly?  I think the advantage of what I propose is that if we need to do something like this for other images from the VAO down the road, we won't have to make chart changes, etc.

The advantage of this solution is that the VAO doesn't know about the BOM, and only has what it needs.

michael-cico (approved):

michael-cico (approved):

PR #2770, opened 2022-03-20
WIP: VZ-5117.  Allow external-dns to be disabled

michael-cico:
# Description

Allow external DNS to be disabled post-install
- Allowing component-level uninstall; will still be gated by other component validators
- Add Uninstall/PostUninstall hooks to component interface, update mocks and unit tests
- Remove external-dns component disable validation
- Rename install.go --> reconcile_components.go

Fixes VZ-5117.

# Checklist 


pkg/istio/istioctl.go
```diff
@@ -102,6 +102,19 @@ func VerifyInstall(log vzlog.VerrazzanoLogger) (stdout []byte, stderr []byte, er
 	return stdout, stderr, nil
 }
 
+// Uninstall runs istioctl -x uninstall to remove Istio
+func Uninstall(log vzlog.VerrazzanoLogger) (stdout []byte, stderr []byte, err error) {
```
pfmackin, line 106: This will break all pods using sidecars, including applications.

michael-cico: We're preventing Istio uninstall from the webhook.

michael-cico: Actually, I'll remove this, since the `istio_component.go` change just returns an unsupported operation error for install in this PR.

PR #2779, opened 2022-03-21
Update manifest with prod storage default

anders-swanson:


platform-operator/controllers/verrazzano/component/spi/testdata/basicProdMerged.yaml
```diff
@@ -54,6 +54,8 @@ spec:
           value: "4.8Gi"
         - name: nodes.data.requests.storage
           value: "50Gi"
+        - name: nodes.master.requests.storage
```
kminder, line 57: Should these test data files be using a value other than the default to ensure they are actually getting used?

pfmackin (approved):

PR #2785, opened 2022-03-22
VZ-5087 Rebuild keycloak config on upgrade using ephemeral storage

mgianatagh:
# Description

When upgrading an environment configured to use ephemeral storage, rebuild the keycloak configuration to it's default settings.  Previous to this change the upgrade would fail because the Keycloak configuration is lost when the MySQL pod recycles.

A subsequent PR will address the issue of the MySQL pod unexpectedly getting recycled at any time.

Fixes VZ-9999

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -1307,7 +1307,7 @@ func rebuildKeycloakConfiguration(ctx spi.ComponentContext) error {
 	err = loginKeycloak(ctx, cfg, cli)
 	if err != nil {
 		pod.ObjectMeta.Annotations[vzconst.VerrazzanoRestartAnnotation] = strconv.Itoa(int(pod.Generation))
-		err = ctx.Client().Update(context.TODO(), &pod)
+		err = ctx.Client().Delete(context.TODO(), &pod)
 		return err
 	}
 
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -1317,7 +1317,7 @@ func checkConfiguration(ctx spi.ComponentContext) error {
 		return err
 	}
 
-	// Try getting the Keycloak groups.  If they exist, a configuration exists.
+	// Try getting the Keycloak groups.  If they exist, a configuration exists, no need to recreate.
 	_, err = getKeycloakGroups(ctx)
 	if err == nil {
 		return nil
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"text/template"
 
+	vzconst "github.com/verrazzano/verrazzano/pkg/constants"
+
 	"github.com/verrazzano/verrazzano/pkg/bom"
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
 	vzos "github.com/verrazzano/verrazzano/pkg/os"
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -662,7 +663,7 @@ func loginKeycloak(ctx spi.ComponentContext, cfg *restclient.Config, cli kuberne
 	ctx.Log().Debugf("loginKeycloak: Login Cmd = %s", maskPw(loginCmd))
 	stdOut, stdErr, err := k8sutil.ExecPod(cli, cfg, kcPod, ComponentName, bashCMD(loginCmd))
 	if err != nil {
-		ctx.Log().Errorf("Component Keycloak failed retrieving logging into: stdout = %s: stderr = %s", stdOut, stdErr)
+		ctx.Log().Errorf("Component Keycloak failed logging into: stdout = %s: stderr = %s", stdOut, stdErr)
```
michael-cico, line 666: Error message still a bit dangling... "failed logging into", what did we fail logging into?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1276,3 +1277,70 @@ func isKeycloakReady(ctx spi.ComponentContext) bool {
 	prefix := fmt.Sprintf("Component %s", ctx.GetComponent())
 	return status.StatefulSetsAreReady(ctx.Log(), ctx.Client(), statefulset, 1, prefix)
 }
+
+// checkConfiguration - check the Keycloak configuration when using ephemeral storage.
```
michael-cico, line 1281: maybe a better name here, like `checkEphemeralStorageConfiguration`, or `createOrUpdateEphemeralStorageConfig`.

The install flow is doing something similar now... is there any reason why we wouldn't just call the same flow, or is not all of the Keycloak configuration code idempotent?

michael-cico: The install/update flow call `configureKeycloakRealms` in `PostInstall`.  It seems to be behaving, but I'm wondering if it should be more conservative.

Ideally `configureKeycloakRealms` would be idempotent I'd think, and just create the realm objects within if they don't exist.  Update seems to be working so far without issue.

Since this seems to be working I don't want to rock the boat, but it might be good be calling the same logic in all cases (install/update/upgrade).

mgianatagh: The upgrade flow will eventually call the same logic as the install flow, but it is handling a case that would not exist on install.  It is repairing as best it can the keycloak configuration back to the default settings.  Also, in my next PR the checkConfiguration function (I will rename it) is also going to be run periodically to handle the case where MySQL gets recycled at any point in time after an install or upgrade.

mgianatagh: I made the logic common to install, upgrade and update

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
mgianatagh: fixed

michael-cico (approved):

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -498,10 +499,39 @@ func updateKeycloakUris(ctx spi.ComponentContext) error {
...
+	}
+
 	cfg, cli, err := k8sutil.ClientConfig()
 	if err != nil {
 		return err
 	}
+
+	// If ephemeral storage is configured, additional steps may be required to
+	// rebuild the configuration lost due to MySQL pod getting restarted.
+	if (ctx.EffectiveCR().Spec.Components.Keycloak != nil) && (ctx.EffectiveCR().Spec.Components.Keycloak.MySQL.VolumeSource == nil) {
```
pfmackin, line 521: it would be better to look at the mysql pod to see if it using ephemeral storage (emptydir)

PR #2805, opened 2022-03-24
VZ-5370.  Trigger app ingress trait reconciles during DNS update

michael-cico:
# Description

Add Watch on Console/Authrpoxy ingress to force us to reconcile IngressTraits on DNS updates
- If the Ingress changes, map the change to a reconcile of all IngressTrait objects

Fixes VZ-5370

# Checklist 

### DIFF
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller.go
@@ -463,9 +463,6 @@ func (r *Reconciler) mutateGateway(gateway *istioclient.Gateway, trait *vzapi.In
 	if err != nil {
 		return err
 	}
-	if len(gateway.Spec.Servers) > 0 {
-		hosts = appendToConfiguredHosts(hosts, gateway.Spec.Servers[0].Hosts)
-	}
 
 	// Set the spec content.
 	gateway.Spec.Selector = map[string]string{"istio": "ingressgateway"}
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller.go
@@ -97,62 +97,6 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) (err error) {
 	return r.setupWatches()
 }
 
-func (r *Reconciler) setupWatches() error {
-	// Set up a watch on the Console/Authproxy ingress to watch for changes in the Domain name;
-	// - Define a mapping to the existing ingress traits and invoke the IngressTrait Reconciler
-	mapFn := handler.ToRequestsFunc(
-		func(a handler.MapObject) []reconcile.Request {
-			return r.createIngressTraitReconcileRequests()
-		})
-	// setup a watch on the Console ingress
-	return r.Controller.Watch(
-		&source.Kind{
-			Type: &k8net.Ingress{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      constants.VzConsoleIngress,
-					Namespace: constants.VerrazzanoSystemNamespace,
-				},
-			},
-		},
-		&handler.EnqueueRequestsFromMapFunc{
-			ToRequests: mapFn,
-		},
-		predicate.Funcs{
-			UpdateFunc: updateFunc,
-		},
-	)
-}
-
-func updateFunc(updateEvent event.UpdateEvent) bool {
-	oldIngress := updateEvent.ObjectOld.(*k8net.Ingress)
-	newIngress := updateEvent.ObjectNew.(*k8net.Ingress)
-	return !reflect.DeepEqual(oldIngress.Spec.TLS, newIngress.Spec.TLS)
-}
-
-//annotateIngressTraits Adds an annotation to existing IngressTrait objects to force them to reconcile;
-//  this is necessary when the DNS domain has been updated and we need to generate new records for applications using
-//  default hostnames that we generate.  Default hostnames are generated based on the main console/authproxy ingress,
-//  so we do it PostInstall/PostUpgrade of the Verrazzano component, which manages that ingress.
-func (r *Reconciler) createIngressTraitReconcileRequests() []reconcile.Request {
-	requests := []reconcile.Request{}
-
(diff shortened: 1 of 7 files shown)


application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -74,18 +78,79 @@ var (
...
 		For(&vzapi.IngressTrait{}).
-		Complete(r)
+		Build(r)
+	if err != nil {
+		return err
+	}
+	return r.setupWatches()
+}
+
+func (r *Reconciler) setupWatches() error {
```
pfmackin, line 100: setupWatches is only called once when the controller is initialize.  This won't handle ingressTraits that are created after setWatches is called.   The way we did jobs is to setup the watch in the reconcile loop for each VZ CR (a watch per resource), gated by a set of VZ CRs so that it is only done once for CR instance.

michael-cico: I'm not sure I completely follow.  The watch is for the Ingress, which there's only one.  When that is updated and the handler is called we map that event to reconcile all the current IngressTrait objects, which get queued up for the controller.

I took a look at the example controller in the Kubebuilder book it looks like they're setting up watches during the controller setup.

https://book-v1.book.kubebuilder.io/basics/simple_controller.html

I'm wondering though if I should set up the watch on the Ingress types and filter the name in the predicate callback.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -74,18 +78,79 @@ var (
...
+	)
+}
+
+func updateFunc(updateEvent event.UpdateEvent) bool {
+	oldIngress := updateEvent.ObjectOld.(*k8net.Ingress)
+	newIngress := updateEvent.ObjectNew.(*k8net.Ingress)
+	return !reflect.DeepEqual(oldIngress.Spec.TLS, newIngress.Spec.TLS)
+}
+
+//annotateIngressTraits Adds an annotation to existing IngressTrait objects to force them to reconcile;
```
kminder, line 132: Comment is out of date with the code.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -122,6 +122,8 @@ func (c verrazzanoComponent) PostInstall(ctx spi.ComponentContext) error {
 // PostUpgrade Verrazzano-post-upgrade processing
 func (c verrazzanoComponent) PostUpgrade(ctx spi.ComponentContext) error {
 	ctx.Log().Debugf("Verrazzano component post-upgrade")
+	c.HelmComponent.IngressNames = c.GetIngressNames(ctx)
```
kminder, line 125: Is this fixing a particular issue?  A comment would be nice since I can't immediately understand why this is needed.

michael-cico: OK.  Yes, this has to do with an issue of these impls and how Go works. I'll update the code with a comment.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -74,18 +78,79 @@ var (
...
+//  this is necessary when the DNS domain has been updated and we need to generate new records for applications using
+//  default hostnames that we generate.  Default hostnames are generated based on the main console/authproxy ingress,
+//  so we do it PostInstall/PostUpgrade of the Verrazzano component, which manages that ingress.
+func (r *Reconciler) createIngressTraitReconcileRequests() []reconcile.Request {
+	requests := []reconcile.Request{}
+
+	ingressList := vzapi.IngressTraitList{}
```
pfmackin, line 139: var should be ingressTraitList or traitList It confused me that is was called ingressList since there is an ingress type.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
pfmackin: what happens if a new ingresstrait is created?  It won't be mapped, will it?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -122,6 +122,8 @@ func (c verrazzanoComponent) PostInstall(ctx spi.ComponentContext) error {
...
+	c.HelmComponent.Certificates = c.GetCertificateNames(ctx)
```
pfmackin, line 126: I had a comment on this before, i think this is a little fragile.  It would be better if the interface defined GetIngressNames and GetComponentNames, and have each component return nil if there aren't any.  That way the helmComponent code can just call those functions and it forces every component to think about it and opt out.

michael-cico: Agreed...the interface does, and HelmComponent does, but we have to get rid of the defaults in HelmComponent.  I'll address that in a different PR.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
michael-cico: It seems to be handled OK in my manual testing.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
michael-cico: Fixed and pushed

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
michael-cico: Based on Paul's comment, I'm going to address this in a different PR.  We'll need to get rid of the HelmComponent stubs and make each component implement it.

application-operator/controllers/ingresstrait/ingresstrait_controller.go
michael-cico: Fixed and pushed

pfmackin (approved):

PR #2812, opened 2022-03-25
VZ-5414 rename tls secret for verrazzano and keycloak

guoyongzhang:
# Description

rename the tls secret from env-secret to verrazzano-tls and keycloak-tls

Fixes VZ-5414

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak_component.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak_component.go
@@ -190,12 +190,10 @@ func (c KeycloakComponent) getInstallArgs(vz *vzapi.Verrazzano) []vzapi.InstallA
 
 // GetCertificateNames - gets the names of the ingresses associated with this component
 func (c KeycloakComponent) GetCertificateNames(ctx spi.ComponentContext) []types.NamespacedName {
-	var certificateNames []types.NamespacedName
-
-	certificateNames = append(certificateNames, types.NamespacedName{
-		Namespace: ComponentNamespace,
-		Name:      fmt.Sprintf("%s-secret", ctx.EffectiveCR().Spec.EnvironmentName),
-	})
-
-	return certificateNames
+	return []types.NamespacedName{
+		{
+			Namespace: ComponentNamespace,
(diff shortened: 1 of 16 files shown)


platform-operator/controllers/verrazzano/component/keycloak/keycloak_component.go
```diff
@@ -190,12 +190,10 @@ func (c KeycloakComponent) getInstallArgs(vz *vzapi.Verrazzano) []vzapi.InstallA
...
 func (c KeycloakComponent) GetCertificateNames(ctx spi.ComponentContext) []types.NamespacedName {
-	var certificateNames []types.NamespacedName
-
-	certificateNames = append(certificateNames, types.NamespacedName{
-		Namespace: ComponentNamespace,
-		Name:      fmt.Sprintf("%s-secret", ctx.EffectiveCR().Spec.EnvironmentName),
-	})
-
-	return certificateNames
+	return []types.NamespacedName{
```
michael-cico, line 193: can you declare this as a variable and reference it from the component struct? Then we can rid of the code on lines 138-139 and 145-146.  We had to put in some kluge to work around Go-isms around composites and polymorphism.

michael-cico: Note that we can't do this in the verrazzano component as it still needs to compute the ste of ingresses and certificates based on what's enabled.

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-ingress.yaml
```diff
@@ -50,4 +50,4 @@ spec:
   tls:
     - hosts:
         - verrazzano.{{ .Values.config.envName }}.{{ .Values.config.dnsSuffix }}
-      secretName: {{ .Values.config.envName }}-secret
+      secretName: verrazzano-tls
```
michael-cico, line 53: Maybe we can reference this as an overridable value from the helm chart, although I wouldn't bother if we don't do that anywhere else.

michael-cico (approved):

PR #2816, opened 2022-03-25
VZ-5278: Bug fix - Do not overwrite WebLogic domain logHome

fredtibbitts:
# Description

This PR addresses the bug where Verrazzano always sets the WebLogic domain `logHome` field. If the user has already set a `logHome`, Verrazzano overwrites it. With these changes, if `logHome` is already set, Verrazzano will honor it.

I manually tested this on an OKE cluster with a `logHome` set to use a persistent volume. I verified that logs show up correctly in OpenSearch. I will create an end-to-end test for this in a follow-on PR.

Fixes VZ-5278

# Checklist 

### DIFF
--- a/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
+++ b/application-operator/controllers/wlsworkload/weblogicworkload_controller.go
@@ -204,6 +204,8 @@ var specConfigurationWDTConfigMap = []string{specField, "configuration", "model"
 var specMonitoringExporterFields = []string{specField, "monitoringExporter"}
 var specRestartVersionFields = []string{specField, "restartVersion"}
 var specServerStartPolicyFields = []string{specField, "serverStartPolicy"}
+var specLogHomeFields = []string{specField, "logHome"}
+var specLogHomeEnabledFields = []string{specField, "logHomeEnabled"}
 
 // this struct allows us to extract information from the unstructured WebLogic spec
 // so we can interface with the FLUENTD code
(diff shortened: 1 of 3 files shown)


bryantidd (approved):

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -555,16 +581,21 @@ func (r *Reconciler) addLogging(ctx context.Context, log vzlog.VerrazzanoLogger,
...
-	// logHome and logHomeEnabled fields need to be set to turn on logging
-	err = unstructured.SetNestedField(weblogic.Object, getWLSLogHome(name), specField, "logHome")
-	if err != nil {
-		log.Errorf("Failed to set logHome: %v", err)
-		return err
+	// set logHome if it was not already specified in the domain spec
+	if logHome == "" {
+		err = unstructured.SetNestedField(weblogic.Object, getWLSLogHome(name), specLogHomeFields...)
+		if err != nil {
+			log.Errorf("Failed to set logHome: %v", err)
```
pfmackin, line 588: normally we return log.ErrorfNewError(...) so that the returned error is exactly the one that was logged.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
```diff
@@ -504,21 +506,45 @@ func (r *Reconciler) addLogging(ctx context.Context, log vzlog.VerrazzanoLogger,
...
+				volumeMountPath = mount.MountPath
+				volumeName = mount.Name
+				foundVolumeMount = true
+				break
+			}
+		}
+
+		if !foundVolumeMount {
+			// user specified logHome but it's not on any volume, Fluentd sidecar won't be able to collect logs
+			log.Info("Unable to find a volume mount for domain logHome, log collection will not work")
```
pfmackin, line 528: should this be an error?  Also do we allow the domain to be created with this issue?

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
fredtibbitts: That was existing code (and this file has `log.Errorf` all over the place). I think it would be a good idea to submit a separate PR that replaces all of those instances.

application-operator/controllers/wlsworkload/weblogicworkload_controller.go
fredtibbitts: We had a slack conversation about this. Yes, we're allowing the domain creation in this case. I'm going to add to the public docs so that it's clear that using a logHome that doesn't reside on a volume means that logs can't be collected.

I would have liked the log message to be a warning but the logger only has info and error as far as I can tell.

PR #2833, opened 2022-03-29
Revert "Revert "VZ-5087 Repair Keycloak configuration when MySQL restarts and using ephemeral storage""

mgianatagh:
Reverts verrazzano/verrazzano#2831


kminder (approved):
Approved.

PR #2838, opened 2022-03-29
Use Data Streams (OpenSearch)

anders-swanson:
- Verrazzano will now use Data Streams instead of indices in OpenSearch
- External ES test was updated to use OpenSearch instead of ElasticSearch
- Fluentd plugin was updated for OpenSearch


platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
```diff
@@ -863,27 +863,17 @@ data:
     </filter>
 
   es-output.conf: |
-    <match **>
-      @type elasticsearch
-      @id out_elasticsearch
+    <match kubernetes.**kube-** kubernetes.**verrazzano-system** kubernetes.**verrazzano-install** kubernetes.**cattle-** kubernetes.**rancher-** kubernetes.**fleet-** kubernetes.**ingress-nginx** kubernetes.**istio-system** kubernetes.**keycloak** kubernetes.**cert-manager**  kubernetes.**_monitoring_** kubernetes.**_metallb-** kubernetes.**_local-path-storage_** systemd.** fluentd.monitor.metrics>
```
kminder, line 866: Can you add a code comment covering what you are trying to match and specifically not match here?
The reason behind switching from ** to a targeted list probably needs to be captured.

kminder, line 866: Can you add a code comment about the purpose of this match?

tests/e2e/pkg/elasticsearch.go
```diff
@@ -26,19 +26,57 @@ const (
 	ISO8601Layout = "2006-01-02T15:04:05.999999999-07:00"
 )
 
+func GetOpenSearchSystemIndex(name string) string {
```
kminder, line 29: Please add a function comment.
Why does the presence of a dataStreamVersion mean the index name is verrazzano-system?

tests/e2e/pkg/elasticsearch.go
```diff
@@ -26,19 +26,57 @@ const (
...
+	if dataStreamVersion {
+		return "verrazzano-system"
+	}
+	if name == "systemd-journal" {
+		return "verrazzano-systemd-journal"
+	}
+	return "verrazzano-namespace-" + name
+}
+
+func GetOpenSearchAppIndex(namespace string) string {
```
kminder, line 49: Please add a function comment.  Especially some detail around how the VZ version impacts the result.

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
ayaranga: This is to match all "system" namespaces so that the system log records are sent to verrazzano-system data stream. We will add a code comment for the same.

anders-swanson:
@kminder should be updated and ready for re-review

tests/e2e/config/scripts/cert.conf
```diff
@@ -0,0 +1,13 @@
...
+[ req ]
+prompt=no
+distinguished_name = distinguished_name
+req_extensions = extensions
+
+[ distinguished_name ]
+organizationName = ExternalOS
+
+[ extensions ]
+subjectAltName = critical,IP:172.18.0.238
```
anders-swanson, line 13: The OpenSearch server certificate needs to reference the LoadBalancer IP of the master service, since we are exposing the service via LoadBalancer..

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
```diff
@@ -896,14 +887,51 @@ data:
...
+        overflow_action drop_oldest_chunk
+      </buffer>
+    </match>
+    <match **>
+      @type opensearch_data_stream
+      @id out_all
+      @log_level info
+      log_es_400_reason true
+
+      data_stream_name verrazzano-application-${$.kubernetes.namespace_name}
```
kminder, line 913: Is going to produce the correct stream names for system components given the match **

tests/e2e/config/scripts/create-external-os.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [ "$CLUSTER_NUMBER" != "1" ]; then
+  echo "Skipping creating external Elasticsearch on a managed cluster"
+  exit 0
+fi
+
+SCRIPT_DIR=$(cd $(dirname "$0"); pwd -P)
+
+# This corresponds to OpenSearch 1.2.3
+OPENSEARCH_CHART_VERSION="1.5.7"
```
kminder, line 21: Is this version going to matter over time?

anders-swanson: Yes, if there are breaking changes in opensearch we might need to use a new chart version. Previously the external ES charts had used some 7.x version of elasticsearch

platform-operator/helm_config/charts/verrazzano/templates/verrazzano-logging.yaml
anders-swanson: Since this is the second match statement, it will hit anything that the first match did not. The first match is set to match all the system components

kminder (dismissed):
Approved.

tests/e2e/pkg/common.go
```diff
@@ -500,6 +500,17 @@ func CheckNSFinalizerRemoved(ns string, clientset *kubernetes.Clientset) bool {
 	return namespace.Finalizers == nil
 }
 
+func getKubeConfigPath(kubeconfigPath string) string {
+	var err error
+	if kubeconfigPath == "" {
+		kubeconfigPath, err = k8sutil.GetKubeConfigLocation()
+		if err != nil {
+			Log(Error, fmt.Sprintf("Failed to get default kubeconfig path: %v", err))
```
kminder, line 508: Seems like an issue that this failure is logged but not returned.

tests/e2e/pkg/kubernetes.go
```diff
@@ -463,6 +463,19 @@ func IsWebLogicOperatorEnabled(kubeconfigPath string) bool {
 	return *vz.Spec.Components.WebLogicOperator.Enabled
 }
 
+// IsOpenSearchEnabled returns true if the OpenSearch component is not set, or the value of its Enabled field otherwise
+func IsOpenSearchEnabled(kubeconfigPath string) bool {
+	vz, err := GetVerrazzanoInstallResourceInCluster(kubeconfigPath)
+	if err != nil {
+		Log(Error, fmt.Sprintf("Error Verrazzano Resource: %v", err))
+		return true
```
kminder, line 471: Seems like the error should be returned vs assuming the component is enabled.

tests/e2e/pkg/kubernetes.go
```diff
@@ -463,6 +463,19 @@ func IsWebLogicOperatorEnabled(kubeconfigPath string) bool {
...
+	}
+	if vz != nil && vz.Spec.Components.Elasticsearch != nil && vz.Spec.Components.Elasticsearch.Enabled != nil {
+		return *vz.Spec.Components.Elasticsearch.Enabled
+	}
+	return true
```
kminder, line 476: This means OS is enabled by default.  I think it should be disabled by default.  What is the current behavior?

tests/e2e/pkg/kubernetes.go
anders-swanson: that does seem weird

tests/e2e/pkg/kubernetes.go
anders-swanson: If there's no component definition, then it's disabled. This code is wrong.

tests/e2e/pkg/common.go
anders-swanson: i dont really understand the point of this function, seems the entire point is to swallow the error of another function. I will remove it.

kminder (approved):
Approved.

PR #2847, opened 2022-03-31
VZ-5293 allow update during install

shihchang:
# Description

* This is to allow config update during installation or another update operation
* MySQL component currently does not have changeable config in CR and should not allow update. Rerunning MySQL installation very often cause Keycloak failure.

Fixes VZ-5293

# Checklist 

### DIFF
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -165,6 +165,8 @@ type ComponentStatusDetails struct {
 	Version string `json:"version,omitempty"`
 	// The generation of the last VZ resource the Component was reconciled against
 	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
+	// The generation of the last VZ resource the Component is being installed against
+	InstallingGeneration int64 `json:"installingGeneration,omitempty"`
 }
 
 // ConditionType identifies the condition of the install/uninstall/upgrade which can be checked with kubectl wait
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -163,9 +163,9 @@ type ComponentStatusDetails struct {
 	State CompStateType `json:"state,omitempty"`
 	// The version of Verrazzano that is installed
 	Version string `json:"version,omitempty"`
-	// The generation of the last VZ resource the Component was reconciled against
+	// The generation of the last VZ resource the Component was successfully reconciled against
 	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
-	// The generation of the last VZ resource the Component is being reconciled against
+	// The generation of the VZ resource the Component is currently being reconciled against
 	ReconcilingGeneration int64 `json:"reconcilingGeneration,omitempty"`
 }
 
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -165,8 +165,8 @@ type ComponentStatusDetails struct {
 	Version string `json:"version,omitempty"`
 	// The generation of the last VZ resource the Component was reconciled against
 	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
-	// The generation of the last VZ resource the Component is being installed against
-	InstallingGeneration int64 `json:"installingGeneration,omitempty"`
+	// The generation of the last VZ resource the Component is being reconciled against
+	ReconcilingGeneration int64 `json:"reconcilingGeneration,omitempty"`
 }
 
 // ConditionType identifies the condition of the install/uninstall/upgrade which can be checked with kubectl wait
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -689,15 +689,15 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	}
 	if conditionType == installv1alpha1.CondInstallComplete {
 		cr.Status.VerrazzanoInstance = vzinstance.GetInstanceInfo(compContext)
-		if componentStatus.InstallingGeneration > 0 {
-			componentStatus.LastReconciledGeneration = componentStatus.InstallingGeneration
-			componentStatus.InstallingGeneration = 0
+		if componentStatus.ReconcilingGeneration > 0 {
+			componentStatus.LastReconciledGeneration = componentStatus.ReconcilingGeneration
+			componentStatus.ReconcilingGeneration = 0
 		} else {
 			componentStatus.LastReconciledGeneration = cr.Generation
 		}
 	} else {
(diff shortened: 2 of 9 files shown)


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -165,6 +165,8 @@ type ComponentStatusDetails struct {
 	Version string `json:"version,omitempty"`
 	// The generation of the last VZ resource the Component was reconciled against
 	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
+	// The generation of the last VZ resource the Component is being installed against
+	InstallingGeneration int64 `json:"installingGeneration,omitempty"`
```
michael-cico, line 169: Should this be called `ReconcilingGeneration`?  For consistency.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -165,6 +165,8 @@ type ComponentStatusDetails struct {
 	Version string `json:"version,omitempty"`
 	// The generation of the last VZ resource the Component was reconciled against
 	LastReconciledGeneration int64 `json:"lastReconciledGeneration,omitempty"`
+	// The generation of the last VZ resource the Component is being reconciled against
```
michael-cico, line 168: Probably re-word this, something like

```
The generation of the VZ resource the Component is currently being reconciled against
```

So, in the `Ready` state, this will == `LastReconciledGeneration` right?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -165,6 +165,8 @@ type ComponentStatusDetails struct {
 	Version string `json:"version,omitempty"`
 	// The generation of the last VZ resource the Component was reconciled against
```
michael-cico, line 166: Probably this should say 

```
The generation of the last VZ resource the Component was successfully reconciled against
```

i.e., add `successfully`

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -689,7 +689,16 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	}
 	if conditionType == installv1alpha1.CondInstallComplete {
 		cr.Status.VerrazzanoInstance = vzinstance.GetInstanceInfo(compContext)
-		componentStatus.LastReconciledGeneration = cr.Generation
+		if componentStatus.ReconcilingGeneration > 0 {
+			componentStatus.LastReconciledGeneration = componentStatus.ReconcilingGeneration
+			componentStatus.ReconcilingGeneration = 0
```
michael-cico, line 694: OK, this answers my previous question... so nm on that 😄

platform-operator/controllers/verrazzano/install.go
```diff
@@ -143,7 +148,17 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
 	return ctrl.Result{}, nil
 }
 
-func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.ComponentStatusDetails) bool {
+func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.ComponentStatusDetails, name string) bool {
+	vzState := ctx.ActualCR().Status.State
+	if vzState != vzapi.VzStateInstalling && vzState != vzapi.VzStateReady {
+		return false
+	}
+	if name == mysql.ComponentName { // CR.mysql is not allowed yet
```
michael-cico, line 156: maybe a brief explanation here why MySQL isn't allowed yet

platform-operator/controllers/verrazzano/install.go
```diff
@@ -143,7 +148,17 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
 	return ctrl.Result{}, nil
 }
 
-func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.ComponentStatusDetails) bool {
+func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.ComponentStatusDetails, name string) bool {
```
michael-cico, line 151: This check has gotten a bit more complicated, might be good to add a comment or 2 as to what it's doing.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -143,7 +148,17 @@ func (r *Reconciler) reconcileComponents(vzctx vzcontext.VerrazzanoContext) (ctr
...
+	vzState := ctx.ActualCR().Status.State
+	if vzState != vzapi.VzStateInstalling && vzState != vzapi.VzStateReady {
```
michael-cico, line 153: Why do we need this check?

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -130,7 +132,8 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 			for _, comp := range registry.GetComponents() {
 				compName := comp.Name()
 				componentStatus := cr.Status.Components[compName]
-				if componentStatus != nil {
+				effectiveCR, _ := transform.GetEffectiveCR(cr)
```
michael-cico, line 135: I think it would be cheaper here to move the effective CR computation outside the loop.  Otherwise you'll be doing the same merge 15 times.  We only need to compute it once.

i.e.

```
			effectiveCR, _ := transform.GetEffectiveCR(cr)
                         // maybe requeue on error?
                            ...
                            if componentStatus != nil && (effectiveCR != nil && comp.IsEnabled(effectiveCR)) {
                               ...
```

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
shihchang: fixed the comment.
When transit to Ready, LastReconciledGeneration is set to ReconcilingGeneration and ReconcilingGeneration is set to 0 to indicate the component is not being reconciled

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
shihchang: fixed

platform-operator/controllers/verrazzano/install.go
shihchang: added a comment - Current VerrazzanoSpec or CRD does not define VerrazzanoSpec.components.mysql. MySQL Config update is not allowed yet.

platform-operator/controllers/verrazzano/install.go
shihchang: added

platform-operator/controllers/verrazzano/install.go
shihchang: to avoid interrupting upgrade flow. changed to 
```
	if vzState == vzapi.VzStateUpgrading || vzState == vzapi.VzStatePaused {
		return false
	}
```

platform-operator/controllers/verrazzano/upgrade.go
shihchang: fixed

michael-cico (approved):

PR #2868, opened 2022-04-06
VZ-4588: make root ca name prettier

wmhopkins:
# Description

Tweak the method used to generate random suffix for name, to use only lower-case letters, rather than full alpha-numeric with upper and lower case. This doesn't need to be cryptographically secure or unguessable, so making it prettier at the expense of randomness seems OK.

Fixes VZ-4588

# Checklist 

### DIFF
--- a/pkg/security/password/password.go
+++ b/pkg/security/password/password.go
@@ -14,8 +14,8 @@ import (
 const mask = "******"
 
 const (
-	modeAlphaNum   = iota
-	modeAlphaLower = iota
+	modeAlphaNum   = 0
+	modeAlphaLower = 1
 )
 
 //GeneratePassword will generate a password of length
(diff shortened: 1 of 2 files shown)


guoyongzhang (approved):

pkg/security/password/password.go
```diff
@@ -42,6 +61,16 @@ func makeAlphaNumeric(input string) (string, error) {
 	return reg.ReplaceAllString(input, ""), nil
 }
 
+// makeAlphaLower removes all non-lower-case-alpha characters
```
michael-cico, line 64: Nit, comment should probably say "Removes all special chars and lower-cases the result" right?

michael-cico (approved):

PR #2919, opened 2022-04-13
VZ-5494. Skip cert renewal checks on initial install

michael-cico:
# Description

Skip the new cert renewal checks on initial install in the cert-manager component code.  Fixes an edge condition where the certs can take some time to be issued, like when using Let's Encrypt as the cert signer.

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
@@ -161,8 +161,9 @@ func (c certManagerComponent) createOrUpdateClusterIssuer(compContext spi.Compon
 		}
 	}
 	if opResult == controllerutil.OperationResultCreated {
-		// We're in the initial install phase, and created the ClusterIssuer for the first time
-		compContext.Log().Oncef("Initial install, skipping certficate renewal checks")
+		// We're in the initial install phase, and created the ClusterIssuer for the first time,
+		// so skip the renewal checks
+		compContext.Log().Oncef("Initial install, skipping certificate renewal checks")
 		return nil
 	}
 	if err := checkRenewAllCertificates(compContext, isCAValue); err != nil {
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
```diff
@@ -148,18 +148,24 @@ func (c certManagerComponent) createOrUpdateClusterIssuer(compContext spi.Compon
...
 			return compContext.Log().ErrorfNewErr("Failed creating Acme resources: %v", err)
 		}
 	} else {
 		// Create resources needed for CA certificates
-		if _, err = createOrUpdateCAResources(compContext); err != nil {
+		if opResult, err = createOrUpdateCAResources(compContext); err != nil {
 			return compContext.Log().ErrorfNewErr("Failed creating CA resources: %v", err)
 		}
 	}
+	if opResult == controllerutil.OperationResultCreated {
```
pfmackin, line 163: what if use switched from ACME to CA?  Would this be an OperationResultCreated?

wmhopkins (approved):
lgtm.

platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
michael-cico: no, it's an update operation when we switch.  The ClusterIssuer gets updated in place, and OperationResultUpdated is returned.  That was actually the original trigger I had to renew certificates

I'm doing some verification of updates post-install again, and seems to be working OK.  I installed with le-staging, then updated to the default self-signed issuer.  Seems to be holding up.  The LE staging pipeline passed also.

I'm wondering if this might be better as a controller watching for ClusterIssuer updates, or perhaps we can consolidate these kinds of checks/actions into the secrets reconciler he's working on.

PR #2921, opened 2022-04-13
VZ-5224 Verify the Prometheus Operator default image arguments

davisgomes:
# Description

This introduces a test to verify that the Prometheus Operator default images are populated properly.

Fixes VZ-5224

# Checklist 

### DIFF
--- a/tests/e2e/pkg/common.go
+++ b/tests/e2e/pkg/common.go
@@ -329,27 +329,22 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
 	return reflect.DeepEqual(m1, m2)
 }
 
-// SlicesContainSubsectionStrings compares two slices and returns true if the first slice is a subsection of the second slice
-func SlicesContainSubsectionStrings(strings1, strings2 []string) bool {
+// SlicesContainSubsetSubstring returns true if the strings in the first slice are substrings of any string in the second slice
+func SlicesContainSubsetSubstring(strings1, strings2 []string) bool {
 	if len(strings1) >= len(strings2) {
 		return false
 	}
 	if len(strings1) == 0 {
 		return true
 	}
-	// count how many times each string occurs in case there are duplicates
-	m1 := map[string]int32{}
-	for _, s := range strings1 {
-		m1[s]++
-	}
-	m2 := map[string]int32{}
-	for _, s := range strings2 {
-		m2[s]++
-	}
-
-	// See if all keys value pairs in map1 match the pairs in map2
-	for k, v := range m1 {
-		if m2[k] != v {
+	for _, s1 := range strings1 {
+		found := false
+		for _, s2 := range strings1 {
+			if strings.Contains(s1, s2) {
+				found = true
+			}
+		}
+		if !found {
 			return false
 		}
 	}
--- a/tests/e2e/pkg/common.go
+++ b/tests/e2e/pkg/common.go
@@ -337,7 +337,7 @@ func SlicesContainSubsetSubstring(strings1, strings2 []string) bool {
 	for _, s1 := range strings1 {
 		found := false
 		for _, s2 := range strings2 {
-			if strings.Contains(s1, s2) {
+			if strings.Contains(s2, s1) {
 				found = true
 				break
 			}
--- a/tests/e2e/pkg/common.go
+++ b/tests/e2e/pkg/common.go
@@ -329,6 +329,33 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
 	return reflect.DeepEqual(m1, m2)
 }
 
+// SlicesContainSubsectionStrings compares two slices and returns true if the first slice is a subsection of the second slice
+func SlicesContainSubsectionStrings(strings1, strings2 []string) bool {
+	if len(strings1) >= len(strings2) {
+		return false
+	}
+	if len(strings1) == 0 {
+		return true
+	}
+	// count how many times each string occurs in case there are duplicates
+	m1 := map[string]int32{}
+	for _, s := range strings1 {
+		m1[s]++
+	}
+	m2 := map[string]int32{}
+	for _, s := range strings2 {
+		m2[s]++
+	}
+
+	// See if all keys value pairs in map1 match the pairs in map2
+	for k, v := range m1 {
+		if m2[k] != v {
+			return false
+		}
+	}
+	return true
+}
+
 // PolicyRulesEqual compares two RBAC PolicyRules for semantic equality
 func PolicyRulesEqual(rule1, rule2 rbacv1.PolicyRule) bool {
 	if SlicesContainSameStrings(rule1.Verbs, rule2.Verbs) &&
--- a/tests/e2e/pkg/common.go
+++ b/tests/e2e/pkg/common.go
@@ -339,6 +339,7 @@ func SlicesContainSubsetSubstring(strings1, strings2 []string) bool {
 		for _, s2 := range strings2 {
 			if strings.Contains(s1, s2) {
 				found = true
+				break
 			}
 		}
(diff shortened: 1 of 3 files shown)


tests/e2e/verify-install/promstack/promstack_test.go
```diff
@@ -33,6 +35,10 @@ var (
 		"servicemonitors.monitoring.coreos.com",
 		"thanosrulers.monitoring.coreos.com",
 	}
+	expectedPromOperatorArgs = []string{
+		"--prometheus-default-base-image=ghcr.io/verrazzano/prometheus:v2.34.0",
+		"--alertmanager-default-base-image=ghcr.io/verrazzano/alertmanager:v0.24.0",
```
fredtibbitts, line 40: I'm wondering if we should remove the tags? Maybe it's sufficient to just check that the base images are from ghcr.io/verrazzano? That way we don't have to update this code whenever we bump the versions. Alternatively we could pull the image and tag from the BOM.

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,33 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
 	return reflect.DeepEqual(m1, m2)
 }
 
+// SlicesContainSubsectionStrings compares two slices and returns true if the first slice is a subsection of the second slice
+func SlicesContainSubsectionStrings(strings1, strings2 []string) bool {
```
desagar, line 333: I'd suggest renaming to `sliceContainsSubset` since what you are doing is checking that one slice is a subset of another - the word "subsection" doesn't really clarify it.

The comment above the function should be modified to say "first slice is a subset of the second slice, and the strings in the first slice occur the same number of times in the second slice".

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,33 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
...
+	m1 := map[string]int32{}
+	for _, s := range strings1 {
+		m1[s]++
+	}
+	m2 := map[string]int32{}
+	for _, s := range strings2 {
+		m2[s]++
+	}
+
+	// See if all keys value pairs in map1 match the pairs in map2
```
desagar, line 350: this comment would be more useful if it stated the overall purpose rather than talk about the maps - something like "See if the strings in the first slice have the same number of occurrences in the second slice"

tests/e2e/pkg/common.go
davisgomes: Yes, that's the word I was looking for.

tests/e2e/verify-install/promstack/promstack_test.go
```diff
@@ -83,6 +89,26 @@ var _ = t.Describe("Prometheus Stack", Label("f:platform-lcm.install"), func() {
...
+		// GIVEN the Prometheus stack is installed
+		// WHEN we check to make sure the default images are from Verrazzano
+		// THEN we see that the arguments are correctly populated
+		WhenPromStackInstalledIt("should have the correct default images", func() {
+			promStackPodsRunning := func() bool {
+				deployment, err := pkg.GetDeployment(verrazzanoMonitoringNamespace, prometheusOperatorDeployment)
+				if err != nil {
+					AbortSuite(fmt.Sprintf("Deployment %v is not found in the namespace: %v, error: %v", prometheusOperatorDeployment, verrazzanoMonitoringNamespace, err))
+				}
+				for _, container := range deployment.Spec.Template.Spec.Containers {
```
desagar, line 101: this might be useful to be refactored into a separate pkg function e.g. something like: containerHasExpectedArgs(deployment, containerName, expectedArgs)

But that can wait for the moment if you'd like to wrap this up quickly.

davisgomes: I can add that to the Kubernetes pkg file.

tests/e2e/verify-install/promstack/promstack_test.go
davisgomes: I thought about this, but there doesn't seem to be a ton of infrastructure in the bom. I'll look into seeing if I can extract it out.

davisgomes: And I'd have to make a more sophisticated internal function if I wanted to check substrings, which I can do if necessary.

fredtibbitts: There might be some utility functions for other tests that pull things out of the BOM. I wouldn't spend too much time on this, but I always find it annoying when I make a simple upgrade of an image change and then have to update tests to match. I was just trying to avoid that but if it's a lot of work then we can deal with it later.

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,28 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
 	return reflect.DeepEqual(m1, m2)
 }
 
+// SlicesContainSubsetSubstring returns true if the strings in the first slice are substrings of any string in the second slice
+func SlicesContainSubsetSubstring(strings1, strings2 []string) bool {
+	if len(strings1) >= len(strings2) {
```
desagar, line 334: just noticed this - wouldn't equal lengths be okay?

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,28 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
...
+		return false
+	}
+	if len(strings1) == 0 {
+		return true
+	}
+	for _, s1 := range strings1 {
+		found := false
+		for _, s2 := range strings1 {
+			if strings.Contains(s1, s2) {
+				found = true
```
desagar, line 344: you can break as soon as found = true

tests/e2e/pkg/common.go
davisgomes: Yes updating now

davisgomes: In fact this check is unnecessary anyways since we're looping and duplicates are okay

tests/e2e/pkg/common.go
davisgomes: You're right. I was debating on whether to add this but I don't see why not.

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,28 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
...
+	if len(strings1) >= len(strings2) {
+		return false
+	}
+	if len(strings1) == 0 {
+		return true
+	}
+	for _, s1 := range strings1 {
+		found := false
+		for _, s2 := range strings1 {
+			if strings.Contains(s1, s2) {
```
desagar, line 340: s1 has to be a substring of s2, correct? If so they have to be reversed. 

strings.Contains is I think first argument contains second argument. https://pkg.go.dev/strings#Contains

desagar: there's also a HasPrefix (if you want s2 to start with what's in s1)

davisgomes: Yes good catch

desagar (approved):

fredtibbitts (approved):

tests/e2e/pkg/common.go
```diff
@@ -329,6 +329,26 @@ func SlicesContainSameStrings(strings1, strings2 []string) bool {
 	return reflect.DeepEqual(m1, m2)
 }
 
+// SlicesContainSubsetSubstring returns true if the strings in the first slice are substrings of any string in the second slice
```
pfmackin, line 332: this should be in verrazzano/pkg/string/slice.go

stevshap (approved):

stevshap (approved):

PR #2926, opened 2022-04-14
VZ 5650: OpenSearch Node Groups

anders-swanson:
Adds OpenSearch node groups to the API.


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -389,6 +389,20 @@ type ElasticsearchComponent struct {
...
+type OpenSearchNode struct {
+	Name      string                       `json:"name,omitempty"`
+	Replicas  int32                        `json:"replicas,omitempty"`
+	Roles     []vmov1.NodeRole             `json:"roles,omitempty"`
+	Storage   *OpenSearchNodeStorage       `json:"storage,omitempty"`
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+}
+
+type OpenSearchNodeStorage struct {
+	Size string `json:"size"`
```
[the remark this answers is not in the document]
michael-cico: ah, wait, I thought we did support that in the VMI CRD, but maybe not?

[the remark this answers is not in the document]
anders-swanson: I created this as its own struct so we could add in storage class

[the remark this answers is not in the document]
michael-cico: yeah, I think that's a separate discussion/PR then.

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -389,6 +389,20 @@ type ElasticsearchComponent struct {
 	// +patchStrategy=merge,retainKeys
 	ESInstallArgs []InstallArgs                 `json:"installArgs,omitempty" patchStrategy:"merge,retainKeys" patchMergeKey:"name"`
 	Policies      []vmov1.IndexManagementPolicy `json:"policies,omitempty"`
+	Nodes         []OpenSearchNode              `json:"nodes,omitempty"`
+}
+
+//OpenSearchNode specifies a node group in the OpenSearch cluster
+type OpenSearchNode struct {
```
kminder, line 396: From an API consistency perspective should things like `replicas` and `resources` be handled similar to how the 'kubernetes' field is being added to the API of other components?

anders-swanson: Resources is a *corev1.ResourceRequirements, so that's already a drop in from the k8s corev1 API. Replicas I wasn't sure on. Node.replicas felt good enough, since the 'common kubernetes' struct had some other stuff that i didn't want (affinity is handled internally by the vmo coordinator).

michael-cico: Replicas in the Istio case (in our CRD) is just `uint32`.  I think it's fine as-is, unless you wanted to align on the same type for consistency.  Not a blocker for this of course.

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
anders-swanson: This was intentional since it is the same type as the vmi value.

stevshap (approved):

kminder (approved):
Approved.

PR #2933, opened 2022-04-15
VZ-5667 config hash to optimize update

shihchang:
# Description

For each component in effectiveCR, compare the hash sum of the effective-config to avoid unnecessary re-install

Fixes VZ-5667

# Checklist 


platform-operator/controllers/verrazzano/component/mysql/mysql_component.go
```diff
@@ -118,3 +117,7 @@ func (c mysqlComponent) getInstallArgs(vz *vzapi.Verrazzano) []vzapi.InstallArgs
 	}
 	return nil
 }
+
+func (c mysqlComponent) GetConfigHash(_ spi.ComponentContext) string {
```
michael-cico, line 121: Is this to prevent updates?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -314,3 +314,20 @@ func (c verrazzanoComponent) GetCertificateNames(ctx spi.ComponentContext) []typ
 
 	return certificateNames
 }
+
+func (c verrazzanoComponent) GetConfigHash(ctx spi.ComponentContext) string {
+	return vzConfigHash(ctx)
+}
+
+func vzConfigHash(ctx spi.ComponentContext) string {
```
michael-cico, line 322: We may need to do something like this for Keycloak.  the reason being, if the DNS or certificate configurations change, we will need Keycloak to run and update the PKCE client URLs and update them in the realm configuration again.  So Keycloak's "hash" should probably be tied to the hash of the `dns` and `certificate` components as well.

If you change the DNS settings without doing that, my guess is that the auth redirects will break.

platform-operator/controllers/verrazzano/component/mysql/mysql_component.go
shihchang: yes. but there is already a condition for mysql in install.checkConfigUpdated, i will remove this.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
shihchang: add dns and cert-manager to keycloakConfigHash

PR #2956, opened 2022-04-19
VZ-5683-OS-component

mbenmclean:
This change moves Opensearch and Opensearch Dashboard out of the verrazzano component and into its own top level component. It is not a helm component and instead simply calls a createVMI function that creates or updates the OS and OSD components during install or upgrade.
This change likewise removes OS and OSD from the verrazzano createVMI function. It updates the unit tests in each directory according.
It also adds a change to the the createVMI logic by overwriting changes to the master node storage only when the master node replicas are not zero.

# Checklist 


platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -0,0 +1,358 @@
...
+					break
+				}
+			}
+		}
+	}
+
+	return isVerrazzanoSecretReady(ctx)
+}
+
+func findStorageOverride(effectiveCR *vzapi.Verrazzano) (*verrazzano.ResourceRequestValues, error) {
```
anders-swanson, line 197: func is duplicated across packages?

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
```diff
@@ -0,0 +1,345 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	system = "system"
+)
+
+//createVMI instantiates the VMI resource
+func createVMI(ctx spi.ComponentContext) error {
+	if !vzconfig.IsVMOEnabled(ctx.EffectiveCR()) {
```
anders-swanson, line 31: a lot of this func is duplicated across the components

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
```diff
@@ -0,0 +1,345 @@
...
+		vmi.Spec.Kibana = newOpenSearchDashboards(cr)
+		return nil
+	})
+	if err != nil {
+		return ctx.Log().ErrorfNewErr("failed to update VMI: %v", err)
+	}
+	return nil
+}
+
+func newVMI() *vmov1.VerrazzanoMonitoringInstance {
```
anders-swanson, line 81: duplicated function

anders-swanson (changes requested):
Looks like it will work, there's just a ton of duplicated code. Please clean up the duplicated code

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	if !reflect.DeepEqual(oldSetting, newSetting) {
+		return fmt.Errorf("Can not change volume settings for %s", ComponentJSONName)
+	}
+	return nil
+}
+
+func (o opensearchComponent) checkEnabled(old *vzapi.Verrazzano, new *vzapi.Verrazzano) error {
+	// Do not allow disabling of any component post-install for now
+	if vzconfig.IsElasticsearchEnabled(old) && !vzconfig.IsElasticsearchEnabled(new) {
```
anders-swanson, line 214: consider renaming these functions to IsOpenSearchEnabled

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
mbenmclean: Yeah I'm going to fix

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -0,0 +1,358 @@
...
+	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/namespace"
+	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/status"
+	"github.com/verrazzano/verrazzano/platform-operator/internal/vzconfig"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ComponentName is the name of the component
```
kminder, line 29: Orphaned comment?

kminder:
Can you confirm that all of these new files are largely code extracted from other existing files?  Basically meaning that the code is already proven and just be relocated?

mbenmclean:
Yes there's were all taken originally from the verrazzano component. I've made changes since then though mainly deleting stuff or pulling new stuff in

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -381,7 +381,8 @@ type MonitoringComponent struct {
 
 // ElasticsearchComponent specifies the Elasticsearch configuration.
 type ElasticsearchComponent struct {
-	MonitoringComponent `json:",inline"`
+	// +optional
```
bobdonat, line 384: Why is this change needed and only to the ElasticsearchComponent

bobdonat: Without this change there is no need for a new version of zz-generated.deepcopy.go

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
mbenmclean: fixed

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
mbenmclean: I think we decided we were going to remove the monitoring component all together.

bobdonat: We decided to leave as and address as a follow up task

platform-operator/constants/constants.go
```diff
@@ -145,3 +145,18 @@ const InitializeOperation = "initialize"
...
 const ReconcileLoopRequeueInterval = 3 * time.Minute
+
+// VMISecret is the secret used for VMI
+const VMISecret = "verrazzano"
+
+// VMISecret is the secret used for VMI
+const GrafanaSecret = "grafana-admin"
+
+// VMIBackupScrtName is the backup VMI secret
+const VMIBackupScrtName = "verrazzano-backup"
```
bobdonat, line 156: for consistency should spell out Secret instead of Scrt

platform-operator/constants/constants.go
```diff
@@ -145,3 +145,18 @@ const InitializeOperation = "initialize"
...
+
+// ObjectstoreAccessKey is used for the VMI backup secret
+const ObjectstoreAccessKey = "object_store_access_key"
+
+// ObjectstoreAccessSecretKey is used for the VMI backup secret
+const ObjectstoreAccessSecretKey = "object_store_secret_key"
```
bobdonat, line 162: Should these be ObjectStore... instead of Objectstore

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -0,0 +1,287 @@
...
+				}
+			}
+		}
+	}
+
+	return common.IsVMISecretReady(ctx), nil
+}
+
+// isOpenSearchReady VMI components ready-check
+func isOpenSearchReady(ctx spi.ComponentContext) bool {
```
bobdonat, line 110: This function looks identical to isOpenSearchInstalled

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,228 @@
...
+		return ctrlerrors.RetryableError{
+			Source:    ComponentName,
+			Operation: "Check if certificates are ready",
+		}
+	}
+	return nil
+}
+
+// PostUpgrade OpenSearch post-upgrade processing
+func (o opensearchComponent) PostUpgrade(ctx spi.ComponentContext) error {
```
bobdonat, line 130: This function is identical to previous function

platform-operator/controllers/verrazzano/component/verrazzano/vmi.go
```diff
@@ -44,11 +38,11 @@ func createVMI(ctx spi.ComponentContext) error {
 	if err := appendVerrazzanoValues(ctx, values); err != nil {
 		return ctx.Log().ErrorfNewErr("failed to get Verrazzano values: %v", err)
 	}
-	storage, err := findStorageOverride(ctx.EffectiveCR())
+	storage, err := common.FindStorageOverride(ctx.EffectiveCR())
```
anders-swanson, line 41: Since this createVMI function is duplicated across the three comopnents with minor changes, can you make a function like:
```
createVMI(ctx spi.ComponentContext, updateFunc func() error) error { ... }
```


The update func would do the body of the createOrUpdate that is different between the components.

platform-operator/constants/constants.go
mbenmclean: fixed

platform-operator/constants/constants.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
mbenmclean: They're not, they call a DoesExist function vs an IsReady function

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: Same as above

platform-operator/internal/k8s/status/deployment_ready.go
```diff
@@ -46,6 +46,22 @@ func DeploymentsAreReady(log vzlog.VerrazzanoLogger, client clipkg.Client, names
 	return true
 }
 
+// DoDeploymentsExist check that the named deployments have the minimum number of specified replicas ready and available
```
bobdonat, line 49: What is the purpose of this function?  The description is not accurate and the function above may do what is needed.

platform-operator/internal/k8s/status/statefulset_ready.go
```diff
@@ -47,6 +47,22 @@ func StatefulSetsAreReady(log vzlog.VerrazzanoLogger, client client.Client, name
 	return true
 }
 
+// DoStatefulSetsExist check that the named statefulsets have the minimum number of specified replicas ready and available
```
bobdonat, line 50: What is the purpose of this function? The description is not accurate and the function above may do what is needed.

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
anders-swanson: So the difference between these two function calls - could we refactor to pass these functions in as args?

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
anders-swanson: The common logic can be put in a function

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
bobdonat: but the rest of the logic is identical.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+// ComponentJSONName is the json name of the OpenSearch-Dashboards component in CRD
+const ComponentJSONName = "opensearch-dashboards"
+
+type opensearchDashboardsComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch-Dashbaords component
+func (d opensearchDashboardsComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName}
```
bobdonat, line 38: OSD needs to be dependent on VMO
the istio and nginx dependencies will need to be removed

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+// ComponentJSONName is the josn name of the opensearch component in CRD
+const ComponentJSONName = "opensearch"
+
+type opensearchComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch component
+func (o opensearchComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName}
```
bobdonat, line 38: OS needs to be dependent on VMO
the istio and nginx dependencies will need to be removed

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
mbenmclean: Let me double check

mbenmclean: Combined code

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: fixec

platform-operator/internal/k8s/status/deployment_ready.go
mbenmclean: It checks if the deployment exists. I fixed the comment

platform-operator/internal/k8s/status/statefulset_ready.go
mbenmclean: It checks if the deployment exists. I fixed the comment

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -0,0 +1,287 @@
...
+	esMasterStatefulset = "vmi-system-es-master"
+)
+
+var (
+	// For Unit test purposes
+	execCommand = exec.Command
+)
+
+// isOpenSearchInstalled checks if OpenSearch has been installed yet
+func isOpenSearchInstalled(ctx spi.ComponentContext) (bool, error) {
```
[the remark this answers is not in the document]
mbenmclean: This is normally implemented as part of the HelmComponent interface, but this is not a helm component. So yes we need it it is checked as part of the upgrade loop.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// IsReady component check
+func (o opensearchComponent) IsReady(ctx spi.ComponentContext) bool {
+	return isOpenSearchReady(ctx)
+}
+
+// PostInstall - post-install, clean up temp files
+func (o opensearchComponent) PostInstall(ctx spi.ComponentContext) error {
+	ctx.Log().Debugf("OpenSearch component post-upgrade")
```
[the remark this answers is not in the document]
anders-swanson: still is wrong debug string

platform-operator/controllers/verrazzano/component/opensearchdashboards/vmi.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common"
+
+	vmov1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+)
+
+var updateFunc common.VMIMutateFunc = func(ctx spi.ComponentContext, storage *common.ResourceRequestValues, vmi *vmov1.VerrazzanoMonitoringInstance, existingVMI *vmov1.VerrazzanoMonitoringInstance) error {
```
anders-swanson, line 14: Any reason why this is a var?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
bobdonat: This is half done this way.  If we are going to remove use of MonitoringComponent then we should remove it all.  This can be done in a follow-up PR.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,196 @@
...
+
+type opensearchComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch component
+func (o opensearchComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName, vmo.ComponentName}
+}
+
+// GetMinVerrazzanoVersion returns the minimum Verrazzano version required by the OpenSearch component
+func (o opensearchComponent) GetMinVerrazzanoVersion() string {
```
bobdonat, line 42: This minimum version can be removed

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+// ComponentJSONName is the josn name of the opensearch component in CRD
+const ComponentJSONName = "opensearch"
+
+type opensearchComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch component
+func (o opensearchComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName, vmo.ComponentName}
```
bobdonat, line 38: opensearch needs to work without istio and nginx dependencies

platform-operator/controllers/verrazzano/component/vmo/vmo.go
```diff
@@ -84,10 +85,10 @@ func ExportVMOHelmChart(ctx spi.ComponentContext) error {
 	return nil
 }
 
-// reassociateResources updates the resources to ensure they are managed by the VMO release/component.  The resource policy
+// ReassociateResources updates the resources to ensure they are managed by the VMO release/component.  The resource policy
 // annotation is removed to ensure that helm manages the lifecycle of the resources (the resource policy annotation is
 // added to ensure the resources are disassociated from the VZ chart which used to manage these resources)
-func reassociateResources(ctx spi.ComponentContext) error {
```
michael-cico, line 90: If this is shared, should it be in the common package?

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
bobdonat: followup task

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
bobdonat: followup task

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -0,0 +1,191 @@
...
+type opensearchDashboardsComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch-Dashbaords component
+func (d opensearchDashboardsComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName, vmo.ComponentName}
+}
+
+// GetMinVerrazzanoVersion returns the minimum Verrazzano version required by the OpenSearch-Dashboards component
+func (d opensearchDashboardsComponent) GetMinVerrazzanoVersion() string {
+	return constants.VerrazzanoVersion1_3_0
```
bobdonat, line 43: same comment as opensearch-component

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+// ComponentJSONName is the json name of the OpenSearch-Dashboards component in CRD
+const ComponentJSONName = "opensearch-dashboards"
+
+type opensearchDashboardsComponent struct{}
+
+// GetDependencies returns the dependencies of the OpenSearch-Dashbaords component
+func (d opensearchDashboardsComponent) GetDependencies() []string {
+	return []string{istio.ComponentName, nginx.ComponentName, vmo.ComponentName}
```
bobdonat, line 38: same comment as opensearch-component

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("Disabling component OpenSearch-Dashboards not allowed")
+	}
+	return nil
+}
+
+// GetIngressNames - gets the names of the ingresses associated with this component
+func (d opensearchDashboardsComponent) GetIngressNames(ctx spi.ComponentContext) []types.NamespacedName {
+	var ingressNames []types.NamespacedName
+
+	if vzconfig.IsKibanaEnabled(ctx.EffectiveCR()) {
```
michael-cico, line 169: This probably doesn't need to be calculated, we did that before when it was part of the VZ component because the set of ingresses could change based on what was enabled.  You can just set the IngressNames field on the component struct most likely.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -0,0 +1,191 @@
...
+
+// PreInstall OpenSearch-Dashboards component pre-install processing; create and label required namespaces, copy any
+// required secrets
+func (d opensearchDashboardsComponent) PreInstall(ctx spi.ComponentContext) error {
+	// create or update  VMI secret
+	if err := common.EnsureVMISecret(ctx.Client()); err != nil {
+		return err
+	}
+	// create or update backup VMI secret
+	if err := common.EnsureBackupSecret(ctx.Client()); err != nil {
```
michael-cico, line 79: Should this just be set up by the VMO?  Or do we want to do it only when a VMI component is enabled?

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -0,0 +1,196 @@
...
+		return fmt.Errorf("Disabling component OpenSearch not allowed")
+	}
+	return nil
+}
+
+// GetIngressNames - gets the names of the ingresses associated with this component
+func (o opensearchComponent) GetIngressNames(ctx spi.ComponentContext) []types.NamespacedName {
+	var ingressNames []types.NamespacedName
+
+	if vzconfig.IsElasticsearchEnabled(ctx.EffectiveCR()) {
```
michael-cico, line 174: same comment here as with the dashboards component.

platform-operator/controllers/verrazzano/component/vmo/vmo_component.go
```diff
@@ -64,12 +66,26 @@ func (c vmoComponent) IsReady(context spi.ComponentContext) bool {
...
 }
 
 // PreUpgrade VMO pre-upgrade processing
 func (c vmoComponent) PreUpgrade(context spi.ComponentContext) error {
 	return common.ApplyCRDYaml(context, config.GetHelmVmoChartsDir())
 }
+
+// Upgrade VMO processing
+func (c vmoComponent) Upgrade(context spi.ComponentContext) error {
+	return c.HelmComponent.Install(context)
```
michael-cico, line 90: why call install here?  I made some changes recently where HelmComponent.Upgrade should be equivalent, plus the call to get the current values if they exist.  If calling the current values returns an "release not found" error we should just not append those to the overrides.

The reason I think we should call Upgrade if possible is that if we don't align the install/upgrade paths before the next release, we'll want to call upgrade at that point.

If it helps, I'm OK with that being a separate PR.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -0,0 +1,253 @@
...
+	return &vmov1.VerrazzanoMonitoringInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      system,
+			Namespace: globalconst.VerrazzanoSystemNamespace,
+		},
+	}
+}
+
+// CreateVMI instantiates the VMI resource
+func CreateVMI(ctx spi.ComponentContext, updateFunc VMIMutateFunc) error {
```
michael-cico, line 53: Should probably call this `CreateOrUpdateVMI`

platform-operator/controllers/verrazzano/component/opensearchdashboards/vmi.go
mbenmclean: I thought that's how you do it to cast the type as VMIMutateFunc.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -0,0 +1,253 @@
...
+		vmi.Spec.CascadingDelete = true
+		return updateFunc(ctx, storage, vmi, existingVMI)
+	})
+	if err != nil {
+		return ctx.Log().ErrorfNewErr("failed to update VMI: %v", err)
+	}
+	return nil
+}
+
+func EnsureVMISecret(cli client.Client) error {
```
michael-cico, line 95: Nit, probably want to add some function comments for this and the others in this file

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: The component interface requires that each component defines that function. I suppose we could return empty string. Might be good practice to set though.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: Yeah that's part of another story. Out of scope for this change

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -0,0 +1,253 @@
...
+			return false
+		}
+		ctx.Log().Debugf("Verrazzano secret not found")
+		return false
+	}
+	return true
+}
+
+// CreateAndLabelVMINamespaces creates and labels the namespaces needed for the VMI resources
+func CreateAndLabelVMINamespaces(ctx spi.ComponentContext) error {
```
michael-cico, line 209: What's getting labeled?  Or is that happening in the CreateVerrazzanoSystemNamespace func?

platform-operator/controllers/verrazzano/component/vmo/vmo.go
mbenmclean: Yeah this was a quick fix while we were trying to get it working. It requires reworking some unit tests to move to common. I'll probably do as a second pass if that's cool.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: Same response

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
bobdonat: I don't think that is true.  Most components do not specify the min version

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: Good point, I'll change.

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
```diff
@@ -0,0 +1,215 @@
...
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	system = "system"
+)
+
+var updateFunc common.VMIMutateFunc = func(ctx spi.ComponentContext, storage *common.ResourceRequestValues, vmi *vmov1.VerrazzanoMonitoringInstance, existingVMI *vmov1.VerrazzanoMonitoringInstance) error {
```
michael-cico, line 20: Probably in line with Anders' comment elsewhere, this probably doesn't need to be a var, it can be passed in as a callback as long as the signature matches the formal parameter signature.

Unless it's a var so it can be overridden during unit testing.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -74,7 +67,23 @@ func NewComponent() spi.Component {
 // PreInstall Verrazzano component pre-install processing; create and label required namespaces, copy any
 // required secrets
 func (c verrazzanoComponent) PreInstall(ctx spi.ComponentContext) error {
-	if err := setupSharedVMIResources(ctx); err != nil {
+	if vzconfig.IsVMOEnabled(ctx.EffectiveCR()) {
+		// Make sure the VMI CRD is installed since the Verrazzano component may create/update
+		// a VMI CR
+		if err := common.ApplyCRDYaml(ctx, config.GetHelmVmoChartsDir()); err != nil {
```
michael-cico, line 73: should this be happening in the VMO component?

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: We only want to do this when there are VMI components that use it I assume. Maybe it should be in the VMO but currently the VPO still does a lot of work setting up the monitoring stack components like this.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/common/vmi.go
mbenmclean: Ok i agree

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
mbenmclean: So just define in place when I call CreateVMI?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
mbenmclean: No because the Verrazzano component has to upgrade first because it needs to apply its network policies, but it also needs the VMI CRD which is in the VMO component. It's kind of hack but it's what we're going with for now.

anders-swanson (dismissed):

platform-operator/controllers/verrazzano/component/opensearch/vmi.go
michael-cico: I think you can define it as a named func, and just pass in the name of the func where needed as an argument.  If the signatures don't match then the compiler will complain.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
michael-cico: no, this makes sense I think.

platform-operator/controllers/verrazzano/component/vmo/vmo.go
michael-cico: Yup, no worries

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
michael-cico: Probably need to move those policies to the VMO chart/comp.  But not for this change.

bobdonat (approved):

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
mbenmclean: Yes because they're helm components so the helm component interface does that for them

PR #2957, opened 2022-04-19
VZ 5704: Acceptance test for node groups

anders-swanson:
Adds simple acceptance test for node groups


tests/e2e/opensearch/topology/topology_suite_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"testing"
+)
+
+func TestTopology(t *testing.T) {
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "VMI Suite")
```
kminder, line 14: The text probably needs to be udpated.

kminder (approved):
I'm not thrilled that these tests were added to the existing Kind tests.  Having said that there is not current viable alternative so this work should not be blocked.
Approved.

PR #2995, opened 2022-04-25
VZ-5760.  Allow updates during upgrade

michael-cico:
# Description

Update the validation around upgrade to
- Allow component updates during upgrade
- Require that first CR update after a VPO update includes an upgrade request

Fixes VZ-5760.

# Checklist 

### DIFF
--- a/platform-operator/apis/verrazzano/v1alpha1/validate.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/validate.go
@@ -6,11 +6,9 @@ package v1alpha1
 import (
 	"context"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"io/fs"
 	"os"
-	"reflect"
 	"strings"
 
 	vzos "github.com/verrazzano/verrazzano/pkg/os"
@@ -131,22 +129,16 @@ func ValidateProfile(requestedProfile ProfileType) error {
 	return nil
 }
 
-// ValidateUpgradeRequest Ensures that for the upgrade case only the version field has changed
+// ValidateUpgradeRequest Ensures hat an upgrade is requested as part of an update if necessary,
+// and that the version of an upgrade request is valid.
+// - an upgrade must be performed with or before the next update to the configuration after the VPO has been updated
 func ValidateUpgradeRequest(current *Verrazzano, new *Verrazzano) error {
-	installedVersion := current.Status.Version
 	currentSpec := current.Spec
 	newSpec := new.Spec
 
 	if !config.Get().VersionCheckEnabled {
 		zap.S().Infof("Version validation disabled")
 		return nil
-	}
-
-	// if the installed version is not == BOM version and newspec versiom isn't set,
-	// reject the update; upgrade needs to happen before any update
-	//
-	if len(newSpec.Version) == 0 {
-		semver.NewSemVersion(installedVersion)
 	}
 
 	// Short-circuit if the version strings are the same
@@ -157,18 +149,41 @@ func ValidateUpgradeRequest(current *Verrazzano, new *Verrazzano) error {
 		// if we get here, the current version is not empty, but the new version is
 		return fmt.Errorf("Requested version is not specified")
 	}
+
+	// if the installed version is not == BOM version and newspec versiom isn't set,
+	// reject the update; upgrade needs to happen before any update
+	bomVersion, err := GetCurrentBomVersion()
+	if err != nil {
+		return err
+	}
+	installedVersion, err := semver.NewSemVersion(current.Status.Version)
+	if err != nil {
+		return err
+	}
+	if len(newSpec.Version) == 0 && !bomVersion.IsEqualTo(installedVersion) {
+		return fmt.Errorf("Upgrade required for update, set version field to %v to upgrade", bomVersion.ToString())
+	}
+
 	if err := ValidateVersion(newSpec.Version); err != nil {
 		// new version is not nil, but we couldn't parse it
 		return err
 	}
 
+	if len(newSpec.Version) == 0 {
+		return nil
+	}
+
 	requestedSemVer, err := semver.NewSemVersion(newSpec.Version)
 	if err != nil {
-		// parse error on new version string
 		return err
 	}
 
-	// Verify that the new version request is > than the currently version
+	if !requestedSemVer.IsEqualTo(bomVersion) {
+		return fmt.Errorf("Requested version %s does not match BOM version %s",
+			requestedSemVer.ToString(), bomVersion.ToString())
+	}
+
+	// Verify that the new version request is > than the current version
 	if len(currentSpec.Version) > 0 {
 		currentSemVer, err := semver.NewSemVersion(currentSpec.Version)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,66 @@ func ValidateProfile(requestedProfile ProfileType) error {
...
 	if !config.Get().VersionCheckEnabled {
 		zap.S().Infof("Version validation disabled")
 		return nil
 	}
-	// Short-circuit if the version strings are the same
-	if currentSpec.Version == newSpec.Version {
+
+	installedVerString := current.Status.Version
+	if len(strings.TrimSpace(installedVerString)) == 0 {
+		// Boundary condition -- likely just created and install hasn't started yet
```
wmhopkins, line 143: Or the user edited the VZ while the install was going, but before the status was updated (not sure when that happens). Could the CR end up with invalid data relative to the install that was performed? I.e., such that the reconciler would get called again with a VZ CR that wasn't really validated, since validation is short-circuited here? Should we disallow edits for this case?

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,66 @@ func ValidateProfile(requestedProfile ProfileType) error {
...
+	if err != nil {
 		return err
 	}
+	if len(newSpec.Version) == 0 && bomVersion.IsGreatherThan(installedVersion) {
+		// Attempted an update before an upgrade has been done, reject the edit
+		return fmt.Errorf("Upgrade required for update, set version field to v%v to upgrade", bomVersion.ToString())
+	}
 
+	// New version is a non-zero-len string, short-circuit if the version strings are the same
+	if currentSpec.Version == newSpec.Version {
```
wmhopkins, line 166: If I'm understanding things, if the two versions are the same, then this is effectively an update. Just above, an empty version string indicates an update, and the update is refused if the current version is less than the BOM version. Seems we should do the same here, if we want to accept both methods of triggering an update.

TBH, it makes more sense to me to reject empty version strings altogether, so that the spec isn't left with an empty version. Seems like specifying the same version but changing other fields would trigger the reconciler to do an update without leaving the spec with an ampty version.

wmhopkins (commented):
mostly looks good, had a couple comments.

platform-operator/apis/verrazzano/v1alpha1/validate.go
michael-cico: If the version string the status field is null, then we never finished an initial install and there's no upgrade that can/needs to happen.  We set the status version field to the current BOM version once the install starts.

For some reason, we get called for an update right after an initial create.  I'm not sure exactly why that happens, which is why I needed to add that guard.

platform-operator/apis/verrazzano/v1alpha1/validate.go
michael-cico: We never set the version field in the spec; that has to be done by the user.  The version in the status section we set to track the installing/installed version.  

The reason for the version field in the spec is so the user has some trigger to start us doing an upgrade.  It's kinda clumsy but it's the only thing we could settle on.  But, we don't require them to set it initially.  And, they can only set one version, whatever is currently in the VPO BOM.

What the check before this part is doing (where the spec field is empty but the status version is < the BOM version) is for a situation where

- the user has updated the VPO (there's a new version available to upgrade to)
- The user has never upgraded (spec version is empty)
- the user makes non-upgrade edit (changes some other configuration)

I'll review the logic a little more and do some more testing.

At some point we need to consolidate the code paths and this won't be necessary; or we may want to keep it as our way of making sure the user is ready to upgrade, and has to do it explicitly (opt-in).  At least until we change how upgrade is done.

wmhopkins: Right, I wasn't suggesting we set the `spec.version`, and I think the intended logic makes sense, but IIUC, it would be possible for someone to trigger an update by keeping the `spec.version` the same as the current VZ while editing other fields. At least, it would pass validation, even if the upgrade hadn't happened yet, because you're just returning nil here, before any of the subsequent checks for `requested.version == bom.version`. Whether the reconciler would do anything with it is another question, I guess, but it would pass validation and get written to etcd, so the reconciler would have to make things right somehow, or leave things in an inconsistent state. I.e., it would have to process an update based on the changed VZ fields. But the version wouldn't have been validated, so the reconciler might be in the position of having to process the update even though there was a required upgrade that hadn't happened yet. Unless I'm missing something.

michael-cico: OK.  I re-worked the logic a bit, and that short-circuit no longer exists.  please re-review.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,77 @@ func ValidateProfile(requestedProfile ProfileType) error {
...
+		return fmt.Errorf("Upgrade required for update, set version field to v%v to upgrade", bomVersion.ToString())
+	}
+	return nil
+}
 
-	// Verify that the new version request is > than the currently version
-	if len(currentSpec.Version) > 0 {
-		currentSemVer, err := semver.NewSemVersion(currentSpec.Version)
+func validateNewVersion(current *Verrazzano, newVerString string, bomVersion *semver.SemVersion) error {
+	newSpecVer, err := semver.NewSemVersion(newVerString)
```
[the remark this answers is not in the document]
michael-cico: It's trimmed before the call into `validateNewVersion`

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,77 @@ func ValidateProfile(requestedProfile ProfileType) error {
...
+		return err
+	}
+	if !newSpecVer.IsEqualTo(bomVersion) {
+		// A newer version is available, the user must opt-in to an upgrade before we allow any edits
+		return fmt.Errorf("Requested version %s does not match BOM version v%s, please upgrade to the current BOM version",
+			newSpecVer.ToString(), bomVersion.ToString())
+	}
+	// Verify that the new version request is > than the current version
+	// - in reality, this should probably never happen
+	currentVerString := strings.TrimSpace(current.Spec.Version)
```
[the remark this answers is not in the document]
michael-cico: we do want to allow updates if the versions match; we're trying to prevent "accidental" or back-door upgrades via an update.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,92 @@ func ValidateProfile(requestedProfile ProfileType) error {
...
+	if len(currentVerString) > 0 {
+		currentSpecVer, err := semver.NewSemVersion(currentVerString)
 		if err != nil {
-			// Unable to parse the current spec version; this should never happen
 			return err
 		}
-		if requestedSemVer.IsLessThan(currentSemVer) {
-			return fmt.Errorf("Requested version %s is not newer than current version %s", requestedSemVer.ToString(), currentSemVer.ToString())
+		if newSpecVer != nil && newSpecVer.IsLessThan(currentSpecVer) {
```
wmhopkins, line 212: I think the underlying pathological case you're trying to detect here is the case that `len(currentVerString) > 0 && currentSpecVer != currentStatusVer`, which isn't precisely what you're checking, but this code is harmless either way and may detect some weird situations, so fine to leave as is.

platform-operator/apis/verrazzano/v1alpha1/validate.go
```diff
@@ -131,49 +129,92 @@ func ValidateProfile(requestedProfile ProfileType) error {
 	return nil
 }
 
-// ValidateUpgradeRequest Ensures that for the upgrade case only the version field has changed
-func ValidateUpgradeRequest(currentSpec *VerrazzanoSpec, newSpec *VerrazzanoSpec) error {
+// ValidateUpgradeRequest Ensures hat an upgrade is requested as part of an update if necessary,
```
pfmackin, line 132: typo hat

PR #3022, opened 2022-04-28
VZ-5680: Allow VMO to install without hard dependencies

bobdonat:
# Description

This pull request allows VMO to be installed if its dependency on nginx is not enabled.  This allows for only VMO/OS/OSD to be installed. 

Dependencies during install are now honored only if the dependency is enabled and the webhook validation to check if dependencies are enabled was removed. 

# Checklist 


anders-swanson (approved):

platform-operator/helm_config/charts/verrazzano-monitoring-operator/templates/configmap.yaml
```diff
@@ -13,4 +13,8 @@ data:
     metricsPort: {{ .Values.monitoringOperator.metricsPort }}
     defaultSimpleCompReplicas: {{ .Values.monitoringOperator.defaultSimpleCompReplicas }}
     defaultPrometheusReplicas: {{ .Values.monitoringOperator.defaultPrometheusReplicas }}
+    {{- if .Values.config.envName }}
+    {{- if .Values.config.dnsSuffix }}
     defaultIngressTargetDNSName: {{ .Values.config.envName }}-verrazzano-ingress.{{ .Values.config.dnsSuffix }}
```
anders-swanson, line 18: is this value used?

bobdonat: VMO uses it when
  vmo.Spec.IngressTargetDNSName == ""

bobdonat: maybe that code is not needed in VMO

michael-cico (approved):

michael-cico:
Will it work standalone without Keycloak?  What is used for auth?

bobdonat:
> Will it work standalone without Keycloak? What is used for auth?

As I understand the idea was that keycloak/auth proxy would not be setup and auth setup was an exercise for the user

mbenmclean:
So there will need to be a subsequent change to add the context to the GetDependencies function and add logic to create optional dependencies, right. Because this change on it's own won't allow the VMI components to get installed without NGINX. They will still try to wait for it.

bobdonat:
> So there will need to be a subsequent change to add the context to the GetDependencies function and add logic to create optional dependencies, right. Because this change on it's own won't allow the VMI components to get installed without NGINX. They will still try to wait for it.

With this change a component will not wait for a component if it is not enabled.  We will have to test the VMI components to see if there are any other changes needed.

PR #3029, opened 2022-04-29
VZ-4939 VAO cert per ingresstrait; move appconfig finalizer to ingresstrait 

guoyongzhang:
# Description

- move appconfig finalizer for deleting ingress cert/secret to IngressTrait
- cert/secret per IngressTrait, instead of per appconfig, while gateway remains per appconfig
- cert/secret name changes
- delete cert/secret with old names from old version of Verrazzano in IngressTrait reconciler
- update todo-list and its end-to-end test with new naming

# Checklist


jmaron99:
I don't think this is the answer.  This finalizer has gone back and forth between the app config and the ingress trait.  At this point the ingress trait is per application, not per ingress trait (you can check on that - look at the naming logic).  In that case the app ingress may get deleted if one of multiple ingress traits are deleted from the app, thus disabling all the other exposed ingress traits.

I think the solution here needs to be some sort of logic to prevent the additional reconciliation requests if the app config is in the process of deletion - perhaps check for a null app config or a deletion timestamp in the app config associated with the ingress.

guoyongzhang:
> 

Things may change since I am working on https://jira.oraclecorp.com/jira/browse/VZ-4939.  "the code that generates the secret name is per gateway, not per app".  If I did this JIRA, this PR would be ok, right?

jmaron99:
> > 
> 
> Things may change since I am working on https://jira.oraclecorp.com/jira/browse/VZ-4939. "the code that generates the secret name is per gateway, not per app". If I did this JIRA, this PR would be ok, right?

yes, it should since there will be a set of resources per ingress trait and the appropriate way to delete them would be a finalizer associated with the ingress trait

application-operator/controllers/appconfig/appconfig_controller.go
```diff
@@ -95,25 +96,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 // doReconcile performs the reconciliation operations for the application configuration
 func (r *Reconciler) doReconcile(ctx context.Context, appConfig *oamv1.ApplicationConfiguration, log vzlog2.VerrazzanoLogger) (ctrl.Result, error) {
-	// If the application configuration no longer exists or is being deleted then cleanup the associated cert and secret resources
+	// the logic to delete cert/secret is moved to the ingress trait finalizer
```
jmaron99, line 99: Wouldn't those older apps also require logic to remove their associated cert/secret resources?  Aren't they named differently in that instance?

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -217,9 +274,9 @@ func (r *Reconciler) coallateAllHostsForTrait(trait *vzapi.IngressTrait, status
 	return allHosts
 }
 
-// getGatewayName will generate a gateway name from the namespace and application name of the provided trait. Returns
+// buildGatewayName will generate a gateway name from the namespace and application name of the provided trait. Returns
 // an error if the app name is not available.
-func getGatewayName(trait *vzapi.IngressTrait) (string, error) {
+func buildGatewayName(trait *vzapi.IngressTrait) (string, error) {
```
jmaron99, line 279: So you're reusing the same gateway for all ingresses, correct?

jmaron99: looks like mutate gateway simply adds the server/ingress to the list of hosts

jmaron99:
Outside the scope of this PR, but I just noticed that there doesn't seem to be any cleanup of gateway/VZ resources when an app config is deleted?  Is that intentional @pfmackin ?

guoyongzhang:
> Do we know if OAM validates the uniqueness of traits in an app config? I assume yes...

If users specify trait names, they are responsible for making them unique; if they don't, we generate trait name like bobby-wls-trait-79b67d9d88.

guoyongzhang:
> Outside the scope of this PR, but I just noticed that there doesn't seem to be any cleanup of gateway/VZ resources when an app config is deleted? Is that intentional @pfmackin ?

Gateways are created with ownerReferences to ApplicationConfiguration.  They will be cascading deleted when appconfig is deleted.
VirtualService has ownerReferences to IngressTrait.

jmaron99 (approved):

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -163,6 +187,39 @@ func (r *Reconciler) doReconcile(ctx context.Context, trait *vzapi.IngressTrait,
 	return r.updateTraitStatus(ctx, trait, status)
 }
 
+// isIngressTraitBeingDeleted determines if the ingress trait is in the process of being deleted.
+// This is done checking for a non-nil deletion timestamp.
+func isIngressTraitBeingDeleted(trait *vzapi.IngressTrait) bool {
+	return trait != nil && trait.GetDeletionTimestamp() != nil
+}
+
+// removeFinalizerIfRequired removes the finalizer from the application configuration if required
```
pfmackin, line 196: comment is wrong, this is not app config

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -163,6 +187,39 @@ func (r *Reconciler) doReconcile(ctx context.Context, trait *vzapi.IngressTrait,
...
+		appName := vznav.GetNamespacedNameFromObjectMeta(trait.ObjectMeta)
+		log.Debugf("Removing finalizer from application configuration %s", appName)
+		trait.Finalizers = vzstring.RemoveStringFromSlice(trait.Finalizers, finalizerName)
+		err := r.Update(ctx, trait)
+		return vzlogInit.ConflictWithLog(fmt.Sprintf("Failed to remove finalizer from application configuration %s", appName), err, zap.S())
+	}
+	return nil
+}
+
+// addFinalizerIfRequired adds the finalizer to the app config if required
```
pfmackin, line 209: comment is wrong, this is not app config

application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -352,6 +431,8 @@ func (r *Reconciler) createOrUseGatewaySecret(ctx context.Context, trait *vzapi.
 	if trait.Spec.TLS != (vzapi.IngressSecurity{}) {
 		secretName = r.validateConfiguredSecret(trait, status)
 	} else {
+		cleanupCert(buildLegacyCertificateName(trait), r.Client, log)
```
pfmackin, line 434: both cleanup calls need to check for error and return if err so reconcile can rerun.  Error should be logged.

PR #3030, opened 2022-04-29
VZ 3337: Istio Tracing

anders-swanson:
If Jaeger is enabled/installed, Istio will detect the Jaeger installation and configure itself to send Envoy traces to Jaeger (all apps in the Istio mesh).
- Envoy does not hot reload configuration. This means the pods would need to be restarted (either by the user, or by verrazzano) to start exporting traces.
- Istio sampling rate is 1% by default. We likely need to follow up another change to allow this to be configurable.


platform-operator/constants/constants.go
```diff
@@ -166,3 +166,9 @@ const VerrazzanoIngressSecret = "verrazzano-tls" //nolint:gosec //#gosec G101
 
 // VerrazzanoLocalCABundleSecret is a secret containing the admin ca bundle
 const VerrazzanoLocalCABundleSecret = "verrazzano-local-ca-bundle" //nolint:gosec //#gosec G101
+
+//KubernetesAppLabel is a label key for kubernetes apps
+const KubernetesAppLabel = "app.kubernetes.io/component"
+
+//JaegerCollectorService is a label value for Jaeger collector
+const JaegerCollectorService = "service-collector"
```
kminder, line 174: Is this our label or a "standard" Jaeger label?  I'm trying to understand if it should be "name spaced".

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1213,19 +1193,42 @@ func (r *Reconciler) watchPods(namespace string, name string, log vzlog.Verrazza
...
 
-	// Watch pods and trigger reconciles for Verrazzano resources when a pod is created
-	err := r.Controller.Watch(
-		&source.Kind{Type: &corev1.Pod{}},
-		mapFn,
-		predicateFunc)
-	if err != nil {
-		return err
+func (r *Reconciler) watchJaegerService(namespace string, name string, log vzlog.VerrazzanoLogger) error {
+	// Watch services and trigger reconciles for Verrazzano resources when the jaeger service is created
```
kminder, line 1200: Probably need to capitalize Jaeger in the comment.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1269,6 +1272,11 @@ func (r *Reconciler) initForVzResource(vz *installv1alpha1.Verrazzano, log vzlog
 		return newRequeueWithDelay(), err
 	}
 
+	if err := r.watchJaegerService(vz.Namespace, vz.Name, log); err != nil {
+		log.Errorf("Failed to set Service watch for Verrazzano CR %s: %v", vz.Name, err)
```
kminder, line 1276: Should the error message be about Jaeger vs Verrazzano?

platform-operator/constants/constants.go
anders-swanson: It's a Jaeger label

kminder: ok

kminder (approved):
Approved

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1185,20 +1173,12 @@ func (r *Reconciler) watchJobs(namespace string, name string, log vzlog.Verrazza
...
-					Name:      name,
-				}},
-			}
-		})
-
-	// Watch pod create
-	predicateFunc := predicate.Funcs{
-		CreateFunc: func(e event.CreateEvent) bool {
+	// Watch pods and trigger reconciles for Verrazzano resources when a pod is created
+	log.Debugf("Watching for pods to activate reconcile for Verrazzano CR %s/%s", namespace, name)
```
pfmackin, line 1177: Why are we doing this?  Seems like a very broad solution to force a full reconcile of all components.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1213,19 +1193,42 @@ func (r *Reconciler) watchPods(namespace string, name string, log vzlog.Verrazza
...
+}
 
-	// Watch pods and trigger reconciles for Verrazzano resources when a pod is created
-	err := r.Controller.Watch(
-		&source.Kind{Type: &corev1.Pod{}},
-		mapFn,
-		predicateFunc)
-	if err != nil {
-		return err
+func (r *Reconciler) watchJaegerService(namespace string, name string, log vzlog.VerrazzanoLogger) error {
```
pfmackin, line 1199: This component specific watch logic seems like it doesn't belong in the overall VZ CR reconcile.  Short term this is probably ok, but per component reconciles would be a better fit.

PR #3045, opened 2022-05-02
OpenSearch network policy

anders-swanson:
Include node groups when detecting if cluster is multi-node for the creation of network policies.


bobdonat (approved):

bobdonat (dismissed):

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -268,6 +268,54 @@ func CheckIngressesAndCerts(ctx spi.ComponentContext, comp spi.Component) error
...
+
+//addNodeGroupReplicas iterates through each OpenSearch node and sums the replicas
+func addNodeGroupReplicas(os *vzapi.ElasticsearchComponent, replicas *int32) {
+	for _, node := range os.Nodes {
+		*replicas += node.Replicas
+	}
+}
+
+//addInstallArgReplicas sums the replicas from master, data, and ingest node install args.
+func addInstallArgReplicas(os *vzapi.ElasticsearchComponent, replicas *int32) error {
```
mbenmclean, line 295: This function feel kind of complicated. If you think this is the best way to implement I'll take your word for it.

platform-operator/controllers/verrazzano/component/common/vmi_test.go
```diff
@@ -171,6 +171,72 @@ func Test_FindStorageOverride(t *testing.T) {
...
+	}
+	oneReplicaVZ := mkVZ(true)
+	oneReplicaVZ.Spec.Components.Elasticsearch.ESInstallArgs = []vzapi.InstallArgs{
+		{
+			Name:  "nodes.master.replicas",
+			Value: "1",
+		},
+	}
+	multiNodeVZ := mkVZ(true)
+	multiNodeVZ.Spec.Components.Elasticsearch.Nodes = []vzapi.OpenSearchNode{
```
mbenmclean, line 194: Maybe add test case with some combinations of master, ingest, and data to make sure it's behaving as expected?

mbenmclean (approved):

platform-operator/controllers/verrazzano/component/common/vmi.go
anders-swanson: We need to loop through the install args and sum the replicas - since they all have string keys i used the switch. the closure func is there to convert the value and sum the replicas.

PR #3061, opened 2022-05-03
VZ 5801: Configure Istio Tracing

anders-swanson:
Configures tracing sampling rate, and whether Istio mesh tracing is enabled. By default, mesh tracing will be disabled.
Moved some code around from CLI args to be in the Istio operator YAML file. 

The user may enable mesh tracing and adjust the sampling rate via istio install args.


kminder (approved):
Approved.

PR #3065, opened 2022-05-04
WIP Support arbitrary helm values in Prometheus pushgateway component

anajosep:
# Description

Fixes VZ-9999

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway.go
+++ b/platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway.go
@@ -6,6 +6,10 @@ package pushgateway
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/pkg/bom"
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/internal/vzconfig"
+	"strconv"
 
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/k8s/status"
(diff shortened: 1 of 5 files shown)


platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway.go
```diff
@@ -47,3 +51,22 @@ func preInstall(ctx spi.ComponentContext) error {
...
+		Key:   "serviceMonitor.enabled",
+		Value: strconv.FormatBool(vzconfig.IsPrometheusEnabled(ctx.EffectiveCR())),
+	})
+
+	return kvs, nil
+}
+
+// GetHelmOverrides appends Helm value overrides for the Prometheus Operator Helm chart
+func GetHelmOverrides(ctx spi.ComponentContext) []vzapi.Overrides {
+	return ctx.EffectiveCR().Spec.Components.PrometheusPushgateway.ValueOverrides
```
davisgomes, line 71: Check to make sure the PrometheusPushgateway pointer is not nil before dereferencing.

platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway_component.go
```diff
@@ -42,6 +41,8 @@ func NewComponent() spi.Component {
 			ImagePullSecretKeyname:  secret.DefaultImagePullSecretKeyName,
 			ValuesFile:              filepath.Join(config.GetHelmOverridesDir(), "prometheus-pushgateway-values.yaml"),
 			Dependencies:            []string{},
+			AppendOverridesFunc:     AppendOverrides,
+			GetHelmValueOverrides:   GetHelmOverrides,
```
davisgomes, line 45: This will conflict with some of the changes we are working on. Whoever merges second will have to update.

fredtibbitts (commented):
The arbitrary helm values is still a work in progress. My feeling is we shouldn't be extending other components to support this until we get it all working with one component. I think we should hold on this.

platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway_component.go
anajosep: Ack, I will rebase once the changes are in.

platform-operator/controllers/verrazzano/component/prometheus/pushgateway/pushgateway.go
kminder: Isn't that what line 71 does?

davisgomes: Yes! That wasn't there when I left this comment.

platform-operator/helm_config/overrides/prometheus-pushgateway-values.yaml
```diff
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 serviceMonitor:
-  enabled: true
+  enabled: false
```
fredtibbitts, line 5: I think the service monitor should be enabled, why is this changing to false?

anajosep: Since prometheus and prometheus operator are disabled by default, creating service monitor when pushgateway alone is enabled might cause errors. So I thought of keeping the default behaviour as false and only enabled when prometheus is enabled or overriden by helm values in CR.

fredtibbitts: Oh I see, I missed that in the code above. That's fine, although I think you should change the check from Prometheus enabled to PrometheusOperator enabled, since the Prometheus Operator install is what creates the ServiceMonitor CRD and the operator consumes the ServiceMonitor resources.

anajosep:
Changes no longer required.

PR #3067, opened 2022-05-04
VZ-5733 Helm values resource monitoring implementation

adiforluls:
# Description

Contains new controller implementations for Secrets and ConfigMaps to monitor file resources for arbitrary helm overrides.

Fixes VZ-5733

# Checklist 


platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,130 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package configmaps
+
+import (
+	"context"
+	installv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
```
davisgomes, line 11: Reorganize imports

platform-operator/controllers/controller_utils_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Kind: "ConfigMap",
+	},
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      testCMName,
+		Namespace: testNS,
+	},
+	Immutable:  nil,
+	Data:       map[string]string{"override": "true"},
+	BinaryData: nil,
+}
```
davisgomes, line 44: Again, I feel that there are some edge cases left out in the unit tests.

platform-operator/controllers/secrets/helm_override_secret.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	installv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (r *VerrazzanoSecretsReconciler) reconcileHelmOverrideSecret(ctx context.Context, req ctrl.Request, vz *installv1alpha1.Verrazzano) (ctrl.Result, error) {
```
davisgomes, line 16: Add function comment

platform-operator/controllers/secrets/helm_override_secret.go
```diff
@@ -0,0 +1,54 @@
...
+			r.log.Errorf("Failed to construct component context: %v", err)
+			return newRequeueWithDelay(), err
+		}
+		if componentName, ok := controllers.VzContainsResource(componentCtx, secret); ok {
+			err := r.updateVerrazzanoForHelmOverrides(componentCtx, componentName)
+			if err != nil {
+				r.log.Errorf("Failed to reconcile ConfigMap: %v", err)
+				return newRequeueWithDelay(), err
+			}
+			r.log.Infof("Updated Verrazzano Resource")
```
davisgomes, line 40: This section looks really similar to the configmap reconciler. Can these be consolidated into one util function?

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,130 @@
...
+		_, err := r.reconcileHelmOverrideConfigMap(ctx, req, vz)
+		if err != nil {
+			zap.S().Errorf("Failed to reconcile ConfigMap: %v", err)
+			return newRequeueWithDelay(), err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *VerrazzanoConfigMapsReconciler) reconcileHelmOverrideConfigMap(ctx context.Context, req ctrl.Request, vz *installv1alpha1.Verrazzano) (ctrl.Result, error) {
```
davisgomes, line 69: Add function comment

davisgomes: Add for the rest. I won't comment on each function.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		componentCtx, err := spi.NewContext(r.log, r.Client, vz, false)
+		if err != nil {
+			r.log.Errorf("Failed to construct component context: %v", err)
+			return newRequeueWithDelay(), err
+		}
+
+		if componentName, ok := controllers.VzContainsResource(componentCtx, configMap); ok {
+			err := r.updateVerrazzanoForHelmOverrides(componentCtx, componentName)
```
shihchang, line 89: Is it sufficient to update/reconcile only the component overridden? Which components will have this arbitrary overrides and do they have dependent components? @michael-cico

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: fixing

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// SetupWithManager creates a new controller and adds it to the manager
+func (r *VerrazzanoConfigMapsReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}).
+		Complete(r)
+}
+
+func (r *VerrazzanoConfigMapsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
adiforluls: fixing, forgot to do in a bunch of places

platform-operator/controllers/configmaps/configmaps_controller_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	config.TestProfilesDir = "../../manifests/profiles"
+	defer func() { config.TestProfilesDir = "" }()
+
+	request := newRequest(testNS, testCMName)
+	reconciler := newConfigMapReconciler(cli)
+	res, err := reconciler.Reconcile(context.TODO(), request)
+
+	asserts.NoError(err)
+	asserts.Equal(false, res.Requeue)
```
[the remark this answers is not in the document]
adiforluls: If the request is from a different namespace, an empty result is returned in such a case with no error. I'll add a test which will expect that no API calls are being made to get the ConfigMap in this case, and another case where request is from the same namespace but the cluster is missing the ConfigMap, that will result in a requeue and generate an error.

platform-operator/controllers/secrets/helm_override_secret.go
```diff
@@ -0,0 +1,54 @@
...
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (r *VerrazzanoSecretsReconciler) reconcileHelmOverrideSecret(ctx context.Context, req ctrl.Request, vz *installv1alpha1.Verrazzano) (ctrl.Result, error) {
+
+	secret := &corev1.Secret{}
+	if vz.Namespace == req.Namespace {
+		if err := r.Get(ctx, req.NamespacedName, secret); err != nil {
+			zap.S().Errorf("Failed to fetch ConfigMap in Verrazzano CR namespace: %v", err)
```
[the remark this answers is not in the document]
adiforluls: Ah good catch

platform-operator/controllers/secrets/helm_override_secret.go
adiforluls: Yeah, this can probably become a utility func

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: This line of code is looking for the ConfigMap in the CR. If it is referenced under a component, the idea then is to meet the criteria to cause the component to re-enter the install flow.
Currently only prometheus-operator has this functionality, other helm components will also have this functionality in the coming sprints, and afaik they are independent. But I think any component that supports value overrides will also share this functionality, so moving forward we can add handling for components that are co-dependent.

platform-operator/controllers/configmaps/configmaps_controller_test.go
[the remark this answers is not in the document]
adiforluls: Made a slight modification, ConfigMap/Secret not found will not result in requeue now

platform-operator/controllers/secrets/secrets_controller_test.go
```diff
@@ -6,9 +6,14 @@ package secrets
...
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"testing"
+	"time"
```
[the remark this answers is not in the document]
adiforluls: Ah alright, I thought I ran goimports but I suspect it only sorted stuff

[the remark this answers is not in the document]
adiforluls: Fixed

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -38,79 +41,62 @@ func (r *VerrazzanoSecretsReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	if req.Name != constants.VerrazzanoIngressSecret || req.Namespace != constants.VerrazzanoSystemNamespace {
-		return ctrl.Result{}, nil
+	if req.Name == constants.VerrazzanoIngressSecret && req.Namespace == constants.VerrazzanoSystemNamespace {
+		return r.reconcileVerrazzanoTLS(ctx, req)
 	}
 
-	// Get the verrazzano ingress secret
-	caSecret := corev1.Secret{}
-	err := r.Get(context.TODO(), req.NamespacedName, &caSecret)
+	vzList := &installv1alpha1.VerrazzanoList{}
```
michael-cico, line 55: maybe break this out into its own reconcile private method, like you did with `reconcileVerrazzanoTLS` above.  not blocking, but maybe a follow-on.

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
```diff
@@ -80,6 +80,16 @@ func (g grafanaComponent) GetJSONName() string {
 	return ComponentJSONName
 }
 
+// GetHelmOverrides returns the Helm overrides for a component
+func (g grafanaComponent) GetHelmOverrides(_ spi.ComponentContext) []vzapi.Overrides {
+	return []vzapi.Overrides{}
+}
+
+// IsMonitoringEnabled indicates if monitoring of override sources is enabled or not for a component
+func (g grafanaComponent) IsMonitoringEnabled(_ spi.ComponentContext) bool {
```
michael-cico, line 89: Name also is a bit vague.  Also, when is monitoring of override sources not enabled?

platform-operator/controllers/secrets/secrets_controller.go
adiforluls: The main processing is being done in a similar fashion, this is just getting verrazzano resource and then calling the private reconcile method

platform-operator/controllers/secrets/helm_override_secret.go
```diff
@@ -0,0 +1,56 @@
...
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// reconcileHelmOverrideSecret looks through the Verrazzano CR for the Secret
+// if the request is from the same namespace as the CR
+func (r *VerrazzanoSecretsReconciler) reconcileHelmOverrideSecret(ctx context.Context, req ctrl.Request, vz *installv1alpha1.Verrazzano) (ctrl.Result, error) {
+
+	secret := &corev1.Secret{}
+	if vz.Namespace == req.Namespace {
```
michael-cico, line 26: Is it a requirement that the secrets override resource be in the same namespace as the VZ resource?

platform-operator/controllers/secrets/secrets_controller_test_utils.go
```diff
@@ -0,0 +1,97 @@
...
+			{
+				Type: vzapi.CondInstallComplete,
+			},
+		},
+		Components: compStatusMap,
+	},
+}
+
+// create component status map required for testing
+func makeVerrazzanoComponentStatusMap() vzapi.ComponentStatusMap {
```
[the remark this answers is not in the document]
adiforluls: Will refactor

platform-operator/controllers/secrets/helm_override_secret.go
adiforluls: Yea it's from the design

platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -52,6 +52,10 @@ type ComponentInfo interface {
 	GetCertificateNames(context ComponentContext) []types.NamespacedName
 	// GetJsonName returns the josn name of the verrazzano component in CRD
 	GetJSONName() string
+	// GetHelmOverrides returns the list of overrides for a component
```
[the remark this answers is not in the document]
adiforluls: Mmmm yea will refactor

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
adiforluls: I'll rename the func to something else

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,150 @@
...
+				if configMap.Finalizers == nil {
+					configMap.Finalizers = append(configMap.Finalizers, constants.KubeFinalizer)
+					err := r.Update(context.TODO(), configMap)
+					if err != nil {
+						return newRequeueWithDelay(), err
+					}
+					return reconcile.Result{Requeue: true}, nil
+				}
+			} else {
+				configMap.Finalizers = nil
```
michael-cico, line 112: I think we should only remove our finalizer, and have a unique finalizer name like we do for the VZ resource (maybe can use the same?).

cc: @kminder @pfmackin

michael-cico: @kminder @pfmackin Also we had some questions about processing CM deletes, esp if there's more than one finalizer.  Right now we just return from the controller if the resource isn't found, but we were thinking that we should always trigger a VZ reconcile in that case.  The reason being that in the very remote possibility of having another non-VZ finalizer attached to the CM, we may trigger a reconcile while the CM still exists, and when the other finalizer is removed we may get another update when it's truly gone.  

We may want to fabricate that scenario, if we decide to just remove our own finalizer and not nil-out the whole field.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Get Verrazzano from the cluster
+	vzList := &installv1alpha1.VerrazzanoList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(""),
+	}
+	err := r.List(ctx, vzList, listOpts...)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			zap.S().Infof("VZ Not found ConfigMap")
```
kminder, line 58: This message could be improved.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *VerrazzanoConfigMapsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	if ctx == nil {
+		ctx = context.TODO()
+	}
+
+	// Get Verrazzano from the cluster
+	vzList := &installv1alpha1.VerrazzanoList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(""),
```
kminder, line 53: Do we only support Verrazzano CRs in the default namespace?  I don't even know.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	if vz.Namespace == req.Namespace {
+		if err := r.Get(ctx, req.NamespacedName, configMap); err != nil {
+			zap.S().Errorf("Failed to fetch ConfigMap in Verrazzano CR namespace: %v", err)
+			// Do not reconcile if the ConfigMap was deleted
+			if errors.IsNotFound(err) {
+				return reconcile.Result{}, nil
+			}
+			return newRequeueWithDelay(), err
+		}
+		if result, err := r.initLogger(*configMap); err != nil {
```
kminder, line 91: There might be a good reason but it struck me as odd that the logger is initialized this late in the processing.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,160 @@
...
+					configMap.Finalizers = append(configMap.Finalizers, constants.KubeFinalizer)
+					err := r.Update(context.TODO(), configMap)
+					if err != nil {
+						return newRequeueWithDelay(), err
+					}
+					return reconcile.Result{Requeue: true}, nil
+				}
+			} else {
+				// Requeue if other finalizers are present
+				if len(configMap.Finalizers) > 1 {
```
kminder, line 115: If some other controller were doing the same thing wouldn't this effectively cause an infinite requeue loop?

platform-operator/controllers/configmaps/configmaps_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// TestConfigMapReconciler tests Reconciler method of Configmap controller.
```
kminder, line 30: Should be using the standard test header description.  Same for all the tests.

platform-operator/controllers/controller_utils_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	compContext := fakeComponentContext(mock, &testVZ)
+
+	mock.EXPECT().Status().Return(mockStatus).AnyTimes()
+	mockStatus.EXPECT().
+		Update(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, vz *vzapi.Verrazzano, opts ...client.UpdateOption) error {
+			return nil
+		})
+	err := UpdateVerrazzanoForHelmOverrides(mock, compContext, "prometheus-operator")
```
kminder, line 91: Isn't a mock.finish() required?

platform-operator/controllers/secrets/helm_override_secret.go
```diff
@@ -0,0 +1,83 @@
...
+		if err := r.Get(ctx, req.NamespacedName, secret); err != nil {
+			zap.S().Errorf("Failed to fetch Secret in Verrazzano CR namespace: %v", err)
+			// Do not retry if secret is deleted
+			if errors.IsNotFound(err) {
+				return reconcile.Result{}, nil
+			}
+			return newRequeueWithDelay(), err
+		}
+
+		if result, err := r.initLogger(*secret); err != nil {
```
kminder, line 39: Same logger init question.

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -40,92 +42,62 @@ func (r *VerrazzanoSecretsReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	caSecret := corev1.Secret{}
-	err := r.Get(context.TODO(), req.NamespacedName, &caSecret)
+	err := r.List(ctx, vzList, listOpts...)
 	if err != nil {
-		// Secret should never be not found, unless we're running while installation is still underway
-		zap.S().Errorf("Failed to fetch secret %s/%s: %v",
-			req.Namespace, req.Name, err)
-		return newRequeueWithDelay(), nil
+		if apierrors.IsNotFound(err) {
+			zap.S().Infof("VZ not found Secret")
```
kminder, line 66: Improve or remove this error message.

platform-operator/controllers/secrets/secrets_controller_test.go
```diff
@@ -159,6 +164,167 @@ func TestMultiClusterNamespaceUnexpectedErr(t *testing.T) {
...
+	reconciler := newSecretsReconciler(cli)
+	res0, err0 := reconciler.Reconcile(context.TODO(), request)
+
+	asserts.NoError(err0)
+	asserts.Empty(res0)
+
+	vz := vzapi.Verrazzano{}
+	err := cli.Get(context.TODO(), types.NamespacedName{Namespace: testNS, Name: testVZName}, &vz)
+	asserts.NoError(err)
+	asserts.Equal(int64(1), vz.Status.Components["prometheus-operator"].ReconcilingGeneration)
```
kminder, line 187: These aren't asserting very much.  Are there really no other expected changes to the state?  I noticed this in other tests also.

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: This can be yanked out. I initially was using it as a debug message

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: Probably any namespace, setting Namespace option to empty implies that the list method is gonna check all namespaces to find verrazzano

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: I'll see if it makes sense to get the resource early to setup the logger

platform-operator/controllers/configmaps/configmaps_controller_test.go
adiforluls: Will fix

platform-operator/controllers/controller_utils_test.go
adiforluls: Ah I missed that, will fix

platform-operator/controllers/secrets/secrets_controller_test.go
adiforluls: These controllers only set `ReconcilingGeneration` to 1 for the target component, after this the Verrazzano controller Reconcile loop will be called to change things further in real scenario, so yeah these are pretty much the only changes that these controllers do.
Basically it's meeting the minimum possible criteria to cause an update to the target component.

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: The actual work starts after the logger has been setup (before this we are only getting Verrazzano and the target resource (CM/Secret), so it made sense for it to remain here.

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: Removed this info

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: By default List looked in all NS so removed ListOption

platform-operator/controllers/configmaps/configmaps_controller.go
adiforluls: Updated this logic to remove the wait

platform-operator/controllers/secrets/secrets_controller_test.go
adiforluls: I evaluated the UTs and added more sensible assertions in couple places, but there are only 3 things to assert in almost all the tests (result, error, expected Verrazzano or CM/Secret update).

kminder: ok

kminder (approved):
Approved.

platform-operator/controllers/configmaps/configmaps_controller.go
```diff
@@ -0,0 +1,162 @@
...
+					configMap.Finalizers = append(configMap.Finalizers, constants.OverridesFinalizer)
+					err := r.Update(context.TODO(), configMap)
+					if err != nil {
+						return newRequeueWithDelay(), err
+					}
+					return reconcile.Result{Requeue: true}, nil
+				}
+			} else {
+				// Requeue if other finalizers are present
+				if configMap.Finalizers != nil && !controllerutil.ContainsFinalizer(configMap, constants.OverridesFinalizer) {
```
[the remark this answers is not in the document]
adiforluls: Well if there are finalizers present but we have removed ours, then we basically retry since the object is still present in the cluster. On retrying we either find out that finalizers which we don't own are still present or we find that the object has been truly deleted (which will trigger a reconcile to unapply deleted values).
If there is only our finalizer then it is removed and the values are unapplied through reconciliation of Verrazzano immediately.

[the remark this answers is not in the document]
kminder: k

PR #3071, opened 2022-05-04
VZ-5810.  Fix noisy controller message part 2

michael-cico:
# Description

Checks for verrazzano-mc namespace existence before doing any other processing in the secrets controller.
- Also contains some minor updates to the VZ logger per previous PR comments.

Fixes VZ-5810.

# Checklist 

### DIFF
--- a/platform-operator/controllers/secrets/secrets_controller.go
+++ b/platform-operator/controllers/secrets/secrets_controller.go
@@ -100,11 +100,6 @@ func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return nil
 	})
 
-	if !r.multiclusterNamespaceExists() {
-		// Multicluster namespace doesn't exist yet, requeue
-		return newRequeueWithDelay(), nil
-	}
-
 	if err != nil {
 		r.log.ErrorfThrottled("Failed to create or update secret %s/%s: %s",
 			constants.VerrazzanoMultiClusterNamespace, constants.VerrazzanoLocalCABundleSecret, err.Error())
--- a/platform-operator/controllers/secrets/secrets_controller.go
+++ b/platform-operator/controllers/secrets/secrets_controller.go
@@ -5,6 +5,7 @@ package secrets
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/types"
 	"time"
 
 	vzctrl "github.com/verrazzano/verrazzano/pkg/controller"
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -93,6 +99,12 @@ func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		mcCASecret.Data["ca-bundle"] = caSecret.Data["ca.crt"]
 		return nil
 	})
+
+	if !r.multiclusterNamespaceExists() {
```
pfmackin, line 103: do we need to do this again?

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -68,6 +69,11 @@ func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 	r.log = log
 
+	if !r.multiclusterNamespaceExists() {
```
pfmackin, line 72: can we just create it?

platform-operator/controllers/secrets/secrets_controller.go
michael-cico: Fixed and pushed

platform-operator/controllers/secrets/secrets_controller.go
wmhopkins: I believe today it's create by the VPO during install, so I don't think it makes sense for the secrets controller to create it. There should be only one owner for the resource, and that owner should probably continue to be the installer.

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -104,6 +111,19 @@ func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+func (r *VerrazzanoSecretsReconciler) multiclusterNamespaceExists() bool {
+	ns := corev1.Namespace{}
+	err := r.Get(context.TODO(), types.NamespacedName{Name: constants.VerrazzanoMultiClusterNamespace}, &ns)
+	if err == nil {
+		return true
+	}
+	if !apierrors.IsNotFound(err) {
+		r.log.ErrorfThrottled("Unexpected error checking for namespace %s: %v", constants.VerrazzanoMultiClusterNamespace, err)
+	}
+	r.log.Progressf("Namespace %s does not exist, nothing to do", constants.VerrazzanoMultiClusterNamespace)
```
wmhopkins, line 123: This message will be wrong if the secrets_controller ends up managing more secrets than just this one.

wmhopkins: Also, adding the check for the namespace and short-circuiting if not doesn't hurt, but it's a bunch more code when we could just rely on the create/update failure to signal the problem -- we could just throttle the existing error message.

wmhopkins (approved):
lgtm.

platform-operator/controllers/secrets/secrets_controller.go
michael-cico: I thought it would be useful to just avoid doing any of the work if the namespace doesn't exist.  

The code will have to change the reconcile does more than manage `verrazzano-mc` secrets, but that's probably an exercise for the future implementer of that.

PR #3079, opened 2022-05-05
VZ-5780 -  When managed cluster is updated to use Let's Encrypt staging certs, make sure CA sync to admin cluster works (i.e. use tls-additional-ca secret when it exists instead of the verrazzano-tls secret)

desagar:
# Description

Fixes VZ-5780

# Checklist 

### DIFF
--- a/application-operator/mcagent/mcagent_clusterca.go
+++ b/application-operator/mcagent/mcagent_clusterca.go
@@ -16,9 +16,10 @@ import (
 )
 
 const (
-	keyCaCrtNoDot = "cacrt"
-	keyCaCrt      = "ca.crt"
-	keyCaBundle   = "ca-bundle"
+	keyCaCrtNoDot   = "cacrt"
+	keyCaCrt        = "ca.crt"
+	keyCaBundle     = "ca-bundle"
+	keyAdditionalCa = "ca-additional.pem"
 )
 
 // Synchronize Secret objects to the local cluster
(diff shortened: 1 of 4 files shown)


application-operator/mcagent/mcagent_clusterca.go
```diff
@@ -122,6 +118,33 @@ func (s *Syncer) syncLocalClusterCA() error {
...
+		Name:      globalconst.AdditionalTLS,
+	}, &localCASecret)
+	if client.IgnoreNotFound(errAddlTLS) != nil {
+		return nil, errAddlTLS
+	}
+
+	if errAddlTLS == nil {
+		return localCASecret.Data[keyAdditionalCa], nil
+	}
+	// additional TLS secret not found, check for Verrazzano TLS secret
```
michael-cico, line 137: Should we just bundle them all if both are present?

We should be cleaning up the things we aren't using on updates, but is it harmful/insecure to include both?

wmhopkins: Well, both will always be present for the staging case, but the ca.crt in verrazzano-tls should be empty. If it's not -- if we get non-empty values from both secrets, then our config is broken, because it's a hybrid of staging and self-signed or custom CA.

To answer the more general question -- is it bad to include all the CAs we can find -- I think the answer is yes. We should only trust LE/staging certs if we're actually configured to use them. The staging environment is explicitly documented as not being as rigorously vetted as the prod environment, and LE specifically warns against trusting the staging CAs in any client that you use for any other purpose. So bad to trust them in your browser, e.g. -- I suspect some of us do -- and we shouldn't have VZ clients trust them if they're not actually in use.

wmhopkins (approved):
lgtm.

wmhopkins (approved):

PR #3099, opened 2022-05-06
NO JIRA: Align Fluentd matcher for system components

anders-swanson:
Observed that the new `verrazzano-monitoring` namespace was being indexed as a Verrazzano application in OpenSearch. I updated the Fluentd matcher to align with the oci logging matcher to fix this.


fredtibbitts (approved):

kminder (approved):
Approved.

PR #3100, opened 2022-05-06
NO JIRA: framework.Log.x() tees to Stdout

anders-swanson:
Calls to framework.Log were only exported to the log db, and not visible in the console. This change tees output from framework.Log to `stdout`.


kminder (approved):
Approved.

PR #3106, opened 2022-05-09
VZ-5464 fluentd update tests

shihchang:
# Description
* add validation for OCI logging and external Opensearch
* updater may return error for testing validation
* add fluentd update to dynamic-config-tests

Fixes VZ-5464

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
+++ b/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
@@ -349,7 +349,7 @@ func validateEntryExist(secret *corev1.Secret, entry string) error {
 	secretName := secret.Name
 	_, ok := secret.Data[entry]
 	if !ok {
-		return fmt.Errorf("missing %s in secret \"%s\"", entry, secretName)
+		return fmt.Errorf("invalid Fluentd configuration, missing %s entry in secret \"%s\"", entry, secretName)
 	}
 	return nil
 }
--- a/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
+++ b/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
@@ -4,9 +4,11 @@
 package verrazzano
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
 
+	globalconst "github.com/verrazzano/verrazzano/pkg/constants"
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/authproxy"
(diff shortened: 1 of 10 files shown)


platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -261,3 +271,85 @@ func (c verrazzanoComponent) GetCertificateNames(ctx spi.ComponentContext) []typ
...
+					return fmt.Errorf("duplicate mount path found: %s; Fluentd by default has mount paths: %v", mountPath, existingFluentdMountPaths)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func validateLogCollector(fluentd *vzapi.FluentdComponent) error {
+	if fluentd.OCI != nil && fluentd.ElasticsearchURL != globalconst.DefaultOpensearchURL && fluentd.ElasticsearchURL != "" {
```
michael-cico, line 321: could the `ElasticsearchURL` checks just be 

```
len(fluentd.ElasticsearchURL) > 0
```

?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
```diff
@@ -261,3 +271,85 @@ func (c verrazzanoComponent) GetCertificateNames(ctx spi.ComponentContext) []typ
...
+		return validateEntryExist(secret, esPasswordKey)
+	}
+	return nil
+}
+
+func validateEntryExist(secret *corev1.Secret, entry string) error {
+	secretName := secret.Name
+	_, ok := secret.Data[entry]
+	if !ok {
+		return fmt.Errorf("missing %s in secret \"%s\"", entry, secretName)
```
michael-cico, line 352: Might be good to have a more descriptive error, i.e.,

```
Invalid Fluentd configuration, missing %s entry in secret \"%s\"
```

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
fredtibbitts: Currently it's one or the other. I don't think it would take much work to support both at the same time, but we don't support both today.

tests/e2e/update/fluentd/fluentd_update_test.go
```diff
@@ -0,0 +1,296 @@
...
+})
+
+var _ = t.Describe("Update Fluentd", Label("f:platform-lcm.update"), func() {
+	t.Describe("fluentd verify", Label("f:platform-lcm.fluentd-verify"), func() {
+		t.It("fluentd default config", func() {
+			validateDaemonset(pkg.VmiESURL, pkg.VmiESInternalSecret, "")
+		})
+	})
+
+	t.Describe("Validate external Opensearch config", Label("f:platform-lcm.fluentd-update-validation"), func() {
```
michael-cico, line 99: Looks like you're covering a bunch of configuration changes here; do we validate the logging backends at all?  What state is the system in after these run?

Main reason I ask is that we will be running the system/app verifications after this stage, hopefully will they find the logs in the right place?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
shihchang: in effectiveCR, ElasticsearchURL="" would be translated to globalconst.DefaultOpensearchURL, so only if ElasticsearchURL != DefaultOpensearchURLand ElasticsearchURL != "" then it is external Opensearch

tests/e2e/update/fluentd/fluentd_update_test.go
shihchang: in this test-suite, we only validate if the state of the system has been correctly re-configured, we do not validate that the fluentd in such state is functioning in this suite as those should have been covered in those suites (muti-cluster-AT and oci-integration)

michael-cico: Technically, it means that these are more like integration tests instead of e2e tests, but I'm OK with that.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go
michael-cico: I thought the default Opensearch URL was our internal endpoint....but this works so 👍

michael-cico (approved):

PR #3107, opened 2022-05-09
VZ-5076 Cleanup errors in AuthProxy log during install

ayaranga:
# Description

Added code to return Service Unavailable error (503) from AuthProxy for 5xx errors from Keycloak. 

Fixes VZ-5076

# Checklist 

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
@@ -697,11 +697,11 @@ data:
         end
         if not res then
             me.info("Failed requesting token from Keycloak: response is nil")
-            me.unauthorized("Error requesting token: response was nil")
+            me.unauthorized("Error requesting token: response is nil")
         end
         if not (res.status == 200) then
             me.info("Failed requesting token from Keycloak: response status code is "..res.status.." and response body is "..res.body)
-            me.unauthorized("Error requesting token: response status code was "..res.status.." and response body was "..res.body)
+            me.unauthorized("Error requesting token: response status code is "..res.status.." and response body is "..res.body)
         end
         local tokenRes = cjson.decode(res.body)
         if tokenRes.error or tokenRes.error_description then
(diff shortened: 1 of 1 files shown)


platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -699,6 +706,10 @@ data:
             me.info("Failed requesting token from Keycloak: response is nil")
             me.unauthorized("Error requesting token: response was nil")
         end
+        if res.status >= 500 then
```
wmhopkins, line 709: While this change addresses the specific issue reported in the Jira, what's wanted here is more general error handling. Anything other than a 200 status code indicates a problem getting the token, and therefore a failure of this method and a failure to authenticate the user. So I would change the condition here to `res.status != 200`.

Also, this should return unauthorized, rather than service_unavailable. While 503 may accurately reflect the nature of the problem for some cases (not necessarily everything >= 500), the context here is that we're trying to authenticate a user. If we fail for any reason, we should always return 401. This tells the user what the high-level problem is, from their POV, without leaking any information that could potentially help someone trying to break into the system. The log message will be enough for an admin to see what's going on.

wmhopkins (changes requested):

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -699,6 +706,10 @@ data:
...
+            me.info("Failed requesting token from Keycloak: response body is "..res.body)
+            me.service_unavailable("Error requesting token: response status code was "..res.status)
```
wmhopkins, line 711: Would be good to be consistent in logging both status and body in both places (I think it ends up in the same log with different levels, so may be redundant when debug is enabled, but there's not much point in logging one without the other.)

wmhopkins (changes requested):

wmhopkins (approved):
lgtm.

anajosep (approved):

PR #3133, opened 2022-05-12
VZ-5781 - Sync registration info from admin to managed cluster so that DNS updates are picked up. Also make sure update to Let's Encrypt Staging admin cluster updates the local ca bundle synced to managed.

desagar:
# Description

When admin cluster DNS and certs are updated, the managed cluster fluentd log pushes should continue to work (they don't always, right now).  The following changes were made:

1. Sync the registration secret entries for ES URL and creds to the managed cluster when they change. That way when there are DNS updates on admin cluster, the new ES URL automatically syncs to managed cluster
2. In secret_controller, make sure that the admin cluster local ca bundle is also updated for the LE staging case (i.e. based on additional TLS secret present in cattle-system). Otherwise updating admin cluster to LE staging certs will not sync the certs to managed cluster.
3. Just added - force a fluentd restart (using an annotation on daemonset) when managed cluster CAs or registration is updated. This restart is not guaranteed to happen when only certs are updated.

Fixes VZ-9999

# Checklist 

### DIFF
--- a/application-operator/mcagent/mcagent_clusterca.go
+++ b/application-operator/mcagent/mcagent_clusterca.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/verrazzano/verrazzano/application-operator/constants"
 	globalconst "github.com/verrazzano/verrazzano/pkg/constants"
+	"github.com/verrazzano/verrazzano/pkg/mcconstants"
 	platformopclusters "github.com/verrazzano/verrazzano/platform-operator/apis/clusters/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -17,13 +18,11 @@ import (
 
 const (
 	keyCaCrtNoDot = "cacrt"
-	keyCaCrt      = "ca.crt"
-	keyCaBundle   = "ca-bundle"
 )
 
 // Synchronize Secret objects to the local cluster
 func (s *Syncer) syncClusterCAs() error {
-	err := s.syncAdminClusterCA()
+	err := s.syncRegistrationFromAdminCluster()
 	if err != nil {
 		s.Log.Errorf("Error syncing Admin Cluster CA: %v", err)
 	}
@@ -34,15 +33,28 @@ func (s *Syncer) syncClusterCAs() error {
 	return nil
 }
 
-// syncAdminClusterCA - synchronize the admin cluster CA cert -- update local copy if admin CA changes
-func (s *Syncer) syncAdminClusterCA() error {
+// syncRegistrationFromAdminCluster - synchronize the admin cluster registration info including
+// CA cert, URLs and credentials -- update local registration if any of those change
+func (s *Syncer) syncRegistrationFromAdminCluster() error {
 
-	// Get the cluster CA secret from the admin cluster
+	// Get the cluster CA secret from the admin cluster - for the CA secret, this is considered
+	// the source of truth
 	adminCASecret := corev1.Secret{}
 	err := s.AdminClient.Get(s.Context, client.ObjectKey{
 		Namespace: constants.VerrazzanoMultiClusterNamespace,
 		Name:      constants.VerrazzanoLocalCABundleSecret,
 	}, &adminCASecret)
+	if err != nil {
+		return err
+	}
+
+	// Get the managed cluster registration secret for THIS managed cluster, from the admin cluster.
+	// This will be used to sync registration information here on the managed cluster.
+	adminRegistrationSecret := corev1.Secret{}
+	err = s.AdminClient.Get(s.Context, client.ObjectKey{
+		Namespace: constants.VerrazzanoMultiClusterNamespace,
+		Name:      getRegistrationSecretName(s.ManagedClusterName),
+	}, &adminRegistrationSecret)
 	if err != nil {
 		return err
 	}
@@ -57,10 +69,19 @@ func (s *Syncer) syncAdminClusterCA() error {
 		return err
 	}
 
-	// Update the local cluster registration secret if the admin CA certs are different
-	if !secretsEqualTrimmedWhitespace(registrationSecret.Data[keyCaBundle], adminCASecret.Data[keyCaBundle]) {
+	// Update the local cluster registration secret if the admin CA certs are different, or if
+	// any of the registration info on admin cluster is different
+	if !byteSlicesEqualTrimmedWhitespace(registrationSecret.Data[mcconstants.AdminCaBundleKey], adminCASecret.Data[mcconstants.AdminCaBundleKey]) ||
+		!registrationInfoEqual(registrationSecret, adminRegistrationSecret) {
 		result, err := controllerutil.CreateOrUpdate(s.Context, s.LocalClient, &registrationSecret, func() error {
-			registrationSecret.Data[keyCaBundle] = adminCASecret.Data[keyCaBundle]
+			// Get CA info from admin CA secret
+			registrationSecret.Data[mcconstants.AdminCaBundleKey] = adminCASecret.Data[mcconstants.AdminCaBundleKey]
+
+			// Get other registration info from admin registration secret for this managed cluster
+			registrationSecret.Data[mcconstants.ESURLKey] = adminRegistrationSecret.Data[mcconstants.ESURLKey]
+			registrationSecret.Data[mcconstants.RegistrationUsernameKey] = adminRegistrationSecret.Data[mcconstants.RegistrationUsernameKey]
+			registrationSecret.Data[mcconstants.RegistrationPasswordKey] = adminRegistrationSecret.Data[mcconstants.RegistrationPasswordKey]
+			registrationSecret.Data[mcconstants.KeycloakURLKey] = adminRegistrationSecret.Data[mcconstants.KeycloakURLKey]
 			return nil
 		})
 		if err != nil {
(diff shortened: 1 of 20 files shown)


platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -39,21 +41,36 @@ func (r *VerrazzanoSecretsReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	caKey := "ca.crt"
+	if isAddnlTLSSecret {
+		caKey = vzconst.AdditionalTLSCAKey
+	}
 
-	// The only secret we care about (for now) is the verrazzano ingress tls secret (verrazzano-tls)
-	if req.Name != constants.VerrazzanoIngressSecret || req.Namespace != constants.VerrazzanoSystemNamespace {
+	if isVzIngressSecret && r.additionalTLSSecretExists() {
```
michael-cico, line 58: Do we not want to sync this bundle if it exists?

pkg/constants/constants.go
```diff
@@ -83,6 +83,9 @@ const VerrazzanoPromInternal = "verrazzano-prom-internal"
 // AdditionalTLS is an optional tls secret that contains additional CA
 const AdditionalTLS = "tls-ca-additional"
 
+// AdditionalTLSCAKey is the key containing the CA in the secret specified by the AdditionalTLS constant
+const AdditionalTLSCAKey = "ca-additional.pem"
```
pfmackin, line 87: what does additional mean in this context?

platform-operator/controllers/clusters/sync_agent_secret.go
```diff
@@ -83,7 +84,7 @@ func (r *VerrazzanoManagedClusterReconciler) syncAgentSecret(vmc *clusterapi.Ver
 	}
 
 	// Load the kubeconfig struct
-	token := secret.Data[TokenKey]
+	token := secret.Data[mcconstants.TokenKey]
```
pfmackin, line 87: Is this the service account token?  If so can that be in the name, as in ServiceAccountTokenKey

platform-operator/controllers/clusters/sync_registration_secret.go
```diff
@@ -123,13 +123,13 @@ func (r *VerrazzanoManagedClusterReconciler) mutateRegistrationSecret(secret *co
...
-		ESURLKey:                []byte(esURL),
-		ESCaBundleKey:           esCaBundle,
-		RegistrationUsernameKey: esUsername,
-		RegistrationPasswordKey: esPassword,
-		KeycloakURLKey:          []byte(keycloakURL),
-		AdminCaBundleKey:        adminCaBundle,
+		mcconstants.ManagedClusterNameKey:   []byte(manageClusterName),
+		mcconstants.ESURLKey:                []byte(esURL),
+		mcconstants.ESCaBundleKey:           esCaBundle,
+		mcconstants.RegistrationUsernameKey: esUsername,
```
pfmackin, line 129: Why is the creds called Registration.., that is confusing to me.  Shouldn't it be ESUsername, ESPassword?

platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -1398,14 +1399,14 @@ func expectSyncRegistration(t *testing.T, mock *mocks.MockClient, name string, e
 		DoAndReturn(func(ctx context.Context, name types.NamespacedName, secret *corev1.Secret) error {
 			if externalES {
 				secret.Data = map[string][]byte{
-					VerrazzanoUsernameKey: []byte(externalUserData),
-					VerrazzanoPasswordKey: []byte(externalPasswordData),
-					FluentdESCaBundleKey:  []byte(externalCaData),
+					mcconstants.VerrazzanoUsernameKey: []byte(externalUserData),
```
pfmackin, line 1402: why is key Verrazzano prefixed by the var is external prefixed.  What exactly are these creds for?

platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -1491,11 +1492,11 @@ func expectSyncManifest(t *testing.T, mock *mocks.MockClient, mockStatus *mocks.
 		Get(gomock.Any(), types.NamespacedName{Namespace: constants.VerrazzanoMultiClusterNamespace, Name: GetRegistrationSecretName(name)}, gomock.Not(gomock.Nil())).
 		DoAndReturn(func(ctx context.Context, name types.NamespacedName, secret *corev1.Secret) error {
 			secret.Data = map[string][]byte{
-				ManagedClusterNameKey:   []byte(clusterName),
-				CaCrtKey:                []byte(caData),
-				RegistrationUsernameKey: []byte(userData),
-				RegistrationPasswordKey: []byte(passwordData),
-				ESURLKey:                []byte(urlData),
+				mcconstants.ManagedClusterNameKey:   []byte(clusterName),
+				mcconstants.CaCrtKey:                []byte(caData),
```
pfmackin, line 1496: what is this CaCrtKey?   I know it is a CA bundle, but what is it used for?   Is there some comments somewhere that explains it?

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -39,21 +41,36 @@ func (r *VerrazzanoSecretsReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// We only care about the CA secret for the cluster - this can come from the verrazzano ingress
```
pfmackin, line 44: I don't understand what this secret is for and why it is not always in verrazzano-system.  It is non-intuitive that we would care about a secret in cattle-system?  Can you add additional comments (i know it is not directly related to your change)

application-operator/mcagent/mcagent_clusterca.go
```diff
@@ -55,24 +67,44 @@ func (s *Syncer) syncAdminClusterCA() error {
...
+		!registrationInfoEqual(registrationSecret, adminRegistrationSecret) {
+		opResult, err = controllerutil.CreateOrUpdate(s.Context, s.LocalClient, &registrationSecret, func() error {
+			// Get CA info from admin CA secret
+			registrationSecret.Data[mcconstants.AdminCaBundleKey] = adminCASecret.Data[mcconstants.AdminCaBundleKey]
+
+			// Get other registration info from admin registration secret for this managed cluster
+			registrationSecret.Data[mcconstants.ESURLKey] = adminRegistrationSecret.Data[mcconstants.ESURLKey]
+			registrationSecret.Data[mcconstants.RegistrationUsernameKey] = adminRegistrationSecret.Data[mcconstants.RegistrationUsernameKey]
+			registrationSecret.Data[mcconstants.RegistrationPasswordKey] = adminRegistrationSecret.Data[mcconstants.RegistrationPasswordKey]
+			registrationSecret.Data[mcconstants.KeycloakURLKey] = adminRegistrationSecret.Data[mcconstants.KeycloakURLKey]
```
desagar, line 85: From interactive review: also copy ES ca bundle here and in secret controller

platform-operator/controllers/clusters/sync_agent_secret.go
desagar: This was just a refactor - I moved the constant to a different package. It looks like it's the service account token. I can refactor if you feel strongly.

platform-operator/controllers/clusters/sync_registration_secret.go
desagar: These are the keys in the registration secret for the managed cluster.

platform-operator/controllers/clusters/vmc_controller_test.go
desagar: Yes, these are all declared in secret_keys.go which I moved out to the mcconstants package so that I can share it with app operator. This CaCrtKey constant for example, is here: https://github.com/verrazzano/verrazzano/blob/4d55b68209d045f587aeb8c29e5173c61b9ed731/pkg/mcconstants/secret_keys.go#L8

application-operator/mcagent/mcagent_clusterca.go
desagar: done - turned out this was needed for fluentd restart to work correctly anyway.

platform-operator/controllers/secrets/secrets_controller.go
desagar: This tls-additional-ca secret is used by the installer when using Let's Encrypt staging certs. It's not clear to me exactly why this is - @michael-cico knows more about it than me. My understanding is - this secret will be present for LE staging certs, and absent in other cases. In LE staging case, it should be considered the source of truth and we should ignore the verrazzano-tls secret.

I've tried to explain in the inline comment in the code, what the secret is used for.

michael-cico: We do create the additional secret for Rancher when LE/staging is configured.  It's a rancher hook for additional bundles for them to load.  When LE/staging is configured the generated cert secrets do not include a CA bundle, as they do with our self-signed/custom CA cases.

So we have to provide the LE staging CA bundle where we intend to use it.  I think @wmhopkins entered an enhancement to `cert-manager` to provide these bundles when LE/staging is in play, and store the required staging CAs in the cert secrets like they do in the self-signed case.  But until then, we have to provide it where needed.

We have talked about creating a cluster-scoped "source of truth" for these bundles, but that's out of the scope of this PR.

michael-cico (approved):

PR #3136, opened 2022-05-12
Relax OS validators

anders-swanson:
Update new VMO image that allows unhealthy node scale down and relax the validator. Validator will only check for argument duplicates now.


stevshap (approved):

kminder (approved):
Approved.

PR #3138, opened 2022-05-12
VZ-5819 Execute registry test in the upgrade-paths pipeline

ayaranga:
# Description

Also modified registry test to read the registry details from BOM for
non-private registry configurations.

Fixes VZ-5819

# Checklist 

### DIFF
--- a/tests/e2e/registry/registry_test.go
+++ b/tests/e2e/registry/registry_test.go
@@ -45,6 +45,7 @@ var listOfNamespaces = []string{
 	"verrazzano-install",
 	"verrazzano-mc",
 	"verrazzano-system",
+	"verrazzano-monitoring",
 }
 
 var t = framework.NewTestFramework("registry")
--- a/tests/e2e/registry/registry_test.go
+++ b/tests/e2e/registry/registry_test.go
@@ -4,10 +4,8 @@
 package registry
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
-	"os/exec"
 	"strings"
 	"time"
 
@@ -22,34 +20,12 @@ import (
 const (
 	waitTimeout     = 2 * time.Minute
 	pollingInterval = 10 * time.Second
-	// Pod Substring for finding the platform operator pod
-	platformOperatorPodNameSearchString = "verrazzano-platform-operator"
 )
 
 var registry = os.Getenv("REGISTRY")
-var imagePrefix = pkg.GetImagePrefix()
 
 // Map that contains the images present in the BOM with the associated registry URL
 var imageRegistryMap = make(map[string]string)
-
-// Struct based on Verrazzano BOM JSON
-type verrazzanoBom struct {
-	Registry   string `json:"registry"`
-	Version    string `json:"version"`
-	Components []struct {
-		Name          string `json:"name"`
-		Subcomponents []struct {
-			Registry   string `json:"registry"`
-			Repository string `json:"repository"`
-			Name       string `json:"name"`
-			Images     []struct {
-				Image            string `json:"image"`
-				Tag              string `json:"tag"`
-				HelmFullImageKey string `json:"helmFullImageKey"`
-			} `json:"images"`
-		} `json:"subcomponents"`
-	} `json:"components"`
-}
 
 // List of namespaces from which all the pods are queried to confirm the images are loaded from the target registry/repo
 var listOfNamespaces = []string{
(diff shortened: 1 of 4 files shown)


tests/e2e/registry/registry_test.go
```diff
@@ -68,14 +94,115 @@ var _ = t.Describe("Private Registry Verification", Label("f:platform-lcm.privat
...
+								fmt.Sprintf("FAIL: The image for the pod %s in initContainers, doesn't start with expected registry URL prefix %s, image name %s", pod.Name, registryURL, pod.Spec.InitContainers[k].Image))
 						}
 					}
 				}
 			})
 	})
+
+// getRegistryURL returns the private registry url if the private registry env is set
+// If private registry not set, the registry url is determined based on what is specified in the BOM
+func getRegistryURL(image string) (string, error) {
```
michael-cico, line 117: why not just use the `pkg.bom` code?

tests/e2e/registry/registry_test.go
```diff
@@ -20,9 +22,34 @@ import (
...
 )
 
 var registry = os.Getenv("REGISTRY")
+var imagePrefix = pkg.GetImagePrefix()
+
+// Map that contains the images present in the BOM with the associated registry URL
+var imageRegistryMap = make(map[string]string)
+
+// Struct based on Verrazzano BOM JSON
+type verrazzanoBom struct {
```
michael-cico, line 36: seems to duplicate the code in `pkg/bom`?

tests/e2e/registry/registry_test.go
```diff
@@ -68,14 +94,115 @@ var _ = t.Describe("Private Registry Verification", Label("f:platform-lcm.privat
...
+				}
+				imageRegistryMap[image.Image] = registry + "/" + subComponent.Repository
+			}
+		}
+	}
+	return nil
+}
+
+// Get the BOM from the platform operator in the cluster and build the BOM structure from it
+func getBOM(vBom *verrazzanoBom) error {
```
michael-cico, line 173: not necessarily for this pass, but maybe we should move this into the `tests/e2e/pkg` code to be shared?

You could also return `(*verrazzanoBOM, error)` instead of requiring the caller to pass in the struct, although I think that could also be just my own subjective view on it.  This way avoids a nil-check at least.

pshirodkar-orcl, line 173: This method is pretty much similar to the one in bom-validator. Instead of copy/pasting, can you please look into creating a public method `getBom` that can be called in both bom-validator and this test.

tests/e2e/registry/registry_test.go
pshirodkar-orcl, line 36: This struct is pretty much similar to the one in bom-validator. Instead of copy/pasting, can you please look into creating a public struct that can be called in both bom-validator and this test.

tests/e2e/registry/registry_test.go
```diff
@@ -68,14 +94,115 @@ var _ = t.Describe("Private Registry Verification", Label("f:platform-lcm.privat
...
+	}
+	globalRegistry := vBom.Registry
+	for _, component := range vBom.Components {
+		for _, subComponent := range component.Subcomponents {
+			for _, image := range subComponent.Images {
+				registry := globalRegistry
+				if len(subComponent.Registry) > 0 {
+					registry = subComponent.Registry
+				}
+				imageRegistryMap[image.Image] = registry + "/" + subComponent.Repository
```
pshirodkar-orcl, line 165: The registry and repository can also be overridden at the image level https://github.com/verrazzano/verrazzano/blob/1f7a53391e3319f3b50fd821ebeaba4d4974d4f1/pkg/bom/bom.go#L78. Can we please account for these fields as well? This will make the test robust if in case the `verrazzanon-bom.json` gets updated containing images that override registry and repository at their level in the future.

tests/e2e/registry/registry_test.go
```diff
@@ -68,14 +94,115 @@ var _ = t.Describe("Private Registry Verification", Label("f:platform-lcm.privat
...
+		if err != nil {
+			return "", err
+		}
+	}
+	// Remove the version from image
+	imageURL := strings.Split(image, ":")[0]
+	imageParts := strings.Split(imageURL, "/")
+	imageName := imageParts[len(imageParts)-1]
+	// If the image is not defined in the bom, return the default registry
+	if imageRegistryMap[imageName] == "" {
```
pshirodkar-orcl, line 134: Is this if condition executed when image belongs to, for example, example apps like hello-helidon etc which are not mentioned in bom?

ayaranga: The registry tests are run only against the Verrazzano system namespaces. This check is added mainly for the verrazzano-operator case where this image was present in the previous versions and not in 1.3. When upgrade happened, this pod was still present and the test failed in my local cluster. I will start a thread if this pod should have been deleted as part of the upgrade. So, just returned the default registry(ghcr.io) instead of failing.

tests/e2e/registry/registry_test.go
ayaranga: Yes, fixed in the latest commit.

tests/e2e/registry/registry_test.go
ayaranga: Refactored the code to use pkg/bom

tests/e2e/registry/registry_test.go
ayaranga: Refactored the code to use pkg/bom. The bom-validator changes can be taken up in a separate PR.

pshirodkar-orcl: Can you please file a ticket to account for bom-validator changes?

tests/e2e/registry/registry_test.go
```diff
@@ -68,14 +70,86 @@ var _ = t.Describe("Private Registry Verification", Label("f:platform-lcm.privat
...
+	for _, component := range bomDoc.Components {
+		for _, subComponent := range component.SubComponents {
+			registry := globalRegistry
+			if len(subComponent.Registry) > 0 {
+				registry = subComponent.Registry
+			}
+			repository := subComponent.Repository
+			for _, image := range subComponent.Images {
+				if len(image.Registry) > 0 {
+					registry = subComponent.Registry
```
pshirodkar-orcl, line 138: This should be `image.Registry` instead.

ayaranga: Yes, you are right. Copy-paste error. Fixed it.

pshirodkar-orcl (approved):

PR #3141, opened 2022-05-12
Convert so more mocks to fakes in the upgrade tests

mgianatagh:
# Description

Another batch of conversions from mock to fake

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/upgrade_test.go
+++ b/platform-operator/controllers/verrazzano/upgrade_test.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	oamapi "github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	//	oamcore "github.com/crossplane/oam-kubernetes-runtime/apis/core"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	"github.com/verrazzano/verrazzano/pkg/helm"
@@ -453,8 +454,7 @@ func TestUpgradeStartedWhenPrevFailures(t *testing.T) {
 				Generation: 2,
 				Finalizers: []string{finalizerName}},
 			Spec: vzapi.VerrazzanoSpec{
-				Version: "0.2.0",
-			},
+				Version: "0.2.0"},
 			Status: vzapi.VerrazzanoStatus{
 				State:      vzapi.VzStateReady,
 				Components: makeVerrazzanoComponentStatusMap(),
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -479,29 +441,21 @@ func TestUpgradeStartedWhenPrevFailures(t *testing.T) {
...
-	mock.EXPECT().
-		Get(gomock.Any(), types.NamespacedName{Namespace: namespace, Name: name}, gomock.Not(gomock.Nil())).
-		DoAndReturn(func(ctx context.Context, name types.NamespacedName, verrazzano *vzapi.Verrazzano) error {
-			verrazzano.TypeMeta = metav1.TypeMeta{
-				APIVersion: "install.verrazzano.io/v1alpha1",
-				Kind:       "Verrazzano"}
-			verrazzano.ObjectMeta = metav1.ObjectMeta{
-				Namespace:  name.Namespace,
-				Name:       name.Name,
+	_ = vzapi.AddToScheme(k8scheme.Scheme)
```
kminder, line 448: Seems like this setup code gets repeated often.  Are there subtle differences?  If not perhaps this should be made into a shared utility function probably just for this file.

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -625,49 +560,27 @@ func TestUpgradeCompleted(t *testing.T) {
...
+		rbac.NewClusterRoleBinding(&verrazzanoToUse, name, getInstallNamespace(), buildServiceAccountName(name)),
+	).Build()
 
 	config.TestProfilesDir = "../../manifests/profiles"
 	defer func() { config.TestProfilesDir = "" }()
 
 	// Create and make the request
 	request := newRequest(namespace, name)
-	reconciler := newVerrazzanoReconciler(mock)
+	reconciler := newVerrazzanoReconciler(c)
```
kminder, line 573: Some of these are left with almost no assertions.  Was there nothing in the mock EXPECT setup that actually checks some related behavior?  Honestly it doesn't really look like it but I thought I ask.

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -727,46 +633,23 @@ func TestUpgradeCompletedMultipleReconcile(t *testing.T) {
...
```
kminder, line 646: This test description says 
"THEN ensure a condition with type UpgradeCompleted is added"
where is that assertion?

platform-operator/controllers/verrazzano/upgrade_test.go
mgianatagh: If there was an assertion in the mock EXPECT, I did preserve it.  Otherwise I did not add any new assertions, I assumed the existing ones were sufficient.

platform-operator/controllers/verrazzano/upgrade_test.go
mgianatagh: Either the comments for the test were incorrect, or these condition checks were never added.  I will take a look through the tests I have converted so far.  Maybe there was some missing assertion checks in the tests.

platform-operator/controllers/verrazzano/upgrade_test.go
mgianatagh: it is repeated for most tests, some have variations.  I can add a utility function to handle all the common cases.

kminder (approved):
Approved.

PR #3142, opened 2022-05-13
VZ-5842 - Acceptance tests for DNS updates to OCI Public Zone

shashankkumarIITB:
# WIP
# Description

- Fixes VZ-5842
- Added acceptance tests for DNS updates from Wildcard DNS to OCI public zone
- Added acceptance tests for CertManager updates from self-signed to Let'sEncrypt
- Created the `post-install` Jenkinsfile to execute post-install VZ CR update tests
- Renamed files for existing DNS and CertManager tests 

# How the post-install pipeline behaves?
- The tests make use of `verrazzano-post-install-updates` pipeline.
- Select the type of cluster to be used for testing - `kind`or `oke`. Default is `kind`.
- Select the type of OCI DNS type - `GLOBAL` or `PRIVATE`. Default is `GLOBAL`.
- Select the type of ACME CertManager environment type - `staging` or `production`. Default is `staging`.
- Default Verrazzano `dev` profile with Wildcard DNS and default CertManager configuration is installed in the cluster along with some test applications.
- OCI DNS Zone of the specified type is created.
- Validation tests are run against the successful VZ installation.
- Updates are applied to the Verrazzano resource : 
1. Wildcard DNS -> OCI DNS (type as selected earlier)
2. Self-signed CertManager -> Let'sEncrypt ACME (environment type as specified earlier)
- Validation tests are run again to ensure that are changes are picked up by Verrazzano and the deployed applications.
- Cluster the deleted and the resources are cleaned up.

# Checklist 


ci/oke-ocidns/Jenkinsfile
```diff
@@ -56,6 +56,9 @@ pipeline {
         choice (description: 'Specifies  DNS scope. Values: GLOBAL, PRIVATE. Default: GLOBAL',name: 'DNS_SCOPE',
             // 1st choice is the default value
             choices: [ "GLOBAL","PRIVATE" ])
+        // TODO: Change default to 'no' for running DNS update tests
+        choice (description: 'Run DNS update tests - Wildcard DNS to OCI DNS', name: 'RUN_DNS_UPDATE_TESTS',
```
michael-cico, line 60: I think making it optional means we'd have to configure another instance of a periodic run for it, just to execute the update scenario.  It might save some resources to just always do it based on a particular scenario.

michael-cico: We could set this flag based on the condition that either `magicdns` or `kind` is configured, and still use it where needed for zone creation, etc.  That would mean that we always run an update scenario though in those cases.

ci/oke-ocidns/Jenkinsfile
```diff
@@ -447,6 +451,28 @@ pipeline {
...
+                runGinkgo('update/ocidns')
+            }
+            post {
+                always {
+                    archiveArtifacts artifacts: '**/coverage.html,**/logs/*,**/test-cluster-dumps/**', allowEmptyArchive: true
+                    junit testResults: '**/*test-result.xml', allowEmptyResults: true
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
shashankkumarIITB: Will have a look at that Jenkinsfile and make the changes.

ci/post-install-updates/Jenkinsfile
```diff
@@ -0,0 +1,1064 @@
+// Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
michael-cico, line 1: Since the intent of this is to replace the existing OKE/OCI DNS pipelines, we should perhaps name the directory for this just something like `oci-dns`?  

At some point we may also make this suite work with KIND, and we can remove those cases from the dynamic-config pipeline, so this pipeline will handle OCI DNS cases for all k8s platforms and OCI DNS configurations.

PR #3145, opened 2022-05-13
VZ 5882: Fire watch using flag

anders-swanson:
Reconciler needs to keep track of whether a watch has fired, to reconcile ready components.

### DIFF
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -6,8 +6,11 @@ package verrazzano
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/istio"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/keycloak"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	vzctrl "github.com/verrazzano/verrazzano/pkg/controller"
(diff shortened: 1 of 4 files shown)


mgianatagh (dismissed):

platform-operator/controllers/verrazzano/install.go
```diff
@@ -173,6 +174,10 @@ func checkConfigUpdated(ctx spi.ComponentContext, componentStatus *vzapi.Compone
 	if componentStatus.ReconcilingGeneration > 0 {
 		return ctx.ActualCR().Generation > componentStatus.ReconcilingGeneration
 	}
+	// The component should be reconciled if it is watched
+	if r.IsWatchedComponent(name) {
```
pfmackin, line 178: I think this needs to be before the previous block where we check generation, otherwise it will never execute if the component generation is > 0

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -1226,6 +1232,7 @@ func (r *Reconciler) watchJaegerService(namespace string, name string, log vzlog
 			service := e.Object.(*corev1.Service)
 			if service.Labels[vzconst.KubernetesAppLabel] == vzconst.JaegerCollectorService {
 				log.Debugf("Jaeger service %s/%s created", service.Namespace, service.Name)
+				r.AddWatch(istio.ComponentJSONName)
```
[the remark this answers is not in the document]
anders-swanson: Yes, I want to reconcile istio if jaeger was installed

platform-operator/controllers/verrazzano/install.go
anders-swanson: Yep, you are right

pfmackin (approved):

PR #3168, opened 2022-05-17
Tvlatas/agent oke region lock

agv426:
# Description

This change will lock all jobs running in a pipeline to run in the same region (any AD in the region). The initial region selected to run in will be propagated along to all triggered jobs.

This also will have the OKE cluster run in the same region as the Jenkins runner.

That will limit us to running in regions where we have Jenkins runners. These are the large 3-AD regions: IAD, PHX, LHR, and FRA.

We will not be using the smaller remote regions for OKE clusters after this change.

# Checklist 


ci/private-registry/Jenkinsfile
```diff
@@ -28,6 +33,7 @@ pipeline {
     }
 
     parameters {
+        choice (description: 'OCI region to run Jenkins agent and OKE clusters in', name: 'LOCK_REGION', choices: availableRegions)
```
kminder, line 36: Should the description indicate the value is actually ignored?

tests/e2e/config/scripts/terraform/cluster/create-cluster.sh
```diff
@@ -29,6 +29,8 @@ done
 # retry 3 times, 30 seconds apart
 tries=0
 MAX_TRIES=3
+export TF_LOG=DEBUG
```
kminder, line 32: Are these intentional?

kminder (approved):
Approved.

ci/private-registry/Jenkinsfile
agv426: Yeah I can update it. Was debating whether to add it now or not. Intention is that eventually we can honor it, when we look at per-region registry support and where the related jobs are run (we want to do this in parallel with other test jobs in the periodics but some release related stuff needs to be worked out)

tests/e2e/config/scripts/terraform/cluster/create-cluster.sh
agv426: yeah, we're tracking down an OKE cluster creation issue that mostly seems to impact OCI integration tests but is intermittent in other tests. The debug helps to tell if it is this issue or something else (and also we need the debug for the service & TF folks)

stevshap (approved):
refreshing kevin's review

PR #3265, opened 2022-05-27
VZ-5909 Add Values field to overrides

mbenmclean:
This change adds a values field to the install overrides. This field enables the user to provide inline YAML so they don't have to provide a configmap or secret to configure overrides. This change also extends the overrides test in the dynamic config tests to test this new value.

# Checklist 


tests/e2e/update/overrides/overrides_test.go
```diff
@@ -206,6 +224,9 @@ var _ = t.Describe("Post Install Overrides", func() {
 })
 
 func deleteOverrides() {
+	m := PrometheusOperatorDefaultModifier{}
+	update.UpdateCR(m)
```
adiforluls, line 228: So this will set overrides to nil, now deleting cm/secret won't have any effect as they aren't referenced in CR anymore. So maybe modify the associated func to not nil out the overrides if removing inline values, and nil out overrides when it's invoked in AfterSuite, or use a separate modifier to only delete inline values.

mbenmclean: Ok I will change

mbenmclean: Fixed

davisgomes (approved):

PR #3273, opened 2022-05-31
Scchen/vz 6014 admin cm update

shihchang:
# Description
- test update cert-manager config of admin-cluster
- assert ca secret sync and managed Fluentd
- make upgrade stages optional

Fixes VZ-6014

# Checklist 

### DIFF
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -460,9 +460,11 @@ pipeline {
                         runGinkgoAdmin('update/fluentdextes')
                     }
                 }
-                stage("verify-managed-cluster-cert-update") {
+                stage("verify-cert-manager-update") {
                     steps {
-                        runGinkgoAdmin('update/certmc')
+                        script {
+                            parallel verifyCertManagerUpdate()
+                        }
                     }
                 }
                 stage('verify-install') {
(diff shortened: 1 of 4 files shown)


ci/multicluster/Jenkinsfile
```diff
@@ -418,11 +419,13 @@ pipeline {
                     }
                 }
                 stage("upgrade-platform-operator") {
+                    when { expression { return params.UPGRADE_VERRAZZANO == true } }
```
michael-cico, line 422: I'm not sure this will work; don't we install from an earlier version in the install stages?

shihchang: Yes, it is in the getVerrazzanoOperatorYaml() method https://github.com/verrazzano/verrazzano/blob/scchen/VZ-6014_admin_cm_update/ci/multicluster/Jenkinsfile#L762 when [ "true" == "${params.EXTERNAL_ELASTICSEARCH}" ] || [ "false" == "${params.UPGRADE_VERRAZZANO}" ] it will install the operator built from the commit which does not need upgrade. Otherwise (UPGRADE_VERRAZZANO=true & EXTERNAL_ELASTICSEARCH=false) it will install the earlier release

PR #3288, opened 2022-06-02
VZ-6020: Remove "old" node exporter

fredtibbitts:
# Description

This PR removes the installation of the "old" node exporter from the Verrazzano helm chart. There is a new Prometheus Operator node exporter that replaces it.

Note that prior to this change, the new node exporter was disabled by default and had to be explicitly enabled. Now the new node exporter is enabled by default if the Prometheus component is enabled, so that we preserve the previous behavior.

This PR will not be merged until the work to fully migrate over to the new Prometheus Operator stack is complete.

Implements VZ-6020

# Checklist 


platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -42,10 +42,12 @@ const (
 	tmpFileCreatePattern = tmpFilePrefix + "*." + tmpSuffix
 	tmpFileCleanPattern  = tmpFilePrefix + ".*\\." + tmpSuffix
 
-	fluentDaemonset       = "fluentd"
-	nodeExporterDaemonset = "node-exporter"
+	fluentDaemonset = "fluentd"
 
 	prometheusDeployment = "vmi-system-prometheus-0"
+
+	monitoringNamespace = "monitoring"
```
jrosinsk, line 49: Will the monitoring namespace continue to exist after we move to the new namespace?  if so, we might want to make this a constant, but I guess this is only needed for the node-exporter cleanup.

pkg/constants/constants.go
```diff
@@ -56,9 +56,6 @@ const KeycloakNamespace = "keycloak"
 // RancherSystemNamespace - the Rancher cattle-system namespace
 const RancherSystemNamespace = "cattle-system"
 
-// VerrazzanoMonitoringNamespace - the keycloak namespace
-const VerrazzanoMonitoringNamespace = "monitoring"
-
```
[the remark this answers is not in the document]
fredtibbitts: There is already another constant for `VerrazzanoMonitoringNamespace = verrazzano-monitoring` in `platform-operator/pkg/constants/constants.go`, and having the two is confusing, so I removed this one since it's no longer needed and added a package-local constant so it's just defined where it's needed.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
fredtibbitts: I chose not to delete the `monitoring` namespace, since it's a fairly generic name and a customer may have their own resources there. Verrazzano will no longer do anything with the `monitoring` namespace, except for this code to clean up node exporter, as you point out.

fredtibbitts:
Closed since the work was merged in https://github.com/verrazzano/verrazzano/pull/3456.

PR #3300, opened 2022-06-06
VZ-6006 monitorChanges skips all component updates

mbenmclean:
This change adds logic in the component install that will skip the install for a component during update if monitorChanges is set to false. Previously, monitorChanges set to false would skip updates to the ConfigMaps and Secrets referenced in the overrides. This extends this functionality to inline value overrides, install args, and any other component spec updates.

This change also fixes a bug where after an upgrade, and update will kick off another upgrade because of a bug in comparing the spec and status version strings. This was fixed by using the SemVer CompareTo function to compare the two version as semver objects.

Fixes VZ-6006

# Checklist 


kminder (approved):
Approved.

PR #3301, opened 2022-06-06
VZ-6056 multicluster dns update tests

shihchang:
# Description
Multicluster DNS update tests
- test admin-cluster dns update, assert managed Fluentd OpenSearch url
- test managed-cluster dns update, assert admin Prometheus scrapeTargests

Fixes VZ-6056

# Checklist 

### DIFF
--- a/ci/multicluster/Jenkinsfile
+++ b/ci/multicluster/Jenkinsfile
@@ -464,6 +464,13 @@ pipeline {
                         runGinkgoAdmin('update/fluentdextes')
                     }
                 }
+                stage("verify-dns-update") {
+                    steps {
+                        script {
+                            parallel verifyDNSUpdate()
+                        }
+                    }
+                }
                 stage("verify-cert-manager-update") {
                     steps {
                         script {
@@ -1527,6 +1534,17 @@ def verifyUpgrade() {
         },
         "Verify Metrics": {
             runGinkgoRandomize('metrics/syscomponents')
+        },
+    ]
+}
+
+def verifyDNSUpdate() {
+    return [
+        "verify-admin-cluster-dns-update": {
+            runGinkgoAdmin('update/dnsac')
+        },
+        "verify-managed-cluster-dns-update": {
+            runGinkgoAdmin('update/dnsmc')
         },
     ]
 }
--- a/tests/e2e/update/dnsac/dnsac_test.go
+++ b/tests/e2e/update/dnsac/dnsac_test.go
@@ -19,11 +19,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-const (
-	domainN = "nip.io"
-	domainS = "sslip.io"
-)
-
 var (
 	t               = framework.NewTestFramework("update dns")
 	adminCluster    *multicluster.Cluster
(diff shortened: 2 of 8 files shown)


ci/multicluster/Jenkinsfile
```diff
@@ -1531,6 +1538,17 @@ def verifyUpgrade() {
     ]
 }
 
+def verifyDNSUpdate() {
+    return [
```
michael-cico, line 1542: Do we leave the sample app up and running during the updates?  Or do we only run the sample app tests after the install/updates?

ci/multicluster/Jenkinsfile
```diff
@@ -464,6 +464,13 @@ pipeline {
                         runGinkgoAdmin('update/fluentdextes')
                     }
                 }
+                stage("verify-dns-update") {
```
michael-cico, line 467: So the flow of this whole suite is

- install and/or upgrade
- run some updates
- run verify/app tests

Is that right?  Wondering if we need to be able to run some verify tests before the updates to verify the initial install.  Doesn't have to be this pass though.

ci/multicluster/Jenkinsfile
shihchang: Yes, all the update tests run after the deployment of all sample apps but before sample app tests. Sample apps are deployed and tested in two different stages and all the update tests are between those two stages: https://build.verrazzano.io/job/verrazzano-multi-cluster-acceptance-tests/job/scchen%252FVZ-6056_multicluster_dns_update/

michael-cico: when I look at a Blue Ocean view of the run, it looks like the app tests only run after the updates occur... I don't see them getting deployed before the updates?

https://build.verrazzano.io/blue/organizations/jenkins/verrazzano-multi-cluster-acceptance-tests/detail/scchen%2FVZ-6056_multicluster_dns_update/9/pipeline

tests/e2e/update/dnsac/dnsac_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/verrazzano/verrazzano/pkg/test/framework"
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/tests/e2e/multicluster"
+	"github.com/verrazzano/verrazzano/tests/e2e/pkg"
+	"github.com/verrazzano/verrazzano/tests/e2e/update"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	domainN = "nip.io"
```
michael-cico, line 23: might be more descriptive to call these `nipDomain`/`sslipDomain`, `domainN` I kinda figured, but had to go verify.

tests/e2e/update/dnsac/dnsac_test.go
```diff
@@ -0,0 +1,159 @@
...
+})
+
+// updateAdminClusterDNS switch dns config to sslip.io or nip.io
+func updateAdminClusterDNS() string {
+	adminVZ := adminCluster.GetCR(true)
+	var oldDNS = adminVZ.Spec.Components.DNS
+	var newDNS *vzapi.DNSComponent = nil
+	var domainOld, domainNew = domainN, domainS
+	oldEsIng := pkg.GetSystemOpenSearchIngressURL(adminCluster.KubeConfigPath)
+	if isDefaultDNS(oldDNS) {
```
michael-cico, line 88: so all of this checking is because we can't be guaranteed of the starting point right?  nip.io vs. sslip.io?

ci/multicluster/Jenkinsfile
shihchang: There is a 'Deploy Sample Application' stage https://github.com/verrazzano/verrazzano/blob/master/ci/multicluster/Jenkinsfile#L406 runs before upgrade and update tests; that deploySampleApp stage deploy the multicluster/hello-helidon-sample https://github.com/verrazzano/verrazzano/blob/master/ci/multicluster/Jenkinsfile#L1453

tests/e2e/update/dnsac/dnsac_test.go
shihchang: fixed

tests/e2e/update/dnsac/dnsac_test.go
shihchang: The pipeline currently always start with nip.io. This test however is able to switch between nip and sslip. If the test is started with nip, it switch to sslip, verify the state propagation, and then switch back to nip. If the pipeline is changed to start with sslip, it should continue to work.

PR #2863, opened 2022-04-05
JIRA: VZ-5488 support upgrades with ephemeral storage or authproxy transition 

jmaron99:
# Description

This PR addresses two upgrade scenarios:
- An upgrade with ephemeral storage from VZ 1.0.x to 1.3+
- An upgrade from authproxy as resource in VZ chart to a version in which authproxy is installed by a separate chart

Specific Modifications:
- VZ component disassociates authproxy in upgrade code prior to proceeding with upgrading.  It also marks the authproxy associated resources as resources to keep (rather than delete)
- The Authproxy component removes the "keep" resource policy annotation since it wants to manage those resources in all instances (install and upgrade)
- VZ component modified to only run the ES post upgrade test if non-ephemeral storage in use (ephemeral storage during upgrade prevents the retention of the index between upgrades)

Fixes VZ-5488

# Checklist 

As the author of this PR, I have:

- [x] Checked that I included or updated copyright and license notices in all files that I altered
- [x] Added or updated unit tests for any new functions I added
- [ ] Added or updated integration tests if appropriate
- [x] Added or updated acceptance tests if appropriate

Code reviewer, please confirm this PR:

- [ ] Addressed the requirement and meets the acceptance criteria
- [ ] Does not introduce unrelated or spurious changes
- [ ] Does not introduce any unapproved dependency
- [ ] Makes sense and it easy to understand, and/or difficult areas of code are clearly documented so that they can be understood

PR #3305, opened 2022-06-06
VZ-5035: Avoid calls to pre-install in a reconcile loop when there is a failure during install

anajosep:
# Description

Pre-install is called during installation even after the component is installed. Fix is to have an additional state in the state transistion, so that pre-install does not execute again when there is an error in the install phase.

Fixes VZ-5035

# Checklist 


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -177,6 +177,9 @@ const (
 	// CondPreInstall means an install about to start.
 	CondPreInstall ConditionType = "PreInstall"
 
+	// CondPreInstallComplete means an install about to start.
```
pfmackin, line 180: means pre-install has completed.

Note the PreInstall condition above has "means an install is about to start"

anajosep: Corrected the code comment as suggested.
`
// CondPreInstall means pre-install is in progress.
	
// CondPreInstallComplete means pre-install has completed.
`

michael-cico: General question, is the real problem that some of the pre-install impls are not idempotent?  I think that moving from Pre-install to Installing is about the same as having a "preinstall complete" state.  If something happens while trying to set "Preinstall Complete" will we still end up back in the PreInstall() call?

anajosep: I think the idea is to have distinct states for pre-install, install and post-install phases in the state machine. Ideal thing would be to ensure that each of these phases are idempotent, but if in case, they are not, this design would give some sort of protection. Also, the chances of hitting an error during install phase is more compared to when updating the status in VZ CR.

michael-cico: We might still be in the same boat; but I'm guessing you've seen an improvement with this?

I think if we transition from pre-install --> installing, it's not much different than adding this new state.  In both of these cases, we end up with potential for the status not updating properly and re-entering things call, and same with all the other state-transitions/calls.  Or is my thinking off here?

michael-cico: Just read the bug (probably should have done that first).  I think this will end up in the same situation, the status update can still fail entering the new state right?  and we'll be back into the call. 

Might be better to fix the authproxy call, is that possible?

anajosep: Authproxy Preinstall function is made idempotent by [PR-2863](https://github.com/verrazzano/verrazzano/pull/2863/files#diff-827454a8c1f40bedad959d33dd8913d28d524b44ba8ea0dc121d508440ceb109) which was merged on 6th April 2022

Since the root cause of the issue was already fixed, I feel there is no need to introduce a new state in the installer state machine to solve the same problem. I am planning to close this PR and the bug (as not reproducible)

PR #3317, opened 2022-06-07
VZ-6049 - Refactor scrape config editing into a separate package so that it can be shared

desagar:
# Description

Fixes VZ-9999

# Checklist 

### DIFF
--- a/pkg/scrapeconfigutils/edit_scrapeconfig.go
+++ b/pkg/scrapeconfigutils/edit_scrapeconfig.go
@@ -1,3 +1,6 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
 package scrapeconfigutils
 
 import (
--- a/pkg/scrapeconfigutils/edit_scrapeconfig.go
+++ b/pkg/scrapeconfigutils/edit_scrapeconfig.go
@@ -1,0 +1,51 @@
+package scrapeconfigutils
+
+import (
+	"github.com/Jeffail/gabs/v2"
+	"sigs.k8s.io/yaml"
+)
+
+const jobNameKey = "job_name"
+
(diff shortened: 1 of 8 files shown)


pkg/scrapeconfigutils/edit_scrapeconfig.go
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package scrapeconfigutils
+
+import (
+	"github.com/Jeffail/gabs/v2"
+	"sigs.k8s.io/yaml"
+)
+
+const jobNameKey = "job_name"
```
fredtibbitts, line 11: This constant is also defined in `sync_prometheus.go`, should we move it to the global `constants.go` so it's defined in one place?

pkg/scrapeconfigutils/edit_scrapeconfig.go
```diff
@@ -0,0 +1,54 @@
...
+
+// EditScrapeJob edits a scrape config that and adds or replaces the specified job with the new scrape
```
fredtibbitts, line 13: Nit, wording: "that and adds"

fredtibbitts (approved):
Just a couple of very minor things, otherwise LGTM.

pkg/scrapeconfigutils/edit_scrapeconfig.go
desagar: Yes, I agree. I didn't want to take on that refactoring here because those are not the only 2 spots - it's hard-coded in a number of places and would affect a lot of files. 

The VerrazzanoMonitoringNamespace constant is also defined in 2 places - I'll see if I can do a separate PR for just these refactors.

pkg/scrapeconfigutils/edit_scrapeconfig.go
desagar: thanks, will fix in next PR since I likely will be in that code anyway.

PR #3330, opened 2022-06-09
No Jira: fix MC External ES upgrade

mbenmclean:
Previously the MC External ES test didn't actually upgrade, but hacked an upgrade by comparing version string v1.2.3 and 1.2.3. I fixed this so now version strings v1.2.3 and 1.2.3 will equivalent. If UPGRADE_VERRAZZANO is enabled now for MC External ES, the pipeline will fail. This change has MC External ES upgrade from v1.2.0 when UPGRADE_VERRAZZANO is enabled.

### DIFF
--- a/ci/JenkinsfileTestTrigger
+++ b/ci/JenkinsfileTestTrigger
@@ -165,7 +165,8 @@ pipeline {
                                         string(name: 'INCLUDED_TESTS', value: params.INCLUDED_TESTS),
                                         string(name: 'EXCLUDED_TESTS', value: params.EXCLUDED_TESTS),
                                         booleanParam(name: 'EMIT_METRICS', value: params.EMIT_METRICS),
-                                        string(name: 'CONSOLE_REPO_BRANCH', value: params.CONSOLE_REPO_BRANCH)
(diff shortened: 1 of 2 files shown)


ci/JenkinsfileTestTrigger
```diff
@@ -165,7 +165,8 @@ pipeline {
                                         string(name: 'INCLUDED_TESTS', value: params.INCLUDED_TESTS),
                                         string(name: 'EXCLUDED_TESTS', value: params.EXCLUDED_TESTS),
                                         booleanParam(name: 'EMIT_METRICS', value: params.EMIT_METRICS),
-                                        string(name: 'CONSOLE_REPO_BRANCH', value: params.CONSOLE_REPO_BRANCH)
+                                        string(name: 'CONSOLE_REPO_BRANCH', value: params.CONSOLE_REPO_BRANCH),
+                                        string(name: 'VERSION_FOR_INSTALL', value: 'v1.2.0'),
```
michael-cico, line 169: how does this change handle both the `v1.2.0` and `1.2.0` case?  i.e, if that was a problem, if someone manually enters "1.2.0" it still works?

PR #3340, opened 2022-06-09
VZ-6083: Add -f argument for vz install command

bobdonat:
# Description

This pull request adds the -f argument to vz install command. This allows a user to specify an ordered list of yaml files containing vz custom resources.

# Checklist 


tools/vz/cmd/helpers/overlay.go
```diff
@@ -0,0 +1,80 @@
...
+		vzYaml, err = overlayVerrazzano(vzYaml, string(readBytes))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	vz := &vzapi.Verrazzano{}
+	err := yaml.Unmarshal([]byte(vzYaml), &vz)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unmarshal yaml into a verrazzano resource: %s", err.Error())
```
mgianatagh, line 39: The end user may not understand what this means.  Maybe the error should say something along the lines that there were errors combining the files passed on the command line.

tools/vz/cmd/helpers/overlay.go
```diff
@@ -0,0 +1,80 @@
...
+func overlayVerrazzano(baseYAML string, overlayYAML string) (string, error) {
+	if strings.TrimSpace(baseYAML) == "" {
+		return overlayYAML, nil
+	}
+	if strings.TrimSpace(overlayYAML) == "" {
+		return baseYAML, nil
+	}
+	baseJSON, err := yaml.YAMLToJSON([]byte(baseYAML))
+	if err != nil {
+		return "", fmt.Errorf("YAMLToJSON error in base: %s\n%s", err, baseJSON)
```
mgianatagh, line 61: another instance where this may not have much meaning to the end user.  I suggest taking a look at each new error being returned that would be printed out on the console and see if they can be more helpful to the end user.

tools/vz/cmd/install/install.go
```diff
@@ -51,8 +51,12 @@ vz install -f base.yaml -f custom.yaml --set profile=prod --log-format json`
 	verrazzanoPlatformOperatorWait = 1
 )
 
+var vpoWaitRetries = 60
 var logsEnum = cmdhelpers.LogFormatSimple
 
+// Use with unit testing
+func resetVpoWaitRetries() { vpoWaitRetries = 60 }
```
mgianatagh, line 58: I wonder if we could use cobra command options that are hidden to accomplish this instead?  I am not sure if that approach would be more of a hack than using this function.  At runtime we could have the test dynamically add a flag to the command object that contains the vpo wait retry count, and have the install code use it if found.  The command option would never be useable by an end user because it would only exist for the duration of the test.

mgianatagh (approved):
left some feedback that could be handled in the next follow-up PR

tools/vz/cmd/helpers/overlay.go
bobdonat: Would something like this be better?
"Failed to convert yaml to json in ..."

tools/vz/cmd/helpers/overlay.go
bobdonat: Would this be better?
"Failed to create a verrazzano install resource: %s"

tools/vz/cmd/install/install.go
bobdonat: seems like that could introduce a possible failure in the install code which is not possible with the current hack

stevshap (dismissed):

tools/vz/cmd/helpers/overlay.go
mgianatagh: maybe the same error as you suggested above "Failed to create a verrazzano install resource: %s"

tools/vz/cmd/install/install.go
mgianatagh: ok

mgianatagh (approved):

tools/vz/cmd/helpers/overlay.go
bobdonat: Fixed

PR #3383, opened 2022-06-16
JIRA: VZ-6180 check for component enablement before installing associated resources

jmaron99:
# Description

Component helm charts were installing resources from other components without any regard for whether those other components are actually installed.

Fixes VZ-6180

# Checklist 


platform-operator/helm_config/charts/verrazzano-authproxy/templates/monitors.yaml
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-
+  {{- if .Values.prometheusEnabled }}
```
pfmackin, line 3: I think this means the ServiceMonitor resource will never be create if you install without prometheus, then later install prometheus.  Even so, this is an improvement.

pfmackin (approved):

platform-operator/helm_config/charts/verrazzano-authproxy/templates/monitors.yaml
jmaron99: actually - it's a good catch.  that should be "prometheusOperator" since it's the resource that understands ServiceMonitor...

pfmackin (approved):

michael-cico (approved):

PR #3410, opened 2022-06-22
updated olcne cr and renamed it

bhavaniravichandran:
When installing verrazzano on olcne env in OCI, found that the CR that the public doc links to need to be updated. This PR updates intsall-olcne.yaml . The file is renamed to use the newer ocne name.

### DIFF
--- a/platform-operator/config/samples/install-ocne.yaml
+++ b/platform-operator/config/samples/install-ocne.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
+# Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 #
 # This install resource sample installs the "dev" profile for Verrazzano on an OLCNE cluster.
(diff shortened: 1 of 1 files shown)


platform-operator/config/samples/install-ocne.yaml
```diff
@@ -26,19 +32,15 @@ spec:
...
+        type: NodePort
+        ports:
+        - name: https
+          port: 443
+          nodePort: 32443
+          protocol: TCP
+          targetPort: 8443
       istioInstallArgs:
       - name: gateways.istio-ingressgateway.externalIPs
         valueList:
```
guoyongzhang, line 46: same as the other IP

platform-operator/config/samples/install-ocne.yaml
```diff
@@ -1,21 +1,27 @@
-# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
guoyongzhang, line 1: copyright should stay with 2020, 2022

platform-operator/config/samples/install-ocne.yaml
```diff
@@ -1,21 +1,27 @@
...
-  name: my-olcne-install
+  name: my-ocne-install
 spec:
-  environmentName: uni
+  environmentName: myenv
   profile: dev
   components:
     dns:
       external:
         suffix: abc.def.xyz.com
```
guoyongzhang, line 16: In the OLCNE docs, the example for domain is example.com.  We should ebeb consistent here too.

guoyongzhang (approved):

PR #3431, opened 2022-06-27
Add uninstall state machine

pfmackin:
Add top-level and component level state machine for uninstall.   

# Checklist 

### DIFF
--- a/platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
+++ b/platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
@@ -148,6 +148,18 @@ func (g grafanaComponent) PostInstall(ctx spi.ComponentContext) error {
 	return common.CheckIngressesAndCerts(ctx, g)
 }
 
+func (g grafanaComponent) PreUninstall(context spi.ComponentContext) error {
+	return nil
+}
+
+func (g grafanaComponent) Uninstall(context spi.ComponentContext) error {
+	return nil
+}
+
+func (g grafanaComponent) PostUninstall(context spi.ComponentContext) error {
+	return nil
+}
+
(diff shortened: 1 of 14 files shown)


platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
```diff
@@ -148,6 +148,18 @@ func (g grafanaComponent) PostInstall(ctx spi.ComponentContext) error {
 	return common.CheckIngressesAndCerts(ctx, g)
 }
 
+func (g grafanaComponent) PreUninstall(context spi.ComponentContext) error {
```
michael-cico, line 151: Should we just have a single uninstall method?  or is it too complicated to avoid calling Helm twice for an uninstall, etc.

platform-operator/controllers/verrazzano/uninstall_component.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// componentUninstallContext has the Uninstall context for a Verrazzano component Uninstall
+type componentUninstallContext struct {
+	state componentUninstallState
+}
+
+// UninstallComponents will Uninstall the components as required
+func (r *Reconciler) UninstallComponents(log vzlog.VerrazzanoLogger, cr *installv1alpha1.Verrazzano, tracker *UninstallTracker) (ctrl.Result, error) {
+	spiCtx, err := spi.NewContext(log, r.Client, cr, r.DryRun)
```
michael-cico, line 49: Should we declare an "uninstall" operation that we can set on the compcontext, for logging?

pfmackin: i added one but looks like i didn't use it.  Let me fix that

PR #3447, opened 2022-06-28
VZ-5850: Velero component onboarding

theabmitra:
Velero operator needs to be installed with verrazzano at install time
This task will track new component api spec that needs to be introduced to install this
This spec will allow users flexibility to enable/disable the component


platform-operator/controllers/verrazzano/component/velero/velero_installer.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	"go.uber.org/zap"
+	"io"
+	"os"
+	"os/exec"
+	"time"
+)
+
+type bashCommand struct {
+	Timeout      time.Duration `json:"timeout"`
```
michael-cico, line 17: Is this configurable?

platform-operator/controllers/verrazzano/component/velero/velero_installer.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type VeleroImage struct {
+	VeleroImage                    string `json:"velero"`
+	VeleroPluginForAwsImage        string `json:"velero-plugin-for-aws"`
+	VeleroResticRestoreHelperImage string `json:"velero-restic-restore-helper"`
+}
+
+// Generic method to execute shell commands
+func genericRunner(bcmd *bashCommand, log vzlog.VerrazzanoLogger) *runnerResponse {
```
michael-cico, line 35: might be good to consolidate this with the other runner utilities up under the main `pkg` directories.    This is general enough that we may want to use it with our `runner` package.

Not necessarily for this pass however.

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func getValueFromResourceReqs(rq *v1.ResourceRequirements, resourceCategory, resourceType string) string {
+	valueNotConfigured := true
+	switch strings.ToLower(resourceCategory) {
+	case "limits":
+		if rq.Limits != nil {
+			switch resourceType {
+			case "cpu":
+				if rq.Limits.Cpu() != nil {
```
michael-cico, line 77: might be more concise to just initialize with the default value and then replace that if configured, e.g.,

```
value := getDefaultValues(...)
if rq.Limits.Cpu() != nil {
   value = rq.Limits.Cpu.String()
}
return value
```

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
```diff
@@ -0,0 +1,251 @@
...
+func componentInstall(ctx spi.ComponentContext) error {
+
+	args, err := buildInstallArgs()
+	if err != nil {
+		ctx.Log().Errorf("Unable to build installargs %v", zap.Error(err))
+		return err
+	}
+	var vcmd bashCommand
+	var veleroInstallResponse *runnerResponse
+	vcmd.Timeout = time.Second * 600
```
michael-cico, line 167: I think one thing with the timeout is it's arbitrary; 10 mins seems like a lot, but under some conditions it may not be.  

How long do we typically expect this to take?

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
```diff
@@ -0,0 +1,251 @@
...
+	bcmd = append(bcmd, veleroBin, installCli)
+	bcmd = append(bcmd, veleroImageCli, args.VeleroImage)
+	bcmd = append(bcmd, pluginImageCli, args.VeleroPluginForAwsImage)
+	bcmd = append(bcmd, resticCli, volSnapshotEnableCli, nosecret, noDefaultBackup)
+	bcmd = append(bcmd, veleroCPURequestCmd, veleroCPULimitCmd, veleroMemRequestCmd, veleroMemLimitCmd)
+	bcmd = append(bcmd, resticCPURequestCmd, resticCPULimitCmd, resticMemRequestCmd, resticMemLimitCmd)
+	vcmd.CommandArgs = bcmd
+
+	if os.Getenv("DEV_TEST") != "True" {
+		veleroInstallResponse = genericRunner(&vcmd, ctx.Log())
```
michael-cico, line 188: Might be better to follow the other conventions for unit testing and allow replacing the real `genericRunner` with a test version, and then let the test reset it, rather than env-checks.  I'll defer to @pfmackin and @kminder on this one.

I've tended to also lean towards using interfaces, although it's hair-splitting.  There's a `runner` package we use for Helm and Istioctl we might want to merge your impl with.  The timeout is nice, but I think without configurability it might not be useful.

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
```diff
@@ -0,0 +1,251 @@
...
+
+	return nil
+}
+
+func buildInstallArgs() (VeleroImage, error) {
+	args := map[string]interface{}{
+		"namespace": constants.VeleroNameSpace,
+	}
+	var vi VeleroImage
+	bomFile, err := bom.NewBom(config.GetDefaultBOMFilePath())
```
michael-cico, line 213: we have a package for querying the BOM, in `pkg.bom`.  You can mirror the HelmComponent code that uses that, or possibly we can refactor some of that code so it can be shared if necessary.

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
```diff
@@ -0,0 +1,164 @@
...
+func (v veleroComponent) GetJSONName() string {
+	return ComponentJSONName
+}
+
+// GetHelmOverrides returns the Helm override sources for a component
+func (v veleroComponent) GetOverrides(_ *vzapi.Verrazzano) []vzapi.Overrides {
+	return []vzapi.Overrides{}
+}
+
+// MonitorOverrides indicates whether monitoring of Helm override sources is enabled for a component
```
michael-cico, line 111: Is there anything we need to do here to support external overrides?

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
```diff
@@ -0,0 +1,164 @@
...
+func (v veleroComponent) ValidateUpdate(_, _ *vzapi.Verrazzano) error {
+	return nil
+}
+
+func (v veleroComponent) PreUninstall(_ spi.ComponentContext) error {
+	return nil
+}
+
+func (v veleroComponent) Uninstall(_ spi.ComponentContext) error {
+	return nil
```
michael-cico, line 155: this will need to get implemented eventually, but you may need to pick up Paul's recent fix that requires components to opt-in to the newer (in-development) VPO uninstall hooks.

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (v veleroComponent) PostUninstall(_ spi.ComponentContext) error {
+	return nil
+}
+
+func (v veleroComponent) IsOperatorUninstallSupported() bool {
+	return true
```
michael-cico, line 163: actually, I would make this `false` for now until we shake out the new uninstall work in progress.

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
michael-cico: nm, looks like from below you picked it up already.

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
michael-cico: Looks like we're not implementing it yet, which is fine.

platform-operator/internal/vzconfig/enabled.go
```diff
@@ -179,3 +179,11 @@ func IsApplicationOperatorEnabled(vz *vzapi.Verrazzano) bool {
 	}
 	return true
 }
+
+// IsVeleroEnabled returns false only if Velero Operator is explicitly disabled in the CR
```
michael-cico, line 183: Comment is wrong, in this case it returns `true` only when explicitly **enabled** in the CR

platform-operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
```diff
@@ -232,6 +232,11 @@ function delete_fluentd {
   fi
 }
 
+function delete_velero {
+  log "Uninstall velero"
+  velero uninstall --namespace velero --force || err_return $? "Could not delete Velero"
```
michael-cico, line 237: Just curious, does this delete their CRDs also?  Or just the deployment resources and configmap?

platform-operator/thirdparty/manifests/velero/velero-configmap.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
michael-cico, line 1: this might be better under `platform-operator/manifests`.  We started that a while ago with the intent of moving all charts, etc under there.  Plus, is this our own file we created, or something we got from Velero or another 3rdparty?

@pfmackin what do you think?

tests/e2e/config/scripts/install-verrazzano-kind.yaml
```diff
@@ -25,3 +25,22 @@ spec:
       enabled: true
     prometheusNodeExporter:
       enabled: true
+    velero:
```
michael-cico, line 28: We may want to revisit how we end up running this in the e2e tests; not sure if we want to always have it on.

However, we will definitely want it enabled for the private registry scenarios, to make sure it works in that case.  The images will get pushed, but we'll want to have it installed and verified there also.

michael-cico: Are the e2e tests and where we want to run those for this still undecided?

platform-operator/controllers/verrazzano/component/velero/velero_installer.go
theabmitra: make sense

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
theabmitra: yes, this can be refactored

platform-operator/controllers/verrazzano/component/velero/velero_installer.go
theabmitra: Not from VZ api today, but i have exposed so this can be done in future .

tests/e2e/config/scripts/install-verrazzano-kind.yaml
theabmitra: no, this will be reverted. i am using this as a test for verification of the ginkgo tests

platform-operator/scripts/uninstall/uninstall-steps/3-uninstall-verrazzano.sh
michael-cico: OK.  I think with the newer install stuff, we're planning on not removing CRDs, at least by default.  We may have to consider what to do there, it velero supports that.  On the other hand, I'm guessing only velero uses/consumes those CRDs, so it might be fine either way.

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
theabmitra: normally the os.exec is mocked with a env variable . I have used interfaces for mocking http in a separate PR , but for this method its a bit overkill i felt. hence i kept it simple .

platform-operator/controllers/verrazzano/component/velero/velero_installer.go
theabmitra: NA

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
theabmitra: NA, moving to helm based installer

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
theabmitra: NA, moving to helm based installer

platform-operator/controllers/verrazzano/component/velero/velero_operator.go
theabmitra: NA, moving to helm based installer

platform-operator/internal/vzconfig/enabled.go
theabmitra: Fixed

michael-cico (dismissed):

platform-operator/Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN yum update -y \
     && yum install -y kubectl-1.20.11-4.el7 \
     && yum install -y helm-3.8.0-1.el7 \
     && yum install -y istio-istioctl-1.13.2-1.el7 \
+    && yum install -y velero-1.8.1-1.el7 \
```
kminder, line 35: Is this still required now that the helm chart is being used?

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
```diff
@@ -0,0 +1,130 @@
...
+func NewComponent() spi.Component {
+	return veleroHelmComponent{
+		helm.HelmComponent{
+			ReleaseName:             ComponentName,
+			JSONName:                ComponentJSONName,
+			ChartDir:                filepath.Join(config.GetThirdPartyDir(), ChartDir),
+			ChartNamespace:          ComponentNamespace,
+			IgnoreNamespaceOverride: true,
+			SupportsOperatorInstall: true,
+			MinVerrazzanoVersion:    constants.VerrazzanoVersion1_3_0,
```
kminder, line 55: Is 1_3_0 correct?

platform-operator/thirdparty/charts/velero/velero-install.yaml
```diff
@@ -0,0 +1,3771 @@
+apiVersion: v1
```
kminder, line 1: This file seems odd.  Doesn't it duplicate content from the crds directory?

platform-operator/verrazzano-bom.json
```diff
@@ -614,6 +614,33 @@
...
+    {
+      "name": "velero",
+      "subcomponents": [
+        {
+          "repository": "verrazzano",
+          "name": "velero",
+          "images": [
+            {
+              "image": "velero",
+              "tag": "v1.8.1",
```
kminder, line 627: Do we not have images tagged with version-timestamp-commithash yet?

theabmitra: OLCNE builds are pushed to an internal repo and they dont support commit hash. We have to manually pick up that build and publish it here .

platform-operator/Dockerfile
theabmitra: i will be cleaning that up. Its not needed any more

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
theabmitra: changed to 1.4

platform-operator/thirdparty/charts/velero/velero-install.yaml
theabmitra: cleaned up

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
theabmitra: NA

platform-operator/internal/k8s/status/daemonset_ready.go
```diff
@@ -19,6 +19,12 @@ import (
 
 // DaemonSetsAreReady Check that the named daemonsets have the minimum number of specified nodes ready and available
 func DaemonSetsAreReady(log vzlog.VerrazzanoLogger, client client.Client, namespacedNames []types.NamespacedName, expectedNodes int32, prefix string) bool {
+	resticPodLabel := map[string]string{
```
michael-cico, line 22: Should this be here, or handled as a one-off in the velero component?  It seems odd to embed component specific logic in a shared utility func.

platform-operator/internal/k8s/status/deployment_ready.go
```diff
@@ -5,19 +5,26 @@ package status
...
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
 )
 
 // DeploymentsAreReady check that the named deployments have the minimum number of specified replicas ready and available
 func DeploymentsAreReady(log vzlog.VerrazzanoLogger, client clipkg.Client, namespacedNames []types.NamespacedName, expectedReplicas int32, prefix string) bool {
+	veleroPodLabel := map[string]string{
```
michael-cico, line 21: same comment as above.

platform-operator/internal/k8s/status/daemonset_ready.go
michael-cico: Just do custom ready checks in the component I mean, instead of this.  Or make this general enough to handle the matchlabels or something.

michael-cico: Met with Abhishek briefly, we'll go with this for now, possibly we can look at making the utilities more flexible without disrupting the code base too much via a separate PR.

michael-cico (approved):

kminder (dismissed):
Approved.

pfmackin (approved):

PR #3455, opened 2022-06-29
VZ-6113 Changes for parsing jaeger logs by fluentd

anajosep:
VZ-6113 - Changes in fluentd config to parse logs from Jaeger Operator and Jaeger Components (query, ui, collector)


fredtibbitts (approved):
I can't comment on the Jaeger parsing specifics but the changes generally seem fine to me.

tests/e2e/logging/system/system_logging_test.go
```diff
@@ -512,6 +518,26 @@ func validateNodeExporterLogs() bool {
 		noExceptions)
 }
 
+func validateJaegerCollectorLogs() bool {
+	return validateElasticsearchRecords(
+		logLevelElasticsearchRecordValidator,
+		func() (string, error) { return pkg.GetOpenSearchSystemIndex(monitoringNamespace) },
```
desagar, line 524: This namespace (which is declared with value "monitoring" at the top of this file), no longer exists.

davisgomes: Same comment from me.

platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
```diff
@@ -707,6 +707,35 @@ data:
...
+    # filter to parse jaeger operator and jaeger resource log files
+    <filter kubernetes.**jaeger-operator-**verrazzano-monitoring_jaeger-**>
+      @type parser
+      @id jaeger
+      key_name log
+      reserve_data true
+      emit_invalid_record_to_error true
+      <parse>
+        @type multi_format
+        # zap log format for jaeger components
```
kminder, line 719: The reference to jaeger here should be capitalized.

tests/e2e/logging/system/system_logging_test.go
anajosep: This namespace is using for the getting the index name in OS. From v1.3, data streams are being used and all the system logs are stored in a single data stream `verrazzano-system` and for older versions, the old index name would be used which is `verrazzano-system-monitoring`. `monitoring` is the right value to be used here. For v1.3+ this value is ignored and it always returns the index to look for as `verrazzano-system`

`if usingDataStreams {
		return "verrazzano-system", nil
	}
`

platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
anajosep: Corrected the commented lines to have `Jaeger`

fredtibbitts (approved):

PR #3456, opened 2022-06-29
VZ-5861,VZ-5863 Switch ingress to use new Prometheus, migrate from old Prometheus to new in install/upgrade and tests

desagar:


fredtibbitts (approved):

kminder (approved):
Approved after meeting to review.

PR #3461, opened 2022-06-30
Changes to enable custom rancher dashboard

lake-of-dreams:
- Env variable `CATTLE_UI_OFFLINE_PREFERRED` needs to be set to `true` for rancher to source all assets from the embedded dashoard and not  navigate to https://releases.rancher.com
- rancher versions will be changed in this PR once the BFS pipeline related [PRs](https://github.com/verrazzano/rancher/pull/36) are approved and merged

### DIFF
--- a/platform-operator/thirdparty/charts/rancher/values.yaml
+++ b/platform-operator/thirdparty/charts/rancher/values.yaml
@@ -38,9 +38,9 @@ debug: false
 restrictedAdmin: false
 
 # Extra environment variables passed to the rancher pods.
-# extraEnv:
-# - name: CATTLE_TLS_MIN_VERSION
-#   value: "1.0"
+extraEnv:
+- name: CATTLE_UI_OFFLINE_PREFERRED
+  value: "true"
 
 # Fully qualified name to reach your Rancher server
 # hostname: rancher.my.org
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -125,13 +125,13 @@
           "images": [
             {
               "image": "rancher",
-              "tag": "20220628131948-d06f15118",
+              "tag": "v2.6.4-20220504205847-7131173f2",
               "helmFullImageKey": "rancherImage",
               "helmTagKey": "rancherImageTag"
             },
             {
               "image": "rancher-agent",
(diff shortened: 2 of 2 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -125,13 +125,13 @@
           "images": [
             {
               "image": "rancher",
-              "tag": "v2.6.4-20220504205847-7131173f2",
+              "tag": "20220628131948-d06f15118",
```
kminder, line 128: Is this temporary?

lake-of-dreams: yes its just for testing, the final value of tag will have both

platform-operator/thirdparty/charts/rancher/values.yaml
```diff
@@ -38,9 +38,9 @@ debug: false
 restrictedAdmin: false
 
 # Extra environment variables passed to the rancher pods.
-# extraEnv:
-# - name: CATTLE_TLS_MIN_VERSION
-#   value: "1.0"
+extraEnv:
+- name: CATTLE_UI_OFFLINE_PREFERRED
```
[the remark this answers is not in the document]
lake-of-dreams: I explained this in another comment on wiki. Pasting it here again:

- The index page is set as UIDashboardIndex  in https://github.com/verrazzano/rancher/blob/oracle-build-from-source-v2.6.4/pkg/ui/handler.go#L27
- In the same line, we specify the local dashboard path in image as  UIDashboardPath.We also specify a setting UIOfflinePreferred in the same line.
- Now in UI handler code in https://github.com/rancher/steve/blob/master/pkg/ui/handler.go#L118 , the home page or index page of Rancher dashboard is loaded either from - https://releases.rancher.com/dashboard/release-2.6.4/index.html (when UIOfflinePreferred has value as "dynamic" which is the default value) or
- from local dashboard path that we embed in the rancher image at build time ( When UIOfflinePreferred has value as "true").
- In this PR,  we change the value of UIOfflinePreferred setting (which is driven from CATTLE_UI_OFFLINE_PREFERRED env variable) as true and therefore all dashboard content is always loaded from embedded local custom dashboard only

PR #3479, opened 2022-07-04
VZ-5666: Verrazzano clusterrole changes to allow hook role permissions.

theabmitra:


kminder:
Update the PR description with change details.

kminder (approved):
Approved.

PR #3481, opened 2022-07-05
VZ-6251: Verrazzano bug report - initial implementation

bhatpmk:
Implements the initial version of the command: vz bug-report. In this change, the command captures the following resources from the cluster as JSON files.

- Verrazzano resource
- Logs from verrazzano-platform-operator, verrazzano-monitoring-operator and verrazzano-application-operator pods
- Workloads (Deployment and ReplicaSet, StatefulSet, Daemonset), pods, events, ingress and services from verrazzano-system namespace

The output of the command bug-report can be extracted and used as the capture-dir for the "vz analyze" command


tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func runCmdBugReport(cmd *cobra.Command, args []string, vzHelper helpers.VZHelper) error {
+	bugReportFile, err := cmd.PersistentFlags().GetString(constants.BugReportFileFlagName)
+	if err != nil {
+		return fmt.Errorf("error fetching flag: %s", err.Error())
+	}
+
+	// Validate the report file format
+	if !strings.HasSuffix(bugReportFile, constants.BugReportFileExtn) {
```
agv426, line 44: Do we want to interpret the file extension here and fail if it is not ".tar.gz"?

What are other CLI utilities doing in that situation? I'm not sure that kubectl has a situation where an output filename is specified?

For example, other patterns we could follow:

1. Add a ".tar.gz" extension if it is not supplied on the filename. That would make it clear the resulting file was a .tar.gz file, and also would support them just giving up the filename without the extension.
2. Just use the filename supplied. That would cause confusion though, not suggesting we do this (IMO)

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
+	"github.com/verrazzano/verrazzano/tools/vz/test/helpers"
+	"io/ioutil"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"os"
+	"testing"
+)
+
+func TestBugReportWithoutAnyFlag(t *testing.T) {
```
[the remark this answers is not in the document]
bhatpmk: Done

tools/vz/cmd/bugreport/bugreport.go
bhatpmk: The validation is removed now. The CLI will create the bug report using the user specified input.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -0,0 +1,63 @@
...
+	return cmd
+}
+
+func runCmdBugReport(cmd *cobra.Command, args []string, vzHelper helpers.VZHelper) error {
+	bugReportFile, err := cmd.PersistentFlags().GetString(constants.BugReportFileFlagName)
+	if err != nil {
+		return fmt.Errorf("error fetching flag: %s", err.Error())
+	}
+
+	// Fail if the bugReportFile already exists
```
kminder, line 43: Seems like this should also be able to check to ensure the file can be written and fail up front with a nice message if not.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -40,9 +40,10 @@ func runCmdBugReport(cmd *cobra.Command, args []string, vzHelper helpers.VZHelpe
 		return fmt.Errorf("error fetching flag: %s", err.Error())
 	}
 
-	// Validate the report file format
-	if !strings.HasSuffix(bugReportFile, constants.BugReportFileExtn) {
-		return fmt.Errorf("unsupported report-file: %s, set a .tar.gz file", bugReportFile)
+	// Fail if the bugReportFile already exists
+	_, err = os.Stat(bugReportFile)
+	if !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
bhatpmk: I will address this in a follow-up change

tools/vz/cmd/bugreport/bugreport.go
bhatpmk: I will address this in a follow-up change

bhatpmk:
VZ-6251

PR #3432, opened 2022-06-27
VZ-5953 CLI Uninstall Command error handling

stevshap:
Fixes VZ-5953

Adds error handling to prevent uninstall logs from writing to a closed channel.

PR #3489, opened 2022-07-05
NO JIRA: Reduce sleep interval when waiting during CLI uninstall

bobdonat:
This pull request changes sleep intervals to be 1 second instead of 5 seconds.  When the sleep intervals were 5 seconds the unit tests were taking 15+ seconds to complete.  Also, moved a couple of sleeps to end of loops instead of start of loops.

### DIFF
--- a/tools/vz/cmd/uninstall/uninstall.go
+++ b/tools/vz/cmd/uninstall/uninstall.go
@@ -43,9 +43,9 @@ vz uninstall --timeout 30m`
 )
 
 // Number of retries after waiting a second for uninstall pod to be ready
-const uninstallDefaultWaitRetries = 60
+const uninstallDefaultWaitRetries = 300
 
-const verrazzanoUninstallJobDetectWait = 5
+const verrazzanoUninstallJobDetectWait = 1
 
 var uninstallWaitRetries = uninstallDefaultWaitRetries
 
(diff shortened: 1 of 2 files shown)


stevshap (approved):

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -199,6 +192,13 @@ func getUninstallPodName(c client.Client, vzHelper helpers.VZHelper, jobName str
 			return "", fmt.Errorf("Waiting for %s, failed to list pods: %s", jobName, err.Error())
 		}
 		if len(podList.Items) == 0 {
+			retryCount++
+			if retryCount > uninstallWaitRetries {
+				return "", fmt.Errorf("Waiting for %s, %s pod not found in namespace %s", jobName, jobName, vzconstants.VerrazzanoInstallNamespace)
+			}
+			time.Sleep(verrazzanoUninstallJobDetectWait * time.Second)
```
mgianatagh, line 199: I though the sleeps were moved to the beginning of the loop to ensure the unit tests don't fail due to the vz resource being instantly deleted and the timeout value not being reached.  I think we want to put this back.

bobdonat: I moved the two sleeps related to waiting for the verrazzano-install pod.  I did not change the sleep that is checking the vz resource.  I think the change should be fine.

bobdonat: I think this was the previous PR https://github.com/verrazzano/verrazzano/pull/3432 that addressed the vz resource being instantly deleted.

mgianatagh (approved):

PR #3493, opened 2022-07-05
Sdosapat/vz 6255

saidu941:

Validation:
```
[19:41:37 bash vz]$make cli
GO111MODULE=on GOPRIVATE=github.com/verrazzano/* go install -ldflags "-X 'github.com/verrazzano/verrazzano/tools/vz/cmd/version.gitCommit=8eb4a519f422bb27e2e6f5f7987f2251785a0a37' -X 'github.com/verrazzano/verrazzano/tools/vz/cmd/version.buildDate=2022-07-05T19:41:44Z' -X 'github.com/verrazzano/verrazzano/tools/vz/cmd/version.cliVersion=1.4.0'" ./...
[19:42:02 bash vz]$vz
The vz tool is a command-line utility that allows Verrazzano operators to query and manage a Verrazzano environment

Usage:
  vz [command]

Available Commands:
  analyze     Analyze cluster
  help        Help about any command
  install     Install Verrazzano
  status      Status of the Verrazzano installation and access endpoints
  uninstall   Uninstall Verrazzano
  upgrade     Upgrade Verrazzano
  version     Verrazzano version information

Flags:
      --context string      The name of the kubeconfig context to use
  -h, --help                help for vz
      --kubeconfig string   Path to the kubeconfig file to use

Use "vz [command] --help" for more information about a command.
[19:42:10 bash vz]$

```

### DIFF
--- a/tools/vz/pkg/helpers/vzsanitize.go
+++ b/tools/vz/pkg/helpers/vzsanitize.go
@@ -1,0 +1,118 @@
+//Copyright (C) 2022, Oracle and/or its affiliates.
+//Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package helpers
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+var regexList = []string{}
+var regexToReplacementMap = make(map[string]string)
+
+const redactIpAddress = "REDACTED-IP4-ADDRESS"
+const ipv4Regex = "[[:digit:]]{1,3}\\.[[:digit:]]{1,3}\\.[[:digit:]]{1,3}\\.[[:digit:]]{1,3}"
+
+func initRegexToReplacementMap() {
+	regexToReplacementMap[ipv4Regex] = redactIpAddress
+}
+
+// GetMatchingFiles returns the filenames for files that match a regular expression.
+func GetMatchingFiles(rootDirectory string, fileMatchRe *regexp.Regexp) (fileMatches []string, err error) {
+	if len(rootDirectory) == 0 {
+		return nil, fmt.Errorf("GetMatchingFiles requires a rootDirectory")
+	}
+
+	if fileMatchRe == nil {
+		return nil, fmt.Errorf("GetMatchingFiles requires a regular expression")
+	}
+
+	walkFunc := func(fileName string, fileInfo os.FileInfo, err error) error {
+		if !fileMatchRe.MatchString(fileName) {
+			return nil
+		}
+		if !fileInfo.IsDir() {
+			fileMatches = append(fileMatches, fileName)
+		}
+		return nil
+	}
+
+	err = filepath.Walk(rootDirectory, walkFunc)
+	if err != nil {
+		return nil, err
+	}
+	return fileMatches, err
+}
+
+func SanitizeDirectory(rootDirectory string, replaceFile bool) {
+	initRegexToReplacementMap()
+	fileMatches, err := GetMatchingFiles(rootDirectory, regexp.MustCompile(".*"))
+	check(err)
+	for _, eachFile := range fileMatches {
+		SanitizeFile(eachFile, replaceFile)
+	}
+}
+
+func SanitizeFile(path string, replaceFile bool) error {
+	fmt.Println("path provided is: ", path)
+	input, err := os.Open(path)
+	check(err)
+	defer input.Close()
+
+	outFile, err := os.Create(path + "_tmpfoo")
+	check(err)
+	defer outFile.Close()
+
+	br := bufio.NewReader(input)
+	for {
+		l, _, err := br.ReadLine()
+
+		if err == io.EOF || l == nil {
+			break
+		}
+		check(err)
+
+		sanitizedLine := sanitizeEachLine(string(l))
+		outFile.WriteString(sanitizedLine + "\n")
+	}
+
+	//originalFile, err := os.ReadFile(path)
+	//check(err)
+	//fmt.Println("Original file is: \n", string(originalFile))
+
+	//modifiedFile, err := os.ReadFile(path + "_tmpfoo")
+	//check(err)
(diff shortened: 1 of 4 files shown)


kminder:
Cleanup the PR description and title.

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -0,0 +1,120 @@
...
+		if err == io.EOF || l == nil {
+			break
+		}
+		check(err)
+
+		sanitizedLine := sanitizeEachLine(string(l))
+		outFile.WriteString(sanitizedLine + "\n")
+	}
+
+	//originalFile, err := os.ReadFile(path)
```
kminder, line 89: Cleanup dead code.

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -0,0 +1,120 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+var regexList = []string{}
+var regexToReplacementMap = make(map[string]string)
+
+const redactIpAddress = "REDACTED-IP4-ADDRESS"
```
kminder, line 18: Have you investigated replacing these with something that includes an irreversible hash of the IPv4 address so that they could correlated?

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -0,0 +1,120 @@
...
+	initRegexToReplacementMap()
+	fileMatches, err := GetMatchingFiles(rootDirectory, regexp.MustCompile(".*"))
+	check(err)
+	for _, eachFile := range fileMatches {
+		SanitizeFile(eachFile, replaceFile)
+	}
+}
+
+//Sanitize the given file, replaceFile boolean is to whether inplace replacement or to a new file
+func SanitizeFile(path string, replaceFile bool) error {
```
kminder, line 66: Is there any way to sanitize files as they are written instead of afterwards?

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	err = filepath.Walk(rootDirectory, walkFunc)
+	if err != nil {
+		return nil, err
+	}
+	return fileMatches, err
+}
+
+// This will iterate over the given root directory to sanitize each file
```
kminder, line 54: Please follow Godoc conventions.
https://go.dev/blog/godoc

tools/vz/pkg/helpers/vzsanitize.go
agv426: Yeah, it was in the original thinking, the script Saidu started with was just being used to sanitize data for our unit testing (which didn't need to be correlated), so I don't think he was looking at that in this round.

One way hashing them though should be pretty simple to add in here.

One related question is whether we also need to separately generate a mapping file that the customer would retain and not send along to support. The mapping file would need to be tied to the generated bug report in some way so a customer could use that to tell which IP/Host/etc... the hashed value really mapped back to. The report and mapping could both contain a shared generated UUID, the name of the mapping file would need to be supplied or derived from the filename supplied for the report. Support would never see the customers internal redacted info or the mapping file, the customer would always be in control of that.

I don't think a mapping file is required for support to be able to correlate, it would mainly be useful if support needed to communicate an issue with a specific IP/Host back to the customer (ie: Host XXXXX has an issue, the customer could use that to map the XXXX has back to their actual host, etc...). So I think we can defer a mapping file for now, and consider adding it in the future.

tools/vz/pkg/helpers/vzsanitize.go
agv426: Yeah, we could do that. I'll discuss with Bhat and Saidu in the sync up this morning.

I think that we could sanitize the data as we retrieve data from the cluster. That would be the safest thing, that code will be shared and doing that would ensure that nothing would be leaked inadvertently (ie: the data in-memory would be redacted as well during analysis)

tools/vz/pkg/helpers/vzsanitize.go
kminder: I don't think the mapping file is important.  If that communication is important we could just tell the customer that the IP that hashed to XXXXXX via MD5, SHA1 or whatever had some issue.  So the key is to use some well known hashing function.

tools/vz/pkg/helpers/vzsanitize.go
saidu941: Updated the PR

tools/vz/pkg/helpers/vzsanitize.go
saidu941: Yes, this is just a util function (currently not used)
func SanitizeALine() is currently being used, which does sanitizes as they're written

tools/vz/pkg/helpers/vzsanitize.go
saidu941: Updated the comments following the doc
Please lmk if it's not yet completed

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -0,0 +1,123 @@
...
+	err = filepath.Walk(rootDirectory, walkFunc)
+	if err != nil {
+		return nil, err
+	}
+	return fileMatches, err
+}
+
+// SanitizeDirectory will iterate over the given root directory to sanitize each file
+// ReplaceFile is to whether replace the file with sanitized content or create a new file with _tmpfoo suffix
+func SanitizeDirectory(rootDirectory string, replaceFile bool) {
```
agv426, line 56: Is this effectively dead code now that we are sanitizing on input? ie: SanitizeDirectory and GetMatchingFiles

agv426 (dismissed):

tools/vz/pkg/helpers/vzsanitize.go
saidu941: Updated

agv426 (approved):

PR #3494, opened 2022-07-06
VZ-6251: Check whether report file can be created before actually collecting the info from the cluster

bhatpmk:
This PR adds validations to determine whether the bug report file can actually be created. If there is an existing file with the same name, or if the user doesn't have the permission to create the file, an appropriate error is returned with this change.


agv426 (approved):

kminder (approved):
Approved.

bhatpmk:
VZ-6251

PR #3498, opened 2022-07-06
VZ-6228 introduce the cert-manager uninstall step

davisgomes:
PR to move the cert-manager uninstall to the go component.


platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -834,6 +838,51 @@ func cleanupUnusedResources(compContext spi.ComponentContext, isCAValue bool) er
...
+// uninstallCertManager is the implementation for the cert-manager uninstall step
+// this removes cert-manager ConfigMaps from the cluster and after the helm uninstall, deletes the namespace
+func uninstallCertManager(compContext spi.ComponentContext) error {
+	// Delete the kube-system cert-manager configMaps [controller, caInjector]
+	var controllerCM v1.ConfigMap
+	controllerCM.SetName(controllerConfigMap)
+	controllerCM.SetNamespace(constants.KubeSystem)
+	err := compContext.Client().Delete(context.TODO(), &controllerCM)
+	if crtclient.IgnoreNotFound(err) != nil {
+		return compContext.Log().ErrorfNewErr("Failed to delete the ConfigMap %s/%s: %v", constants.KubeSystem, controllerConfigMap, err)
```
bobdonat, line 850: Will this fail the uninstall?  Wondering if this should be a warning and we continue on.

platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
```diff
@@ -152,6 +153,19 @@ func (c certManagerComponent) PostUpgrade(compContext spi.ComponentContext) erro
...
+func (c certManagerComponent) Uninstall(compContext spi.ComponentContext) error {
+	if compContext.IsDryRun() {
+		compContext.Log().Debug("cert-manager Uninstall dry run")
+		return nil
+	}
+	err := c.HelmComponent.Uninstall(compContext)
+	if err != nil {
+		return err
+	}
+	return uninstallCertManager(compContext)
```
michael-cico, line 166: One option here is to call `uninstallCertManager` from the `PostUninstall`, and then you don't have to call the Helm impl directly here.  But 6 of one 1/2 dozen the other.

pfmackin, line 166: i think this should be in post uninstajl.  Uninstall should just use the default helm component uninstall.

pfmackin: I agree with the caveat that this approach might be faster since the uninstall_component.go doesn't call post uninstall until after helm reports that the release is uninstalled.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -834,6 +838,51 @@ func cleanupUnusedResources(compContext spi.ComponentContext, isCAValue bool) er
...
+	_, err = controllerutil.CreateOrUpdate(context.TODO(), compContext.Client(), &certNS, func() error {
+		certNS.Finalizers = []string{}
+		return nil
+	})
+	if err != nil {
+		return compContext.Log().ErrorfNewErr("Failed to remove the finalizer for the namespace %s: %v", constants.CertManagerNamespace, err)
+	}
+
+	// Delete the cert-manager namespace now that the finalizers have been removed
+	err = compContext.Client().Delete(context.TODO(), &certNS)
```
michael-cico, line 877: Do we have to delete the `ClusterIssuer`?  I think that's cluster-scoped right?

We also may want to move namespace deletion to a globally scoped uninstall operation, since this shares a namespace with `external-dns`.  It won't bite us now, since this is a full uninstall, but when we get to being able to dynamically enable/disable components it might then.

@pfmackin what do you think?

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
bobdonat: Agreed.  I read the code wrong.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
pfmackin: I think it is ok to delete namespace and clusterissuer here.   However, in the case where multiple products use the same namespace (verrazzano system), maybe we should defer NS deletion to a post-uninstall global  state

platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
pfmackin: Since we don't want helm uninstall to be called multiple times when some other failure occurs (not related to helm) i think this should be moved to postUninstall as Mike mentioned.

davisgomes: Okay, I will make this change

pfmackin (approved):

PR #3501, opened 2022-07-06
NO JIRA: revert move of sleep timers during CLI uninstall

bobdonat:
The pull request reverts the move of sleep timers during CLI install.  Unit tests can intermittently fail.


stevshap (approved):
approved

mgianatagh (approved):

PR #3507, opened 2022-07-06
VZ 5363: Handle error for metrics trait

anders-swanson:
handle error when removing metrics trait finalizers.

### DIFF
--- a/application-operator/controllers/metricstrait/metricstrait_controller.go
+++ b/application-operator/controllers/metricstrait/metricstrait_controller.go
@@ -292,7 +292,7 @@ func (r *Reconciler) reconcileTraitDelete(ctx context.Context, trait *vzapi.Metr
 	// Only remove the finalizer if all related resources were successfully updated.
 	if !status.ContainsErrors() {
 		if err := r.removeFinalizerIfRequired(ctx, trait, log); err != nil {
-			return ctrl.Result{Requeue: true}, err // the caller always does a requeue if there is an error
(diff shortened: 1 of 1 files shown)


application-operator/controllers/metricstrait/metricstrait_controller.go
```diff
@@ -291,7 +291,9 @@ func (r *Reconciler) reconcileTraitDelete(ctx context.Context, trait *vzapi.Metr
 	status := r.deleteOrUpdateObsoleteResources(ctx, trait, &reconcileresults.ReconcileResults{}, log)
 	// Only remove the finalizer if all related resources were successfully updated.
 	if !status.ContainsErrors() {
-		r.removeFinalizerIfRequired(ctx, trait, log)
+		if err := r.removeFinalizerIfRequired(ctx, trait, log); err != nil {
+			return ctrl.Result{Requeue: true}, err // the caller always does a requeue if there is an error
```
pfmackin, line 295: this should requeue after delay or else it can pin the cpu if the error keeps happening

anders-swanson: The caller ignores the ctrl.Result value, but i guess there is no harm adding it here too.

pfmackin: we have a common function NewRequeueWithDelay i think it is

pfmackin (approved):

PR #3508, opened 2022-07-06
VZ-6358, VZ-6359.  Move prom operator and node exporter uninstall to VPO

michael-cico:


platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -221,3 +238,8 @@ func (r *Reconciler) deleteSecret(log vzlog.VerrazzanoLogger, namespace string,
 	}
 	return nil
 }
+
+//deleteNamespaces Cleans up any namespaces shared by multiple components
+func (r *Reconciler) deleteNamespaces(log vzlog.VerrazzanoLogger) error {
+	return namespace.DeleteNamespace(log, r.Client, promoperator.ComponentNamespace)
```
bobdonat, line 244: Shouldn't this use the delete functions we recently added to resources.go

platform-operator/internal/k8s/namespace/namespace.go
```diff
@@ -44,6 +46,20 @@ func CreateVerrazzanoMultiClusterNamespace(client client.Client) error {
 	return CreateAndLabelNamespace(client, globalconst.VerrazzanoMultiClusterNamespace, false, false)
 }
 
+//DeleteNamespace Patch out any finalizers and delete the specified namespace
```
pfmackin, line 49: we have Delete in pkg/resources.  Maybe this can just wrap that call

platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -208,6 +220,11 @@ func (r *Reconciler) isMC(log vzlog.VerrazzanoLogger) (bool, error) {
 	return true, nil
 }
 
+//uninstallCleanup Perform the final cleanup of shared resources, etc not tracked by individal component uninstalls
+func (r *Reconciler) uninstallCleanup(log vzlog.VerrazzanoLogger) error {
+	return r.deleteNamespaces(log)
```
pfmackin, line 225: can't we just call namespace.DeleteNamespace(log, r.Client, promoperator.ComponentNamespace) from here?

pfmackin (approved):

platform-operator/controllers/verrazzano/uninstall.go
michael-cico: Yes, I changed the internal helper to use that now.  Pushed.

platform-operator/internal/k8s/namespace/namespace.go
michael-cico: yes, just pushed that change.

platform-operator/controllers/verrazzano/uninstall.go
michael-cico: we can, but I broke it this way so that the namespace deletes will be in a separate call, as those grow in number.

We'll also be doing other cleanup here I think.

platform-operator/internal/k8s/namespace/namespace.go
```diff
@@ -48,16 +47,16 @@ func CreateVerrazzanoMultiClusterNamespace(client client.Client) error {
...
-	})
-	if err != nil {
-		return log.ErrorfNewErr("Failed to remove the finalizer for the namespace %s: %v", nsName, err)
+	res := resource.Resource{
+		Name:   nsName,
+		Client: cli,
+		Object: &corev1.Namespace{},
+		Log:    log,
+	}
+	if err := res.RemoveFinalizers(); err != nil {
```
bobdonat, line 56: I just merged a change related to this.  You can make a single call RemoveFinalizersAndDelete function call now.

bobdonat (approved):

PR #3509, opened 2022-07-07
VZ-6079 Changes for exposing Jaeger URL as part of VZ CR

anajosep:
VZ-6079 Changes for exposing Jaeger URL

Provide SSO support for Jaeger UI URL and expose it throug VZ CR in the status fields along with the URLs of other components.


platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -512,6 +514,8 @@ data:
             serverUrl = me.makeVmiBackendUrl('es-ingest', '9200')
         elseif backend == 'kiali' then
             serverUrl = me.makeVmiBackendUrl(backend, '20001')
+        elseif backend == 'jaeger' then
+            serverUrl = 'http://jaeger-operator-jaeger-query.verrazzano-monitoring.svc.cluster.local:16686'
```
ayaranga, line 518: For the first cut, this maybe fine. Eventually, it would be better to get query service similar to what we do in our istio component to get collector service https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/istio/jaeger.go#L52-L67

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -96,3 +124,17 @@ func (c jaegerOperatorComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzap
 	}
 	return c.validateJaegerOperator(new)
 }
+
+// createOrUpdateKialiResources create or update related Kiali resources
```
ayaranga, line 128: nit: Update comment to reflect Jaeger

ayaranga (dismissed):

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
ayaranga: Thinking more, the Jaeger Operator name is used in multiple places (authpolicy, service account name..). It is better we add a validation not to change the name than addressing in all the places.

michael-cico: To @anajosep 's point, it may be hard to do that since this is lua code.  All the other services are hard-coded.  

You'd have to maybe use Go templates and massage the lua script code in the configmap at build time, which is beyond the scope of this PR I think.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -39,14 +40,16 @@ func resetWriteFileFunc() {
...
+	deploymentName        = "jaeger-operator"
+	tmpFilePrefix         = "jaeger-operator-overrides-"
+	tmpSuffix             = "yaml"
+	tmpFileCreatePattern  = tmpFilePrefix + "*." + tmpSuffix
+	tmpFileCleanPattern   = tmpFilePrefix + ".*\\." + tmpSuffix
+	jaegerSecName         = "verrazzano-jaeger-secret"
+	jaegerCreateField     = "jaeger.create"
+	jaegerHostName        = "jaeger"
+	jaegerCertificateName = "jaeger-tls"
+	openSearchURL         = "http://verrazzano-authproxy-elasticsearch.verrazzano-system.svc.cluster.local:8775"
```
michael-cico, line 52: Does Jaeger only run in a cluster with OS?  i.e., will it run in a managed cluster where OS is not present?  Or is this just the Jaeger UI that will only run in admin clusters?

michael-cico (dismissed):

wmhopkins (changes requested):
In general looks fine to me, but would like to see a utility method to generate the backend URL, see my comment to that effect.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
pfmackin: that is a good question, this won't work on managed cluster.  I do think Jaeger should be on the managed cluster

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -349,3 +354,76 @@ func isCreateJaegerInstance(ctx spi.ComponentContext) (bool, error) {
...
+				SecretName: "jaeger-tls",
+			},
+		}
+		ingress.Spec.Rules = []networkv1.IngressRule{ingRule}
+
+		if ingress.Annotations == nil {
+			ingress.Annotations = make(map[string]string)
+		}
+		ingress.Annotations["kubernetes.io/tls-acme"] = "true"
+		ingress.Annotations["nginx.ingress.kubernetes.io/proxy-body-size"] = "6M"
```
kminder, line 407: Hard coding some of these seems like a problem.
Is there no way to source these from Helm values?

kminder:
Can you please confirm that this Jaeger status and URL information looks ok with the new VZ CLI?

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -425,6 +425,8 @@ data:
         else
             if backend == 'verrazzano' then
                 check_host = backend..me.hostSuffix
+            elseif backend == 'jaeger' then
```
wmhopkins, line 428: Are we sure we don't have any in-cluster user cases? Don't think the "if in_cluser == true" block above will work for things that aren't running in verrazzano-system namespace.

wmhopkins (dismissed):

anajosep:
> Can you please confirm that this Jaeger status and URL information looks ok with the new VZ CLI?

The Jaeger URL was missing in the template. I fixed that and now the output looks like below:
`
$ vz status
Verrazzano Status
  Name: dev
  Namespace: default
  Version: 1.4.0-local+79fe9cde
  State: Ready
  Profile: dev
  Access Endpoints:
    Console URL: https://verrazzano.default.172.18.0.231.nip.io
    Grafana URL: https://grafana.vmi.system.default.172.18.0.231.nip.io
    Jaeger URL: https://jaeger.default.172.18.0.231.nip.io
    Keycloak URL: https://keycloak.default.172.18.0.231.nip.io
    Kiali URL: https://kiali.vmi.system.default.172.18.0.231.nip.io
    Kibana URL: https://kibana.vmi.system.default.172.18.0.231.nip.io
    OpenSearch URL: https://elasticsearch.vmi.system.default.172.18.0.231.nip.io
    Prometheus URL: https://prometheus.vmi.system.default.172.18.0.231.nip.io
    Rancher URL: https://rancher.default.172.18.0.231.nip.io
`

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
anajosep: Only the Jaeger Query/UI component is exposed via the Ingress through Auth proxy. The `jaeger-collector`  component is what other services inside the cluster communicate to and it does not go via the Verrazzano auth proxy. `jaeger-collector` component is protected through appropriate Authorization policy and Network Policy. There aren't any ingresses for the `jaeger-collector`, it receives the spans from services inside the cluster and pushes it to storage (OpenSearch).  `jaeger-query` and `jaeger-ui` reads data from OS and displays it.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
anajosep: I followed the approach followed by other ingress resources like prometheus.

ayaranga (approved):

PR #3515, opened 2022-07-07
VZ-6321 Component uninstall for Rancher

davisgomes:
The uninstall of Rancher and relevant resources is now manged by the component.


platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,165 @@
...
+	// List all the namespaces that need to be cleaned from Rancher components
+	nsList := corev1.NamespaceList{}
+	err := ctx.Client().List(context.TODO(), &nsList)
+	if err != nil {
+		return ctx.Log().ErrorfNewErr("Failed to list the Rancher namespaces: %v", err)
+	}
+
+	// For Rancher namespaces, run the system tools uninstaller
+	for _, ns := range nsList.Items {
+		matches, err := regexp.MatchString("^cattle-|^local|^p-|^user-|^fleet|^rancher", ns.Name)
```
bobdonat, line 41: Can't this pick up namespace that are not rancher namespaces?  Is there a selector that can be used instead

pfmackin, line 41: can we search using labels instead of names?

platform-operator/config/scripts/run.sh
```diff
@@ -62,5 +62,6 @@ elif [ "${MODE}" == "UNINSTALL" ]; then
   dump-uninstall-logs 0
 else
   # Run the operator
+  create-kubeconfig
```
kminder, line 65: Why is this just required now?

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
kminder, line 1: Just 2022?

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// deleteRoleResources delete the Rancher role objects: ClusterRole, ClusterRoleBinding
+func deleteRoleResources(ctx spi.ComponentContext) error {
+	clusterRoleMatch := "cattle.io|app:rancher|rancher-webhook|fleetworkspace-|fleet-|gitjob"
```
kminder, line 108: Not a new problem but this really isn't great.  I imagine there could be non-Rancher/non-Verrazzano fleet installations.

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// For Rancher namespaces, run the system tools uninstaller
+	for _, ns := range nsList.Items {
+		matches, err := regexp.MatchString("^cattle-|^local|^p-|^user-|^fleet|^rancher", ns.Name)
+		if err != nil {
+			return ctx.Log().ErrorfNewErr("Failed to verify that namespace %s is a Rancher namespace: %v", ns.Name, err)
+		}
+		if matches {
+			args := []string{"remove", "-c", "/home/verrazzano/kubeconfig", "--namespace", ns.Name, "--force"}
+			cmd := osexec.Command(rancherSystemTool, args...) //nolint:gosec //#nosec G204
```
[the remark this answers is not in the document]
davisgomes: This is a blocking call that takes about 1-2 minutes

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
davisgomes: Yes, correct

platform-operator/config/scripts/run.sh
davisgomes: The system tool needs the kubeconfig for the uninstall. The system tool binary is added in the Dockerfile.

platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -107,8 +108,18 @@ func (r *Reconciler) reconcileUninstall(log vzlog.VerrazzanoLogger, cr *installv
...
 			}
-			tracker.vzState = vzStateUninstallDone
+			tracker.vzState = vzStateUninstallCleanup
 
+		case vzStateUninstallCleanup:
+			spiCtx, err := spi.NewContext(log, r.Client, cr, r.DryRun)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			err = r.uninstallCleanup(spiCtx, log)
```
michael-cico, line 118: we probably don't need a separate `log` arg since it's in the context.

platform-operator/config/scripts/run.sh
michael-cico: since it's referenced in the VPO deployment env now in this PR, I'd just always create the kubeconfig, and get rid of the `INSTALL` part of the switch.  We probably should have gotten rid of that months ago.

When we get rid of the uninstall job we'll also get rid of the `NOOP` part, since we that was to have dry-run type Job executions to test out the VPO job management code.

platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -208,6 +219,11 @@ func (r *Reconciler) isMC(log vzlog.VerrazzanoLogger) (bool, error) {
 	return true, nil
 }
 
+// uninstallCleanup Perform the final cleanup of shared resources, etc not tracked by individal component uninstalls
+func (r *Reconciler) uninstallCleanup(ctx spi.ComponentContext, log vzlog.VerrazzanoLogger) error {
+	return rancher.PostUninstall(ctx)
```
michael-cico, line 224: Should we be calling component code directly from here?

michael-cico: I guess it's not that weird... but we seem to be calling `rancher.PostUninstall` from the component and here.

platform-operator/controllers/verrazzano/uninstall.go
davisgomes: fixed!

platform-operator/controllers/verrazzano/uninstall.go
davisgomes: I think it's okay if it runs twice because it should be agnostic to cluster state and essentially no-op if the post uninstall has occurred

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,227 @@
...
+	namespaceNames := []string{
+		"^cattle-",
+		"^local",
+		"^p-",
+		"^user-",
+		"^fleet",
+		"^rancher",
+	}
+	for _, ns := range nsList.Items {
+		matches, err := regexp.MatchString(strings.Join(namespaceNames, "|"), ns.Name)
```
pfmackin, line 51: is there any label or annotation that gives us a hint that the namespace is owned by rancher?

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -0,0 +1,227 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// deleteMatchingResources delete the Rancher objects that need to match a string: ClusterRole, ClusterRoleBinding, RoleBinding, PersistentVolumes
+func deleteMatchingResources(ctx spi.ComponentContext) error {
+	// list of matching prefixes for cluster roles, cluster role bindings
```
pfmackin, line 135: Same comment, is there an annotation or label we can check to make sure this is really a rancher resource?

pfmackin (dismissed):

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -144,6 +144,7 @@ func deleteMatchingResources(ctx spi.ComponentContext) error {
 		"pod-impersonation-helm-op-",
 		"cattle-unauthenticated",
 		"default-admin-",
+		"proxy-",
```
michael-cico, line 147: is this too broad?  Should we look for the specific resource names?

michael-cico (approved):

platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -224,9 +227,12 @@ func (r *Reconciler) isMC(log vzlog.VerrazzanoLogger) (bool, error) {
 	return true, nil
 }
 
-//uninstallCleanup Perform the final cleanup of shared resources, etc not tracked by individal component uninstalls
-func (r *Reconciler) uninstallCleanup(log vzlog.VerrazzanoLogger) error {
-	return r.deleteNamespaces(log)
+// uninstallCleanup Perform the final cleanup of shared resources, etc not tracked by individual component uninstalls
+func (r *Reconciler) uninstallCleanup(ctx spi.ComponentContext) error {
+	if err := rancher.PostUninstall(ctx); err != nil {
```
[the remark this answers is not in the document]
michael-cico: ok

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
davisgomes: I tried to include as much as possible without including the generated characters for each resource i.e. proxy-h425i3

PR #3521, opened 2022-07-07
VZ 6023: Fluentd prometheus metrics

anders-swanson:
Enables prometheus metrics to be scraped from Fluentd.


kminder (approved):
Approved.

PR #3529, opened 2022-07-08
VZ-6228 Update the delete functions to use global package

davisgomes:
The cert-manager uninstall delete function now uses the global resource delete package.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
@@ -25,6 +25,7 @@ import (
 	certv1client "github.com/jetstack/cert-manager/pkg/client/clientset/versioned/typed/certmanager/v1"
 	"github.com/verrazzano/verrazzano/pkg/bom"
 	"github.com/verrazzano/verrazzano/pkg/constants"
+	vzresource "github.com/verrazzano/verrazzano/pkg/k8s/resource"
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
 	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
 	"github.com/verrazzano/verrazzano/pkg/security/password"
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -841,37 +842,87 @@ func cleanupUnusedResources(compContext spi.ComponentContext, isCAValue bool) er
...
+	}.Delete()
+	if err != nil {
 		return err
 	}
 
 	// Delete the ClusterIssuer created by Verrazzano
-	if err := deleteObjectWithLog(compContext, verrazzanoClusterIssuerName, vzconst.DefaultNamespace, &v1.ConfigMap{}); err != nil {
+	err = vzresource.Resource{
+		Name:      verrazzanoClusterIssuerName,
+		Namespace: vzconst.DefaultNamespace,
```
bobdonat, line 870: Are clusterissuers namespace scoped?  I don't think a namespace is needed here.

platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -841,37 +842,87 @@ func cleanupUnusedResources(compContext spi.ComponentContext, isCAValue bool) er
...
+		Object:    &v1.Secret{},
+		Log:       compContext.Log(),
+	}.Delete()
+	if err != nil {
 		return err
 	}
 
 	// Remove finalizers from the cert-manager namespace to avoid hanging namespace deletion
 	certNs := v1.Namespace{}
-	err := compContext.Client().Get(context.TODO(), types.NamespacedName{Name: ComponentNamespace}, &certNs)
```
bobdonat, line 874: I have a common function for removing a finalizer but it is not merged.  We can change this afterwards to use the common function

davisgomes: Sounds good

bobdonat (approved):

PR #3531, opened 2022-07-08
VZ-6353: Update component condition timestamp in VZ status

mbenmclean:
This change updates the timestamp in the component conditions in the VZ CR status if the condition already exists. It does not change the order of the status, just simply replaces the old condition data with the new condition data. Functionally speaking, all this does is update the lastTransitionTime field.

### DIFF
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -711,6 +711,7 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 func appendConditionIfNecessary(log vzlog.VerrazzanoLogger, compStatus *installv1alpha1.ComponentStatusDetails, newCondition installv1alpha1.Condition) []installv1alpha1.Condition {
 	for _, existingCondition := range compStatus.Conditions {
 		if existingCondition.Type == newCondition.Type {
+			existingCondition.LastTransitionTime = newCondition.LastTransitionTime
 			return compStatus.Conditions
 		}
 	}
--- a/platform-operator/controllers/verrazzano/controller.go
+++ b/platform-operator/controllers/verrazzano/controller.go
@@ -709,9 +709,9 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 }
 
 func appendConditionIfNecessary(log vzlog.VerrazzanoLogger, compStatus *installv1alpha1.ComponentStatusDetails, newCondition installv1alpha1.Condition) []installv1alpha1.Condition {
-	for _, existingCondition := range compStatus.Conditions {
+	for i, existingCondition := range compStatus.Conditions {
 		if existingCondition.Type == newCondition.Type {
-			existingCondition.LastTransitionTime = newCondition.LastTransitionTime
+			compStatus.Conditions[i] = newCondition
 			return compStatus.Conditions
 		}
 	}


platform-operator/controllers/verrazzano/upgrade_component.go
```diff
@@ -127,6 +130,9 @@ func (r *Reconciler) upgradeSingleComponent(spiCtx spi.ComponentContext, upgrade
 		case compStateUpgradeDone:
 			compLog.Oncef("Component %s has successfully upgraded", compName)
 			upgradeContext.state = compStateEnd
+			if err := r.updateComponentStatus(compContext, "Upgrade complete", installv1alpha1.CondUpgradeComplete); err != nil {
```
kminder, line 133: Haven't look at the other status messages but should this be "Upgrade completed" or are all the messages consistently inconsistent (i.e. Upgrade started vs Upgrade complete).

mbenmclean: Yes I modeled it off install which is inconsistent. Install has Install started and Install complete. I think we should perhaps change install to be consistent. There's no reason to keep carrying forward this inconsistency. I don't think contents of the message is used in any material way.

mbenmclean:
> 

Yes that is correct. I can see how that might present a problem for your metrics. Because if you were to scrape at that moment then the started would come after the complete. Perhaps you could add a check to ignore the value if the started time comes after the completed time. Or would the missing datapoint (or 0) cause issues in the timeseries?

GeorgeAeillo:
> 

I could add a check that would be able to check if the start time comes after the completed time for the metrics. I don't know if that would create an issue with a 0 for the time series because I could always have the value of the install/upgrade be the previous label if that case occurs in the scrape.

michael-cico:
I was going to have a similar comment to George's.  Should we reset the conditions array completely when we transition into a non-terminal ConditionType?  i.e., if it's "UpgradeStarted" or "InstallStarted, we clear out the array and before setting the new value/timestamp?

Not sure if there's a convention for this in the k8s docs.

mbenmclean:
I'm not sure. Wiping out the condition array seems like it would have issues as well.

stevshap:
Do we want to introduce the "Update" status (instead of just setting the condition as a flavor of "install")?

platform-operator/controllers/verrazzano/upgrade_component.go
pfmackin, line 133: Move line 132 (upgradeContext.state = compStateEnd) to after updateComponentStatus, in case it fails.  The status should not change if there is a possible error

platform-operator/controllers/verrazzano/upgrade_component.go
```diff
@@ -85,6 +85,9 @@ func (r *Reconciler) upgradeSingleComponent(spiCtx spi.ComponentContext, upgrade
 			if installed {
 				compLog.Oncef("Component %s is installed and will be upgraded", compName)
 				upgradeContext.state = compStatePreUpgrade
+				if err := r.updateComponentStatus(compContext, "Upgrade started", installv1alpha1.CondUpgradeStarted); err != nil {
```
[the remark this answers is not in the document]
mbenmclean: Sounds good

platform-operator/controllers/verrazzano/upgrade_component.go
mbenmclean: Sounds good

mbenmclean:
Steven we decided that we would not add an update condition for this change. Also, that wouldn't necessarily solve the problem George and I are discussing.

PR #3534, opened 2022-07-11
VZ-5339: Enable access logs for Authproxy

adiforluls:
This change is to enable access logs for Authproxy.
Fixes VZ-5339


tests/e2e/logging/system/system_logging_test.go
```diff
@@ -301,7 +301,7 @@ func validateAuthProxyLogs() bool {
 	}
 	exceptions = append(exceptions, istioExceptions...)
 	return validateElasticsearchRecords(
-		allElasticsearchRecordValidator,
+		noLevelElasticsearchRecordValidator,
```
kminder, line 304: Why was this change required?

adiforluls: `error_log` in NGINX records log level but not `access_log` it seems going by their docs (they have examples of specifying log severity level for error logs but not access logs).

PR #3537, opened 2022-07-11
VZ-6195 Verrazzano InstallComplete Status Fix 

absBitm:
Verrazzano install must set condition InstallComplete only when the component is really ready

### DIFF
--- a/platform-operator/controllers/verrazzano/component/nginx/nginx.go
+++ b/platform-operator/controllers/verrazzano/component/nginx/nginx.go
@@ -42,7 +42,7 @@ func isNginxReady(context spi.ComponentContext) bool {
 	}
 	prefix := fmt.Sprintf("Component %s", context.GetComponent())
 	ip, err := vzconfig.GetIngressIP(context.Client(), context.EffectiveCR())
-	log.Progressf("GetIngressIP : %s , error : %v", ip, err)
+	fmt.Println("GetIngressIP : , error : ", ip, err)
 	return err == nil && status.DeploymentsAreReady(context.Log(), context.Client(), deployments, 1, prefix)
 }
 
--- a/platform-operator/controllers/verrazzano/component/nginx/nginx.go
+++ b/platform-operator/controllers/verrazzano/component/nginx/nginx.go
@@ -41,7 +41,9 @@ func isNginxReady(context spi.ComponentContext) bool {
 		},
 	}
 	prefix := fmt.Sprintf("Component %s", context.GetComponent())
-	return status.DeploymentsAreReady(context.Log(), context.Client(), deployments, 1, prefix)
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -41,7 +41,8 @@ func isNginxReady(context spi.ComponentContext) bool {
 		},
 	}
 	prefix := fmt.Sprintf("Component %s", context.GetComponent())
-	return status.DeploymentsAreReady(context.Log(), context.Client(), deployments, 1, prefix)
+	_, err := vzconfig.GetIngressIP(context.Client(), context.EffectiveCR())
```
michael-cico, line 44: do we need to check the error here?

michael-cico: OK I see it's done on the return below.

platform-operator/controllers/verrazzano/component/nginx/nginx.go
```diff
@@ -41,7 +41,8 @@ func isNginxReady(context spi.ComponentContext) bool {
...
+	return status.DeploymentsAreReady(context.Log(), context.Client(), deployments, 1, prefix) && err == nil
```
michael-cico, line 45: You might want to do the `err == nil` check before the deployments check, otherwise we might do a deployment check for nothing.  If `err` is not `nil` it should short-circuit the conditional I believe.  Or you could do an error check and return false before the deployment check as an alternative, which might be more explicit.

michael-cico: i.e.,

```
if err != nil {
  return err
}
// then do deployment check
```

absBitm: it worked thanks

platform-operator/controllers/verrazzano/component/nginx/nginx_test.go
```diff
@@ -225,7 +246,17 @@ func TestIsNGINXNotReady(t *testing.T) {
...
+		Spec: vzapi.VerrazzanoSpec{
+			EnvironmentName: "myenv",
+			Components: vzapi.ComponentSpec{
+				Ingress: &vzapi.IngressNginxComponent{
+					Type: vzapi.LoadBalancer,
+				},
+			},
+		},
+	}
+	assert.False(t, isNginxReady(spi.NewFakeContext(fakeClient, vz, false)))
```
michael-cico, line 259: I'm wondering if we should have a separate test case for where the deployments are ready but the IP is not resolved yet.  In this test case both conditions are false.

michael-cico (approved):

PR #3543, opened 2022-07-12
VZ-5012: Activate OCI and OKE Driver in Rancher on install

lake-of-dreams:
Changes to activate the OCI and OKE driver in Rancher using Rest API, related tests and acceptance test


platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -274,3 +265,45 @@ func parseTokenResponse(resp *http.Response) (string, error) {
 	}
 	return tokenResponse.Token, nil
 }
+
+func (r *RESTClient) ActivateOCIDriver() error {
```
kminder, line 269: Function comments on all of these please.

platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -274,3 +265,45 @@ func parseTokenResponse(resp *http.Response) (string, error) {
...
+	resp, err := r.invokeRancherEndpoint(rancherActivateOCIDriverURLPath, http.MethodPost, nil)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed activating Rancher OCI Driver, url %s: %v", fmt.Sprintf("https://%s%s", r.hostname, rancherActivateOCIDriverURLPath), resp.Status)
```
kminder, line 276: Don't recall if we decided about following the golang standard of lower case messages or not.  Please check.  Is there any potential that this type of error should be throttled?  I don't recall if the standard is to log at the time of failure.  Can you please check with Paul.

platform-operator/controllers/verrazzano/component/common/rancher.go
lake-of-dreams: added

platform-operator/controllers/verrazzano/component/common/rancher.go
lake-of-dreams: Have made the messages lower case any way, @pfmackin you may review please.
On error handling, I just followed what other calls like [this](https://github.com/verrazzano/verrazzano/pull/3543/files#diff-808c0b10ae4a004d72be681d4faf058161451b13d032e7ce4eb509c0ec0bbf8aL241) are doing and I guess since we are using retryablehttp client, calls are retried anyway?

kminder:
Can you please investigate the use of the Rancher CRs to do this instead of using the APIs.  Using the CRs will be far more resilient.

PR #3549, opened 2022-07-12
VZ-6327: Move Velero uninstall to Go code

bobdonat:
This pull request moves the uninstall of Velero from the install scripts to Go code.

### DIFF
--- a/pkg/k8s/resource/resource.go
+++ b/pkg/k8s/resource/resource.go
@@ -5,8 +5,10 @@ package resource
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/k8s/resource/resource.go
```diff
@@ -30,3 +32,22 @@ func (r Resource) Delete() error {
 	}
 	return nil
 }
+
+// RemoveFinalizers remove all finalizers from a resource
```
pfmackin, line 36: not a blocker but can you add a follow up unit test

pfmackin (approved):

pkg/k8s/resource/resource.go
bobdonat: my Istio PR has the unit test for this

pkg/k8s/resource/resource.go
```diff
@@ -30,3 +32,22 @@ func (r Resource) Delete() error {
...
+func (r Resource) RemoveFinalizers() error {
+	val := reflect.ValueOf(r.Object)
+	kind := val.Elem().Type().Name()
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: r.Namespace, Name: r.Name}, r.Object)
+	if client.IgnoreNotFound(err) != nil {
+		return r.Log.ErrorfNewErr("Failed to get the resource of type %s, named %s/%s: %v", kind, r.Object.GetNamespace(), r.Object.GetName(), err)
+	} else if err == nil {
```
michael-cico, line 43: can this just be a simple `else` clause?  I think the `IgnoreNotFound` check is good enough ?

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
```diff
@@ -136,3 +139,21 @@ func (v veleroHelmComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Ve
...
+// PostUninstall processing for Velero
+func (v veleroHelmComponent) PostUninstall(context spi.ComponentContext) error {
+	res := resource.Resource{
+		Name:   ComponentNamespace,
+		Client: context.Client(),
+		Object: &corev1.Namespace{},
+		Log:    context.Log(),
+	}
+	// Remove finalizers from the velero namespace to avoid hanging namespace deletion
+	err := res.RemoveFinalizers()
```
michael-cico, line 152: Maybe it might be good to have a `res.ForceDelete()` method that removed the finalizers on the object as part of the cleanup.  Less duplicated code in other places.

pkg/k8s/resource/resource.go
davisgomes: If the error is `NotFound`, and it was only an `else` clause, then we would try to update an object that does not exist.

bobdonat: yes, I was just going to say the same thing

platform-operator/controllers/verrazzano/component/velero/velero_operator_component.go
bobdonat: One function that does both the remove of the finalizer and delete of the resource?

bobdonat: OK, if I do that in a followup PR?

davisgomes: Yea this would be nice to have since we often combine both

pfmackin: normally you want to only remove the finalizer if delete hangs or the controller that handles the resource has been uninstalled

PR #3552, opened 2022-07-12
VZ 6413: Merge component overrides

anders-swanson:


platform-operator/controllers/verrazzano/transform/testdata/merged.yaml
```diff
@@ -0,0 +1,23 @@
...
+  components:
+    rancher:
+      overrides:
+        - values:
+            affinity: {}
+        - values:
+            replicas: 1
+    prometheusOperator:
+      overrides:
+        - values:
```
jmaron99, line 14: so the merge yields two separate "values" maps?  I guess that works?

michael-cico (approved):

platform-operator/controllers/verrazzano/transform/testdata/merged.yaml
anders-swanson: The profile values are appended to the actual CR, then the strategic merge is applied (actual CR is a patch onto profile). This gets around the issues we have with strategic merge and the overrides API.

PR #3565, opened 2022-07-13
VZ-6322: Move verrazzano component uninstall to Go code

bobdonat:
This pull request moves the verrazzano component uninstall from the scripts to Go code.
The uninstall script included deleting of clusterroles/clusterrolebindings/secrets that were not needed so the equivalent is not in the Go code.


platform-operator/scripts/uninstall/uninstall-steps/1-uninstall-istio.sh
```diff
@@ -19,14 +19,6 @@ trap 'rc=$?; rm -rf ${TMP_DIR} || true; _logging_exit_handler $rc' EXIT
 CONFIG_DIR=$INSTALL_DIR/config
 
 function finalize() {
-  # Removing possible reference to verrazzano in clusterroles and clusterrolebindings
```
pfmackin, line 22: shouldn't we delete 1-uninstall-istio.sh since Istio is now uninstalled in go code?  Maybe a followup PR

michael-cico: yeah, probably delete this and the other one.

michael-cico (approved):

PR #3566, opened 2022-07-13
VPO Metrics Implementation

GeorgeAeillo:
- Implemented metrics that output the latest install and upgrade time of a given component based on the VZ CR
- Implemented metrics that give the reconcile counter, reconcile error count, and the duration of a reconcile (in summary format)
- The functions that populate these metrics are called in the reconcile function of the VPO
- The file that contains structs and helper functions is metricsexporter/metricsexporter.go 
- The file that contains other public and helper functions is metricsexporter/metricsexporter_utils.go
- Unit tests for these functions can be found in the metricsexporter/metricsexporter_test.go file and in the verrazzano/controller_test.go file 

Checklist

- [x]  Added or updated acceptance tests if appropriate


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -78,11 +79,14 @@ var unitTesting bool
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;watch;list;create;update;delete
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	// Get the Verrazzano resource
+	startTime := time.Now().UnixMilli()
+	defer metricsexporter.CollectReconcileMetricsTime(startTime)
```
mbenmclean, line 83: Can probably make these one line

mbenmclean: Unless this startTime var is used elsewhere

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -708,6 +717,7 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	return r.updateVerrazzanoStatus(log, cr)
 }
 
+//Test Comment
```
mbenmclean, line 720: Remove comment

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,20 +1,365 @@
...
+	grafanaInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_grafana_install_duration_seconds",
+		Help: "The duration of the latest installation of the grafana component in seconds",
+	})
+	coherenceInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_coherence_install_duration_seconds",
+		Help: "The duration of the latest installation of the coherence component in seconds",
+	})
+	mySQLInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_my_sql_install_duration_seconds",
```
mbenmclean, line 95: should be mysql, not my_sql

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,20 +1,365 @@
...
+	grafanaUpgradeTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_grafana_upgrade_duration_seconds",
+		Help: "The duration of the latest upgrade of the grafana component in seconds",
+	})
+	coherenceUpgradeTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_coherence_upgrade_duration_seconds",
+		Help: "The duration of the latest upgrade of the coherence component in seconds",
+	})
+	mySQLUpgradeTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_my_sql_upgrade_duration_seconds",
```
mbenmclean, line 199: should be mysql, not my_sql

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,20 +1,365 @@
...
+		Name: "vpo_reconcile_duration_seconds",
+		Help: "The duration of each reconcile call in the Verrazzano-platform-operator in seconds"},
+		[]string{"reconcile_index"},
+	)
+	reconcileErrorCounterMetric = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "vpo_error_reconcile_counter",
+		Help: "The number of times the reconcile function has returned an error in the Verrazzano-platform-operator",
+	})
+	verrazzanoAuthproxyInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_authproxy_install_duration_seconds",
```
mbenmclean, line 35: Not sure you need the vz_ in front of all these.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,20 +1,365 @@
...
+	prometheusNodeExporterInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_prometheus_node_exporter_install_duration_seconds",
+		Help: "The duration of the latest installation of the prometheus-node-exporter component in seconds",
+	})
+	jaegerOperatorInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_jaeger_operator_install_duration_seconds",
+		Help: "The duration of the latest installation of the jaeger-operator component in seconds",
+	})
+	verrazzanoConsoleInstallTimeMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "verrazzano_console_install_duration_seconds",
```
mbenmclean, line 131: just console

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+}
+func CollectReconcileMetricsError() {
+	reconcileErrorCounterMetric.Add(1)
+}
+
+func AnalyzeVZCR(CR vzapi.Verrazzano) {
+	//Get the VZ CR Component Map (Store it in this function, so the state does not change)
+	mapOfComponents := CR.Status.Components
+	for componentName, componentStatusDetails := range mapOfComponents {
+		possibleInstallCompletionTime := ""
```
mbenmclean, line 387: change this to `var possibleInstallCompletionTime string`
change the other ones too

mbenmclean: also probably don't need the possible at the front of the var name

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: I thought we were adding the vz in front of the VZ metrics, so the user would be able to differentiate the go runtime metrics from the metrics that VZ is generating? Should I still remove all together, or do something else in the name if we still want to indicate that these are VZ generated metrics

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Do we want to keep the verrazzano, so than that way the user will know that the metric was generated by us?

platform-operator/controllers/verrazzano/upgrade_test.go
```diff
@@ -1075,10 +1075,20 @@ func TestUpgradeComponentWithBlockingStatus(t *testing.T) {
...
+		}).AnyTimes()
 	mockComp.EXPECT().IsInstalled(gomock.Any()).Return(true, nil).AnyTimes()
 	mockComp.EXPECT().PreUpgrade(gomock.Any()).Return(nil).Times(1)
 	mockComp.EXPECT().Upgrade(gomock.Any()).Return(fmt.Errorf("Upgrade in progress")).AnyTimes()
 	mockComp.EXPECT().Name().Return("testcomp").Times(1).AnyTimes()
+	mockStatus.EXPECT().
+		Update(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, verrazzano *vzapi.Verrazzano, opts ...client.UpdateOption) error {
+			return nil
+		}).AnyTimes()
```
davisgomes, line 1091: Is this expect added twice? You should only need it once if it works `AnyTimes()`

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+			installCompletionInSecondsUnix := installCompletionInSeconds.Unix()
+			if installStartInSecondsUnix <= installCompletionInSecondsUnix {
+				totalDurationOfInstall := (installCompletionInSecondsUnix - installStartInSecondsUnix)
+				_, ok := installMetricsMap[componentName]
+				if ok {
+					installMetricsMap[componentName].Set(float64(totalDurationOfInstall))
+				}
+				if !ok {
+					zap.S().Errorf("Component %s does not have metrics in the install metrics map", componentName)
+				}
```
davisgomes, line 419: Usually, nested if statements is an anti-pattern in go. You might see a note in the auto check about cognitive complexity. I also see some duplication in this function, is there a way you could break this up and use the same function for install and upgrade?

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+			}
+			if status.Type == vzapi.CondUpgradeStarted {
+				UpgradeStartTime = status.LastTransitionTime
+			}
+			if status.Type == vzapi.CondUpgradeComplete {
+				UpgradeCompletionTime = status.LastTransitionTime
+			}
+		}
+		if InstallStartTime != "" && InstallCompletionTime != "" {
+			installStartInSeconds, _ := time.Parse(time.RFC3339, InstallStartTime)
```
mbenmclean, line 407: I would break this block up into another function and call it from here. It will help with congitive complexity.

mbenmclean: Also, the install and upgrade look very similar. You should be able to write a generic function that can do both and call it with parameters instead of repeating all this code.

mbenmclean: Also, in the function comment, can you include a detailed description of exactly what the function does. (ei: this function calculates these two values,  does this if value a is true and does this other thing if value b is false, that short of thing.) I'm finding this very difficult to read.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+					upgradeMetricsMap[componentName].Set(float64(totalDurationOfUpgrade))
+				}
+				if !ok {
+					zap.S().Errorf("Component %s does not have metrics in the upgrade metrics map", componentName)
+				}
+			}
+		}
+	}
+}
+func registerMetricsHandlersHelper() error {
```
mbenmclean, line 440: Please add a function comment for all these functions

platform-operator/metricsexporter/metricsexporter_test.go
```diff
@@ -0,0 +1,285 @@
...
+//THEN the function updates the reconcileCounterMetric by 1 and creates a new time for that reconcile in the reconcileLastDurationMetric
+func TestCollectReconcileMetrics(t *testing.T) {
+	assert := asserts.New(t)
+	tests := []struct {
+		name                   string
+		expectedIncrementValue float64
+	}{
+		{
+			name:                   "Test that reoncile counter is incremented by one when function is called",
+			expectedIncrementValue: float64(1),
```
mbenmclean, line 28: I'm not sure why there are 3 test cases here. They all have the same expectedIncrementValue, just with different names. What is different between them?

platform-operator/metricsexporter/metricsexporter_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	asserts "github.com/stretchr/testify/assert"
+	installv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+)
+
+//TestCollectReconcileMetrics tests the CollectReconcileMetrics fn
+//GIVEN a call to CollectReconcileMetrics
```
mbenmclean, line 17: There is no call to CollectReconcileMetrics. There is a call to CollectReconcileMetricsTime. Is that what you meant?

platform-operator/metricsexporter/metricsexporter_test.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+}
+
+//TestErrorCounterMetric tests the TestErrorCounterMetric fn
+//GIVEN a call to TestErrorCounterMetric
+//WHEN the function is called
+//THEN the function increments the reconcile error counter metric
+func TestErrorCounterMetric(t *testing.T) {
+	assert := asserts.New(t)
+	tests := []struct {
```
mbenmclean, line 62: You don't need to have a tests list with only one item. This is for when you have multiple items you want to test. Please remove the for loop from this test if you are only running one test case.

platform-operator/metricsexporter/metricsexporter_test.go
```diff
@@ -0,0 +1,285 @@
...
+			assert.Equal(tt.expectedErrorIncrementValue, errorCounterAfter-errorCounterBefore)
+		})
+	}
+}
+
+// TestAnalyzeVZCR tests the AnalyzeVZCR fn
+// GIVEN a call to AnalyzeVZCR
+// WHEN a VZ CR with or without timestamps is passed to the fn
+// THEN the function properly updates or does nothing to the component's metric
+func TestAnalyzeVZCR(t *testing.T) {
```
mbenmclean, line 85: This is a well written test. Good Job!

stevshap:
Comments should be formatted as:
// TestCollectReconcileMetrics...
not
//TestCollectReconcileMetrics...

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -186,7 +188,12 @@ func TestInstall(t *testing.T) {
 			// Create and make the request
 			request := newRequest(namespace, name)
 			reconciler := newVerrazzanoReconciler(mock)
+			reconcileCounterBefore := metricsexporter.GetReconcileCounterMetric()
 			result, err := reconciler.Reconcile(nil, request)
+			reconcileCounterAfter := metricsexporter.GetReconcileCounterMetric()
+			// Test the counter function here
```
stevshap, line 194: Don't think this comment is needed

platform-operator/metricsexporter/metricsexporter_test.go
stevshap: +1!

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+			upgradeCompletionInSeconds, _ := time.Parse(time.RFC3339, UpgradeCompletionTime)
+			upgradeCompletionInSecondsUnix := upgradeCompletionInSeconds.Unix()
+			if upgradeStartInSecondsUnix <= upgradeCompletionInSecondsUnix {
+				totalDurationOfUpgrade := (upgradeCompletionInSecondsUnix - upgradeStartInSecondsUnix)
+				_, ok := upgradeMetricsMap[componentName]
+				if ok {
+					upgradeMetricsMap[componentName].Set(float64(totalDurationOfUpgrade))
+				}
+				if !ok {
+					zap.S().Errorf("Component %s does not have metrics in the upgrade metrics map", componentName)
```
michael-cico, line 434: We should probably be using the VerrazzanoLogger where possible.

michael-cico: you can pass that in from the controller as the first arg to this func, and do the same with the other funcs.

platform-operator/metricsexporter/metricsexporter_test.go
GeorgeAeillo: There is not anything different between them, I will change them that function to be only one test case.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -23,3 +368,103 @@ func InitalizeMetricsEndpoint() {
...
+}
+func CollectReconcileMetricsError() {
+	reconcileErrorCounterMetric.Add(1)
+}
+
+func AnalyzeVZCR(CR vzapi.Verrazzano) {
+	//Get the VZ CR Component Map (Store it in this function, so the state does not change)
+	mapOfComponents := CR.Status.Components
+	for componentName, componentStatusDetails := range mapOfComponents {
+		var InstallCompletionTime string = ""
```
michael-cico, line 387: thes vars should all be camel case probably.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,20 +1,365 @@
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// PreInstall called multiple times but only successfully completed once
-//
+var (
```
michael-cico, line 19: another nit, might be good to move all of the metrics definitions/structures to a separate file, to distinguish them from the public/exported functions.  Consumers of the package will be interested in the exported functions/interfaces/objects, plus it will just make things easier to consume visually.

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Thanks! I passed in the VZ logger into that func. In the other functions, a VZ logger either isn't created, or has the possibility of not being created, so a zap logger was passed instead.

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Thanks for your feedback! In my most recent commit, I split up the original file into a metricsexporter internal file and a metricsexporter public function file.

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+		Help: fmt.Sprintf("The duration of the latest installation of the %s component in seconds", name),
+	})
+	m.LatestUpgradeDuration = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: fmt.Sprintf("vz_%s_upgrade_duration_seconds", name),
+		Help: fmt.Sprintf("The duration of the latest upgrade of the %s component in seconds", name),
+	})
+}
+
+// This function creates and populates a MetricComponent struct
+func initalizeMetricsComponent(name string) metricComponent {
```
anders-swanson, line 33: What's the point of having this function and the `init` function?

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+		[]string{"reconcile_index"},
+	)
+	reconcileErrorCounterMetric = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "vpo_error_reconcile_counter",
+		Help: "The number of times the reconcile function has returned an error in the Verrazzano-platform-operator",
+	})
+	// The metricsMap keys are the names of the components as they appear in the VZ CR
+	metricsMap = map[string]metricComponent{
+		"verrazzano-authproxy":            authproxyMetrics,
+		"oam-kubernetes-runtime":          oamMetrics,
```
anders-swanson, line 96: can you save some screen space by saying:
`"oam-kubernetes-runtime": initializeMetricsComponent("oam")`?

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+
+	registry = prometheus.DefaultRegisterer
+)
+
+// This function is used to determine whether a durationTime for a component metric should be set and what the duration time is
+// If the start time is greater than the completion time, the metric will not be set
+// After this check, the function calculates the duration time and tries to set the metric of the component
+// If the component's name is not in the metric map, an error will be raised to prevent a seg fault
+func metricParserHelperFunction(log vzlog.VerrazzanoLogger, componentName string, startTime string, completionTime string, typeofOperation string) {
+	startInSeconds, _ := time.Parse(time.RFC3339, startTime)
```
anders-swanson, line 138: Why are we ignoring the errors for `time.Parse`?

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+	if startInSecondsUnix >= completionInSecondsUnix {
+		return
+	}
+	totalDuration := (completionInSecondsUnix - startInSecondsUnix)
+	_, ok := metricsMap[componentName]
+	if !ok {
+		log.Errorf("Component %s does not have metrics in the metrics map", componentName)
+		return
+	}
+	if typeofOperation == "upgrade" {
```
anders-swanson, line 151: You can use a typed value for `typeOfOperation`. Something like:

```
type metricsOperation string
const (
	operationInstall metricsOperation = "install"
	operationUpgrade metricsOperation = "upgrade"
)

func metricParserHelperFunction(log vzlog.VerrazzanoLogger, componentName string, startTime string, completionTime string, operation metricsOperation) {
 ...
}
```

This is going to force callers to pass in a value of type `metricsOperation`. It's a similar concept to an Enum in other languages.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: The initialize function creates a struct while the init function populates it. I tried having the initialization and population on the same line, but was running into errors.

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+	for k := range metricsMap {
+		metricsList = append(metricsList, metricsMap[k].LatestInstallDuration)
+		metricsList = append(metricsList, metricsMap[k].LatestUpgradeDuration)
+	}
+	metricsList = append(metricsList, reconcileCounterMetric, reconcileErrorCounterMetric, reconcileLastDurationMetric)
+	return metricsList
+}
+
+var (
+	authproxyMetrics              = initalizeMetricsComponent("authproxy")
```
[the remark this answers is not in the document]
anders-swanson: Did you consider using a struct to wrap these so they are cohesively bound to a single object?

platform-operator/metricsexporter/metricsexporter_internal.go
davisgomes: Here's an example in the uninstaller: https://github.com/verrazzano/verrazzano/blob/a59b79bff5fc590d1e51ab948b64db1532ffe145/platform-operator/controllers/verrazzano/uninstall.go#L28

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+}
+func registerMetricsHandlersHelper(log *zap.SugaredLogger) error {
+	var errorObserved error
+	for metric, i := range failedMetrics {
+		err := registry.Register(metric)
+		if err != nil {
+			log.Errorf("Failed to register metric index %v for VPO", i)
+			errorObserved = err
```
anders-swanson, line 165: What happens if you get an error while registering the metric in two or more separate loop iterations? It looks like the first error will be ignored.

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+// This function is used to determine whether a durationTime for a component metric should be set and what the duration time is
+// If the start time is greater than the completion time, the metric will not be set
+// After this check, the function calculates the duration time and tries to set the metric of the component
+// If the component's name is not in the metric map, an error will be raised to prevent a seg fault
+func metricParserHelperFunction(log vzlog.VerrazzanoLogger, componentName string, startTime string, completionTime string, typeofOperation string) {
+	startInSeconds, _ := time.Parse(time.RFC3339, startTime)
+	startInSecondsUnix := startInSeconds.Unix()
+	completionInSeconds, _ := time.Parse(time.RFC3339, completionTime)
+	completionInSecondsUnix := completionInSeconds.Unix()
+	if startInSecondsUnix >= completionInSecondsUnix {
```
anders-swanson, line 142: What does it mean if this statement is true? Is that certain to be impossible?

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+			delete(failedMetrics, metric)
+		}
+	}
+	return errorObserved
+}
+func registerMetricsHandlers(log *zap.SugaredLogger) {
+	initializeFailedMetricsArray() // Get list of metrics to register initially
+	// loop until there is no error in registering
+	for err := registerMetricsHandlersHelper(log); err != nil; err = registerMetricsHandlersHelper(log) {
+		log.Errorf("Failed to register some metrics for VPO: %v", err)
```
anders-swanson, line 177: Will this generate multiple error messages for the same error? E.g., a message on L164 when the metric fails to register, and then here on L177 when `registerMetricsHandlers` receives the error returned by `registerMetricsHandlersHelper`?

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: There is a possibility where the start time for a new install operation is written, while the completion time for the older install operation is currently in the conditions as well. In this case, we want to keep the old metric that we have for the install and wait until the conditions for the new operations fully get written.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter_public_functions.go
```diff
@@ -0,0 +1,94 @@
...
+			log.Errorf("Failed to start metrics server for verrazzano-platform-operator: %v", err)
+		}
+	}, time.Second*3, wait.NeverStop)
+}
+
+// This function increments the reconcile counter metric and calculates the duration time for a reconcile
+// It assigns the duration time to a gauge vector with the key in the gauge vector being the current reconcile index
+// Once the duration time is set, the reconcile index is incremented
+// Could put a log but debug level
+func CollectReconcileMetricsTime(startTime int64, log *zap.SugaredLogger) {
```
anders-swanson, line 35: Suggest to have startTime be a `time.Time` value. Then you do not have to worry about the unit when doing time arithmetic. ---> `durationTime := time.Now.Sub(startTime).Seconds() // Or whatever time unit you need`

platform-operator/metricsexporter/metricsexporter_internal.go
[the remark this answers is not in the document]
anders-swanson: The typical way to do this would be to create & inject the object at program startup, so it is present in any containing objects that call its methods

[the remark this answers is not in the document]
anders-swanson: The struct would need the metrics map

platform-operator/metricsexporter/metricsexporter_internal.go
```diff
@@ -0,0 +1,185 @@
...
+		Help: "The duration of each reconcile call in the Verrazzano-platform-operator in seconds"},
+		[]string{"reconcile_index"},
+	)
+	reconcileErrorCounterMetric = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "vpo_error_reconcile_counter",
+		Help: "The number of times the reconcile function has returned an error in the Verrazzano-platform-operator",
+	})
+	// The metricsMap keys are the names of the components as they appear in the VZ CR
+	metricsMap = map[string]metricComponent{
+		"verrazzano-authproxy":            authproxyMetrics,
```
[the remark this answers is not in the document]
anders-swanson: if you type `istio.ComponentName` it's a string that holds the value "istio"

platform-operator/metricsexporter/metricsexporter_internal.go
anders-swanson: it's generally best to not make such assumptions. What happens if this is called with a value that for some reason doesn't parse (maybe we cannot think of the reason until it happens)

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: Yep! Thanks!

platform-operator/metricsexporter/metricsexporter_internal.go
anders-swanson: could you put a log.Debug here that succinctly says something along those lines?

platform-operator/metricsexporter/metricsexporter_public_functions.go
```diff
@@ -0,0 +1,94 @@
...
+		if installStartTime != "" && installCompletionTime != "" {
+			metricParserHelperFunction(log, componentName, installStartTime, installCompletionTime, "install")
+		}
+		if upgradeStartTime != "" && upgradeCompletionTime != "" {
+			metricParserHelperFunction(log, componentName, upgradeStartTime, upgradeCompletionTime, "upgrade")
+		}
+	}
+}
+func GetErrorCounterMetric() float64 {
+	return testutil.ToFloat64(reconcileErrorCounterMetric)
```
anders-swanson, line 84: Why are we using `testutil` package outside of a test?

platform-operator/metricsexporter/metricsexporter_internal.go
zabdulre: The error will not be logged explicitly, you are right. This can be switched pretty easily to log the error for all iterations, do you think that would be good?

platform-operator/metricsexporter/metricsexporter_public_functions.go
anders-swanson: From the `testutil.ToFloat64` function doc:
```
// Use this function with caution. It is computationally very expensive and thus
// not suited at all to read values from Metrics in regular code. This is really
// only for testing purposes, and even for testing, other approaches are often
// more appropriate (see this package's documentation).
```

platform-operator/metricsexporter/metricsexporter_internal.go
zabdulre: This logs the error while the helper points to the metric which failed. But I think you are right, it would be better to have one comprehensive log statement

platform-operator/metricsexporter/metricsexporter_public_functions.go
```diff
@@ -0,0 +1,94 @@
...
+
+			}
+			if status.Type == vzapi.CondUpgradeStarted {
+				upgradeStartTime = status.LastTransitionTime
+			}
+			if status.Type == vzapi.CondUpgradeComplete {
+				upgradeCompletionTime = status.LastTransitionTime
+			}
+		}
+		if (installStartTime == "" || installCompletionTime == "") && (upgradeStartTime == "" || upgradeCompletionTime == "") {
```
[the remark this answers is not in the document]
GeorgeAeillo: That is a logic bug, that should have been a break, I believe, since I want it to skip to the next component Name.

platform-operator/metricsexporter/metricsexporter_internal.go
anders-swanson: suggest to collect all the failed metrics as you loop through and return them all as an error like "registration failed for metrics [x, y, z]"

anders-swanson: And only emit a log message from higher on the stack

platform-operator/metricsexporter/metricsexporter_internal.go
anders-swanson: A common pattern in go is something like:
```
func NewMetricComponent(name string) *metricComponent {
    return &metricComponent{
        LatestInstallDuration: prometheus.NewGauge(prometheus.GaugeOpts{
		Name: fmt.Sprintf("vz_%s_install_duration_seconds", name),
		Help: fmt.Sprintf("The duration of the latest installation of the %s component in seconds", name),
	}),
        LatestUpgradeDuration: prometheus.NewGauge(prometheus.GaugeOpts{
		Name: fmt.Sprintf("vz_%s_upgrade_duration_seconds", name),
		Help: fmt.Sprintf("The duration of the latest upgrade of the %s component in seconds", name),
	}),
    }
}
```

platform-operator/metricsexporter/metricsexporter_internal.go
zabdulre: That would be really good, just one thing is that each metric has its own unique error message. Should I not print each metrics' error maybe?

platform-operator/metricsexporter/metricsexporter_public_functions.go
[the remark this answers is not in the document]
anders-swanson: If this should be a break, if the break necessary? The subsequent conditionals would handle it?

platform-operator/metricsexporter/metricsexporter_internal.go
anders-swanson: the `errors.Wrap` function may provide the functionality needed here to wrap multiple errors. I assuming once you encounter and error, you still want to continue processing the rest of the loop?

zabdulre: yes that is the plan

zabdulre: Addressed in newest version :)

platform-operator/metricsexporter/metricsexporter_public_functions.go
GeorgeAeillo: Thanks! I changed it to a Time.Time value, however I will still have to divide the number of milliseconds by 1000, as Prometheus wants base units emitted, so instead of outputting 2 milliseconds, I would output 0.002 seconds.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: Thanks! I just did.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter_public_functions.go
GeorgeAeillo: Those functions are being used in controller_test.go. The reason that they are in the external file is that the controller_test.go is a different file. How should I resolve this?

anders-swanson: Well, I really do recommend to use a struct to wrap the metrics stuff. If that was the case, you could instantiate an instance of that struct in the test suite.
If there is a reason you cannot do that or are adverse to doing that, you need to make public Getters for the package level variables, or make the package level variables public.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: I updated my error with Zaid's error fixes.

platform-operator/metricsexporter/metricsexporter_internal.go
GeorgeAeillo: I updated my error handling with Zaid's error fixes.

platform-operator/metricsexporter/metricsexporter_internal.go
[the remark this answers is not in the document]
GeorgeAeillo: Thanks! I added it.

platform-operator/metricsexporter/metricsexporter_public_functions.go
GeorgeAeillo: I made a struct to wrap the metrics and instantiated an instance of the struct per test function. For the test functions that are called by external packages, they cannot be declared in a test file. Davis and I tried this, but it did not work.

platform-operator/metricsexporter/metricsexporter_public_functions.go
```diff
@@ -0,0 +1,94 @@
...
+func CollectReconcileMetricsTime(startTime int64, log *zap.SugaredLogger) {
+	reconcileCounterMetric.Add(float64(1))
+	durationTime := (float64(time.Now().UnixMilli() - startTime)) / 1000.0
+	metric, _ := reconcileLastDurationMetric.GetMetricWithLabelValues(strconv.Itoa(reconcileIndex))
+	metric.Set(float64(durationTime))
+	log.Debugf("Time duration metric updated with label %v", reconcileIndex)
+	reconcileIndex = reconcileIndex + 1
+}
+func CollectReconcileMetricsError(log *zap.SugaredLogger) {
+	reconcileErrorCounterMetric.Add(1)
```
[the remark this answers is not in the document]
GeorgeAeillo: I combined all of the package level variables into one struct, so there is only one package level variable.

platform-operator/metricsexporter/metricsexporter_public_functions.go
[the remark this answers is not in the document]
GeorgeAeillo: It should be continue, as I want the next iteration of the loop to execute, not for it to execute out of the loop completely.

platform-operator/metricsexporter/metricsexporter_public_functions.go
anders-swanson: Do something like this:
```
// in public functions file
func GetErrorCounterMetric() {
  return reconcileErrorCounterMetric
}

// in test file
floatVal := testuil.ToFloat64(GetErrorCounterMetric)
```

GeorgeAeillo: Thanks! I just implemented this

platform-operator/metricsexporter/metricsexporter_public_functions.go
GeorgeAeillo: Thanks! I just added a constant for the number of milliseconds in a second.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,90 @@
...
+	durationMetricMap      map[metricName]*DurationMetric
+	metricsComponentMap    map[metricName]*MetricsComponent
+}
+type SimpleCounterMetric struct {
+	metric prometheus.Counter
+}
+
+func (c *SimpleCounterMetric) Inc(log *zap.SugaredLogger, err error) {
+	c.metric.Inc()
+	if err != nil {
```
anders-swanson, line 34: Why are you passing in an error and logging it?

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,90 @@
...
-//InitalizeMetricsEndpoint creates and serves a /metrics endpoint at 9100 for Prometheus to scrape metrics from
-func InitalizeMetricsEndpoint() {
-	go wait.Until(func() {
-		http.Handle("/metrics", promhttp.Handler())
-		err := http.ListenAndServe(":9100", nil)
-		if err != nil {
-			zap.S().Errorf("Failed to start metrics server for verrazzano-platform-operator: %v", err)
-		}
-	}, time.Second*3, wait.NeverStop)
+type MetricsExporter struct {
```
anders-swanson, line 11: FYI in go you can declare multiple types in a `type` block like this:
```
type (
  a struct {}

  b struct {}
  /// etc.
)
```

It's not required, just another way to express types.

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: If the error counter is being incremented due to an error, this would log the error. Should I assume that those errors in the reconcile function are being logged elsewhere?

platform-operator/metricsexporter/metricsexporter_external.go
```diff
@@ -0,0 +1,359 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package metricsexporter
```
anders-swanson, line 4: I am confused on the distinction between metricsexporter.go and metricsexporter_external.go - they both have public and private members, what is 'external' supposed to mean in this context?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -74,11 +75,30 @@ var unitTesting bool
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;watch;list;create;update;delete
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	// Get the Verrazzano resource
+	zapLogForMetrics := zap.S()
```
davisgomes, line 78: You can consider initializing the logger with this controller using zap.S().With(controllerName)

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -74,11 +75,30 @@ var unitTesting bool
...
+	// Error is defined is
```
davisgomes, line 79: fix comment

platform-operator/metricsexporter/metricsexporter_external.go
GeorgeAeillo: The metricsexporter.go defines the structs and the member functions of the structs while the external defines everything else. Should I change it to metricsexporter_structs.go and metricsexporter.go?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -74,11 +75,30 @@ var unitTesting bool
...
+	counterMetricObject, err := metricsexporter.GetSimpleCounterMetric(metricsexporter.ReconcileCounter)
+	if err != nil {
+		zapLogForMetrics.Error(err)
+	}
+	counterMetricObject.Inc(zapLogForMetrics, err)
```
davisgomes, line 84: It's not clear why you're passing in the error. Is this an error metrics? If so, maybe the variable name should reflect that.

davisgomes: I see it's not the error counter. Does this require the passing of an error?

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -187,7 +191,13 @@ func TestInstall(t *testing.T) {
 			// Create and make the request
 			request := newRequest(namespace, name)
 			reconciler := newVerrazzanoReconciler(mock)
+			reconcileCounterMetric, err := (metricsexporter.GetSimpleCounterMetric(metricsexporter.ReconcileCounter))
```
davisgomes, line 194: Extra braces I believe are unnecessary

platform-operator/metricsexporter/metricsexporter.go
davisgomes: The error should be logged where it happens, so I think this log is unnecessary.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,90 @@
...
+func (d *DurationMetric) TimerStop() {
+	d.timer.ObserveDuration()
+}
+
+type MetricsComponent struct {
+	latestInstallDuration *SimpleGaugeMetric
+	latestUpgradeDuration *SimpleGaugeMetric
+}
+
+func (m *MetricsComponent) getInstall() *SimpleGaugeMetric {
```
davisgomes, line 85: Maybe adding duration to these names would make it more clear what they do (i.e. `getInstallDuration()`

platform-operator/metricsexporter/metricsexporter_test.go
```diff
@@ -0,0 +1,294 @@
...
+		expectedErrorIncrementValue: float64(1),
+	}
+	t.Run(test.name, func(t *testing.T) {
+		reconcileErrorCounterObject, err := GetSimpleCounterMetric(ReconcileError)
+		assert.NoError(err)
+		reconcileErrorCounterBefore := testutil.ToFloat64(reconcileErrorCounterObject.Get())
+		reconcileErrorCounterObject.Inc(logForTest, nil)
+		reconcileErrorCounterAfter := testutil.ToFloat64(reconcileErrorCounterObject.Get())
+		assert.Equal(test.expectedErrorIncrementValue, reconcileErrorCounterAfter-reconcileErrorCounterBefore)
+	})
```
davisgomes, line 75: Can you combine these functions into one. Right now it's kind of wasting the for loop formatting by splitting it into two tests

platform-operator/metricsexporter/metricsexporter.go
anders-swanson: This function is not handling the error, nor is it recording it in any way. Logging is a side effect that should be expressed where the error actually occurs.

platform-operator/metricsexporter/metricsexporter_external.go
davisgomes: These look like utility functions to the struct. A typical file naming format we use for cases like this is `metricsexporter.go` which contains the struct definition and member functions and `metricsexporter_utils.go` which handles non-member functions

platform-operator/metricsexporter/metricsexporter_external.go
```diff
@@ -0,0 +1,359 @@
...
+		log.Debug("Component %s is not updated, as there is an ongoing operation in progress")
+		return nil
+	}
+	totalDuration := (completionInSecondsUnix - startInSecondsUnix)
+	_, ok := MetricsExp.internalData.metricsComponentMap[componentName]
+	if !ok {
+		log.Errorf("Component %s does not have metrics in the metrics map", componentName)
+		return nil
+	}
+	if typeofOperation == operationUpgrade {
```
davisgomes, line 217: Could you use global constants for this check? I think there is a global variable for the operation install and upgrade

platform-operator/controllers/verrazzano/controller.go
GeorgeAeillo: Something like this zapLogForMetrics := zap.S().With("verrazzano controller")?

GeorgeAeillo: There is a verrazzano logger created later in the code, but I didn't want to use it because there is a possibility that the function could return and it not be created.

platform-operator/metricsexporter/metricsexporter_external.go
GeorgeAeillo: Thanks! I changed it to `metricsexporter_utils.go`

platform-operator/controllers/verrazzano/controller.go
davisgomes: Yes. we only log errors at the source

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -74,11 +75,30 @@ var unitTesting bool
...
+	errorCounterMetricObject, err := metricsexporter.GetSimpleCounterMetric(metricsexporter.ReconcileError)
+	if err != nil {
+		zapLogForMetrics.Error(err)
+	}
+
+	reconcileDurationMetricObject, err := metricsexporter.GetDurationMetric(metricsexporter.ReconcileDuration)
```
anders-swanson, line 90: 1. Is there any error handling that we need to do here? Logging the error and continuing can potentially cause panics.
2. Is there any reason we need to create these every reconcile run? Can they not be injected into the controller object?

platform-operator/controllers/verrazzano/controller.go
anders-swanson: See my other comment on the function definition. Errors should not be handled this way.

platform-operator/metricsexporter/metricsexporter_external.go
```diff
@@ -0,0 +1,359 @@
...
+		return fmt.Errorf("error in parsing completion time %s for operation %s for component %s", completionTime, typeofOperation, componentName)
+	}
+	completionInSecondsUnix := completionInSeconds.Unix()
+	if startInSecondsUnix >= completionInSecondsUnix {
+		log.Debug("Component %s is not updated, as there is an ongoing operation in progress")
+		return nil
+	}
+	totalDuration := (completionInSecondsUnix - startInSecondsUnix)
+	_, ok := MetricsExp.internalData.metricsComponentMap[componentName]
+	if !ok {
```
anders-swanson, line 213: typically validation like this would go at the start of a function

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -97,6 +117,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		ControllerName: "verrazzano",
 	})
 	if err != nil {
+		errorCounterMetricObject.Inc(zapLogForMetrics, err)
```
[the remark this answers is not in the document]
GeorgeAeillo: No, the object will not be nil, but a dummy metric variable will be created each time that it runs, so that way reconcile will continue to operate and run as normal and not be dependent on metric failure. Since I am enforcing that a type of metricName is passed in as a parameter to GetSimpleCounterMetric, this will only occur if a developer adds a metric name as a constant in the metricsexporter_utils.go file, but does not add the corresponding metric. Previously, I had the function returning an error at this point instead of using a dummy metric that is not registered, should I revert back to that implementation?

platform-operator/metricsexporter/metricsexporter_external.go
```diff
@@ -0,0 +1,359 @@
...
+
+			}
+			if status.Type == vzapi.CondUpgradeStarted {
+				upgradeStartTime = status.LastTransitionTime
+			}
+			if status.Type == vzapi.CondUpgradeComplete {
+				upgradeCompletionTime = status.LastTransitionTime
+			}
+		}
+		if (installStartTime == "" || installCompletionTime == "") && (upgradeStartTime == "" || upgradeCompletionTime == "") {
```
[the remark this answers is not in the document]
GeorgeAeillo: Thanks!

platform-operator/controllers/verrazzano/controller.go
GeorgeAeillo: Thanks! I edited this member function for this object, so that it will not accept a log and errors anymore.

platform-operator/metricsexporter/metricsexporter_external.go
```diff
@@ -0,0 +1,359 @@
...
+		},
+	}
+}
+
+func initSimpleCounterMetricMap() map[metricName]*SimpleCounterMetric {
+	return map[metricName]*SimpleCounterMetric{
+		ReconcileCounter: {
+			prometheus.NewCounter(prometheus.CounterOpts{
+				Name: "vpo_reconcile_counter",
+				Help: "The number of times the reconcile function has been called in the Verrazzano-platform-operator",
```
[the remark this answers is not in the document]
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Thanks! I fixed this function's parameters.

platform-operator/controllers/verrazzano/controller.go
GeorgeAeillo: A dummy metric is created each reconcile run, so a panic will not be created. The reconcile duration metrics are not being created every reconcile run, they are only be creating once and a pointer is being returned to access them. What do you mean by injecting them into the controller object?

platform-operator/metricsexporter/metricsexporter_external.go
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Thanks!

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Cool! I did not know that.

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -187,7 +191,13 @@ func TestInstall(t *testing.T) {
...
+			asserts.NoError(err)
+			reconcileCounterBefore := testutil.ToFloat64(reconcileCounterMetric.Get())
 			result, err := reconciler.Reconcile(nil, request)
```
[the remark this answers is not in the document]
GeorgeAeillo: Do you mean assert no error with the other assert statements in that function?

platform-operator/metricsexporter/metricsexporter_test.go
GeorgeAeillo: I combined them into one test.

platform-operator/controllers/verrazzano/controller_test.go
GeorgeAeillo: I edited the extra braces out.

platform-operator/controllers/verrazzano/controller.go
GeorgeAeillo: I edited the log to have the name of controller, there wasn't a name of a constant, so I just did "verrazzano".

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/prometheus/client_golang/prometheus/testutil"
```
stevshap, line 10: these imports need to be cleaned up

stevshap: "context"
"fmt"
"sync"
"testing"
"time"

should be grouped together at the top

stevshap: and then the rest of the imports grouped underneath. see certmanager_component.go as an example

GeorgeAeillo: Thanks!

davisgomes (dismissed):
Approved with a few follow up comments that can be fixed in the next PR:
- [platform-operator/metricsexporter/metricsexporter.go](https://github.com/verrazzano/verrazzano/pull/3566/files#diff-7c3033608fd923571995b3f5047517eabd71bf34b3de6ccf0e781ccd728f4d97) lines 16-17: please put those comments above the decalration, I think they're a bit too long to be inline
- Please add comments to all functions giving a simple description of their functionality

platform-operator/controllers/verrazzano/controller_test.go
```diff
@@ -6,24 +6,23 @@ package verrazzano
...
+	clustersapi "github.com/verrazzano/verrazzano/platform-operator/apis/clusters/v1alpha1"
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	helm2 "github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/helm"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/registry"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/rbac"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
+	"github.com/verrazzano/verrazzano/platform-operator/metricsexporter"
 	"github.com/verrazzano/verrazzano/platform-operator/mocks"
```
stevshap, line 27: no new line

GeorgeAeillo: Thanks! I just edited it.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,105 @@
...
-		if err != nil {
-			zap.S().Errorf("Failed to start metrics server for verrazzano-platform-operator: %v", err)
-		}
-	}, time.Second*3, wait.NeverStop)
+type MetricsExporter struct {
+	internalConfig configuration
+	internalData   data
+}
+
+// thisMetric array will be automatically populated with all the metrics from each map. Metrics not included in a map can be added to thisMetric array for registration.
```
mbenmclean, line 15: I don't see a thisMetric array declared anywhere. Is this comment old?

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,105 @@
...
+// //thisMetric map will be automatically populated with all metrics which were not registered correctly. Metrics in thisMetric map will be retried periodically.
```
mbenmclean, line 16: I don't see a thisMetric map declared anywhere. Is this comment old?

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -1,25 +1,105 @@
...
+func (c *SimpleCounterMetric) Get() prometheus.Counter {
+	return c.metric
+}
+
+type SimpleGaugeMetric struct {
+	metric prometheus.Gauge
+}
+
+// This member function sets a SimpleGaugeMetric to a user provided float64 number
+
```
mbenmclean, line 54: delete extra lines

platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -0,0 +1,370 @@
...
+		var upgradeCompletionTime string
+		var upgradeStartTime string
+		var installStartTime string
+		for _, status := range componentStatusDetails.Conditions {
+			if status.Type == vzapi.CondInstallStarted {
+				installStartTime = status.LastTransitionTime
+			}
+			if status.Type == vzapi.CondInstallComplete {
+				installCompletionTime = status.LastTransitionTime
+
```
mbenmclean, line 295: delete extra line

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Yes, those comments were old.

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Yes, that comment was old.

platform-operator/metricsexporter/metricsexporter.go
GeorgeAeillo: Thanks! I deleted them.

platform-operator/metricsexporter/metricsexporter_utils.go
GeorgeAeillo: Thanks! I deleted it.

michael-cico (approved):

PR #3579, opened 2022-07-14
VZ-6418.  Remove uninstall job.

michael-cico:


pfmackin (approved):

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -433,32 +401,6 @@ func (r *Reconciler) deleteServiceAccount(ctx context.Context, log vzlog.Verrazz
...
-	} else if err != nil {
-		log.Errorf("Failed to get install cluster role binding %s: %v", binding.Name, err)
-		return err
-	}
-
-	return nil
-}
-
 // deleteClusterRoleBinding deletes the cluster role binding
 func (r *Reconciler) deleteClusterRoleBinding(ctx context.Context, log vzlog.VerrazzanoLogger, vz *installv1alpha1.Verrazzano) error {
```
bobdonat, line 405: can this function be removed

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -534,54 +476,6 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-
-	if err := r.setUninstallCondition(log, jobFound, vz); err != nil {
-		return err
-	}
-
-	return nil
-}
-
 // buildUninstallJobName returns the name of an uninstall job based on Verrazzano resource name.
 func buildUninstallJobName(name string) string {
```
bobdonat, line 480: This function should be able to be removed

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -389,34 +385,6 @@ func (r *Reconciler) doesOCIDNSConfigSecretExist(vz *installv1alpha1.Verrazzano)
...
-	} else if err != nil {
-		log.Errorf("Failed to get install service account %s: %v", buildServiceAccountName(vz.Name), err)
-		return err
-	}
-
-	return nil
-}
-
 // deleteServiceAccount deletes the service account used for install
 func (r *Reconciler) deleteServiceAccount(ctx context.Context, log vzlog.VerrazzanoLogger, vz *installv1alpha1.Verrazzano, namespace string) error {
```
bobdonat, line 389: is this function still needed

platform-operator/controllers/verrazzano/controller.go
michael-cico: we can't we need it to clean up any old jobs during upgrade I think, right @pfmackin ?

platform-operator/controllers/verrazzano/controller.go
michael-cico: Same comment as for the other, I think we need this to clean up any old jobs.

platform-operator/controllers/verrazzano/controller.go
michael-cico: same comment

PR #3586, opened 2022-07-14
VZ-6437: Velero namespace change

theabmitra:
Today velero install happens in a namespace called velero. This could collide with another install of velero if its been installed out of band. 

We will be installing velero in verrazzano-backup namespace.


kminder (dismissed):
Approved.

kminder (approved):
Approved.

PR #3614, opened 2022-07-18
VZ-6267 HA for NGINX

mbenmclean:
This change implements HA for NGINX. This involves adding autoscaling and affinity for the ingress controller, and increasing the replicas and adding affinity for the default backend. This change also modifies one of the update e2e tests to use overrides instead of install args.


platform-operator/controllers/verrazzano/component/spi/component_context_test.go
```diff
@@ -161,6 +162,7 @@ func TestContextProfilesMerge(t *testing.T) {
 			a.Equal(test.actualCR, *context.ActualCR(), "Actual CR unexpectedly modified")
 			a.NotNil(context.EffectiveCR(), "Effective CR was nil")
 			a.Equal(vzapi.VerrazzanoStatus{}, context.EffectiveCR().Status, "Effective CR status not empty")
+			fmt.Println(context.EffectiveCR())
```
mgianatagh, line 165: was this line meant to be removed?

mbenmclean: Oh yeah I'll fix

PR #3622, opened 2022-07-19
VZ-6251: Include workloads for all the Verrazzano components in the bug-report

bhatpmk:
This PR updated the bug-report command as below -
1. Capture the workloads for all the Verrazzano components
2. The flag --report-file is optional now, if not specified the command attempts to create the report with the name bug-report.tar.gz under the current directory
3.  The json file containing the Verrazzano resource is renamed from verrazzano_resources.json to verrazzano-resources.json
4. The error messages when the command fails to get one or more resources, are redirected to standard error

Contributes to VZ-6251


agv426 (approved):

bhatpmk:
VZ-6251

PR #3623, opened 2022-07-19
VZ-5012: Replicate logic of oci/oke driver activation in Rancher PostUpgrade

lake-of-dreams:
In https://github.com/verrazzano/verrazzano/pull/3543, some code was added to [rancher_component.PostInstall](https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/rancher/rancher_component.go#L271) to activate the drivers using. Rancher API. 
Now when VZ is upgraded from a lower version to 1.4, this code is being called while doing reconciliation at a later time, after the upgrade has finished. A [test](https://github.com/verrazzano/verrazzano/blob/master/tests/e2e/verify-infra/restapi/rancher_url_test.go#L130) was also added to verify the effect of these changes.
However, this test fails in upgrade pipeline when the reconciliation logic kicks off after the test has timeout, passes otherwise.

This PR to implement activation of drivers in PostUpgrade stage also

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
@@ -299,3 +299,38 @@ func (r rancherComponent) MonitorOverrides(ctx spi.ComponentContext) bool {
 	}
 	return false
 }
+
+// PostUpgrade activates OCI and OKE drivers in Rancher
+func (r rancherComponent) PostUpgrade(ctx spi.ComponentContext) error {
+	c := ctx.Client()
+	log := ctx.Log()
+	vz := ctx.EffectiveCR()
+	password, err := common.GetAdminSecret(c)
+	if err != nil {
+		return log.ErrorfThrottledNewErr("Failed getting Rancher admin secret: %s", err.Error())
+	}
+
+	rancherHostName, err := getRancherHostname(c, vz)
+	if err != nil {
+		return log.ErrorfThrottledNewErr("Failed getting Rancher hostname: %s", err.Error())
+	}
+
+	rest, err := common.NewClient(c, rancherHostName, password)
+	if err != nil {
+		return log.ErrorfThrottledNewErr("Failed getting Rancher client: %s", err.Error())
+	}
+
+	if err := rest.ActivateOCIDriver(); err != nil {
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -299,3 +299,42 @@ func (r rancherComponent) MonitorOverrides(ctx spi.ComponentContext) bool {
...
 	return false
 }
+
+// PostUpgrade activates OCI and OKE drivers in Rancher
+func (r rancherComponent) PostUpgrade(ctx spi.ComponentContext) error {
+	c := ctx.Client()
+	log := ctx.Log()
+	vz := ctx.EffectiveCR()
+	password, err := common.GetAdminSecret(c)
+	if err != nil {
```
michael-cico, line 309: If this is shared, can it be refactored into a common func/method and called from both spots?

lake-of-dreams: refactored the common code

michael-cico (approved):

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -299,3 +292,62 @@ func (r rancherComponent) MonitorOverrides(ctx spi.ComponentContext) bool {
...
+
+	if err := rest.SetAccessToken(); err != nil {
+		return nil, log.ErrorfThrottledNewErr("Failed setting Rancher access token: %s", err.Error())
+	}
+
+	return rest, nil
+}
+
+// activateDrivers activates the oci nodeDriver and oraclecontainerengine kontainerDriver
+func activateDrivers(log vzlog.VerrazzanoLogger, rest *common.RESTClient) error {
```
pfmackin, line 343: What does this do for kind based clusters?  Also, are we always activating OKE drivers for OCI?  What about OCNE, will this have any side effects?

lake-of-dreams: Please see the jira vz-5012 and driver documentation at https://rancher.com/docs/rancher/v2.6/en/admin-settings/drivers/#cluster-drivers , these are used to manage _hosted_ k8s clusters from Rancher

PR #3636, opened 2022-07-20
VZ-6418.  Fix cp error in webhook-init mode

michael-cico:
Follow-up for VZ-6418, fix an error trying to create the Kubeconfig when the VPO runs in webhook-init mode.
- Also removes the now obsolete `VZ_INSTALL_IMAGE` env var

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
@@ -29,9 +29,6 @@ spec:
           args:
             - --zap-log-level=info
             - --init-webhooks=true
-          env:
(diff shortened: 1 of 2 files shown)


platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             - --init-webhooks=true
           env:
             - name: MODE
-              value: RUN_OPERATOR
+              value: WEBHOOK_INIT
```
pfmackin, line 34: where do we use or check for this value?

pfmackin (dismissed):

bhatpmk (approved):

PR #3647, opened 2022-07-21
Save generated files in post-success hook of Build Images stage

michael-cico:
Move save-generated files step to success hook in `Build Images` stage, so that it is always created on a successful image build.  Right now if any of the parallel stages fails it won't push the operator manifests to ObjectStore.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -224,6 +224,8 @@ pipeline {
                             }
                         }
                         success {
+                            echo "Saving generated files"
+                            saveGeneratedFiles()
                             script {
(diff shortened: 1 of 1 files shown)


anders-swanson (approved):

Jenkinsfile
```diff
@@ -224,6 +224,8 @@ pipeline {
                             }
```
agv426, line 224: Should we change the stage name to include that it saves the generated files as well, so folks know that when it is done the files will be there

anders-swanson (approved):

agv426 (approved):

PR #3661, opened 2022-07-22
Redirect stdout and stderr to files, inside the bug report

bhatpmk:
This PR includes change to
1. write the standard out and err to files inside the bug report
2. distinguish the warning message to validate the contents of the bug report, with other info messages displayed while capturing the resources
3. display the error message to go through the standard err, only when there is at least an error reported by the command, but the execution continued


tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -162,3 +166,30 @@ func checkExistingFile(bugReportFile string) error {
...
+func displayWarning(successMessage string, helper helpers.VZHelper) {
+	// This might be the efficient way, but does the job of displaying a formatted message
+
+	// Draw a line to differentiate the warning from the info message
+	count := len(successMessage)
+	if len(successMessage) < minLineLength {
+		count = minLineLength
+	}
+
+	sep := ""
```
agv426, line 180: I think we can use [strings.repeat](https://pkg.go.dev/strings#Repeat) to do this:
sep := strings.repeat("-", count)

agv426 (approved):

bhatpmk:
Contributes to VZ-6251

PR #3684, opened 2022-07-25
VZ-6428 Update the prometheus-adapter image

davisgomes:


platform-operator/verrazzano-bom.json
```diff
@@ -490,7 +490,7 @@
           "images": [
             {
               "image": "prometheus-adapter",
-              "tag": "v0.9.1-3",
+              "tag": "v0.9.1-20220725190037-de39a05e",
```
bobdonat, line 493: why is a different convention being used for the tag here

bobdonat: should it be v0.9.1-4

davisgomes: It's because the image required a rebuild, so I rebuilt it from the feature branch and kept the tag. I'm not sure we can rebuild the release branch unless I merge. Those images have the nice formatting based on the release spec.

davisgomes: I think I might be able to rebuild the release. I'm attempting that now

bobdonat (approved):

PR #3688, opened 2022-07-25
VZ-6532.  Fix Rancher uninstall resiliency issues

michael-cico:
- Call Rancher PostUninstall() from global cleanup hook in all cases
- Uninstall pipeline tweaks to clean up log gathering and variable usages

### DIFF
--- a/platform-operator/controllers/verrazzano/uninstall.go
+++ b/platform-operator/controllers/verrazzano/uninstall.go
@@ -239,11 +239,6 @@ func (r *Reconciler) deleteMCResources(ctx spi.ComponentContext) error {
 		if err := r.deleteSecret(ctx.Log(), vzconst.VerrazzanoSystemNamespace, vzconst.MCAgentSecret); err != nil {
 			return err
 		}
-
-		// Run Rancher Post Uninstall to delete the Rancher resources on the managed cluster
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -298,9 +293,21 @@ func (r *Reconciler) uninstallCleanup(ctx spi.ComponentContext) (ctrl.Result, er
...
+	// Run Rancher Post Uninstall to delete the Rancher resources on the managed cluster
+	if err := r.runRancherPostInstall(ctx); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	return r.deleteNamespaces(ctx.Log())
 }
 
+func (r *Reconciler) runRancherPostInstall(ctx spi.ComponentContext) error {
+	if found, comp := registry.FindComponent(rancher.ComponentName); found {
```
pfmackin, line 305: I think this needs to run on the managed cluster which is why it was in the MC logic before.   With this change it will not run on the managed cluster.

pfmackin: my bad this is fine

pfmackin (approved):

PR #3702, opened 2022-07-27
VZ-6425: Restore prod profile to have a single replica for most components

mgianatagh:
Restore prod profile to have a single replica for most components.  In some cases a component requires more than one (e.g. OpenSearch).  An alternative approach will be used to select HA settings which will include higher replica counts.


anders-swanson (approved):
LGTM

stevshap (approved):

PR #3703, opened 2022-07-27
VZ-6502 Support for Jaeger SPM feature

ayaranga:
Following prerequisites are done so that the user can enable Jaeger
Service Performance Monitoring feature by just setting the Helm value
`jaeger.spec.query.metricsStorage.type` to `prometheus` in the Verrazzano CR:
- Set Jaeger query Prometheus server to Verrazzano managed Prometheus by
  default
- Update Prometheus authorization and network policies to allow Jaeger
  to access Prometheus

### DIFF
--- a/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
+++ b/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
@@ -609,6 +609,22 @@ func createOrUpdatePrometheusAuthPolicy(ctx spi.ComponentContext) error {
 						},
 					}},
 				},
+				{
+					// allow Jaeger to access Prometheus
+					From: []*securityv1beta1.Rule_From{{
+						Source: &securityv1beta1.Source{
+							Principals: []string{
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -609,6 +609,22 @@ func createOrUpdatePrometheusAuthPolicy(ctx spi.ComponentContext) error {
 						},
 					}},
 				},
+				{
+					// allow Jaeger to access Prometheus
+					From: []*securityv1beta1.Rule_From{{
```
kminder, line 614: I'm not asking for a change here but rather just a question.
Could this be done via Helm values?  I imagine that could require a Helm chart change?

fredtibbitts (approved):

wmhopkins (approved):
lgtm.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
ayaranga: Will be taken up as part of VZ-6603.

PR #3707, opened 2022-07-27
VZ-6521 (subtask of story VZ-6448) Use Rancher URL as server address (with corresponding CA) in the admin kubeconfig provided to managed clusters. Functionality is behind feature flag pending work for VZ-6449.

desagar:

### DIFF
--- a/platform-operator/controllers/clusters/sync_agent_secret.go
+++ b/platform-operator/controllers/clusters/sync_agent_secret.go
@@ -30,6 +30,11 @@ func setConfigFunc(f func() (*rest.Config, error)) {
 	getConfigFunc = f
 }
 
+// rancherBasedKubeConfigEnabled - feature flag for whether we support
+// Rancher based Kubeconfig - introduced for VZ-6448 Populate VMC with managed cluster data
+// Should be removed and enabled by default when story VZ-6449
+var rancherBasedKubeConfigEnabled = false
+
 // Create an agent secret with a kubeconfig that has a token allowing access to the managed cluster
 // with restricted access as defined in the verrazzano-managed-cluster role.
 // The code does the following:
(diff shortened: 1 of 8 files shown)


platform-operator/controllers/clusters/sync_agent_secret.go
```diff
@@ -152,6 +193,31 @@ func (r *VerrazzanoManagedClusterReconciler) mutateAgentSecret(secret *corev1.Se
 	return nil
 }
 
+// getRancherCACert returns the certificate authority data from Rancher's TLS secret
+func (r *VerrazzanoManagedClusterReconciler) getRancherCACert() (string, error) {
+	ingressSecret := corev1.Secret{}
+
+	errAddlTLS := r.Client.Get(context.TODO(), client.ObjectKey{
```
fredtibbitts, line 200: why not just use `err` here?

platform-operator/controllers/clusters/vmc_controller.go
```diff
@@ -340,6 +341,17 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
 	return r.Status().Update(ctx, vmc)
 }
 
+// getVerrazzanoResource gets the installed Verrazzano resource in the cluster (of which only one is expected)
+func (r *VerrazzanoManagedClusterReconciler) getVerrazzanoResource() (*v1alpha1.Verrazzano, error) {
```
fredtibbitts, line 345: I'm surprised we don't have a utility function for this, but I don't see one.

platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/Jeffail/gabs/v2"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	constants2 "github.com/verrazzano/verrazzano/application-operator/constants"
```
fredtibbitts, line 20: Minor nit, I'm not crazy about the `constants2` since it doesn't help the person looking at it to know what constants. The only other place we import this and rename it is in tests and we use `aocnst` which isn't great either.

platform-operator/controllers/clusters/sync_agent_secret.go
```diff
@@ -88,17 +104,29 @@ func (r *VerrazzanoManagedClusterReconciler) syncAgentSecret(vmc *clusterapi.Ver
 	return nil
 }
 
+func (r *VerrazzanoManagedClusterReconciler) buildKubeConfigUsingRancherURL(serviceAccountSecret corev1.Secret) (*vzk8s.KubeConfig, error) {
```
fredtibbitts, line 107: Add a function comment?

fredtibbitts (commented):
Looks good, just a few very minor comments.

platform-operator/controllers/clusters/vmc_controller.go
desagar: do you think there's a different place it should go? any suggestions?

fredtibbitts: I do not, I think this is fine.

platform-operator/controllers/clusters/vmc_controller_test.go
desagar: fixed - I also found that there was already a constant in mcconstants. I removed the extra constant and refactored all the places that use it.

fredtibbitts (approved):

PR #3715, opened 2022-07-28
VZ-6537 Updated console version to uptake the fix in console code

anajosep:
VZ-6537 - Fix for bug in the Verrazzano console to show only the instance URLs which are enabled in the VZ CR.

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -191,7 +191,7 @@
             },
             {
               "image": "console",
-              "tag": "1.4.0-20220722132540-d8a58fc",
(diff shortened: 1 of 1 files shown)


anajosep:
### Rancher component disabled in VZ CR
Screenshot 2022-08-01 at 11 26 55 AM
### Rancher, Kiali, Jaeger  disabled in VZ CR
Screenshot 2022-08-01 at 12 06 06 PM
### Grafana, Prometheus, Jaeger, Kiali disabled in VZ CR
Screenshot 2022-08-01 at 12 47 16 PM

lake-of-dreams (approved):

PR #3719, opened 2022-07-28
VZ-6401: Upgrade Helm to v3.9.2

fredtibbitts:
This PR upgrades Helm to 3.9.2, in both the Verrazzano Platform Operator Dockerfile and the go.mod (the versions need to be in sync). Upgrading pulled in several new versions of golang packages and I believe I have added (or confirmed) that all of the right replace directives are in place to pin the versions. I regenerated the third party license file and it looks like most of the changes are in the ordering of the copyright notices.

Also I made another change to remove a direct dependency on `k8s.io/utils` that I do not think we have approval to use.


kminder (approved):
Approved.

PR #3722, opened 2022-07-28
Sdosapat/vz 6503 cluster dump

saidu941:


Jenkinsfile
```diff
@@ -525,7 +525,7 @@ pipeline {
 
     post {
         always {
-            archiveArtifacts artifacts: '**/coverage.html,**/logs/**,**/verrazzano_images.txt,**/*-cluster-dump/**', allowEmptyArchive: true
+            archiveArtifacts artifacts: '**/coverage.html,**/logs/**,**/verrazzano_images.txt,**/*-cluster-dump/**,**/*-cluster-snapshot/**', allowEmptyArchive: true
```
agv426, line 528: I still see *-cluster-dump/** included as well, is something still creating the old directory name or can those now be removed? (I see this in multiple jenkinsfiles, so the question is for all of them not just this one)

agv426, line 528: In some of these I see **/*-cluster-snapshot/**, and in others I see **/*cluster-snapshot*/**. I think these were inconsistent in spots before, though the updated variants here are not all aligned with what was in the original specific Jenkinsfiles.

Maybe we should make them all consistent with one pattern that always catches them **/*cluster-snapshot*/** seems like it should do that across the board.

agv426 (approved):

PR #3725, opened 2022-07-28
VZ-6420: Add test to ensure CRDs are not deleted after uninstall

bobdonat:
This pull request adds an acceptance test to verify that CRDs are not deleted after an uninstall.  This test is run during the uninstall resiliency AT.

### DIFF
--- a/tests/e2e/uninstall/verifycrds/verify_crds_test.go
+++ b/tests/e2e/uninstall/verifycrds/verify_crds_test.go
@@ -1,7 +1,7 @@
 // Copyright (c) 2022, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-package verify_crds
+package verifycrds
 
 import (
 	"fmt"
--- a/tests/e2e/uninstall/verifycrds/verify_crds_test.go
+++ b/tests/e2e/uninstall/verifycrds/verify_crds_test.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 )
 
+// Expected verrazzano.io CRDs after uninstall
 var verrazzanoiocrds = map[string]bool{
 	"ingresstraits.oam.verrazzano.io":                              false,
 	"loggingtraits.oam.verrazzano.io":                              false,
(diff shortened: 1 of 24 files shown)


tests/e2e/uninstall/verifycrds/verify_crds_test.go
```diff
@@ -0,0 +1,204 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/verrazzano/verrazzano/pkg/test/framework"
+	"github.com/verrazzano/verrazzano/tests/e2e/pkg"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"strings"
+)
+
+// Expected verrazzano.io CRDs after uninstall
+var verrazzanoiocrds = map[string]bool{
+	"ingresstraits.oam.verrazzano.io":                              false,
```
davisgomes, line 17: Are any of the values ever true? Do we need the map value?

tests/e2e/uninstall/verifycrds/verify_crds_test.go
```diff
@@ -0,0 +1,204 @@
...
+	})
+})
+
+// checkCRds checks for both expected CRDs and unexpected CRDs for a given CRDs suffix (for example, verrazzano.io)
+func checkCrds(crds *apiextv1.CustomResourceDefinitionList, expectdCrds map[string]bool, suffix string) {
+	unexpectedCrd := false
+	for _, crd := range crds.Items {
+		_, ok := expectdCrds[crd.Name]
+		if ok {
+			expectdCrds[crd.Name] = true
```
davisgomes, line 174: Would it be easier to append this value to an unexpected found and loop through that later? I get why you implemented the map for O(1) lookup time. Would a "Set" (`map[string]struct{}`) with a  work better? O(1) lookup time, but the struct takes up zero bytes. Not that big of a deal if your implementation is easier.

davisgomes: the format of the map would be
```
var istioiocrds = map[string]bool{
	"authorizationpolicies.security.istio.io":  {},
        ...
}
```

tests/e2e/uninstall/verifycrds/verify_crds_test.go
bobdonat: The value is initialized to false and set to true when a match is found

tests/e2e/uninstall/verifycrds/verify_crds_test.go
bobdonat: I am not clear on what you are suggesting.

pfmackin (approved):

davisgomes (approved):

PR #3738, opened 2022-08-01
VZ-6604: Extend vz analyze for live cluster analysis

bhatpmk:
Extends _vz analyze_ command to analyze the live cluster, when the flag _--capture-dir_ is not provided


tools/vz/pkg/analysis/internal/util/report/report.go
```diff
@@ -249,8 +249,14 @@ func GenerateHumanReport(log *zap.SugaredLogger, reportFile string, reportFormat
 		if len(sourcesWithoutIssues) > 0 {
 			_, err = fmt.Fprintf(writeOut, "\n\n")
 		}
-		for _, source := range sourcesWithoutIssues {
-			_, err = fmt.Fprintf(writeOut, "INFO: No issues detected or to report for %s\n", source)
+		if len(sourcesWithoutIssues) == 1 {
+			// This is a workaround to avoid printing the source when analyzing the live cluster, although it impacts the
+			// regular use case with a directory containing a single cluster snapshot
+			_, err = fmt.Fprintf(writeOut, "INFO: No issues detected or to report for the cluster\n")
```
kminder, line 255: Is this INFO in the Fprintf consistent with other log output?

tools/vz/pkg/helpers/vzhelper.go
```diff
@@ -129,3 +130,12 @@ func getAllComponents(vzRes vzapi.Verrazzano) []string {
 	}
 	return compSlice
 }
+
+// isDirEmpty returns whether the directory is empty or not
+func IsDirEmpty(directory string, filesToIgnore int) bool {
```
kminder, line 135: Is an int for filesToIgnore really sufficient for this type of generic utility?

tools/vz/pkg/analysis/internal/util/report/report.go
bhatpmk: It is consistent with how the analysis tool reports the issues. Probably there is a scope to improve that throughout the analysis tool.
cc. @agv426

tools/vz/pkg/helpers/vzhelper.go
bhatpmk: I have made it a private function, as it might not be useful for other cases

tools/vz/pkg/analysis/internal/util/report/report.go
agv426: I think that this is the only spot that was using "INFO", as part of the report output. So this was originally meant to be part of the "report", not a log message (ie: it always showed up in the report, it wasn't controlled by the logging level). We can just change it up if we want to get rid of it from the report output.

For general logging, I think it is still using log.Infof log.Errorf for generating informational and error log messages.

bhatpmk: Done, updated the message

agv426 (approved):

bhatpmk:
VZ-6604

PR #3739, opened 2022-08-01
VZ-6425: Add ha.yaml as example and update HA pipeline to use it

mgianatagh:
Provide an example configuration file, named ha.yaml, that has the suggested configuration to install a highly available prod environment.  Update the rolling upgrade pipeline to use the ha.yaml configuration.

Fix some tests to work with the ha.yaml configuration.

### DIFF
--- a/tests/e2e/verify-install/promstack/promstack_test.go
+++ b/tests/e2e/verify-install/promstack/promstack_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Jeffail/gabs/v2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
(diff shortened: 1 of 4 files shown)


tests/e2e/verify-install/promstack/promstack_test.go
```diff
@@ -98,7 +99,28 @@ func areOverridesEnabled() bool {
...
 		return false
 	}
-	return vz.Spec.Components.PrometheusOperator != nil && len(vz.Spec.Components.PrometheusOperator.ValueOverrides) > 0
+
+	promOper := vz.Spec.Components.PrometheusOperator
+	if promOper == nil || len(promOper.ValueOverrides) == 0 {
+		return false
+	}
+
+	// Look for the operator value
```
fredtibbitts, line 108: Can you elaborate this comment? What's the operator value? It's not clear to me what this code block is checking for. What's special about `podAnnotations.override`?

mgianatagh: Yes, I will update the comment.  The problem is that the test was previously assuming that if any override values were specified, then the override below must have been specified.  I changed to test to explicitly look for what it was expecting.  This test failed when I used the ha.yaml file because it was not setting this override, but is setting other unrelated ones.

prometheusOperator:
  podAnnotations:
    override: "true"

fredtibbitts: I see, thanks.

fredtibbitts (approved):

PR #3740, opened 2022-08-01
Update OS tests to create ISM policies during prepare AT phase

anajosep:
PR to uptake the latest changes in VMO containing the fix for (VZ-6538) - OpenSearch single node cluster (dev profile) goes to unstable (yellow) state, when ISM policies are enabled .


kminder (approved):
Approved.

PR #3741, opened 2022-08-01
Fix an issue to add ingress-controller to list of components/namespaces map

saidu941:

Summary of changes:
> Changes to ComponentName to Namespaces map to remove wrong componentName and update with right one (ingress-controller)


kminder:
Please update the description with a more detail about the problem that this change fixes.
Is there a ticket for this that can be referenced?

pkg/constants/constants.go
```diff
@@ -151,7 +151,7 @@ const (
 	ExternalDNS                   = "external-dns"
 	VerrazzanoApplicationOperator = "verrazzano-application-operator"
 	CoherenceOperator             = "coherence-operator"
-	IngressController             = "ingress-controller-ingress-nginx-controller"
+	IngressController             = "ingress-controller"
 	IngressDefaultBackend         = "ingress-controller-ingress-nginx-defaultbackend"
```
bhatpmk, line 155: We don't have a component for the default backend. The constants are defined only for the components installed by Verrazzano, probably you can remove IngressDefaultBackend

bhatpmk: As IngressDefaultBackend is not used by ComponentNameToNamespacesMap, I think you can keep the variable as it is.

bhatpmk (approved):

PR #3753, opened 2022-08-02
Add VMC status condition for managed CA retrieved from Rancher, add cluster id to Rancher registration status of VMC

desagar:

### DIFF
--- a/platform-operator/controllers/clusters/rancher.go
+++ b/platform-operator/controllers/clusters/rancher.go
@@ -85,13 +85,16 @@ var rancherHTTPClient requestSender = &httpRequestSender{}
 
 // registerManagedClusterWithRancher registers a managed cluster with Rancher and returns a chunk of YAML that
 // must be applied on the managed cluster to complete the registration.
-func registerManagedClusterWithRancher(rc *rancherConfig, clusterName string, log vzlog.VerrazzanoLogger) (string, string, error) {
-	log.Oncef("Registering managed cluster in Rancher with name: %s", clusterName)
-
-	clusterID, err := importClusterToRancher(rc, clusterName, log)
-	if err != nil {
-		log.Errorf("Failed to import cluster to Rancher: %v", err)
-		return "", "", err
+func registerManagedClusterWithRancher(rc *rancherConfig, clusterName string, rancherClusterID string, log vzlog.VerrazzanoLogger) (string, string, error) {
+	clusterID := rancherClusterID
+	var err error
+	if clusterID == "" {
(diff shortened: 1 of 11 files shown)


platform-operator/controllers/clusters/rancher.go
```diff
@@ -52,6 +52,12 @@ const (
 	rancherClusterStateActive = "active"
 )
 
+var (
```
fredtibbitts, line 55: These variables aren't used in `rancher.go`. They should be moved to `vmc_controller_test.go`.

platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -106,44 +95,64 @@ func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Cont
 
 // syncCACertSecret gets the CA cert from the managed cluster (if the cluster is active) and creates or updates the CA cert secret.
 // If there is a CA cert on the managed cluster, then this function returns the secret name containing the CA cert.
```
fredtibbitts, line 97: Comment is no longer accurate.

platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -106,44 +95,64 @@ func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Cont
...
+	clusterID := vmc.Status.RancherRegistration.ClusterID
 	if len(clusterID) == 0 {
-		return "", nil
+		return nil
+	}
+	rc, err := newRancherConfig(r.Client, r.log)
+	if err != nil {
+		return err
+	}
+	if rc == nil || len(vmc.Spec.CASecret) > 0 {
```
fredtibbitts, line 107: `newRancherConfig` fetches multiple cluster resources and makes an HTTP call to Rancher, so it would probably be good to avoid all of that when the CA secret is already set by moving this check above the call to `newRancherConfig`.

platform-operator/controllers/clusters/rancher.go
desagar: done

platform-operator/controllers/clusters/sync_manifest_secret.go
desagar: good catch, fixed

platform-operator/controllers/clusters/sync_manifest_secret.go
desagar: done

fredtibbitts (dismissed):

fredtibbitts (approved):

PR #3755, opened 2022-08-02
VZ 6615: Retry updating the CR

anders-swanson:
Fixes issue where test may fail due to resource conflict when updating the Verrazzano CR.

### DIFF
--- a/tests/e2e/update/fluentd/fluentd_update.go
+++ b/tests/e2e/update/fluentd/fluentd_update.go
@@ -42,7 +42,7 @@ func (u *FluentdModifier) ModifyCR(cr *vzapi.Verrazzano) {
 }
 
 func ValidateUpdate(m update.CRModifier, expectedError string) {
-	gomega.Expect(func() bool {
+	gomega.Eventually(func() bool {
 		err := update.UpdateCR(m)
 		if err != nil {
 			pkg.Log(pkg.Info, fmt.Sprintf("Update error: %v", err))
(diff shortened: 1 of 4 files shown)


tests/e2e/update/fluentd/fluentd_update.go
```diff
@@ -42,7 +42,7 @@ func (u *FluentdModifier) ModifyCR(cr *vzapi.Verrazzano) {
 }
 
 func ValidateUpdate(m update.CRModifier, expectedError string) {
-	gomega.Expect(func() bool {
+	gomega.Eventually(func() bool {
```
mbenmclean, line 45: Why not just use the UpdateCRWithRetries here instead of using an eventually block?

mbenmclean (approved):

tests/e2e/update/fluentd/fluentd_update.go
```diff
@@ -42,7 +42,7 @@ func (u *FluentdModifier) ModifyCR(cr *vzapi.Verrazzano) {
...
 		err := update.UpdateCR(m)
```
pfmackin, line 46: shouldn't do update with retries also?

tests/e2e/update/updater.go
```diff
@@ -91,7 +82,7 @@ func defaultKubeConfigPath() string {
 // If the update fails, it retries by getting the latest version of CR and applying the same
 // update till it succeeds or timesout.
 func UpdateCRWithRetries(m CRModifier, pollingInterval, timeout time.Duration) {
-	RetryUpdate(m, defaultKubeConfigPath(), true, pollingInterval, timeout)
+	RetryUpdate(m, defaultKubeConfigPathOrDie(), true, pollingInterval, timeout)
```
pfmackin, line 85: This is most likely failing because of a stale resourceVersion that will only be resolved if you get a fresh CR.  Is that done in RetryUpdate?

tests/e2e/update/fluentd/fluentd_update.go
anders-swanson: This test wants to inspect the error. The pkg function doesn't have the same interface.

tests/e2e/update/fluentd/fluentd_update.go
anders-swanson: This test wants to inspect the error. The pkg function doesn't have the same interface.

tests/e2e/update/updater.go
anders-swanson: Each time it gets a fresh CR

pfmackin (approved):

PR #3756, opened 2022-08-03
VZ-6665: Replace file name in the supporting data with general message for live cluster analysis

bhatpmk:
This PR updates the analysis report, by replacing the path for the supporting data with a more general message for live cluster analysis in "supportingData" section of the analysis report. The general message provides enough information about the pod or event in a namespace, providing evidence of an issue. There is no change in the report for the static cluster analysis.


agv426 (approved):

bhatpmk:
VZ-6665

PR #3762, opened 2022-08-03
VZ-6360 Added Selector MatchLabels & MatchExpressions mechanism for Helidon Workload CRD 

absBitm:
Added Selector Matchlabels & Matchexpressions mechanism for helidon workload


arjav-desai (approved):
LGTM

application-operator/controllers/helidonworkload/helidonworkload_controller_test.go
```diff
@@ -233,6 +234,16 @@ func TestReconcileCreateHelidon(t *testing.T) {
 				helidonTestContainer,
 			},
 		},
+		Selector: metav1.LabelSelector{
```
lake-of-dreams, line 237: @absBitm How is this test testing that `MatchLabels` and `MatchExpressions` are indeed passed on to the resulting `Deployment` ? If not, should it not have an equivalent integration test to verify this since it was a bug?

application-operator/controllers/helidonworkload/helidonworkload_controller.go
```diff
@@ -218,7 +217,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, workload vzapi.VerrazzanoH
 
 // convertWorkloadToDeployment converts a VerrazzanoHelidonWorkload into a Deployment.
 func (r *Reconciler) convertWorkloadToDeployment(workload *vzapi.VerrazzanoHelidonWorkload, log vzlog.VerrazzanoLogger) (*appsv1.Deployment, error) {
-
+	if workload.Spec.DeploymentTemplate.Selector.MatchLabels == nil {
```
lake-of-dreams, line 220: - Does the `workload` here refers to a `VerrazzanoHelidonWorkload`  object currently in k8s cluster? 
- if yes, and since you are updating it here, will this be _synced_ with the original object that is in cluster i.e. since you added `MatchLabels` to it, will this reflect in cluster on this object?
- if not, should we not avoid _mutating_ the incoming function argument?

application-operator/controllers/helidonworkload/helidonworkload_controller_test.go
absBitm: we can see the match labels and match expressions during debug/ running the test function
Screenshot 2022-08-04 at 1 57 43 PM

lake-of-dreams: I guess this should be testable in one of the pipelines in Jenkins, either via unit test or integration test

application-operator/controllers/helidonworkload/helidonworkload_controller.go
absBitm: * Yes  workload here refers to a VerrazzanoHelidonWorkload object
* Yes its getting synced with the original object that is in the cluster

PR #3764, opened 2022-08-03
VZ-6500 Verify IP address exist in expected Ingresses

davisgomes:
This adds an extra post-install verification step to ensure that all the existing Ingresses have Load Balancer IP addresses.


platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -255,6 +259,14 @@ func (i istioComponent) PostInstall(compContext spi.ComponentContext) error {
 	if err := createEnvoyFilter(compContext.Log(), compContext.Client()); err != nil {
 		return err
 	}
+
+	// istioctl verify-install does not check that the Load Balancer service has an external IP address,
+	// so we have to check this manually to get a useful error message
+	_, err := verifyIstioIngressGatewayIP(compContext.Client(), compContext.EffectiveCR())
```
michael-cico, line 265: should these be in the `IsReady` check?

davisgomes: I didn't want to impede the post install steps, but I think you're right. I'll move it there.

platform-operator/internal/vzconfig/ingress.go
```diff
@@ -84,29 +82,11 @@ func GetServiceType(cr *vzapi.Verrazzano) (vzapi.IngressType, error) {
...
-			// In case of OLCNE, the Status.LoadBalancer.Ingress field will be empty, so use the external IP if present
-			ingressIP = svc.Spec.ExternalIPs[0]
-		} else if len(svc.Status.LoadBalancer.Ingress) > 0 {
-			ingressIP = svc.Status.LoadBalancer.Ingress[0].IP
-		} else {
-			return "", fmt.Errorf("No IP found for service %v with type %v", svc.Name, serviceType)
-		}
-	}
-	return ingressIP, nil
+	return GetExternalIP(client, serviceType, vpoconst.NGINXControllerServiceName, globalconst.IngressNamespace)
```
michael-cico, line 89: Maybe we should rename the func `GetServiceType` called before this to be `GetIngressServiceType` so it's clearer where/what we're getting it from?

I'd say we should probably move the code that gets the Istio external IP to this package, but so far I don't think anything else needs it except the app operator.   We'd have to promote that logic to the top-level `pkg` so it could be shared with the app operator though.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -376,6 +376,14 @@ func (i istioComponent) PostUpgrade(context spi.ComponentContext) error {
 		return err
 	}
 
+	// istioctl verify-install does not check that the Load Balancer service has an external IP address,
+	// so we have to check this manually to get a useful error message
+	_, err = verifyIstioIngressGatewayIP(context.Client(), context.EffectiveCR())
```
michael-cico, line 381: had a similar comment below, maybe make part of the `IsReady` check?

platform-operator/internal/vzconfig/ingress.go
davisgomes: I'll rename the function. I considered moving the Istio external IP function here, but all of these functions are pertaining to (and hardcoded to check) the nginx ingress alone. I thought it would be more confusing to add Istio checks here as well. That was why I created the shared `GetExternalIP` function that can be used for both components.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
davisgomes: Done!

platform-operator/internal/vzconfig/ingress.go
davisgomes: It might be nice to use this in the app operator, but I think that would require a bit more work than the scope of this JIRA presents. I'm not yet sure what areas of the app operator would require this function call and how to federate the implementation.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -385,3 +389,31 @@ func removeTempFiles(log vzlog.VerrazzanoLogger) {
 		log.Errorf("Unexpected error removing temp files: %v", err.Error())
 	}
 }
+
+// getIngressServiceType returns the service type of the Istio ingress service for use in checking its status
+func getIngressServiceType(vz *vzapi.Verrazzano) (vzapi.IngressType, error) {
```
pfmackin, line 394: Can you rename to getIstioIngressGatewayServiceType.  I know it is long but it will avoid confusion.  Same comment for other places where you are specifically looking at the Istio Ingress Gateway

davisgomes: Done

platform-operator/internal/vzconfig/ingress.go
pfmackin: If this is generic to Kubernetes ingress then it should be in pkg/k8s/ingress/ingress.go. Can you move this file as a follow up PR?

pfmackin (approved):

PR #3768, opened 2022-08-04
VZ-5010: Enable Keycloak/Rancher SSO

lake-of-dreams:
- Configures a new client `rancher` in Keycloak with client secret
- Configures a `keycloakoidc` rancher `AuthConfig` CR that sets up the keycloak OIDC provider
- Configures a rancher `User` CR that configures a corresponding user in Rancher
- Configures a rancher `GlobalRoleBinding` CR that assigns admin privs to Verrazzano user
- Entire code in `keycloak` module is refactored to not spin up a `kubectl` process locally in container to execute commands in keycloak pod but rather use the go client. Also refactored all the related tests which earlier spawned bash processes even during testing and had a env variable based way to inject the success/failure condition. Also removed duplicacy from code.
- Also refactored the `k8sutil.ExecPod` utility that will invoke commands in pods, earlier a `fake` for this could only produce one result for any kind of invocation and was problematic when multiple such calls are involved, Changed this to be able to return different `stdout` , `stderr` and `err` based upon the actual command (which eventually becomes the query of the url created by go client) 
- Introduced a func `GetURLForIngress` for getting host url from ingress
-  removed the need of platform-operator/scripts/install/update-kiali-redirect-uris.sh script which was being invoked from keycloak configuration code, instead it was entirely doable using the `-b` parameter on `kcadm` cli
- Changed a test utility func `EventuallyGetRancherURL` to `EventuallyGetURLForIngress` to be able to use at multiple places
- Added logic to disable/enable auth providers

TBD: Logic to disable/enable keycloak vs local authProvider based on setting from rancher component in VZ CR


pkg/k8sutil/k8sutil.go
```diff
@@ -328,3 +331,13 @@ func GetDynamicClientInCluster(kubeconfigPath string) (dynamic.Interface, error)
...
 }
+
+// GetHostFromIngress returns the url for an Ingress
+func GetURLForIngress(client client.Client, name string, namespace string) (string, error) {
+	var ingress = &networkingv1.Ingress{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, ingress)
+	if err != nil {
+		return "", fmt.Errorf("unable to fetch ingress %s/%s, %v", name, namespace, err)
+	}
+	return fmt.Sprintf("https://%s", ingress.Spec.Rules[0].Host), nil
```
kminder, line 342: Should the protocol (i.e. https) be passed in for a generic utility like this?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -285,6 +259,144 @@ const pgClient = `
...
+        "saml.authnstatement": "false",
+        "display.on.consent.screen": "false",
+        "saml.onetimeuse.condition": "false"
+      },
+      "authenticationFlowBindingOverrides": {},
+      "fullScopeAllowed": true,
+      "nodeReRegistrationTimeout": -1,
+      "protocolMappers": [
+        {
+          "id": "8994f0ee-b490-49ae-8793-ee72df7e4b6d",
```
kminder, line 310: Should this and other IDs be generated?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1419,3 +1398,149 @@ func upgradeStatefulSet(ctx spi.ComponentContext) error {
...
+
+	err = t.Execute(&b, &data)
+	if err != nil {
+		return "", err
+	}
+
+	return b.String(), nil
+}
+
+// GetRancherClientSecretFromKeycloak returns the secret from rancher client in keycloak
```
kminder, line 1429: Use title case for products in comments.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -267,6 +267,10 @@ func (r rancherComponent) PostInstall(ctx spi.ComponentContext) error {
 		return log.ErrorfThrottledNewErr("Failed removing Rancher bootstrap secret: %s", err.Error())
 	}
 
+	if err := configureKeycloakOIDCProvider(ctx); err != nil {
+		return log.ErrorfThrottledNewErr("failed configuring keycloak oidc provider: %s", err.Error())
```
kminder, line 271: OIDC?

lake-of-dreams: its abbreviation for OpenID Connect, the Keycloak OIDC provider `AuthConfig` in Rancher is also called `keycloakoidc`

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
lake-of-dreams: removed IDs

pkg/k8sutil/k8sutil.go
lake-of-dreams: implemented

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
lake-of-dreams: fixed

platform-operator/helm_config/charts/verrazzano-platform-operator/crds/clusters.verrazzano.io_verrazzanomanagedclusters.yaml
```diff
@@ -70,7 +70,6 @@ spec:
               rancherRegistration:
                 properties:
                   clusterID:
-                    description: ClusterID is the Rancher cluster ID for this cluster
```
fredtibbitts, line 73: Why did you remove this description?

lake-of-dreams: excluded the file itself

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -267,6 +269,36 @@ func (r rancherComponent) PostInstall(ctx spi.ComponentContext) error {
...
+		}
+	}
+
+	if vz.Status.Version == "" || (vzStatusVer.IsGreatherThan(ver140) || vzStatusVer.IsEqualTo(ver140)) {
+		enableKeycloak := vz.Spec.Components.Rancher == nil || vz.Spec.Components.Rancher.AuthtType == v1alpha1.Keycloak
+		if err := disableOrEnableAuthProvider(ctx, AuthConfigKeycloak, enableKeycloak); err != nil {
+			return log.ErrorfThrottledNewErr("failed enabling keycloak oidc provider: %s", err.Error())
+		}
+
+		enableLocal := (vz.Spec.Components.Rancher != nil && vz.Spec.Components.Rancher.AuthtType == v1alpha1.Local) || !enableKeycloak
```
michael-cico, line 296: Might be easier on the eye if code like

```
(vz.Spec.Components.Rancher != nil && vz.Spec.Components.Rancher.AuthtType == v1alpha1.Local)
```

was in a helper method like `isLocalAuthEnabled()`, and same for the above KC auth enabled check.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -302,14 +334,44 @@ func (r rancherComponent) PostUpgrade(ctx spi.ComponentContext) error {
...
+		}
+	}
+
+	if vzSpecVer != nil && (vzSpecVer.IsGreatherThan(ver140) || vzSpecVer.IsEqualTo(ver140)) {
+		enableKeycloak := vz.Spec.Components.Rancher != nil && vz.Spec.Components.Rancher.AuthtType == v1alpha1.Keycloak
+		if err := disableOrEnableAuthProvider(ctx, AuthConfigKeycloak, enableKeycloak); err != nil {
+			return log.ErrorfThrottledNewErr("failed changing state of keycloak oidc provider: %s", err.Error())
+		}
+
+		enableLocal := (vz.Spec.Components.Rancher != nil && vz.Spec.Components.Rancher.AuthtType == v1alpha1.Local) || !enableKeycloak
```
michael-cico, line 361: these have to be mutually exclusive right?  local vs KC auth?  I feel like this can be simplified somehow; i.e.

```
if version >= 1.4.0 {
  if KC-auth {
     // enable local
     // disable KC
     return ...
  } 
}
// enable local auth
// disable KC auth
return ...
```

Also, I think a lot of this code in the PostInstall should be put in a helper-method or something, esp if you need to do something similar in the upgrade path.

michael-cico: Obviously the above is just some hand-waving, just wondering if the checks are duplicated somehow can be simplified.

fredtibbitts (commented):
Some general questions, because I can't tell by looking at the code.

1. Should the default be "local" to preserve the existing behavior?
2. What happens during upgrade? Does the auth provider get changed to Keycloak because that's the default?
3. What if Keycloak is disabled, does the default then become "local"?

tests/e2e/verify-infra/restapi/rancher_url_test.go
```diff
@@ -38,7 +38,7 @@ var _ = t.Describe("rancher", Label("f:infra-lcm",
 				}
 
 				api := pkg.EventuallyGetAPIEndpoint(kubeconfigPath)
-				rancherURL := pkg.EventuallyGetRancherURL(t.Logs, api)
+				rancherURL := pkg.EventuallyGetURLForIngress(t.Logs, api, "cattle-system", "rancher", "https")
```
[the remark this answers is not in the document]
lake-of-dreams: So currently there is no mechanism for tests to know whether they are running in a upgrade flow or otherwise so I have put some basic tests in place:

1. For version < 1.4.0, will check that local auth config is enabled
2. For version > 1.4.0, will check that one of local or keycloak auth config is enabled because not in all upgrade scenarios will the keycloak be enabled by default

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: refactored this

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: refactored further please

lake-of-dreams:
@fredtibbitts 
>Should the default be "local" to preserve the existing behavior?
What happens during upgrade? Does the auth provider get changed to Keycloak because that's the default?
What if Keycloak is disabled, does the default then become "local"?

1. As per requirement from Jira, all installs greater than equal to 1.4.0 will have Keycloak as default Auth provider
2. During upgrade there are couple of scenarios
- If from a lower version than 1.4.0 to 1.4.0 or higher, Auth provider will NOT be changed to Keycloak unless the rancher component has `authType` as `kecyloak` in the VZ CR
- if from a version greater than equal to 1.4.0 to a higher version, unless the `authType` is changed in VZ CR while upgrading, it will keep the previous. However if while upgrading the `rancher` section is removed completely, then it will make Keycloak as the auth provider
3. If Keycloak is disabled, only local authProvider be enabled.

tests/e2e/verify-infra/restapi/rancher_url_test.go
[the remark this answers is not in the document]
michael-cico: yeah I think our current pipelines are not flexible enough yet.  The 3 upgrade scenarios are 

1. preserve the local auth
2. user updates to KC auth on upgrade, or post-upgrade
3. User updates to KC with the upgrade

We might be able to achieve both by having the upgrade preserve it in the 1.x -> 1.4+ scenario, and then do a post-upgrade update to KC auth.  But I think that's a separate PR as you say.  The basic upgrade path covers (1) and the post-upgrade update covers (2).  I think (3) is probably a nice-to-have, but requires more flexibility the pipeline and also may not be worth the resources for an extra run.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -405,5 +353,107 @@ func configureKeycloakOIDCProvider(ctx spi.ComponentContext) error {
...
+		version = vzTargetVersion
+	} else {
+		version = vzSourceVersion
+	}
+
+	return version == nil || (version.IsGreatherThan(ver140) || version.IsEqualTo(ver140)), nil
+}
+
+func isKeycloakAuthEnabled(isUpgrade bool, vz *vzapi.Verrazzano, ver140 *semver.SemVersion, vzSourceVersion *semver.SemVersion, vzTargetVersion *semver.SemVersion) bool {
+	if vz.Spec.Components.Keycloak != nil && vz.Spec.Components.Keycloak.Enabled != nil && !*vz.Spec.Components.Keycloak.Enabled {
```
michael-cico, line 401: there's an `IsKeycloakEnabled()` func in `platform-operator/internal/vzconfig/enabled.go` you could use here

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -405,5 +353,107 @@ func configureKeycloakOIDCProvider(ctx spi.ComponentContext) error {
...
+	if isUpgrade {
+		version = vzTargetVersion
+	} else {
+		version = vzSourceVersion
+	}
+
+	return version == nil || (version.IsGreatherThan(ver140) || version.IsEqualTo(ver140)), nil
+}
+
+func isKeycloakAuthEnabled(isUpgrade bool, vz *vzapi.Verrazzano, ver140 *semver.SemVersion, vzSourceVersion *semver.SemVersion, vzTargetVersion *semver.SemVersion) bool {
```
michael-cico, line 400: Should probably add some comments in this func to make it clear what it's doing, i.e. what the conditions for returning `true` are.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -405,5 +353,107 @@ func configureKeycloakOIDCProvider(ctx spi.ComponentContext) error {
...
+	if vz.Spec.Components.Rancher != nil && vz.Spec.Components.Rancher.AuthtType == v1alpha1.Keycloak {
+		return true
+	}
+
+	if vz.Spec.Components.Rancher == nil {
+		if !isUpgrade {
+			return true
+		}
+
+		if vzSourceVersion != nil && (vzSourceVersion.IsGreatherThan(ver140) || vzSourceVersion.IsEqualTo(ver140)) && vzTargetVersion != nil && (vzTargetVersion.IsGreatherThan(vzSourceVersion) || vzTargetVersion.IsEqualTo(vzSourceVersion)) {
```
michael-cico, line 414: nit-pick, might be useful to add `IsGreaterThanOrEqualTo` methods on the `SemVer` type.

wmhopkins (dismissed):
lgtm.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: implemented

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: added comments

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: added

lake-of-dreams:
As per feedback from @gilbode , now the `local` auth provider will not be disabled anymore while enabling the `keycloak` auth provider. Changes implemented.

michael-cico (approved):

platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
```diff
@@ -608,6 +608,9 @@ type RancherComponent struct {
 	// +optional
 	Enabled          *bool `json:"enabled,omitempty"`
 	InstallOverrides `json:",inline"`
+	// KeycloakAuthEnabled specifies whether the Keycloak Auth provider is enabled.  Default is false.
+	// +optional
+	KeycloakAuthEnabled *bool `json:"keycloakAuthEnabled,omitempty"`
```
anders-swanson, line 613: FYI if we are adding fields to the API, we need to implement the conversion.

PR #3769, opened 2022-08-04
Enable notifications for uninstall resiliency suite, more infra tests

michael-cico:


ci/uninstall/JenkinsfileUninstallSuite
```diff
@@ -214,7 +214,7 @@ pipeline {
                     slackSend ( channel: "$SLACK_ALERT_CHANNEL", message: "Job Failed - \"${env.JOB_NAME}\" build: ${env.BUILD_NUMBER}\n\nView the log at:\n ${env.BUILD_URL}\n\nBlue Ocean:\n${env.RUN_DISPLAY_URL}\n\nSuspects:\n${SUSPECT_LIST}" )
```
agv426, line 214: I wouldn't send this to the ALERT channel, just send it to the CI channel.

We generally are only sending the pipeline level failures to the ALERT channel, and individual failures to the CI channel when we send those to reduce overloading the ALERT channel.

michael-cico: This is a suite pipeline, that kicks off 5 individual uninstall runs.  There's no slack alerts in the downstream invocations.  Should this still go to the CI channel?  And should I add alerts from the individual runs to the CI channel as well?

agv426: I didn't realize this was a separate pipeline unrelated to the release decision testing. Given it is an independent suite unrelated to that it makes sense to send to the ALERT channel

agv426 (approved):

PR #3774, opened 2022-08-04
VZ-6685 mysql backup fix

theabmitra:
Re-introduce backup hook configuration after it was removed as a result of mysql HA implementation


platform-operator/thirdparty/charts/mysql/templates/configurationFiles-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.configurationFiles }}
```
jmaron99, line 1: I'm not necessarily against the idea of augmenting the existing templates, but it does introduce a complexity to picking up newer versions of the chart etc.  Can you think of a way to use existing chart mechanisms to accomplish the addition of the backup config script?

jmaron99 (approved):

PR #3775, opened 2022-08-05
Fix WhenJaegerOperatorEnabledIt function in Jaeger helidon test

anajosep:
There was a bug in ` WhenJaegerOperatorEnabledIt` function implementation. This caused the tests to be unstable. Deva had fixed this issue in another suite (https://github.com/verrazzano/verrazzano/pull/3773). Implementing the same fix in the `jaeger/helidon` suite as well.


lake-of-dreams (approved):
Approved. A separate task should be filed to refactor duplicate WhenJaegerOperatorEnabledIt function in two files after fixing the cycling dep

agv426 (approved):

PR #3786, opened 2022-08-08
VZ-6698 - update HA pipeline to use prod profile

mgianatagh:
- Update the HA test pipeline to be based on a prod profile instead of dev
- Support longer install timeout (sometimes ha install takes longer than 30 minutes)
- Skip stages after a failure

### DIFF
--- a/ci/scripts/prepare_jenkins_at_environment.sh
+++ b/ci/scripts/prepare_jenkins_at_environment.sh
@@ -25,6 +25,7 @@ WILDCARD_DNS_DOMAIN=${2:-"x=nip.io"}
 KIND_NODE_COUNT=${KIND_NODE_COUNT:-1}
 TEST_OVERRIDE_CONFIGMAP_FILE="./tests/e2e/config/scripts/pre-install-overrides/test-overrides-configmap.yaml"
 TEST_OVERRIDE_SECRET_FILE="./tests/e2e/config/scripts/pre-install-overrides/test-overrides-secret.yaml"
+INSTALL_TIMEOUT_VALUE=${INSTALL_TIMEOUT:-30m}
 
 cd ${GO_REPO_PATH}/verrazzano
 echo "tests will execute" > ${TESTS_EXECUTED_FILE}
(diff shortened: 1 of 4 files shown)


ci/scripts/prepare_jenkins_at_environment.sh
```diff
@@ -25,6 +25,7 @@ WILDCARD_DNS_DOMAIN=${2:-"x=nip.io"}
 KIND_NODE_COUNT=${KIND_NODE_COUNT:-1}
 TEST_OVERRIDE_CONFIGMAP_FILE="./tests/e2e/config/scripts/pre-install-overrides/test-overrides-configmap.yaml"
 TEST_OVERRIDE_SECRET_FILE="./tests/e2e/config/scripts/pre-install-overrides/test-overrides-secret.yaml"
+INSTALL_TIMEOUT_VALUE=${INSTALL_TIMEOUT:-30m}
```
stevshap, line 28: What is this doing? Why not just use `INSTALL_TIMEOUT` here?

mgianatagh: INSTALL_TIMEOUT may not be defined, so in that case it defaults to 30m

tests/e2e/ha/pods.go
```diff
@@ -22,12 +23,15 @@ func EventuallyPodsReady(log *zap.SugaredLogger, cs *kubernetes.Clientset) {
 			return false
 		}
 
+		// Assume all pods are ready.  If debug enabled, log status of each pod that is not ready yet
+		returnValue := true
```
stevshap, line 27: How is this different than the code already there?

stevshap (approved):

tests/e2e/ha/pods.go
mgianatagh: The difference is that when debug is enabled, the test will log all pods not ready or completed on each iteration.  If any get logged then a value of false needs to be returned.  The previous code was returning false on the first pod it found not ready.

ci/scripts/prepare_jenkins_at_environment.sh
stevshap: Gotcha

tests/e2e/ha/pods.go
stevshap: Makes sense

PR #3793, opened 2022-08-08
reduce a specific internal webhook error in Install log

knelson7:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
@@ -175,7 +175,8 @@ func (c certManagerComponent) createOrUpdateClusterIssuer(compContext spi.Compon
 	} else {
 		// Create resources needed for CA certificates
 		if opResult, err = createOrUpdateCAResources(compContext); err != nil {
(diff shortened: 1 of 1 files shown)


stevshap (approved):

platform-operator/controllers/verrazzano/component/certmanager/certmanager_component.go
```diff
@@ -175,7 +175,8 @@ func (c certManagerComponent) createOrUpdateClusterIssuer(compContext spi.Compon
 	} else {
 		// Create resources needed for CA certificates
 		if opResult, err = createOrUpdateCAResources(compContext); err != nil {
-			return compContext.Log().ErrorfNewErr("Failed creating CA resources: %v", err)
+			compContext.Log().Oncef("Failed creating CA resources: %v", err)
+			return err
```
desagar, line 179: I'd suggest returning an error that has the same message, so that the caller still gets the context information that it was CA resource creation that failed. 

You can create an error object with a message by doing this: 
return fmt.Errorf(message)

You should be able to create the message once with fmt.Sprintf and use it for both the Oncef and the fmt.Errorf

PR #3797, opened 2022-08-08
Added support to generate bug-report along with cluster-dump for Jenkins jobs

saidu941:
This PR includes changes to most of the JenkinsFiles where we enable the jenkins job to generate the bug-report along with the cluster dump (when failed or cluster dump on success is enabled)

Summary of changes:
> Changes to most of the jenkinsFiles include adding vz command, archiving artifact for bug-report, enable bug-report command along with running cluster-dump script


ci/chaos/Jenkinsfile
```diff
@@ -715,6 +716,9 @@ def runGinkgoNamespace(testSuitePath, skipDeploy, skipUndeploy, namespace) {
 def dumpK8sCluster(dumpDirectory) {
     sh """
         ${GO_REPO_PATH}/verrazzano/tools/scripts/k8s-dump-cluster.sh -d ${dumpDirectory} -r ${dumpDirectory}/cluster-snapshot/analysis.report
+        ${GO_REPO_PATH}/vz bug-report --report-file ${dumpDirectory}/bug-report.tar.gz
+        mkdir bug-report
```
agv426, line 720: I think the bug-report directory should be explicitly placed, rather than in whatever the current working directory happens to be at the time this is being called. Maybe we can use the dumpDirectory here as the base level, and have a sub-dir for bug-report and another for cluster snapshot. ie:

${dumpDirectory}/snapshot
${dumpDirectory}/bug-report

This comment applies to the general pattern

ci/kind/Jenkinsfile
```diff
@@ -93,7 +93,7 @@ pipeline {
         // Used for dumping cluster from inside tests
         DUMP_KUBECONFIG="${KUBECONFIG}"
         DUMP_COMMAND="${GO_REPO_PATH}/verrazzano/tools/scripts/k8s-dump-cluster.sh"
-        VZ_COMMAND="${GO_REPO_PATH}/verrazzano/tools/scripts/test-call-vz.sh"
```
agv426, line 96: the test-call-vz.sh script is no longer used, we should delete it

agv426 (approved):

PR #3798, opened 2022-08-08
GetEffectionVersion to use for docs link

saidu941:
This PR adds changes to analysis files, added a util function to gett effective version based on the environment variable and the current build information

This is to address VZ-6280, this will fix create the versioned links instead of /latest/


tools/vz/pkg/analysis/internal/util/report/action.go
```diff
@@ -33,27 +36,46 @@ func (action *Action) Validate(log *zap.SugaredLogger) (err error) {
 	return nil
 }
 
+func GetDevelopmentVersion() string {
+	fileName := "../../../../../../../.verrazzano-development-version"
```
agv426, line 40: We can't read that file at CLI runtime, the customer will have a CLI binary may not have any sources and even if they did the current working directory could be anywhere and that relative reference would not work.

The CLI hands in the version, get it from vz/cmd/version/version.go

PR #3799, opened 2022-08-08
VZ-6456: Sync Rancher clusters with VMCs

fredtibbitts:
This PR adds synchronizing of clusters in Rancher with VMCs. When clusters are added in Rancher, we create corresponding empty VMC resources. For any VMC resources that do not have a corresponding Rancher cluster, we delete the VMC.

### DIFF
--- a/platform-operator/controllers/clusters/rancher.go
+++ b/platform-operator/controllers/clusters/rancher.go
@@ -546,7 +546,7 @@ func doRequest(req *http.Request, rc *rancherConfig, log vzlog.VerrazzanoLogger)
 
 		// retry any HTTP 500 errors
 		if resp != nil && resp.StatusCode >= 500 && resp.StatusCode <= 599 {
-			log.Infof("HTTP status %v executing HTTP request %v, retrying", resp.StatusCode, req)
+			log.ErrorfThrottled("HTTP status %v executing HTTP request %v, retrying", resp.StatusCode, req)
 			return false, err
 		}
 
--- a/platform-operator/controllers/clusters/rancher.go
+++ b/platform-operator/controllers/clusters/rancher.go
@@ -6,7 +6,7 @@ package clusters
 import (
 	"bytes"
 	"context"
-	"crypto/md5"
+	"crypto/md5" //nolint:gosec //#gosec G501 // package used for caching only, not security
 	"crypto/tls"
 	"fmt"
 	"io"
@@ -207,7 +207,7 @@ func getAllClustersInRancher(rc *rancherConfig, log vzlog.VerrazzanoLogger) ([]s
 	reqURL := rc.baseURL + clustersPath
 	headers := map[string]string{"Authorization": "Bearer " + rc.apiAccessToken}
 
-	hash := md5.New()
+	hash := md5.New() //nolint:gosec //#gosec G401
 	clusterNames := []string{}
 	for {
 		response, responseBody, err := sendRequest(http.MethodGet, reqURL, headers, "", rc, log)
--- a/platform-operator/controllers/clusters/rancher.go
+++ b/platform-operator/controllers/clusters/rancher.go
@@ -6,6 +6,7 @@ package clusters
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"crypto/tls"
 	"fmt"
 	"io"
@@ -199,6 +200,56 @@ func getClusterIDFromRancher(rc *rancherConfig, clusterName string, log vzlog.Ve
 	}
 
 	return httputil.ExtractFieldFromResponseBodyOrReturnError(responseBody, "data.0.id", "unable to find clusterId in Rancher response")
+}
+
+// getAllClustersInRancher returns a slice of the cluster names registered with Rancher
+func getAllClustersInRancher(rc *rancherConfig, log vzlog.VerrazzanoLogger) ([]string, []byte, error) {
+	reqURL := rc.baseURL + clustersPath
+	headers := map[string]string{"Authorization": "Bearer " + rc.apiAccessToken}
+
+	hash := md5.New()
+	clusterNames := []string{}
+	for {
+		response, responseBody, err := sendRequest(http.MethodGet, reqURL, headers, "", rc, log)
+		if response != nil && response.StatusCode != http.StatusOK {
+			return nil, nil, fmt.Errorf("Unable to get clusters from Rancher, response code: %d", response.StatusCode)
+		}
+
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// parse the response and iterate over the items
+		jsonString, err := gabs.ParseJSON([]byte(responseBody))
+		if err != nil {
+			return nil, nil, err
+		}
+
+		var items []interface{}
+		var ok bool
+		if items, ok = jsonString.Path("data").Data().([]interface{}); !ok {
+			return nil, nil, fmt.Errorf("Unable to find expected data in Rancher clusters response: %v", jsonString)
+		}
+
+		for _, item := range items {
+			i := item.(map[string]interface{})
+			if name, ok := i["name"]; ok {
+				clusterNames = append(clusterNames, name.(string))
+			}
+		}
+
+		// add this response body to the hash
+		io.WriteString(hash, responseBody)
+
+		// if there is a "next page" link then use that to make another request
+		if reqURL, err = httputil.ExtractFieldFromResponseBodyOrReturnError(responseBody, "pagination.next", ""); err != nil {
+			break
+		}
+	}
+
+	// unfortunately Rancher does not support ETags, so we return a hash of the response bodies which allows the caller to know if
+	// there were any changes to the clusters
+	return clusterNames, hash.Sum(nil), nil
 }
 
 // isManagedClusterActiveInRancher returns true if the managed cluster is active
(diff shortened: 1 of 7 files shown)


platform-operator/controllers/clusters/sync_rancher_clusters_test.go
```diff
@@ -0,0 +1,173 @@
...
+		}).Build()
+}
+
+// expectHTTPCalls mocks the HTTP calls we expect the Rancher client to make
+func expectHTTPCalls(t *testing.T, httpMock *mocks.MockRequestSender, testPaging bool) {
+	asserts := assert.New(t)
+
+	// expect an HTTP request to fetch the admin token from Rancher
+	httpMock.EXPECT().
+		Do(gomock.Not(gomock.Nil()), gomock.Not(gomock.Nil())).
```
desagar, line 139: We can be more specific here than gomock.Not(gomock.Nil())), which makes it much easier when trying to debug issues where mock calls don't match.
I wrote a custom matcher that can compare the URL in one of the other tests recently. See example here: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/clusters/vmc_controller_test.go#L1042

platform-operator/controllers/clusters/rancher.go
```diff
@@ -249,6 +251,56 @@ func getClusterIDFromRancher(rc *rancherConfig, clusterName string, log vzlog.Ve
...
+
+		for _, item := range items {
+			i := item.(map[string]interface{})
+			if name, ok := i["name"]; ok {
+				clusterNames = append(clusterNames, name.(string))
+			}
+		}
+
+		// add this response body to the hash
+		io.WriteString(hash, responseBody)
```
desagar, line 291: does WriteString write cumulatively?

platform-operator/controllers/clusters/sync_rancher_clusters_test.go
fredtibbitts: ok thanks, I'll check that out.

platform-operator/controllers/clusters/sync_rancher_clusters.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// if the Rancher response did not change, there is nothing to do
+	if bytes.Equal(previousHash, clustersResponseHash) {
+		log.Debug("Rancher clusters response did not change, nothing to sync")
+		return
+	}
+
+	// for every cluster (ignoring "local") make sure a VMC exists
+	r.ensureVMCs(clusterNames, log)
```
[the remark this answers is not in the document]
fredtibbitts: I am labeling the VMCs, I used the label "app.kubernetes.io/created-by" since that's a well-known label. I guess we should be consistent though.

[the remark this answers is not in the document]
fredtibbitts: Regarding the cluster ID, I'm not sure about the naming. I need to give it some more thought.

platform-operator/controllers/clusters/rancher.go
fredtibbitts: Yes it's cumulative.

platform-operator/controllers/clusters/sync_rancher_clusters_test.go
fredtibbitts: I updated the unit tests to use your custom matchers.

platform-operator/controllers/clusters/sync_rancher_clusters.go
```diff
@@ -0,0 +1,197 @@
...
+	for _, rancherCluster := range rancherClusters {
+		// ignore the "local" cluster
+		if rancherCluster.name == localClusterName {
+			continue
+		}
+
+		// note that we use the Rancher cluster id as the name of the VMC, since the cluster id is unique and the name
+		// in Rancher is not
+		cr := &clustersv1alpha1.VerrazzanoManagedCluster{}
+		if err := r.Get(context.TODO(), types.NamespacedName{Name: rancherCluster.id, Namespace: constants.VerrazzanoMultiClusterNamespace}, cr); err != nil {
```
davisgomes, line 119: Could you use `createOrUpdate` here? It seems like what this block is emulating.

fredtibbitts: Your comment inspired me to look at other options again. I was thinking maybe a server-side apply or something like that could be used to create the spec and status in one shot, but after more research and a few experiments, I can't find any way to do it.

So the status has to be updated separately no matter what; `createOrUpdate` won't update the status. Since I don't want to update the VMC, I only want to create one if it doesn't exist, I don't think `createOrUpdate` really adds value or makes it much simpler. I think I'm going to leave it as is.

davisgomes (approved):

platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -47,38 +50,47 @@ func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Cont
...
+				sb.WriteString(rancherYAML)
+			}
 		}
 	}
 
 	// create/update the manifest secret with the YAML
 	_, err = r.createOrUpdateManifestSecret(vmc, sb.String())
 	if err != nil {
-		return err
+		return false, err
```
desagar, line 93: Technically, maybe we should return vmcWaitingForClusterID in these err cases as well, instead of false, in case something decides to look at that value even if there's an error.

desagar (approved):

platform-operator/controllers/clusters/rancher.go
```diff
@@ -201,6 +207,65 @@ func getClusterIDFromRancher(rc *rancherConfig, clusterName string, log vzlog.Ve
...
+		}
+
+		var items []interface{}
+		var ok bool
+		if items, ok = jsonString.Path("data").Data().([]interface{}); !ok {
+			return nil, nil, fmt.Errorf("Unable to find expected data in Rancher clusters response: %v", jsonString)
+		}
+
+		for _, item := range items {
+			i := item.(map[string]interface{})
```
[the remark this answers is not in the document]
fredtibbitts: Can't hurt, I'll add a check.

platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -47,38 +50,47 @@ func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Cont
...
+	// attempt to fetch the registration manifest YAML
+	vzVMCWaitingForClusterID := false
+
+	if vmc.Labels != nil && vmc.Labels[createdByLabel] == createdByVerrazzano && len(vmc.Status.RancherRegistration.ClusterID) == 0 {
+		r.log.Progressf("Waiting for Verrazzano-created VMC named %s to have a cluster id in the status before attempting to fetch Rancher registration manifest", vmc.Name)
+		vzVMCWaitingForClusterID = true
 	} else {
-		var rancherYAML string
-		rancherYAML, clusterID, err = registerManagedClusterWithRancher(rc, vmc.Name, vmc.Status.RancherRegistration.ClusterID, r.log)
+		// register the cluster with Rancher - the cluster will show as "pending" until the
```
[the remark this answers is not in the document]
fredtibbitts: Correct.

platform-operator/controllers/clusters/sync_manifest_secret.go
fredtibbitts: I think the general rule in Golang is that if an error is returned the other return arguments can't be trusted, but it's easy enough to return that value so I'll make that change.

platform-operator/controllers/clusters/sync_rancher_clusters_test.go
```diff
@@ -0,0 +1,251 @@
...
+	asserts.Equal(clusterID1, cr.Status.RancherRegistration.ClusterID)
+
+	err = r.Get(context.TODO(), types.NamespacedName{Name: clusterName2, Namespace: constants.VerrazzanoMultiClusterNamespace}, cr)
+	asserts.NoError(err)
+	asserts.Equal(createdByVerrazzano, cr.Labels[createdByLabel])
+	asserts.Equal("true", cr.Labels[vzconst.VerrazzanoManagedLabelKey])
+	asserts.Equal(clusterID2, cr.Status.RancherRegistration.ClusterID)
+
+	// the pre-existing VMC that is not labeled (so not auto-created) should still be here
+	err = r.Get(context.TODO(), types.NamespacedName{Name: preExistingClusterNotLabeled, Namespace: constants.VerrazzanoMultiClusterNamespace}, cr)
```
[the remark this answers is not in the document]
fredtibbitts: Will do.

platform-operator/controllers/clusters/sync_rancher_clusters.go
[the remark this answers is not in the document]
fredtibbitts: Just to close the loop on this, we determined that Rancher cluster names are indeed unique, so we are now naming the VMC with the same name as the Rancher cluster. The cluster ID is set in the VMC status.

platform-operator/controllers/clusters/rancher.go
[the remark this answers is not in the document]
fredtibbitts: Done.

platform-operator/controllers/clusters/sync_rancher_clusters_test.go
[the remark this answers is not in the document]
fredtibbitts: Done.

PR #3802, opened 2022-08-08
VZ 6569: v1beta1 spoke implementation

anders-swanson:
Implement API spoke for v1beta1 API


platform-operator/apis/verrazzano/v1beta1/convert_from_istio.go
```diff
@@ -0,0 +1,228 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+// Adapts code from istio_cr.go to convert the Istio component into a YAML document.
+
+package v1beta1
+
```
michael-cico, line 7: this looks like it was copied from the Istio component...maybe we could relocate this code somewhere common?  Although might be tough to avoid import cycles.  Maybe the top-level `pkg` would work.

Or can the corresponding component code be removed with these API conversions?

michael-cico (approved):

platform-operator/apis/verrazzano/v1beta1/convert_from_istio.go
anders-swanson: The component code will be heavily modified/removed once the client is updated

platform-operator/apis/verrazzano/v1alpha1/verrazzano_conversion.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2020, 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package v1alpha1
+
+// Hub Because v1alpha1 is the storage version, is is used as the Hub interface for conversion
+func (v *Verrazzano) Hub() {}
```
pfmackin, line 7: why are you keeping the old version as the storage version?  I would would expect the newer version, but don't have any proof of that.

anders-swanson: However if we have v1alpha1 as the storage version, it means implementing conversion again if we change the storage version, so this may not be best. Will do some more research on it.

pfmackin: I think this is what you are talking about
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

Do not add a new API version of an existing resource and make it the preferred version in the same release, and do not make it the storage version. The latter is necessary so that a rollback of the apiserver doesn't render resources in etcd undecodable after rollback.

platform-operator/apis/verrazzano/v1beta1/convert_from.go
```diff
@@ -0,0 +1,777 @@
...
+	authProxyReplicasKey = "replicas"
+	authProxyAffinityKey = "affinity"
+)
+
+type expandInfo struct {
+	leftMargin int
+	key        string
+}
+
+func (in *Verrazzano) ConvertFrom(srcRaw conversion.Hub) error {
```
pfmackin, line 42: Is this converting directly to v1beta1 or to an internal format

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

The conversion process is logically a "star" with the internal form at the center. Every versioned API can be converted to the internal form (and vice-versa), but versioned APIs do not convert to other versioned APIs directly.

anders-swanson: This is the kubebuilder conversion method where you choose one API as the Hub, and the other APIs are the Spokes.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_conversion.go
anders-swanson: Yeah, that's it. Thanks

pfmackin (approved):

PR #3804, opened 2022-08-09
VZ-6707 Fixed bug VZ goes to InstallComplete even when Jaeger components are not ready

anajosep:
The PR is to fix VZ-6707 where the VZ status goes to `InstallComplete` state even when Jaeger components (jaeger-query, jaeger-collector) pods does not reach the `Running` state with healthy containers. The fix is to wait till the Jaeger component pods are ready to make the VZ status to ready.


anders-swanson:
@anajosep looks like there is some issue with duplication

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -135,6 +135,25 @@ func isJaegerOperatorReady(ctx spi.ComponentContext) bool {
...
+		},
+	}
+	jaegerCollectorDeployment := []types.NamespacedName{
+		{
+			Name:      JaegerCollectorDeploymentName,
+			Namespace: ComponentNamespace,
+		},
+	}
+	prefix := fmt.Sprintf("Component %s", ctx.GetComponent())
+	return status.DeploymentsAreReady(ctx.Log(), ctx.Client(), jaegerQueryDeployment, 1, prefix) &&
```
pshirodkar-orcl, line 153: Both deployments can be checked in a single call to `status.DeploymentsAreReady` method - refer https://github.com/verrazzano/verrazzano/blob/7afe909fc479bf3cc1e32ba70348f4de46ad1db0/platform-operator/controllers/verrazzano/component/certmanager/certmanager.go#L407

anajosep: Fixed it in the latest commit.

ayaranga (dismissed):

pshirodkar-orcl (approved):

PR #3814, opened 2022-08-09
VZ-6476 VZ-6716 Jaeger Tracing in a Mulitcluster environment

ayaranga:
The following changes are done:
- VMC controller code changes to populate all the required fields in MC registration secret for the Jaeger instance on the managed cluster to connect to OpenSearch storage used by the Jaeger instance on the Admin cluster
- MC Agent on managed cluster to create/update a Jaeger CR based on the Jaeger storage information available in the MC registration secret
- Add verrazzano_cluster collector tag appropriately


application-operator/mcagent/mcagent_jaeger.go
```diff
@@ -0,0 +1,200 @@
...
+// jaegerData needed for template rendering
+type jaegerData struct {
+	ClusterName   string
+	OpenSearchURL string
+	SecretName    string
+	MutualTLS     bool
+}
+
+// A template to define Jaeger CR for creating a Jaeger instance in managed cluster.
+const jaegerManagedClusterCRTemplate = `
```
desagar, line 42: which NS will the CR be created in? I think it would be better to not use the default namespace.

application-operator/mcagent/mcagent_jaeger.go
```diff
@@ -0,0 +1,200 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secName,
+			Namespace: jaegerNamespace,
+		},
+	}
+	if _, err := controllerruntime.CreateOrUpdate(context.TODO(), s.LocalClient, secret, func() error {
+		if secret.Data == nil {
+			secret.Data = make(map[string][]byte)
+		}
+		if _, exists := mcSecret.Data[mcconstants.JaegerOSUsernameKey]; exists {
```
desagar, line 179: Shouldn't we be setting the fields to exactly what is in registration secret, even if the value is empty?

If the registration info changed (for example, CA cert is no longer needed so that field is removed), then this will not change it in the managed cluster jaeger secret, because now that field would not exist and we will just keep the old value in the Jaeger secret on managed cluster.

pkg/constants/constants.go
```diff
@@ -120,9 +120,16 @@ const TestPrometheusJobScrapeInterval = "0s"
 // TestPrometheusJob - Name of a test prometheus scraper job
 const TestPrometheusScrapeJob = "test_job"
 
-// Default Opensearch URL
+// Default OpenSearch URL
 const DefaultOpensearchURL = "http://verrazzano-authproxy-elasticsearch:8775"
 
+// Default Jaeger OpenSearch URL
+const DefaultJaegerOSURL = "http://verrazzano-authproxy-elasticsearch.verrazzano-system.svc.cluster.local:8775"
```
desagar, line 127: same question as I had asked before - why do we need a separate constant? verrazzano-auth-proxy-elasticsearch is in verrazzano-system, so this full URL should work for all the cases. 

Not specifying verrazzano-system.svc.cluster.local is just a shorthand when it is in the same namespace. Even for same NS, it should work if you specify the full URL.

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,180 @@
...
+func (r *VerrazzanoManagedClusterReconciler) getJaegerOpenSearchConfig(vzList *vzapi.VerrazzanoList) (*jaegerOpenSearchConfig, error) {
+	jc := &jaegerOpenSearchConfig{}
+	// Get the OpenSearch URL, secret and other details configured in Jaeger spec
+	jsc, err := r.getJaegerSpecConfig(vzList)
+	if err != nil {
+		return jc, r.log.ErrorfNewErr("Failed to fetch the Jaeger spec details from VZ CR %v", err)
+	}
+	// If Jaeger instance creation is disabled in the VZ CR, then just return
+	if !jsc.jaegerCreate {
+		r.log.Infof("Jaeger instance creation is disabled in the VZ CR. Skipping to configure the Jaeger" +
```
desagar, line 55: minor wording change would be good here. 
"Jaeger instance creation is disabled in the Verrazzano CR. Skipping multicluster Jaeger configuration."

Similarly for the next Info log on line 61.

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,180 @@
...
+			" details for Multicluster.")
+		return jc, nil
+	}
+	// If OpenSearch storage is not configured for the Jaeger instance, then just return
+	if jsc.storageType != "elasticsearch" {
+		r.log.Infof("A Jaeger instance with OpenSearch storage is not configured. Skipping to configure the Jaeger" +
```
desagar, line 61: Maybe this should be a warning level log? (also see wording suggestion above)

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,180 @@
...
+			" details for Multicluster.")
+		return jc, nil
+	}
+
+	jaegerSecret, err := r.getSecret(jaegerNamespace, jsc.secName, true)
```
desagar, line 66: Is the jaegerSecret always required? It looks like we only use it in the else case i.e. when OSURL is not the default one. 
If it's not required, then maybe we should ignore notfound error here?

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	jc.username = jaegerSecret.Data[mcconstants.JaegerOSUsernameKey]
+	jc.password = jaegerSecret.Data[mcconstants.JaegerOSPasswordKey]
+	return jc, nil
+}
+
+// getJaegerSpecConfig returns the OpenSearch URL, secret and other details configured for Jaeger instance
+func (r *VerrazzanoManagedClusterReconciler) getJaegerSpecConfig(vzList *vzapi.VerrazzanoList) (jaegerSpecConfig, error) {
+	jsc := jaegerSpecConfig{}
+	// what to do when there is more than one Verrazzano CR
```
desagar, line 98: you can remove this comment - we don't currently plan to allow more than one Verrazzano CR. The function you copied from has the comment because it was written before we knew that we would not allow more than one.

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+		}
+	}
+	return jsc, nil
+}
+
+// canUseVZOpenSearchStorage determines if Verrazzano's OpenSearch can be used as a storage for Jaeger instance.
+// As default Jaeger uses Authproxy to connect to OpenSearch storage, check if Keycloak component is also enabled.
+func canUseVZOpenSearchStorage(vz vzapi.Verrazzano) bool {
+	if (vz.Spec.Components.Elasticsearch != nil && vz.Spec.Components.Elasticsearch.Enabled != nil &&
```
desagar, line 174: I believe the right way to check for other components enabled is:

vzconfig.IsElasticsearchEnabled(vz)
vzconfig.IsKeycloakEnabled(vz)

We should use vzconfig because want only a small number of places to have the knowledge of whether a component is enabled by default (what to do when it is nil etc). Otherwise if we change the default behavior for a component, we will have to change every place that has this logic.

Example of this usage: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/fluentd/fluentd.go#L35

application-operator/mcagent/mcagent_jaeger.go
ayaranga: Yes, it will be created in verrazzano-monitoring namespace. Later in this file, we are setting the namespace explicitly. I will add it here as well to avoid confusion.

application-operator/mcagent/mcagent_jaeger.go
ayaranga: We do not want to set the empty fields as well. To address the concern that you shared, modified the code to create the secret Data freshly all the time.

pkg/constants/constants.go
ayaranga: The hardcoded URL "http://verrazzano-authproxy-elasticsearch:8775" is being used in many places in fluentd component and there are certain conditional blocks checking for this value. Without analyzing the impact completely, I did not want to make a change to the existing URL. We can take this up in a separate PR if required.

platform-operator/controllers/clusters/jaeger.go
ayaranga: Thanks. Done.

platform-operator/controllers/clusters/jaeger.go
ayaranga: Did the wording change as suggested above. I don't see a warning level log in VMC log.

platform-operator/controllers/clusters/jaeger.go
ayaranga: Yes, jaegerSecret is always required to get the username and password details for both the use cases. See the subsequent lines after the else block.

platform-operator/controllers/clusters/jaeger.go
ayaranga: Ok, thanks. Done.

platform-operator/controllers/clusters/jaeger.go
ayaranga: My bad. I had used this in Jaeger Operator VPO code as well. Somehow this has slipped from my mind. Thanks for pointing this out.

pkg/constants/constants.go
desagar: ok fine sounds good

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,181 @@
...
+func (r *VerrazzanoManagedClusterReconciler) getJaegerOpenSearchConfig(vzList *vzapi.VerrazzanoList) (*jaegerOpenSearchConfig, error) {
+	jc := &jaegerOpenSearchConfig{}
+	// Get the OpenSearch URL, secret and other details configured in Jaeger spec
+	jsc, err := r.getJaegerSpecConfig(vzList)
+	if err != nil {
+		return jc, r.log.ErrorfNewErr("Failed to fetch the Jaeger spec details from the Verrazzano CR %v", err)
+	}
+	// If Jaeger instance creation is disabled in the VZ CR, then just return
+	if !jsc.jaegerCreate {
+		r.log.Infow("Jaeger instance creation is disabled in the Verrazzano CR. Skipping multicluster Jaeger" +
```
desagar, line 56: I just realized that this is being done in the VMC reconciler. That will get called repeatedly and these messages will be printed over and over. 

We should change this message and the next Infow one to use r.log.Once() instead so that they are not printed repeatedly

platform-operator/controllers/clusters/jaeger.go
```diff
@@ -0,0 +1,181 @@
...
+	jc.username = jaegerSecret.Data[mcconstants.JaegerOSUsernameKey]
+	jc.password = jaegerSecret.Data[mcconstants.JaegerOSPasswordKey]
+	return jc, nil
+}
+
+// getJaegerSpecConfig returns the OpenSearch URL, secret and other details configured for Jaeger instance
+func (r *VerrazzanoManagedClusterReconciler) getJaegerSpecConfig(vzList *vzapi.VerrazzanoList) (jaegerSpecConfig, error) {
+	jsc := jaegerSpecConfig{}
+	for _, vz := range vzList.Items {
+		if isJaegerOperatorEnabled(vz) {
```
desagar, line 100: one of the code smells being reported is in this nested if statement. I think we can make one improvement here to the nesting, which may or may not fix the actual code smell. That is to replace the if isJaegerOperatorEnabled block with this:

if !isJaegerOperatorEnabled(vz) {
  continue
}

platform-operator/controllers/clusters/jaeger.go
ayaranga: I mistook the comment in vzlog.go and thought Infow logs only once
```
// Infow formats a message and logs it once at Info log level
	Infow(msg string, keysAndValues ...interface{})
```
Will change it to  r.log.Once(). Thanks.

desagar (approved):

PR #3816, opened 2022-08-09
VZ-6632 - Add verrazzano-managed=true label on managed cluster resources created by Verrazzano

desagar:

### DIFF
--- a/tests/e2e/pkg/keycloak.go
+++ b/tests/e2e/pkg/keycloak.go
@@ -7,10 +7,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"go.uber.org/zap"
 	"net/http"
 	"net/url"
 	"strings"
+
+	"go.uber.org/zap"
 
 	"io/ioutil"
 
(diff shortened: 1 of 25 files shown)


mgianatagh (approved):

tests/e2e/pkg/keycloak.go
```diff
@@ -216,7 +217,9 @@ func (c *KeycloakRESTClient) DeleteUser(userRealm string, userID string) error {
 		return fmt.Errorf("invalid response")
 	}
 	defer response.Body.Close()
-	if response.StatusCode != 200 {
+	// all responses in the 200 range are acceptable
+	// in practice, this call returns 204 (No Content) for success
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
```
fredtibbitts, line 222: Are these keycloak.go changes supposed to be included in this PR? Just making sure this is intentional.

fredtibbitts (approved):
This looks fine to me. The only question I have is if we should be using an existing label, like "app.kubernetes.io/created-by". If we're already using the verrazzano-managed label in other places then it's probably better to be consistent and just use that.

tests/e2e/pkg/keycloak.go
desagar: they are not specifically related to this. I made this change in a recent PR which fixed some jaeger tests, but that was reverted by Anand, so I'm including this here.

desagar:
> This looks fine to me. The only question I have is if we should be using an existing label, like "app.kubernetes.io/created-by". If we're already using the verrazzano-managed label in other places then it's probably better to be consistent and just use that.

We are using verrazzano-managed to label application namespaces right now, including ones created by VerrazzanoProjects. I figured I could extend its use to these resources as well. I'll merge this now and we can discuss potentially using a different label. I know there is some work planned for standardizing labels.

PR #3817, opened 2022-08-09
Updated velero container images to use datetime, commit based tags

anajosep:
Velero images in ghcr.io did not have tags of the format `v<version>-datetime-commit_id`. The current PR is to update the BOM file with the correct image format. The respective images are correctly tagged and pushed to ghcr.io already.


theabmitra (approved):
lgtm

PR #3831, opened 2022-08-10
Replace cluster snapshot with bug report for test failures

bhatpmk:
This PR replaces cluster snapshot with bug-report when the test is instrumented to collect cluster information on failures. Capturing a full cluster snapshot is determined by a system property, defining the environment variable (with default value as 'false') will be done as a separate change. The additional namespace(s) set by the test is supplied for the flag --include-namespaces of CLI bug-report.


agv426 (approved):

bhatpmk:
VZ-6695

PR #3832, opened 2022-08-10
VZ-6709 update the Prometheus adapter and Prometheus pushgateway image tags for OL8 rebuild

davisgomes:


platform-operator/verrazzano-bom.json
```diff
@@ -513,7 +513,7 @@
           "images": [
             {
               "image": "prometheus-pushgateway",
-              "tag": "v1.4.2",
+              "tag": ":v1.4.2-20220810135136-71204144",
```
fredtibbitts, line 516: Bad ":" character

fredtibbitts (changes requested):

platform-operator/verrazzano-bom.json
davisgomes: Fixed

fredtibbitts (approved):

PR #3848, opened 2022-08-11
VZ-6732 - Update velero images with image tags using datetime and short commit id

anajosep:
VZ-6732 - BFS builds for Velero related projects were not using the image tags with datetime-short commit id. The latest builds fixed this issue. In order to re-build the images new build had to be triggered after updating the rpm spec version. Docker files were modified to use the latest patched version of Go 1.17 in both OL7 and OL8.


theabmitra (approved):
lgtm

PR #3854, opened 2022-08-12
VZ-6735 Fix Istio CR to use default Jaeger collector URL managed by VZ at bootstrap

anajosep:
### Problem  statement
Currently when Istio is bootstrapped, it uses the default `zipkin` tracing URL as `zipkin.istio-system.svc.cluster.local`. When Jaeger operator is enabled, the default Jaeger instance created and managed by VPO (via Jaeger Operator) is accessible at **`jaeger-operator-jaeger-collector.verrazzano-monitoring.svc.cluster.local`**

VPO watches for any new Jaeger collector service in the cluster and triggers a reconcile of all VZ components when it detects an Jaeger collector service.

Since Jaeger is the last service that is created, all pods that were created prior to it requires a restart to pick up the latest tracing configuration (collector_url, threshold, tls settings)
### Action
- Whether Jaeger is enabled or not, Istio bootstraps with the Jaeger collector service URL (this is known beforehand) corresponding to the default Jaeger instance managed by VPO (via Jaeger Operator). The Istio tracing flag is disabled until user explicitly enables it throug Istio `InstallArgs`.
- VPO would no longer watch for Jaeger collector service in the cluster and trigger a reconciliation. This should resolve some of the instabilities caused in the Istio System.
- If user wants to use a custom Jaeger Instance managed directly via the Jaeger operator, then Istio install args should be overridden using `meshConfig.defaultConfig.tracing.zipkin.address` to provide the correct value. VPO would no longer derive the URL auto magically.


pkg/constants/constants.go
```diff
@@ -130,6 +130,9 @@ const DefaultJaegerOSURL = "http://verrazzano-authproxy-elasticsearch.verrazzano
 // #nosec
 const DefaultJaegerSecretName = "verrazzano-jaeger-secret"
 
+// JaegerInstanceName is the name of the the default Jaeger instance
+const JaegerInstanceName = "jaeger-operator-jaeger"
```
ayaranga, line 134: Replace `JaegerInstanceName` in `platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go` also with this constant

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -213,10 +213,10 @@ var _ = t.Describe("VMI", Label("f:infra-lcm"), func() {
 				)
 			})
 
-		t.It("Elasticsearch health should be green", func() {
-			Eventually(elasticHealth, elasticWaitTimeout, elasticPollingInterval).Should(BeTrue(), "cluster health status not green")
-			Eventually(elasticIndicesHealth, elasticWaitTimeout, elasticPollingInterval).Should(BeTrue(), "indices health status not green")
-		})
+		//t.It("Elasticsearch health should be green", func() {
```
anders-swanson, line 216: commented code?

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -443,23 +443,23 @@ func elasticConnected() bool {
...
-
-func elasticIndicesHealth() bool {
-	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
-	if err != nil {
-		t.Logs.Errorf("Failed to get default kubeconfig path: %s", err.Error())
-		return false
-	}
-	return elastic.CheckIndicesHealth(kubeconfigPath)
-}
+//func elasticHealth() bool {
```
anders-swanson, line 446: commented code?

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -4,74 +4,34 @@
...
-func zipkinPort(service v1.Service) int32 {
-	for _, port := range service.Spec.Ports {
-		if port.Name == "http-zipkin" {
-			return port.Port
-		}
-	}
-	return 9411
+	// During istio bootstrap, if Jaeger operator is not enabled, or if the Jaeger services are not created yet,
+	// use the collector URL of the default Jaeger instance that would eventually be created.
+	collectorURL := fmt.Sprintf("%s-%s.%s.svc.cluster.local:%d",
```
desagar, line 21: This should be changed to have a trailing "." after the word "local"

platform-operator/controllers/verrazzano/component/istio/istio_cr_test.go
```diff
@@ -521,9 +543,88 @@ var cr4 = &vzapi.IstioComponent{
...
+			Name:  "meshConfig.enableTracing",
+			Value: "true",
+		},
+		{
+			Name:  "meshConfig.defaultConfig.tracing.sampling",
+			Value: "100",
+		},
+		{
+			Name:  "meshConfig.defaultConfig.tracing.zipkin.address",
+			Value: "jaeger-collector.foo.svc.cluster.local:5555",
```
desagar, line 561: Correspondingly change the "local" to "local." (add trailing ".")

pkg/constants/constants.go
anajosep: Done. Using the global constant for getting this value.

tests/e2e/verify-infra/vmi/vmi_test.go
anajosep: fixed.

tests/e2e/verify-infra/vmi/vmi_test.go
anajosep: fixed.

platform-operator/controllers/verrazzano/component/istio/jaeger.go
anajosep: Updated the new URL with a trailing `.`

platform-operator/controllers/verrazzano/component/istio/istio_cr_test.go
anajosep: Fixed

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -103,6 +103,7 @@ const jaegerCreateTemplate = `jaeger:
       options:
         es:
           server-urls: {{.OpenSearchURL}}
+          num-replicas: 0
```
desagar, line 106: why is this zero?

anajosep: This value is applicable more for the dev profile. In dev profile OS is a single node cluster. So having replica=1 would make the replica shards to remain un-assigned causing the OS cluster to become unstable (yellow) state. By putting the value to 0, we can avoid this. Will check if there is a way that this can be done dynamically based on the number of OS nodes available.

ayaranga: We need to set this to 1 for the prod case. That's what we do for other indices in OpenSearch. Based on the profile, at least set it to either 0 or 1.

ayaranga: If the user wants a different value, he can always override `jaeger.spec.storage.options.es.num-replicas` in the VZ CR.

ayaranga (dismissed):

ayaranga (approved):

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
anajosep: Fixed the code to derive the replica number based on the number of data nodes of OS cluster.

PR #3855, opened 2022-08-12
VZ-4042 Introduction to VZ Components Version during VZ Install (via BOM) 

absBitm:
Introduction of vz component's version during vz install 
Component's version are feed into bom and fetched from there during vz installations


pkg/bom/bom.go
```diff
@@ -169,6 +170,16 @@ func (b *Bom) GetVersion() string {
 	return b.bomDoc.Version
 }
 
+// GetComponentVersion gets the BOM component's version
+func (b *Bom) GetComponentVersion(component string) string {
+	for _, comp := range b.bomDoc.Components {
```
michael-cico, line 175: Might be good to have a `FindComponent` utility method to look up a component by name, then use that here.

pkg/bom/bom.go
```diff
@@ -169,6 +170,16 @@ func (b *Bom) GetVersion() string {
 	return b.bomDoc.Version
 }
 
+// GetComponentVersion gets the BOM component's version
+func (b *Bom) GetComponentVersion(component string) string {
```
michael-cico, line 174: Need to add a unit test for this also.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -625,6 +627,12 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	// Set the state of resource
 	componentStatus.State = checkCondtitionType(conditionType)
 
+	// Set the version of component
+	bomFile, _ := bom.NewBom(config.GetDefaultBOMFilePath())
```
michael-cico, line 631: I wonder if we should make the BOM object a member of the controller object, and then lazily initialize it on first access?  That way we aren't parsing the BOM on every reconcile, or for every component we call. 

Since this is a method, it makes sense I think to make it a member of the controller struct.

platform-operator/verrazzano-bom.json
```diff
@@ -43,6 +43,7 @@
     },
     {
       "name": "cert-manager",
```
michael-cico, line 45: there's a lot of components missing versions; is this just the first pass?

platform-operator/controllers/verrazzano/controller.go
michael-cico: so, for instance, you can have a `getBom()` method on the controller lazily initializes it and stores it in the controller struct:

```
// Reconciler reconciles a Verrazzano object
type Reconciler struct {
	client.Client
	Scheme            *runtime.Scheme
	Controller        controller.Controller
	DryRun            bool
	WatchedComponents map[string]bool
	WatchMutex        *sync.RWMutex
        bom                      *bom.Bom
}

func (r *Reconciler) getBOM() (*bom.Bom, error) {
   if (r.bom == nil) {
     if r.bom, err = bom.NewBom(config.GetDefaultBOMFilePath()); err != nil {
       return nil, err
     }
   }
   return r.bom, nil
```

You'd have to check for errors on the calling side, etc and requeue if necessary.  An alternative is to initialze the BOM field when the controller is created, and kill the startup in the VPO main.  If the BOM can't be parsed, then we can't do much anyway. Then you can just have `getBOM()` return the `bom.Bom` pointer.

pkg/bom/bom.go
absBitm: done

platform-operator/controllers/verrazzano/controller.go
absBitm: done
I guess we don't need to requeue from calling side

absBitm: I was thinking to use r.BOm as sync.Once, but not required for the first pass

platform-operator/verrazzano-bom.json
absBitm: yes correct

pkg/bom/bom.go
absBitm: carried to pr [3892](https://github.com/verrazzano/verrazzano/pull/3892)

platform-operator/controllers/verrazzano/controller.go
absBitm: carried to pr [3892](carried to pr https://github.com/verrazzano/verrazzano/pull/3892)

PR #3864, opened 2022-08-12
VZ-6742 add Upgrade complete check to avoid Install after Upgrade

davisgomes:
The install loop has been occurring immediately after an Upgrade completes.

I suspect that is due to the "updating" phase of the upgrade not being checked for completeness. If a component is not verified to be ready, the upgrade will complete and an install might start before that component becomes ready. This verification should keep the upgrade alive until all components are ready.

A new addition to this PR is updating the VZ status immediately after an upgrade has completed. This assures that we do not enter another upgrade loop after the upgrade is complete. Instead, we enter the ready state process as the final reconcile loop after upgrade.


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -330,12 +330,18 @@ func (r *Reconciler) ProcUpgradingState(vzctx vzcontext.VerrazzanoContext) (ctrl
 		return result, nil
 	}
 
+	if done, err := r.checkUpgradeComplete(vzctx); !done || err != nil {
```
mbenmclean, line 333: This probably wants to be a throttled log message. If it has to wait for a while it could requeue a bunch of times and spam the logs with this message

davisgomes: I originally had it as `Progressf`. I'll change it back to that.

pfmackin: Progressf is what you want.  We don't want to log an error.

michael-cico, line 333: I feel like this should run before the `reconcileComponents` is done?  Or does that blow the upgrade status/states?

davisgomes: This check verifies that all the components are ready regardless of their upgrade/install method. Placing it here will catch any in post-installs that are still in progress which I suspect is the root of the initial bug.

michael-cico (approved):

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -330,12 +330,18 @@ func (r *Reconciler) ProcUpgradingState(vzctx vzcontext.VerrazzanoContext) (ctrl
...
+		log.Progressf("Upgrade is waiting for all components to enter a Ready state before completion")
+		return newRequeueWithDelay(), err
+	}
+
+	if err := r.updateVzState(log, actualCR, installv1alpha1.VzStateReady); err != nil {
```
pfmackin, line 338: This code is calling update twice in a row (once for checkUgpradeComplete and again for updateVzState.  This will likely hit a conflict error causing a re-queue.  Better if you just modified the VZ CR and did 1 update.

davisgomes: Thanks. I created a new helper function that does both updates.

pfmackin: Don't you need to remove this since it is done in checkUpgradeComplete

davisgomes: I just removed it as a follow up commit.

pfmackin (approved):

PR #3868, opened 2022-08-15
Versioning analysis advice links

saidu941:

### DIFF
--- a/tools/vz/cmd/version/version_test.go
+++ b/tools/vz/cmd/version/version_test.go
@@ -17,7 +17,7 @@ import (
 // TestVersionCmd - check that command reports not implemented yet
 func TestVersionCmd(t *testing.T) {
 
-	CliVersion = "1.2.3"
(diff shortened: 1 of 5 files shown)


tools/vz/cmd/version/version_test.go
```diff
@@ -38,3 +38,20 @@ func TestVersionCmd(t *testing.T) {
...
+	cliVersion = "1.2.3"
+
+	useV8oDoc := os.Getenv("USE_V8O_DOC_STAGE")
+	if useV8oDoc == "true" {
+		assert.True(t, GetEffectiveDocsVersion() == "devel")
+	} else {
+		assert.True(t, GetEffectiveDocsVersion() == "v1.2")
+	}
+
+	t.Setenv("USE_V8O_DOC_STAGE", "true")
```
agv426, line 52: I would not set the environment variable from the tests at all

agv426 (approved):

PR #3871, opened 2022-08-15
Reduce the noisy logs in metrics exporter for application operator

desagar:

### DIFF
--- a/application-operator/metricsexporter/metricsexporter_utils.go
+++ b/application-operator/metricsexporter/metricsexporter_utils.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


application-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -60,10 +62,21 @@ const (
 )
 
 // InitRegisterStart initalizes the metrics object, registers the metrics, and then starts the server
-func InitRegisterStart(log *zap.SugaredLogger) {
+func InitRegisterStart(log *zap.SugaredLogger) error {
+	vlog, err := vzlog.EnsureResourceLogger(&vzlog.ResourceConfig{
+		Name:           "",
```
fredtibbitts, line 67: Seems like these fields should be populated with something reasonable, although I don't know exactly how they are used or if leaving them blank might cause issues.

desagar: this causes some fields to be empty in the log that is printed:
"resource_namespace":"","resource_name":"","controller":"metricsexporter"

I think that's fine because this is the metrics exporter and not a reconciler being called for a specific resource.

fredtibbitts (approved):

PR #3882, opened 2022-08-16
VZ 6569: Conversion

anders-swanson:
Conversion refactor, v1beta1 is used as the storage version.


pfmackin (dismissed):

michael-cico (approved):

michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
```diff
@@ -127,6 +128,11 @@ func (c applicationOperatorComponent) ValidateUpdate(old *vzapi.Verrazzano, new
 	return c.HelmComponent.ValidateUpdate(old, new)
 }
 
+// ValidateUpdate checks if the specified new Verrazzano CR is valid for this component to be updated
+func (c applicationOperatorComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzano, new *installv1beta1.Verrazzano) error {
```
pfmackin, line 132: general question, since this function is in the helm component, why does every component have to override?

pfmackin (approved):

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
anders-swanson: It doesn't, we are going to fill in those stubs with v1beta1 validation implementations, so it doesn't hurt to have them at this point. They are markers

PR #3892, opened 2022-08-18
VZ-4042 Introduction to VZ Components Version during VZ Install (via BOM)

absBitm:
Introducing vz component version during vz install
extension pr of [3855](https://github.com/verrazzano/verrazzano/pull/3855)


pkg/bom/testdata/verrazzano-bom.json
```diff
@@ -43,6 +43,7 @@
     },
     {
       "name": "cert-manager",
```
pfmackin, line 45: You should add a version for every component

platform-operator/verrazzano-bom.json
```diff
@@ -351,6 +359,7 @@
     },
     {
       "name": "kiali-server",
+      "version": "0.0.0",
```
pfmackin, line 362: This is not a valid version

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -670,6 +684,13 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	// Set the state of resource
 	componentStatus.State = checkCondtitionType(conditionType)
 
+	// Set the version of component
```
pfmackin, line 687: This should only be done post-install/update and post-upgrade.  Isn't this function called during those operations?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -670,6 +684,13 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
...
+	if bomFile, err := r.getBOM(); err != nil {
+		if component, er := bomFile.GetComponent(componentName); er != nil {
```
pfmackin, line 689: Don't update components that are disabled

pkg/bom/testdata/verrazzano-bom.json
```diff
@@ -43,6 +43,7 @@
...
+      "version": "v1.7.1",
```
michael-cico, line 46: Looks like we're missing some versions for components, e.g., NGINX, Istio?

platform-operator/verrazzano-bom.json
```diff
@@ -43,6 +43,7 @@
     },
     {
       "name": "cert-manager",
+      "version": "v1.7.1",
```
michael-cico, line 46: Looks like we're missing some versions for components, e.g., NGINX, Istio, external-dns?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -670,6 +684,15 @@ func (r *Reconciler) updateComponentStatus(compContext spi.ComponentContext, mes
 	// Set the state of resource
 	componentStatus.State = checkCondtitionType(conditionType)
 
+	// Set the version of component
+	if conditionType == installv1alpha1.CondInstallComplete || conditionType == installv1alpha1.CondUpgradeStarted {
```
michael-cico, line 688: I think from what @pfmackin said we should set the version on `CondUpgradeComplete`, not `CondUpgradeStarted`

platform-operator/verrazzano-bom.json
absBitm: yes for the first pass, we are only focusing on the component which includes all of the Verrazzano components

michael-cico (approved):

PR #3897, opened 2022-08-18
VZ-6686: Create Secret token (if it does not exist) for VerrazzanoManagedCluster Service Account

pshirodkar-orcl:
Create Secret token (if it does not exist) for VerrazzanoManagedCluster Service Account.
Fixes VZ-6686

Note: When VZ-6449 gets implemented, we will be switching to TokenRequest API for this particular Service Account.

Relates to Kubernetes 1.24 https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade

> The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))

### DIFF
--- a/platform-operator/controllers/clusters/vmc_controller.go
+++ b/platform-operator/controllers/clusters/vmc_controller.go
@@ -192,9 +192,16 @@ func (r *VerrazzanoManagedClusterReconciler) doReconcile(ctx context.Context, lo
 
 func (r *VerrazzanoManagedClusterReconciler) syncServiceAccount(vmc *clustersv1alpha1.VerrazzanoManagedCluster) error {
 	// Create or update the service account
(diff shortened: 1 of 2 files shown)


wmhopkins (approved):
lgtm.

desagar (approved):

PR #3905, opened 2022-08-19
VZ-6564 Add Keycloak groups to Rancher cluster

mbenmclean:
This change adds the Keycloak groups `verrazzao-admins` and `verrazzano-monitors` to the local cluster in Rancher. It does this by creating ClusterRoleTemplateBindings to add RoleTemplates for those groups. It also creates RoleTemplates for the verrazzano-admin and verrazzano-monitor groups.


tests/e2e/verify-infra/restapi/rancher_url_test.go
```diff
@@ -198,7 +198,7 @@ var _ = t.Describe("rancher", Label("f:infra-lcm",
 							return false, fmt.Errorf("verrazzano rancher user global role binding user in invalid")
 						}
 
-						if grbAttributes[rancher.GlobalRoleBindingAttributeRoleName].(string) != rancher.GlobalRoleBindingRoleName {
+						if grbAttributes[rancher.GlobalRoleBindingAttributeRoleName].(string) != rancher.AdminRoleName {
```
lake-of-dreams, line 201: we should probably test that role templates and bindings are created here

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -177,7 +177,7 @@ func removeCRDFinalizers(ctx spi.ComponentContext) {
 
 // deleteMatchingResources delete the Rancher objects that need to match a string: ClusterRole, ClusterRoleBinding, RoleBinding, PersistentVolumes
 func deleteMatchingResources(ctx spi.ComponentContext) error {
-	// list of matching prefixes for cluster roles, cluster role bindings
+	// list of matching prefixes for cluster roles, cluster clusterRole bindings
```
lake-of-dreams, line 180: `cluster clusterRole` ? May be should be just `clusterRole`

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -364,6 +372,51 @@ func configureAuthProviders(ctx spi.ComponentContext, isUpgrade bool) error {
...
+func createOrUpdateRoleTemplates(ctx spi.ComponentContext) error {
+	if err := createOrUpdateRoleTemplate(ctx, VerrazzanoAdminRoleName); err != nil {
+		return err
+	}
+
+	return createOrUpdateRoleTemplate(ctx, VerrazzanoMonitorRoleName)
+}
+
+func createOrUpdateClusterRoleTemplateBindings(ctx spi.ComponentContext) error {
+	groupRolePairs := []groupRolePair{
```
lake-of-dreams, line 384: I don't see the `struct` `groupRolePair` being used anywhere else apart from just looping here. Can't it be implemented using just map[string][]string ?

platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -521,81 +571,106 @@ func configureKeycloakOIDC(ctx spi.ComponentContext) error {
...
+	if clusterRole.Rules != nil && len(clusterRole.Rules) > 0 {
+		data[RoleTemplateAttributeRules] = clusterRole.Rules
+	}
+
+	return createOrUpdateResource(ctx, nsn, GVKRoleTemplate, data)
+}
+
+func createOrUpdateClusterRoleTemplateBinding(ctx spi.ComponentContext, clusterRole string, group string) error {
+	name := fmt.Sprintf("crtb-%s-%s", clusterRole, group)
+	nsn := types.NamespacedName{Name: name, Namespace: ClusterLocal}
```
lake-of-dreams, line 666: are these created in `local` namespace?

platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -521,81 +571,106 @@ func configureKeycloakOIDC(ctx spi.ComponentContext) error {
 	return nil
 }
 
-func createOrUpdateRancherVerrazzanoUser(ctx spi.ComponentContext) error {
+func createOrUpdateResource(ctx spi.ComponentContext, nsn types.NamespacedName, gkv schema.GroupVersionKind, attributes map[string]interface{}) error {
```
lake-of-dreams, line 574: `gkv` should be `gvk`

tests/e2e/verify-infra/restapi/rancher_url_test.go
mbenmclean: This is just a name change. I can add it though if you'd like.

platform-operator/controllers/verrazzano/component/rancher/rancher.go
mbenmclean: Yes they have to be.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
mbenmclean: changed

tests/e2e/verify-infra/restapi/rancher_url_test.go
mbenmclean: added that

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
mbenmclean: fixed

tests/e2e/verify-infra/restapi/rancher_url_test.go
mbenmclean: added

wmhopkins (approved):
lgtm.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: @mbenmclean this is not what you have implemented. You are taking an array of maps with each one containing just one element `[]map[string]string` while what I had suggested was a map with group as key and roles in the value array `map[string][]string` . Both are different data structures. Any reason why you would implement it this way ?

platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -521,83 +596,114 @@ func configureKeycloakOIDC(ctx spi.ComponentContext) error {
 	return nil
 }
 
-func createOrUpdateRancherVerrazzanoUser(ctx spi.ComponentContext) error {
+// createOrUpdateResource creates or updates a Rancher resource used in the Keycloak OIDC integration
+func createOrUpdateResource(ctx spi.ComponentContext, nsn types.NamespacedName, gvk schema.GroupVersionKind, attributes map[string]interface{}) error {
```
lake-of-dreams, line 600: Why was this func added in between and not at the end of file, going to make rebases difficult..

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
mbenmclean: Because I think it's helpful to heave each pair be it's own element you can iterate through. I suppose if I did map[string][]string I could iterate through the indicies and index the []string to get the values. I'm not a big fan of that implementation though because accessing elements gets a little verbose. Is there some reason you favor that implementation?

platform-operator/controllers/verrazzano/component/rancher/rancher.go
mbenmclean: Sorry about that will keep in mind in the future

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
lake-of-dreams: Well its just that when I see some special _structure_ created for containing static data, I would assume that it will be used somewhere else too apart from just looping, for example if we somewhere we were to search in this structure using the `GroupKey` (even that would have been difficult since we will have to traverse all the maps). In test also we are just looping over it and verifying every pair. Also I think we could have used `UnstructuredList` to create these all at once rather than one by one and in that case too, having a map with key as the Group and value as array containing all the roles that we want to assign , would have helped. Similarly in test we could have fetched all of the roles associated with the group and verified. This could have optimized some of the network calls. Nothing major but just wanted to highlight the importance of primitives :)

platform-operator/controllers/verrazzano/component/rancher/rancher.go
lake-of-dreams: sure, no problems

PR #3906, opened 2022-08-19
Ksnelson/vz 6749

knelson7:
two new methods to the ComponentValidator interface and added default implementation for each component in registry

### DIFF
--- a/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
+++ b/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
@@ -9,6 +9,7 @@ import (
 	installv1beta1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1beta1"
 	"path/filepath"
 	"strings"
+
 	vmcv1alpha1 "github.com/verrazzano/verrazzano/platform-operator/apis/clusters/v1alpha1"
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
--- a/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
+++ b/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
@@ -6,9 +6,9 @@ package appoper
 import (
 	"context"
 	"fmt"
+	installv1beta1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1beta1"
 	"path/filepath"
 	"strings"
-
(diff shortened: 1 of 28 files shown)


platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -109,6 +109,10 @@ type ComponentValidator interface {
 	ValidateInstall(vz *vzapi.Verrazzano) error
 	// ValidateUpdate checks if the specified new Verrazzano CR is valid for this component to be updated
 	ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Verrazzano) error
+	// ValidateInstall checks if the specified Verrazzano CR is valid for this component to be installed
+	ValidateInstallV1Beta1(vz *vzapi.Verrazzano) error
```
anders-swanson, line 113: This needs to be the v1beta1.Verrazzano package. vzapi is an import alias for v1alpha1.Verrazzano

anders-swanson: The type of all the parameters for the v1beta1 validate methods must be a v1beta1.Verrazzano. The v1beta1 validating webhook will be passing down objects of type v1beta1.Verrazzano

anders-swanson (approved):
LGTM

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
```diff
@@ -127,6 +127,11 @@ func (c applicationOperatorComponent) ValidateUpdate(old *vzapi.Verrazzano, new
 	return c.HelmComponent.ValidateUpdate(old, new)
 }
 
+// ValidateUpdate checks if the specified new Verrazzano CR is valid for this component to be updated
+func (c applicationOperatorComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzano, new *installv1beta1.Verrazzano) error {
```
michael-cico, line 131: Is this needed, or will it pick up the `HelmComponent` version by default?

michael-cico: Same for the other components as well.

platform-operator/controllers/verrazzano/component/spi/component.go
michael-cico: should the version be an argument and avoid version-specific interface methods, or will it be more complicated than its worth for the number of API versions we may be supporting at any given time (down the road I mean).

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
anders-swanson: These stubs are going to get filled in as we implement component-level validators. It's not strictly necessary at this point but also will not hurt anything

michael-cico (approved):

anders-swanson (approved):
LGTM

PR #3911, opened 2022-08-22
VZ-5010: Update client secret in keycloak authConfig on keycloak reconcile

lake-of-dreams:
In a case when Rancher upgrade finishes before Keycloak, the `clientSecret` field in Keycloak `AuthConfig` CR might still have a stale value. This PR to update the secret on Keycloak reconcile.


platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -75,3 +84,34 @@ func CertPool(certs ...[]byte) *x509.CertPool {
 	}
 	return certPool
 }
+
+func GetRancherMgmtAPIGVKForKind(kind string) schema.GroupVersionKind {
```
kminder, line 88: Should have a function comment

platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -75,3 +84,34 @@ func CertPool(certs ...[]byte) *x509.CertPool {
...
+	return schema.GroupVersionKind{
+		Group:   APIGroupRancherManagement,
+		Version: APIGroupVersionRancherManagement,
+		Kind:    kind,
+	}
+}
+
+func UpdateKeycloakOIDCAuthConfig(ctx spi.ComponentContext, data map[string]interface{}) error {
```
kminder, line 96: Should have a function comment.

kminder (approved):
Approved.

platform-operator/controllers/verrazzano/component/common/rancher.go
lake-of-dreams: added

bhatpmk (approved):

platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -75,3 +84,36 @@ func CertPool(certs ...[]byte) *x509.CertPool {
...
+func GetRancherMgmtAPIGVKForKind(kind string) schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   APIGroupRancherManagement,
+		Version: APIGroupVersionRancherManagement,
+		Kind:    kind,
+	}
+}
+
+//UpdateKeycloakOIDCAuthConfig updates the keycloakoidc AuthConfig CR with specified attributes
+func UpdateKeycloakOIDCAuthConfig(ctx spi.ComponentContext, attributes map[string]interface{}) error {
```
mbenmclean, line 98: Could you refactor this to use the createOrUpdateResource function I created?

mbenmclean: Oh actually maybe not because it is just update, not create

platform-operator/controllers/verrazzano/component/common/rancher_test.go
```diff
@@ -48,6 +50,15 @@ func createAdminSecret() v1.Secret {
 	}
 }
 
+func createKeycloakAuthConfig() unstructured.Unstructured {
```
mbenmclean, line 53: Add a function comment please

platform-operator/controllers/verrazzano/component/common/rancher.go
lake-of-dreams: yes its just update

platform-operator/controllers/verrazzano/component/common/rancher_test.go
lake-of-dreams: do we need function comments on private funcs also? I believe exported function should must have godocs since they show up on godocs.io but for private funcs it should be optional, unless ofcourse it is doing something complex

mbenmclean: Your call, I think some people do it for all function and some do for just exported ones.

lake-of-dreams: yeah I also do document private functions whenever I see someone adding another public functions in future would intend to call it and the name is not reflective of the complex implementation of the function which does not seem to be case here, so probably would keep it as is in this PR.

PR #3916, opened 2022-08-22
Remove the use of ObjectMeta.ClusterName which is deprecated in K8S

desagar:

### DIFF
--- a/application-operator/controllers/metricstrait/metricstrait_utils_test.go
+++ b/application-operator/controllers/metricstrait/metricstrait_utils_test.go
@@ -92,26 +92,6 @@ func Test_copyStringMapEntries(t *testing.T) {
 	assert.Len(target, 2)
 }
 
-// Test_getClusterNameFromObjectMetaOrDefault tests metrics trait utility function getClusterNameFromObjectMetaOrDefault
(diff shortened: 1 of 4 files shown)


application-operator/controllers/metricstrait/metricstrait_utils_test.go
```diff
@@ -292,8 +272,8 @@ func TestCreateServiceMonitorName(t *testing.T) {
 				},
 			},
 			portNum:                    0,
-			expectedServiceMonitorName: "test-app-default-test-namespace-test-comp",
-			expectedLegacyJobName:      "test-app_default_test-namespace_test-comp",
+			expectedServiceMonitorName: "test-app-test-namespace-test-comp",
```
davisgomes, line 275: Are these always the same? Could you merge them into one?

michael-cico (dismissed):

application-operator/controllers/metricstrait/metricstrait_utils_test.go
desagar: There is the underscore vs hyphen because the SM name is a K8S object name. Also the length restriction for the SM name, though I don't know if we're testing that. Will check.

PR #3919, opened 2022-08-23
VZ-6720  Created GetEffectiveCRV1beta1 utility function to handle V1Beta1 VZ CR and using it in component context

anajosep:
VZ-6720 
In order to make VPO to completely shift to the new `v1beta1` based APIs, the component context is modified to have both `v1alpha1` and `v1beta1` versions of the VZ CR. The values of these are derived by calling the appropriate utility methods - `GetEffectiveCR() -  operates on v1alpha1/Verrazzano`
`GetEffectiveCRV1beta1() -  operates on v1beta1/Verrazzano`

Modified the corresponding Component Context creation methods
- `NewContext`
- `NewFakeContext`

Modified the calling code to adopt to the new signature changes in `NewContext` and `NewFakeContext`


platform-operator/controllers/verrazzano/component/spi/component_context.go
```diff
@@ -17,14 +17,13 @@ import (
 var _ ComponentContext = componentContext{}
 
 // NewContext creates a ComponentContext from a raw CR
-func NewContext(log vzlog.VerrazzanoLogger, c clipkg.Client, actualCR *vzapi.Verrazzano, dryRun bool) (ComponentContext, error) {
-
+func NewContext(log vzlog.VerrazzanoLogger, c clipkg.Client, actualCR *vzapi.Verrazzano, actualV1beta1CR *vzapiv1beta1.Verrazzano, dryRun bool) (ComponentContext, error) {
```
anders-swanson, line 20: I am wondering if this should only take the v1beta1 CR as a parameter, and get the v1alpha1 CR via conversion

anders-swanson: Or if actualCR is nil, get the v1alpha1 CR via conversion

anders-swanson: nevermind, i think you are doing it the best way

platform-operator/controllers/verrazzano/transform/merge.go
```diff
@@ -59,6 +61,45 @@ func MergeProfiles(actualCR *vzapi.Verrazzano, profileFiles ...string) (*vzapi.V
 	return &newCR, nil
 }
 
+// MergeProfiles merges a list of v1beta1.Verrazzano profile files with an existing Verrazzano CR.
+// The profiles must be in the Verrazzano CR format
+func MergeProfilesForV1beta1(actualCR *v1beta1.Verrazzano, profileFiles ...string) (*v1beta1.Verrazzano, error) {
+	// First merge the profiles
+	profileStrings, err := appendProfileComponentOverrides(profileFiles...)
```
anders-swanson, line 68: This function unmarshalls YAML into v1alpha1 verrazzanos

platform-operator/controllers/verrazzano/transform/merge.go
```diff
@@ -121,7 +162,40 @@ func GetEffectiveCR(actualCR *vzapi.Verrazzano) (*vzapi.Verrazzano, error) {
...
-//appendComponentOverrides copies the profile overrides over to the actual overrides. Any component that has overrides should be included here.
+// GetEffectiveV1beta1CR Creates an "effective" v1beta1.Verrazzano CR based on the user defined resource merged with the profile definitions
+// - Effective CR == base profile + declared profiles + ActualCR (in order)
+// - last definition wins
+func GetEffectiveV1beta1CR(actualCR *v1beta1.Verrazzano) (*v1beta1.Verrazzano, error) {
+	if actualCR == nil {
+		return nil, nil
+	}
+	// Identify the set of profiles, base + declared
+	profiles := []string{baseProfile, string(vzapi.Prod)}
```
anders-swanson, line 173: should use v1beta1.Prod

anajosep: Fixed.

anders-swanson (approved):

anders-swanson (approved):

platform-operator/controllers/controller_utils.go
```diff
@@ -68,7 +69,7 @@ func ProcDeletedOverride(c client.Client, vz *installv1alpha1.Verrazzano, object
 
 	// DefaultLogger is used since we only need to create a component context and any actual logging isn't being performed
 	log := vzlog.DefaultLogger()
-	ctx, err := spi.NewContext(log, c, vz, false)
+	ctx, err := spi.NewContext(log, c, vz, nil, false)
```
michael-cico, line 72: Might have saved a little trouble adding a new Context constructor that took a v1beta1 actual, and phased it out if that's the plan.

michael-cico: actually, 6 of 1, half dozen the other.  The only way to avoid the signature change ripple effect would be to use a varargs list at the end of CRs or something, which is worse.  Comment retracted :-)

michael-cico (approved):

platform-operator/controllers/controller_utils.go
anders-swanson: There didn't seem to be an elegant way to do this, simple seemed best

PR #3920, opened 2022-08-23
VZ-6801: Build open-source release distribution from periodic job

bhatpmk:
Builds open-source release distribution from periodic job


agv426 (approved):

michael-cico (approved):

ci/scripts/generate_vz_distribution.sh
```diff
@@ -0,0 +1,188 @@
...
+  cp ${GENERATED_BOM_FILE} ${VZ_OPENSOURCE_ROOT}/manifests/verrazzano-bom.json
+
+  # Extract the CLI for Linux AMD64
+  echo "Extract the CLI for Linux AMD64 ..."
+  tar xzf ${VZ_DISTRIBUTION_COMMON}/${VZ_CLI_LINUX_AMD64_TARGZ} -C ${VZ_OPENSOURCE_ROOT}/bin
+
+  # Build distribution for Linux AMD64 architecture
+  echo "Build distribution for Linux AMD64 architecture ..."
+  tar -czf ${VZ_DISTRIBUTION_GENERATED}/${VZ_LINUX_AMD64_TARGZ} -C ${VZ_OPENSOURCE_ROOT} .
+  sha256sum ${VZ_DISTRIBUTION_GENERATED}/${VZ_LINUX_AMD64_TARGZ} > ${VZ_DISTRIBUTION_GENERATED}/${VZ_LINUX_AMD64_TARGZ_SHA256}
```
agv426, line 123: I think we need to do the sha256 from the current directory the file is in so we only have the base filename encoded into the resulting sha256 file.

We changed up the stuff in object storage which had the absolute directory included in it, which causes problems when that path doesn't exist when trying to verify it

bhatpmk: Thank you. I will address this in a follow-up PR immediately.

bhatpmk: The issue is fixed by https://github.com/verrazzano/verrazzano/pull/3929

PR #3921, opened 2022-08-23
VZ-6772 Update E2E test to create a service account token secret

ayaranga:
From k8s 1.24 onwards, when a ServiceAccount is created, no more Secret
is created automatically. Updated E2E test to create a service account
token secret if not exists.

### DIFF
--- a/tests/e2e/pkg/kubernetes.go
+++ b/tests/e2e/pkg/kubernetes.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


tests/e2e/pkg/kubernetes.go
```diff
@@ -1392,13 +1392,20 @@ func GetTokenForServiceAccount(sa string, namespace string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	var secretName string
```
desagar, line 1395: I would prefer that this GetTokenForServiceAccount did not have a side effect of creating the token secret. Instead, since only one test requires this, we should have the test call the CreateServiceAccountTokenSecret function directly.

pshirodkar-orcl (approved):

PR #3923, opened 2022-08-23
VZ-6740 grafana add component changes into v1beta1 version

absBitm:


platform-operator/apis/verrazzano/v1beta1/convert_from.go
```diff
@@ -429,7 +429,9 @@ func convertGrafanaFrom(src *v1alpha1.GrafanaComponent) *GrafanaComponent {
 		return nil
 	}
 	return &GrafanaComponent{
-		Enabled: src.Enabled,
+		Enabled:  src.Enabled,
+		Replicas: src.Replicas,
+		Database: (*DatabaseInfo)(src.Database),
```
anders-swanson, line 434: We shouldn't use casting, instead you should create a v1beta1.Database object and populate it with the values from the v1alpha1.Database

platform-operator/apis/verrazzano/v1beta1/convert_to.go
```diff
@@ -298,7 +298,9 @@ func convertGrafanaTo(in *GrafanaComponent) *v1alpha1.GrafanaComponent {
 		return nil
 	}
 	return &v1alpha1.GrafanaComponent{
-		Enabled: in.Enabled,
+		Enabled:  in.Enabled,
+		Replicas: in.Replicas,
+		Database: (*v1alpha1.DatabaseInfo)(in.Database),
```
anders-swanson, line 303: Here we need to create a v1alpha1.Database and populate it from the v1beta1.Database

platform-operator/apis/verrazzano/v1beta1/convert_to_test.go
```diff
@@ -53,6 +53,11 @@ func TestConvertTo(t *testing.T) {
 				assert.EqualValues(t, v1alpha1Expected.ObjectMeta, v1alpha1Actual.ObjectMeta)
 				assert.EqualValues(t, v1alpha1Expected.Spec, v1alpha1Actual.Spec)
 				assert.EqualValues(t, v1alpha1Expected.Status, v1alpha1Actual.Status)
+				if v1alpha1Expected.Spec.Components.Grafana != nil {
```
anders-swanson, line 56: Can you remove this check, if should be handled by the EqualValues check on line 54

anders-swanson:
Seems some merge conflicts, can you merge & resolve conflicts again?

platform-operator/apis/verrazzano/v1beta1/convert_to.go
absBitm: done

platform-operator/apis/verrazzano/v1beta1/convert_from.go
absBitm: done

platform-operator/apis/verrazzano/v1beta1/convert_to_test.go
absBitm: all testdata does not have Grafana component therefore it was panicking for those testate which doesn't have this Grafana component v1alpha1Expected.Spec.Components.Grafana
eg. platform-operator/apis/verrazzano/v1beta1/testdata/frominstallargs/v1alpha1.yaml

anders-swanson: It's panicking because DataBaseInfo is nil and we are not checking if its nil in the conversion. When doing the conversion for any pointer objects, we need to check if they are nil (otherwise you can get panics).

e.g.,

```
func convertGrafanaFromV1Beta1(in *v1beta1.GrafanaComponent) *GrafanaComponent {
	if in == nil {
		return nil
	}

	var info *DatabaseInfo
	if in.Database != nil {
		info.Host = in.Database.Host
		info.Name = in.Database.Name
	}
	return &GrafanaComponent{
		Enabled:  in.Enabled,
		Replicas: in.Replicas,
		Database: info,
	}
}
```

For both ConvertFrom/ConvertTo it must be done. Then this if statement can be removed, and we can have `assert.EqualValues(t, v1alpha1Expected.Spec, v1alpha1Actual.Spec)` check on line 54 validate the spec entirely.

platform-operator/apis/verrazzano/v1alpha1/convert_from.go
```diff
@@ -313,8 +313,15 @@ func convertGrafanaFromV1Beta1(in *v1beta1.GrafanaComponent) *GrafanaComponent {
 	if in == nil {
 		return nil
 	}
+	var info *DatabaseInfo
+	if in.Database != nil {
```
anders-swanson, line 317: I gave you bad snippet before, this will cause nil pointer dereference
if src.DataBaseInfo is not nil, we must create a new DataBaseInfo object
```
if in.DataBase != nil {
  info = &DataBaseInfo{
    Host: in.Database.Host,
   Name: in.Database.Name,
  }
}
```

for both convert To and convert From it must be done. You are currently trying to assign values to a nil struct, which causes the panic.

absBitm: yes, done

PR #3924, opened 2022-08-23
VZ-6649.  Utility lib for creating a local CAPI management bootstrap cluster

michael-cico:
This represents in-progress changes for Epic VZ-6793, OCNE in a local container runtime.

These are the initial changes to support creating a local CAPI management cluster using KIND with the VZ CLI.  There will be follow-up work to integrate it with a CLI command
- Updated dependencies, 3rdparty license file for KIND 0.14 and CAPI 1.2.0 usage
- A small utility library to support creating a local bootstrap/management cluster with KIND
- Updated generated clients due to updated versions of k8s libraries
- Updated `controller-gen` target to use `go install` to install binary instead of deprecated `go get`


make/generate.mk
```diff
@@ -44,7 +44,7 @@ generate: controller-gen
 .PHONY: controller-gen
 controller-gen:
 ifeq (, $(shell command -v controller-gen))
-	$(GO) get sigs.k8s.io/controller-tools/cmd/controller-gen@${CONTROLLER_GEN_VERSION}
+	$(GO) install sigs.k8s.io/controller-tools/cmd/controller-gen@${CONTROLLER_GEN_VERSION}
```
pfmackin, line 47: why is this changed?

pkg/capi/kind_cluster_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	defer os2.RemoveTempFiles(zap.S(), config.Name())
+	capiclient, err := capiInitFunc("") // TODO: do we need to provide a CAPI config?
+	if err != nil {
+		return err
+	}
+	_, err = capiclient.Init(clusterapi.InitOptions{
+		Kubeconfig: clusterapi.Kubeconfig{
+			Path: config.Name(),
+		},
+		InfrastructureProviders: []string{"docker"},
```
pfmackin, line 36: can you make this a constant.  Also should it be injected or will it always be docker?

pkg/capi/kind_cluster_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	})
+	return err
```
pfmackin, line 38: We should log error here

pkg/capi/kind_cluster_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	config ClusterConfig
+}
+
+func (r *kindClusterManager) GetKubeConfig() (string, error) {
+	return bootstrapProviderImpl.GetKubeconfig(r.config.ClusterName())
+}
+
+func (r *kindClusterManager) Init() error {
+	config, err := r.createKubeConfigFile()
+	if err != nil {
```
pfmackin, line 24: We should log error here

pkg/capi/kind_cluster_manager.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+	defer os2.RemoveTempFiles(zap.S(), config.Name())
+	capiclient, err := capiInitFunc("") // TODO: do we need to provide a CAPI config?
+	if err != nil {
+		return err
```
pfmackin, line 30: We should log error here

pkg/capi/kind_cluster_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	os2 "github.com/verrazzano/verrazzano/pkg/os"
+	"go.uber.org/zap"
+	"io/ioutil"
+	"os"
+	clusterapi "sigs.k8s.io/cluster-api/cmd/clusterctl/client"
+)
+
+// kindClusterManager ClusterLifecycleManager impl for a KIND-based bootstrap cluster
+type kindClusterManager struct {
+	config ClusterConfig
```
pfmackin, line 15: where is this set?

pkg/capi/cluster_lifecycle.go
```diff
@@ -0,0 +1,44 @@
...
+	Type() string
+	ContainerImage() string
+}
+
+// ClusterLifeCycleManager defines the lifecycle operations of a cluster
+type ClusterLifeCycleManager interface {
+	GetConfig() ClusterConfig
+	GetKubeConfig() (string, error)
+	Create() error
+	Init() error
```
pfmackin, line 35: Does Init have to be called before each create?

pkg/capi/capi_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: there should be a capi.go file that corresponds to capi_test.go

pfmackin (approved):

make/generate.mk
michael-cico: I was getting a warning from `go get` that using `get` for binaries is deprecated, and that `go install` is preferred.

pkg/capi/kind_cluster_manager.go
michael-cico: At present, I'm just using that, but we can allow the caller to specify the provider list.  We'll have to for the CLI work that is currently spec'd.

pkg/capi/kind_cluster_manager.go
michael-cico: It's probably better to let the caller do the logging right?  since this is a util lib

pkg/capi/kind_cluster_manager.go
michael-cico: Same comment as above, for a util lib I thought it was better leave that to the caller.

pkg/capi/cluster_lifecycle.go
michael-cico: Init can probably be consolidated with `Create`... wasn't sure if wanted this as a separate step or not.

pkg/capi/capi_test.go
michael-cico: This is strictly to contain shared test code, the stubs etc.  I can move that stuff back into one of the other test files if that helps.

pfmackin (approved):

desagar (approved):

PR #3938, opened 2022-08-24
VZ-6650 - Initial CAPI CLI command and associated refactoring

desagar:


pkg/capi/cluster_lifecycle.go
```diff
@@ -36,9 +45,68 @@ type ClusterLifeCycleManager interface {
 	Destroy() error
 }
 
-// NewBoostrapCluster Creaets a new cluster manager for a local bootstrap cluster
-func NewBoostrapCluster() ClusterLifeCycleManager {
+// NewDefaultBoostrapCluster Creates a new cluster manager for a local bootstrap cluster with default config
```
michael-cico, line 48: I wonder if this is really needed still?  we don't have to decide now I guess.

michael-cico (dismissed):

michael-cico (approved):

pkg/capi/cluster_lifecycle.go
desagar: yeah maybe we can get rid of it eventually.

pfmackin (approved):

PR #3939, opened 2022-08-25
VZ-6666: Uptake Rancher version 2.6.7

pshirodkar-orcl:
Uptake Rancher version 2.6.7.


platform-operator/thirdparty/charts/rancher/Chart.yaml
```diff
@@ -1,16 +1,16 @@
 apiVersion: v1
-name: rancher
+appVersion: v2.6.7
```
pshirodkar-orcl, line 2: Any of the reordering of the elements has not been introduced by me. The reordering of the elements was introduced by following the instructions https://github.com/verrazzano/verrazzano/blob/master/platform-operator/thirdparty/charts/README.md#rancher

`export RANCHER_CHART_VERSION=v2.6.7
rm -rf rancher
helm repo add rancher-stable https://releases.rancher.com/server-charts/stable
helm repo update
helm fetch rancher-stable/rancher --untar=true --version=${RANCHER_CHART_VERSION}`

pshirodkar-orcl: The `rancher/tests` directory has been deleted due it being excluded when following above instructions.

lake-of-dreams:
why does it show `Empty File` for several files like `platform-operator/thirdparty/charts/rancher/templates/deployment.yaml` in the PR?

pshirodkar-orcl:
> why does it show `Empty File` for several files like `platform-operator/thirdparty/charts/rancher/templates/deployment.yaml` in the PR?

Those files remain unchanged and thus do not show as modified.

mgianatagh (approved):

fredtibbitts:
> > why does it show `Empty File` for several files like `platform-operator/thirdparty/charts/rancher/templates/deployment.yaml` in the PR?
> 
> Those files remain unchanged and thus do not show as modified.

I think this can happen if you change the file permissions. The files should not show up in the PR if there were no changes.

fredtibbitts (approved):
In master several of the chart files have the execute bit set, but they do not in this PR, which is why they show up in the PR but with an empty diff. This seems fine to me.

michael-cico (approved):

PR #3946, opened 2022-08-25
VZ 6728: Metrics uses init()

anders-swanson:
Fixup metricsexporter implementation to use golang init() function


davisgomes (dismissed):

anders-swanson:
@davisgomes could you have another look, I forgot i needed to add the VMO image with the same updates.

davisgomes (approved):

PR #3947, opened 2022-08-25
VZ-6650 CAPI delete function in VZ CLI

desagar:

### DIFF
--- a/pkg/capi/bootstrap_provider.go
+++ b/pkg/capi/bootstrap_provider.go
@@ -5,6 +5,7 @@ package capi
 import (
 	"bytes"
 	"io/ioutil"
+	"os"
 
 	kindcluster "sigs.k8s.io/kind/pkg/cluster"
 	kind "sigs.k8s.io/kind/pkg/cmd"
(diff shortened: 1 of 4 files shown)


michael-cico (dismissed):

pfmackin (dismissed):

pkg/capi/bootstrap_provider.go
```diff
@@ -59,7 +62,23 @@ func (k *kindBootstrapProviderImpl) DestroyCluster(clusterName string) error {
 	if err != nil {
 		return err
 	}
-	return provider.Delete(clusterName, kubeconfig)
+	kubePath, err := saveKubeconfigToFile(kubeconfig)
+	if err != nil {
+		return err
+	}
+	return provider.Delete(clusterName, kubePath)
```
michael-cico, line 69: Do we need to add a `defer` statement to delete the kubeconfig?

We probably need to add a command/subcommand to get and store the kubeconfig so it can be re-used.

michael-cico (approved):

PR #3955, opened 2022-08-25
JIRA VZ-6787 mysql operator cluster install

jmaron99:
An implementation for the installation code only with some minor tweaks to upgrade code.


platform-operator/thirdparty/charts/keycloak/values.yaml
```diff
@@ -381,6 +381,10 @@ postgresql:
   networkPolicy:
     enabled: false
 
+mysql:
```
mgianatagh, line 384: Is this a change we made?  Should it instead be added in platform-operator/helm_config/overrides/keycloak-values.yaml

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -287,6 +287,17 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
```
mgianatagh, line 290: Why does the application operator require access to these resources?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1525,6 +1639,51 @@ func GetVerrazzanoUserFromKeycloak(ctx spi.ComponentContext) (*KeycloakUser, err
...
+	execCmd := []string{"bash", "-c", sqlCmd}
+	cfg, cli, err := k8sutil.ClientConfig()
+	if err != nil {
+		return err
+	}
+	mysqlPod, err := getMySQLPod(ctx)
+	if err != nil {
+		return err
+	}
+	stdOut, stdErr, err := k8sutil.ExecPod(cli, cfg, mysqlPod, "mysql", execCmd)
```
pfmackin, line 1666: Is this is trying to create the database?   If so wouldn't we be creating a MySQL CR to do that (just guessing here).

platform-operator/thirdparty/charts/keycloak/values.yaml
jmaron99: There was no mysql values section available and I needed a corresponding change in the stateful set template.  This was the mosts straightforward way (I felt) to enable alternate host names for the db host for keycloak (the host varies depending on whether the router is deployed or not)

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
jmaron99: Not entirely clear but VZ startup fails w/o it

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
jmaron99: this is to setup a keycloak schema to be populated with tables/data when keycloak starts.  Used to be managed with DB init scripts but that facility is not currently available with mysql operator cluster.  We could ask for it.

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
jmaron99: From the keycloak namespace events, w/o this change:

`65s         Warning   FailedCreate           statefulset/mysql                       create Pod mysql-0 in StatefulSet mysql failed error: admission webhook "verrazzano-application-istio-defaulter.verrazzano.io" denied the request: innodbclusters.mysql.oracle.com "mysql" is forbidden: User "system:serviceaccount:verrazzano-system:verrazzano-application-operator" cannot get resource "innodbclusters" in API group "mysql.oracle.com" in the namespace "keycloak"`

stevshap (approved):

mgianatagh (approved):

PR #3960, opened 2022-08-26
VZ-6906 Install MySQL operator before upgrading other components

pfmackin:
During upgrade, install MySQL operator before upgrading other components, since it is needed for MySQL upgrade.    Do this generically so the same can be done for other components as needed.


jmaron99:
If more than one component will need to be installed pre-upgrade will their ordering be important?  Should a "weight" be specified or should we handle that if the need arises?

stevshap:
@jmaron99 I suppose the same question could be asked in a regular upgrade case right? I'm not sure if a hierarchy exists already

stevshap (approved):

pfmackin:
> If more than one component will need to be installed pre-upgrade will their ordering be important? Should a "weight" be specified or should we handle that if the need arises?

pre-upgrade install is the same logic as normal install. Components will be installed in parallel and if they have dependencies then they can explicitly specify them in the component struct or they can handle it in code during the pre-install.

stevshap (dismissed):

stevshap (approved):

PR #3966, opened 2022-08-27
VZ-6902: Include README.md in Verrazzano release distributions

bhatpmk:
This change includes README.md in Verrazzano release distributions. As part of this change, the existing profiles directory is removed and will be added once the final set of profiles are available.


bhatpmk:
VZ-6863

agv426 (approved):

PR #3967, opened 2022-08-27
Fix MySQL operator issues with private registry test

agv426:
Looks like the MySQL operator is not setting the ImagePullSecretKeyname, not sure if that is required, but trying that similar to how other components are setting it (setting like the MySQL component)


agv426:
We are going to modify the repository creation script to set it is public instead, and update the readme. Separately there is a question to follow up on whether the helm override flags can handle this https://dev.mysql.com/doc/mysql-operator/en/mysql-operator-private-registries-helm-operator.html

agv426:
We discussed in the meeting and will merge this

PR #3969, opened 2022-08-29
VZ-6907 Fix for including Jaeger related parameters in the Multicluster registration sync

anajosep:
VZ-6907 - When the DNS suffix is modified in the Admin cluster, say from `nip.io` to `sslip.io`, the corresponding URLs and certs gets modified in the admin cluster and these need to be propagated to all the managed clusters through the registration secret. Jaeger related parameters in the registration secret are not updated as part of the sync flow. Added these parameters to the sync flow.


application-operator/mcagent/mcagent_clusterca.go
```diff
@@ -84,6 +84,12 @@ func (s *Syncer) syncRegistrationFromAdminCluster() (controllerutil.OperationRes
 			registrationSecret.Data[mcconstants.RegistrationPasswordKey] = adminRegistrationSecret.Data[mcconstants.RegistrationPasswordKey]
 			registrationSecret.Data[mcconstants.KeycloakURLKey] = adminRegistrationSecret.Data[mcconstants.KeycloakURLKey]
 			registrationSecret.Data[mcconstants.ESCaBundleKey] = adminRegistrationSecret.Data[mcconstants.ESCaBundleKey]
+			registrationSecret.Data[mcconstants.JaegerOSURLKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSURLKey]
+			registrationSecret.Data[mcconstants.JaegerOSUsernameKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSUsernameKey]
+			registrationSecret.Data[mcconstants.JaegerOSPasswordKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSPasswordKey]
+			registrationSecret.Data[mcconstants.JaegerOSTLSCAKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSTLSCAKey]
+			registrationSecret.Data[mcconstants.JaegerOSTLSCertKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSTLSCertKey]
+			registrationSecret.Data[mcconstants.JaegerOSTLSKey] = adminRegistrationSecret.Data[mcconstants.JaegerOSTLSKey]
```
ayaranga, line 92: Can you update the existing unit tests as well?

desagar: the test has an assertRegistrationInfoEqual - we should assert that these fields are also equal. 

It would be good to add additional testing as well, like maybe a test case where there is a change in Jaeger ES info.

anajosep: Added unit tests to test that the changes to any data in the registration secret, if a sync call is made, then the values of the secrets are synchronized.

desagar (approved):

PR #3974, opened 2022-08-29
VZ-6865, VZ-6802: Update release pipeline for open-source distribution

bhatpmk:
This change updates the release build pipelines for Github release Verrazzano open-source distribution. As part of this change, a new pipeline is added to perform the malware scan of release bundles.

### DIFF
--- a/release/builds/JenkinsfilePostPRT
+++ b/release/builds/JenkinsfilePostPRT
@@ -119,7 +119,6 @@ pipeline {
                     timeout(time: 90, unit: 'MINUTES') {
                         sh """
                             cd ${WORKSPACE}/release/scripts
-                            unset BUNDLE_TO_SCAN
                             ./scan_release_binaries.sh ${params.RELEASE_BRANCH} ${RELEASE_BINARIES_DIR}
                         """
                    }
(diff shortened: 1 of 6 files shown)


release/builds/JenkinsfilePostPRT
```diff
@@ -117,7 +119,8 @@ pipeline {
                     timeout(time: 90, unit: 'MINUTES') {
                         sh """
                             cd ${WORKSPACE}/release/scripts
-                            ./scan_release_binaries.sh ${params.RELEASE_BRANCH} ${RELEASE_BINARIES_DIR} ${params.RELEASE_VERSION}
+                            unset BUNDLE_TO_SCAN
```
agv426, line 122: Doesn't look like unset is needed there

release/scripts/create_github_release.sh
```diff
@@ -25,12 +25,13 @@ EOM
     exit 0
 }
 
-[ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ] || [ "$1" == "-h" ] && { usage; }
+[ -z "$1" ] || [ -z "$2" ] || [ "$1" == "-h" ] || [ -z "$RELEASE_VERSION" ] && { usage; }
```
agv426, line 28: We should call out that the RELEASE_VERSION environment variable needs to be defined in the usage above since it is no longer supplied as an argument

release/builds/JenkinsfilePostPRT
bhatpmk: Done

release/scripts/create_github_release.sh
bhatpmk: Done

agv426 (approved):

bhatpmk:
VZ-6802

PR #3977, opened 2022-08-29
Fix cleanup code so we continue to cleanup the OCNE cluster if the VZ…

agv426:
The OCNE job was not cleaning up the OCNE cluster if the VZ uninstall that was being performed during the job cleanup fails for any reason.


shashankkumarIITB (approved):
LGTM.

PR #3978, opened 2022-08-29
VZ-6921 Upgrade fails when Jaeger Operator component is enabled

anajosep:
VZ-6921 When Verrazzano is upgraded with Jaeger Operator enabled in the CR, then the upgrade fails as the verrazzano component upgrade was failing. This was due to Jaeger Auth and network policy (added as part of the Verrazzano chart templates) are created in newly created `verrazzano-monitoring` namespace. This issue is fixed by creating the required namespace as part of the pre-upgrade part of the Verrazzano component.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/verrazzano_monitoring_test.go
+++ b/platform-operator/controllers/verrazzano/component/common/verrazzano_monitoring_test.go
@@ -4,6 +4,7 @@
 package common
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -11,8 +12,15 @@ import (
 	vzapi "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
+
+var trueValue = true
+var falseValue = false
 
 // TestMutateVerrazzanoMonitoringNamespace tests the MutateVerrazzanoMonitoringNamespace function.
 func TestMutateVerrazzanoMonitoringNamespace(t *testing.T) {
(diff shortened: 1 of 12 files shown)


platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -381,3 +386,35 @@ func removeNodeExporterService(ctx spi.ComponentContext) {
 		}
 	}
 }
+
+// ensureVerrazzanoMonitoringNamespace ensures that the verrazzano-monitoring namespace is created with the right labels.
+func ensureVerrazzanoMonitoringNamespace(ctx spi.ComponentContext) error {
```
desagar, line 391: looks like pretty much the same code exists in the prometheus_utils.go file. I would suggest moving that up into the common directory (looks like that entire file can be moved), and call it verrazzano_monitoring.go or something like that, instead of repeating it.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
```
fredtibbitts, line 10: Nit, this import should be in the lower section.

fredtibbitts (commented):
Agree with @desagar 's comment, re-use the existing code for the namespace.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
```diff
@@ -66,13 +68,18 @@ func isVerrazzanoReady(ctx spi.ComponentContext) bool {
...
-func verrazzanoPreUpgrade(ctx spi.ComponentContext, namespace string) error {
+func verrazzanoPreUpgrade(ctx spi.ComponentContext) error {
 	if err := exportFromHelmChart(ctx.Client()); err != nil {
 		return err
 	}
 	if err := common.EnsureVMISecret(ctx.Client()); err != nil {
 		return err
 	}
+	// Auth policies and Network policies created in the helm chart requires verrazzano-monitoring namespace
+	ctx.Log().Debugf("Creating namespace %s for the Verrazzano component", constants.VerrazzanoMonitoringNamespace)
```
ayaranga, line 79: Should we add a check to create a monitoring namespace only if prometheus or jaeger operator is enabled?

ayaranga (approved):

platform-operator/controllers/verrazzano/component/common/verrazzano_monitoring_test.go
```diff
@@ -54,13 +55,23 @@ func TestMutateVerrazzanoMonitoringNamespace(t *testing.T) {
...
 }
+
+// TestEnsureMonitoringOperatorNamespace asserts the verrazzano-monitoring namespaces can be created
+func TestEnsureMonitoringOperatorNamespace(t *testing.T) {
+	// GIVEN a Verrazzano CR with Jaeger Component enabled,
+	// WHEN we call the EnsureVerrazzanoMonitoringNamespace function,
+	// THEN no error is returned.
+	ctx := spi.NewFakeContext(fake.NewClientBuilder().WithScheme(testScheme).Build(), &vzapi.Verrazzano{}, nil, false)
+	err := EnsureVerrazzanoMonitoringNamespace(ctx)
+	assert.NoError(t, err)
```
desagar, line 76: shouldn't this test also check that the NS exists after the call to EnsureVerrazzanoMonitoringNamespace?

desagar (approved):

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
anajosep: Moved it to the common package an reusing it in multiple places.

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
anajosep: fixed

platform-operator/controllers/verrazzano/component/common/verrazzano_monitoring_test.go
anajosep: Fixed tests and added more scenarios

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano.go
anajosep: fixed

desagar (approved):

PR #3982, opened 2022-08-30
VZ-6649.  Change default bootstrap cluster type to CNE and add default image

michael-cico:


fredtibbitts (approved):

fredtibbitts (approved):

tools/vz/pkg/constants/constants.go
```diff
@@ -175,7 +175,7 @@ const (
 	ClusterNameFlagHelp    = "The name of the cluster - defaults to " + ClusterNameFlagDefault
 
 	ClusterTypeFlagName    = "type"
-	ClusterTypeFlagDefault = "kind"
+	ClusterTypeFlagDefault = "cne"
```
kminder, line 178: This might need to be 'ocne' but this is something that Simon might need to answer.  How is this used?

michael-cico: just to indicate kind/vs CNE.  We can probably remove it, after this change `cne` is the default anyway.  Largely we needed to use vanilla KIND to do testing initially.

kminder: k

desagar (approved):

PR #3987, opened 2022-08-30
VZ CLI get-kubeconfig merges into existing kubeconfig

desagar:

### DIFF
--- a/tools/vz/cmd/cluster/cluster_get_kubeconfig.go
+++ b/tools/vz/cmd/cluster/cluster_get_kubeconfig.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/verrazzano/verrazzano/pkg/capi"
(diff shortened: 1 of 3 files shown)


michael-cico (dismissed):

tools/vz/cmd/cluster/cluster_get_kubeconfig.go
```diff
@@ -76,18 +78,44 @@ func runCmdClusterGetKubeconfig(helper helpers.VZHelper, cmd *cobra.Command, arg
...
+	if err = os.WriteFile(filePath, []byte(kubeconfigContents), 0700); err != nil {
+		fmt.Fprintf(helper.GetOutputStream(), "Failed to write kubeconfig to file %s - %v\n", filePath, err)
+		return err
+	}
+	fmt.Fprintf(helper.GetOutputStream(), "%s %s\n", message, filePath)
+	return nil
+}
+
+func generateTempKubeconfigFilename() string {
+	return fmt.Sprintf("tempkubeconfig_%d", time.Now().UnixNano())
```
fredtibbitts, line 118: The `io/ioutil` package has functions for creating temporary files, might be slightly better than doing this, but I don't know what our `CreateTempFile` function is doing, maybe it adds some value over what `ioutil` provides?

fredtibbitts (approved):
Just the one minor comment, otherwise LGTM.

tools/vz/cmd/cluster/cluster_get_kubeconfig.go
desagar: you're right - here I only need to create an empty file so ioutil.TempFile is sufficient. pushed the change.

michael-cico (approved):

michael-cico (approved):

PR #3997, opened 2022-09-01
VZ-6925: Verify Rancher/Keycloak integration in dynamic config tests

lake-of-dreams:
- Refactored the code that checked access to Rancher and Keycloak and also verified the `AuthConfig` CR
- Verify the access and AuthConfig on dns updates in dynamic config test


kminder (approved):
Approved.

PR #4018, opened 2022-09-02
VZ-6064: Integrating minor release upgrade tests with periodic pipeline 

Abhijithang:
This PR will integrate minor release upgrade tests with the periodic jobs.


ci/JenkinsfilePeriodicTests
```diff
@@ -43,6 +43,16 @@ pipeline {
     }
 
     parameters {
+        string (name: 'N_JOBS_FOR_EACH_BATCH',
+                defaultValue: '3',
+                description: 'This is to define total number of jobs executed in parallel per each batch for Upgrade path minor release pipeline',
+                trim: true)
+        choice (description: 'Predefined config permutations for Verrazzano installation. Prod profile is the default profile for NONE', name: 'VZ_INSTALL_CONFIG',
+                choices: ["NONE", "dev-kind-persistence"])
+        string (name: 'EXCLUDE_RELEASES',
```
pshirodkar-orcl, line 52: this exclude release list is different than https://github.com/verrazzano/verrazzano/blob/670fcdcc01f5eb6669812a848c2dc957d887eb9d/ci/upgrade-paths/JenkinsfileTestTriggerMinorRelease#L44. Any ideas why?

ci/JenkinsfilePeriodicTests
```diff
@@ -43,6 +43,16 @@ pipeline {
     }
 
     parameters {
+        string (name: 'N_JOBS_FOR_EACH_BATCH',
```
agv426, line 46: Why are these controls parameters here rather than explicit values for the test case being run from the periodics?

Periodic runs are generally triggered by CRON trigger, not a human, and the jobs that are triggered are generally given explicit values for the specific run.

Folks who need to test a specific job in a branch generally are going to run that specific job directly and specify parameters for that specific job.

A periodic test run uses a ton of resources, these are not runs that developers should be running from branches in general except when making broad/cross-cutting changes. Even then, I think they would be running with the same settings the periodics use normally and wouldn't be tweaking parameter values for the upgrade tests.

ci/JenkinsfilePeriodicTests
Abhijithang: Adding the upgrade tests incrementally, for now, 1.3.x is clean, and I see one intermittent issue with 1.2.x and 1.1.x. So working on a fix for earlier releases, and before that, I don't want to hold back 1.3.x.

ci/JenkinsfilePeriodicTests
Abhijithang: Sure, it's not a necessity and the default value will be used. I removed the parameter.

ci/JenkinsfilePeriodicTests
pshirodkar-orcl: Excluding 1.2.x releases does not seem the right approach to me since those are still supported.

Abhijithang: Or I could remove this Param from periodic job and update the minor release path job to keep it in sync. What do you think?

agv426 (approved):

kminder (approved):
Approved.

PR #4021, opened 2022-09-06
VZ-6961: Use Verrazzano distributions from private registry tests

bhatpmk:
This PR updates the private-registry job to use the Verrazzano distributions, to validate both lite and full bundles from the periodic job.


ci/JenkinsfilePeriodicTests
```diff
@@ -9,6 +9,9 @@ def agentLabel = env.JOB_NAME.contains('master') ? "phxsmall" : "VM.Standard2.2"
 def TESTS_FAILED = false
 def tarfilePrefix="verrazzano_periodic"
```
bhatpmk, line 10: The variable tarfilePrefix is supplied to update_periodic_on_success.sh, which I have not touched. We need to clean that, when we remove building old bundle

adiforluls (dismissed):

agv426 (dismissed):

adiforluls (approved):
LGTM

PR #4024, opened 2022-09-06
VZ-6948 compare generation for last generation

mbenmclean:
This change adds a lastReconcileGeneration value to short-circuit the VPO reconcile to prevent queued reconcile calls from continuing after we return with no requeue. This should alleviate the issue we are seeing where an install starts after upgrade.


platform-operator/controllers/verrazzano/controller.go
```diff
@@ -71,6 +71,7 @@ var systemNamespaceLabels = map[string]string{
 
 // Set to true during unit testing
 var unitTesting bool
+var lastReconcileGen int64
```
michael-cico, line 74: So we're just tracking this in-memory?  Should this be a first-class citizen in the status field?

michael-cico: i.e. what happens if the VPO restarts?

platform-operator/controllers/verrazzano/upgrade.go
```diff
@@ -152,8 +149,7 @@ func (r *Reconciler) reconcileUpgrade(log vzlog.VerrazzanoLogger, cr *installv1a
 					componentStatus.LastReconciledGeneration = cr.Generation
 				}
 			}
-			// Update the status with the new version and component generations
-			cr.Status.Version = targetVersion
+			// Update the component generations
```
michael-cico, line 152: I guess I'm not sure what's going on here... don't we need to set the version here during upgrade?

michael-cico: OK, is this happening as part of `r.updateVzStatusAndStateAndVersion`?

platform-operator/controllers/verrazzano/controller.go
pfmackin: This is just tracking in memory to handle the case where controller runtime has queued up CRs which causes VZ reconcile to be called  after it finished reconciling.   On restart, that queue is flushed and normal flow happens.

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -142,6 +148,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	log.Oncef("Finished reconciling Verrazzano resource %v", req.NamespacedName)
 	metricsexporter.AnalyzeVerrazzanoResourceMetrics(log, *vz)
 
+	if vz.Status.Conditions[len(vz.Status.Conditions)-1].Type == installv1alpha1.CondUninstallComplete {
```
pfmackin, line 151: You don't need this test and in fact it might cause issues on upgrade.  If you reach this point we are telling the controller runtime that we are done and don't call reconcile again. You should remove lines 151 - 156

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -756,13 +771,15 @@ func conditionToVzState(currentCondition installv1alpha1.ConditionType) installv
 
 // setInstallStartedCondition
 func (r *Reconciler) setInstallingState(log vzlog.VerrazzanoLogger, vz *installv1alpha1.Verrazzano) error {
-	// Set the version in the status.  This will be updated when the starting install condition is updated.
-	bomSemVer, err := installv1alpha1.GetCurrentBomVersion()
-	if err != nil {
-		return err
-	}
+	// Set the version in the status if this is the initial install.  This will be updated when the starting install condition is updated.
```
pfmackin, line 774: Why is this change here?  Seems unrelated to the generation change.

platform-operator/controllers/verrazzano/controller.go
mbenmclean: If the VPO restarts then it won't short circuit, but if everything is working right nothing will happen and it will get reset. This is not necessary for the VPO to work correctly but it's a performance improvement and should alleviate the issue we're seeing where we go straight into an install after an upgrade.

platform-operator/controllers/verrazzano/upgrade.go
mbenmclean: Yeah I think it's more appropriate there. We don't want to get into a situation where the version is updated but we haven't updated the status and state yet.

michael-cico: yeah, looks like a good consolidation

platform-operator/controllers/verrazzano/controller.go
mbenmclean: Paul I'm confused by your comment. We need to update the lastReconcileGen after we're done with reconcile that's how this works. Can you clarify what you mean?

platform-operator/controllers/verrazzano/controller.go
michael-cico: OK, got it, thanks.

platform-operator/controllers/verrazzano/controller.go
mbenmclean: Yeah I was debugging and thought this might be related to an error I was seeing. I can get rid of this change.

mbenmclean: I do think it might be going to leave this in some capacity. I think we want the spec.version field to be the source of truth for version, not the bom. If they run without webhooks on they could conceivably get to a position where the spec.version field is different than what's in the bom. Maybe add an else here where if the spec.version is not empty we set the status.version field to that. There's are all edge cases but that seems like the correct handling of that.

platform-operator/controllers/verrazzano/controller.go
mbenmclean: I've manually testing with dev. How do we force the reconcile. Is it by updating the VZ CR to increment the generation?

PR #4027, opened 2022-09-06
VZ-6956.  Remove obsolete tests and refactor some common code

michael-cico:
Remove the following obsolete test code and refactor some of the common HA code into the common `e2e/pkg` package.

- Removed the `rollingupdate` and `scheduling` test suites from the `ha` suite
- Moved the common code under `tests/e2e/pkg/ha`
- Renamed the `rollingupdate` test dump dir to align with the suite name


mgianatagh (approved):

tests/e2e/pkg/ha/nodes.go
```diff
@@ -66,19 +66,6 @@ func EventuallyEvictNode(cs *kubernetes.Clientset, name string, log *zap.Sugared
 
 		for i := range pods.Items {
 			pod := &pods.Items[i]
-			// Temporarily ignore pods associated with PVs to avoid pods going into pending state
```
fredtibbitts, line 69: EventuallyEvictNode is no longer used, I think the whole function can be removed.

PR #4028, opened 2022-09-06
VZ-6985 Add ocne and ha v1beta1 conversion tests

pfmackin:
Add unit tests to convert the Verrazzano v1beta1 yaml for HA, OCNE, and OCNE HA to v1beta2. This YAML will be used in the OCNE validation manual tests.


platform-operator/apis/verrazzano/v1alpha1/testdata/fromocneha/v1beta1.yaml
```diff
@@ -0,0 +1,116 @@
...
+      type: LoadBalancer
+    istio:
+      overrides:
+        - values:
+            apiVersion: install.istio.io/v1alpha1
+            kind: IstioOperator
+            spec:
+              components:
+                egressGateways:
+                  - enabled: true
```
anders-swanson, line 71: one limitation we have noticed with the istio config is that because ingress/egress gateway settings are Arrays, our YAML merge logic does a replacement merge when combining multiple IsitoOperator objects as it creates the Effective CR. This means Affinity on egress/ingress gateway may be lost if it was defined in the base.yaml, and the user provided CR also defines ingress/egress gateway settings.
This issue always existed, we are just noticing it more now with the newer api

anders-swanson (approved):

PR #4036, opened 2022-09-06
VZ-6930: Fix the pulling of the mysql-operator image with private registry

bobdonat:
This pull request fixes the pulling of the mysql-operator image when a private registry is used.  The change correctly sets the  mysql-operator helm values so that the mysql-operator image can be pulled with an image pull secret.


adiforluls (approved):
LGTM

PR #4037, opened 2022-09-06
VZ-6898: Convert ha.yaml to v1beta1

mgianatagh:
Convert the ha.yaml file to use the v1beta1 API

### DIFF
--- a/examples/ha/ha.yaml
+++ b/examples/ha/ha.yaml
@@ -28,30 +28,10 @@ spec:
       overrides:
         - values:
             controller:
-              affinity:
-                podAntiAffinity:
-                  preferredDuringSchedulingIgnoredDuringExecution:
-                    - podAffinityTerm:
-                        labelSelector:
-                          matchLabels:
-                            app.kubernetes.io/component: controller
-                            app.kubernetes.io/name: ingress-nginx
-                        topologyKey: kubernetes.io/hostname
-                      weight: 100
               autoscaling:
                 enabled: true
                 minReplicas: 2
             defaultBackend:
-              affinity:
-                podAntiAffinity:
-                  preferredDuringSchedulingIgnoredDuringExecution:
-                    - podAffinityTerm:
-                        labelSelector:
-                          matchLabels:
-                            app.kubernetes.io/component: default-backend
-                            app.kubernetes.io/name: ingress-nginx
-                        topologyKey: kubernetes.io/hostname
-                      weight: 100
               replicaCount: 2
     istio:
       overrides:
--- a/examples/ha/ha.yaml
+++ b/examples/ha/ha.yaml
@@ -9,8 +9,9 @@ spec:
   profile: prod
   components:
     authProxy:
-      kubernetes:
-        replicas: 2
+      overrides:
+        - values:
+            replicas: 2
     certManager:
       overrides:
         - values:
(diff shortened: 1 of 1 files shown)


jmaron99:
I have code such as the following in mysql.go:
`	overrides, err := common.GetInstallOverridesYAML(ctx, GetOverrides(ctx.EffectiveCR()).([]vzapi.Overrides))
`
What I'm struggling to understand is at one point will ctx.EffectiveCR() return a v1beta1 CR that will force a change to code like this?  Will the introduction of this file prompt that transition?

stevshap (approved):

examples/ha/ha.yaml
```diff
@@ -1,16 +1,17 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: Can you update both files with the conversion unit tests here https://github.com/verrazzano/verrazzano/tree/master/platform-operator/apis/verrazzano/v1alpha1/testdata/fromha

examples/ha/ha.yaml
```diff
@@ -23,7 +24,7 @@ spec:
       overrides:
         - values:
             replicas: 2
-    ingress:
+    ingressNGINX:
```
pfmackin, line 27: I think this is missing

    ingressNGINX:
      overrides:
        - values:
            controller:
              autoscaling:
                enabled: true
                minReplicas: 2
            defaultBackend:
              replicaCount: 2

examples/ha/ha.yaml
```diff
@@ -42,17 +43,47 @@ spec:
                 pilot:
                   k8s:
                     replicaCount: 2
-      ingress:
-        kubernetes:
-          replicas: 2
-      egress:
-        kubernetes:
-          replicas: 2
+                ingressGateways:
```
pfmackin, line 46: I tihnk we are using overrides for this, it doesn't match the beta1 generated YAML in the unit test

examples/ha/ha.yaml
mgianatagh: It's there

examples/ha/ha.yaml
mgianatagh: this is using overrides

examples/ha/ha.yaml
mgianatagh: The conversion unit test already has the v1alpha1 version of ha.yaml.  We can't use the test generated version as is because of issues of merging lists with values.  The converted output also includes things like empty strings and structs that are not needed in the file we publish to end users.  I have confirmed the differences in ha.yaml I have made agree with the changes in the testdata version of ha.yaml

mgianatagh:
> I have code such as the following in mysql.go:
> ` overrides, err := common.GetInstallOverridesYAML(ctx, GetOverrides(ctx.EffectiveCR()).([]vzapi.Overrides)) `
> What I'm struggling to understand is at one point will ctx.EffectiveCR() return a v1beta1 CR that will force a change to code like this? Will the introduction of this file prompt that transition?

That shouldn't matter because if you get a v1alpha1 representation of the CR, it is expected to work.  An internal conversion takes place from v1beta1 to v1alpha1.

But I also think there are more changes in master regarding v1beta1 support that we need to uptake in the shared branch

PR #4038, opened 2022-09-07
VZ-6962: Retire old release bundle creation and related cleanups

adiforluls:
Updates to image helper script and getting rid of the old bundle.


bhatpmk:
There is a call to ci/scripts/generate_product_zip.sh from the top level Jenkinsfile, which needs to be removed along with the Groovy definitions

adiforluls:
> There is a call to ci/scripts/generate_product_zip.sh from the top level Jenkinsfile, which needs to be removed along with the Groovy definitions

Will address in a follow-up commit

tools/scripts/vz-registry-image-helper.sh
```diff
@@ -44,7 +44,7 @@ Options:
  -b <path>              Bill of materials (BOM) of Verrazzano components; if not specified, defaults to ./verrazzano-bom.json
  -l <archive-dir>       Use the specified imagesDir to load local Docker image tarballs from instead of pulling from
  -i <component>         Include the specified component in the operation (can be repeated for multiple components)
- -f <tarfile>           The name of the tar file to create for downloading and saving Verrazzano images
+ -f <tarfile>           The name of the directory that will be used to save Docker image tarballs locally
```
adiforluls, line 47: We don't generate a compressed tar of images anymore, instead now the new layout consists an image dir containing tarfiles of the images. So changed the pull and save func to just save the tarfiles to the specified dir.

adiforluls:
> There is a call to ci/scripts/generate_product_zip.sh from the top level Jenkinsfile, which needs to be removed along with the Groovy definitions

Fixed

michael-cico (dismissed):

michael-cico (dismissed):

bhatpmk (approved):

PR #4040, opened 2022-09-07
Disable full bundle validation from periodic temporarily

bhatpmk:
Disable the full build validation temporarily, to overcome the "Method code too large" issue from periodic pipeline


adiforluls (approved):
This is a temporary measure that should get the pipeline unblocked. Approved.

PR #4063, opened 2022-09-08
VZ-6959: Update tar command to avoid leading "./" while building the distribution artifacts

bhatpmk:
Updates the tar command to create the distribution artifacts in Verrazzano Lite bundle, to avoid leading "./"


kminder (approved):
Approved.

bhatpmk:
VZ-6959

PR #4064, opened 2022-09-08
aamitra/vz 6798 off jmaron vz6787

theabmitra:


tests/e2e/backup/mysql/mysql_backup_test.go
```diff
@@ -323,6 +341,32 @@ var _ = t.Describe("MySQL Backup and Restore,", Label("f:platform-verrazzano.mys
 
 	})
 
+	t.Context("Flap mysql-operator", func() {
```
kminder, line 344: What does "Flap" mean?

platform-operator/scripts/hooks/mysql-hook.sh
```diff
@@ -44,6 +44,14 @@ function restore() {
   fi
   echo "MySQL is up and ready to receive connections"
 
+
+  mysql -u root -p${MYSQL_ROOT_PASSWORD} -e "SET GLOBAL read_only=0;"
```
kminder, line 48: What does this do and why is it required?

tests/e2e/backup/mysql/mysql_backup_test.go
theabmitra: Flap here implies scaling the operator down followed by up.

platform-operator/scripts/hooks/mysql-hook.sh
ayaranga: In my setup, I see that the DB readonly flag is set to OFF(0) by default. I am using `show variables;` to check the value. Can you verify once again in your setup?

ayaranga: Yes, you are right. I tried restarting (kubectl delete pod mysql-0 -n keycloak) the mysql pod and see that the `read_only` flag gets set to ON initially for sometime and then it gets set to OFF automatically. Looks like the flag is set to ON until some DB operations are completed. As it is set to ON intentionally, in my opinion, we should not be changing this value. It is better to wait until the flag gets set to OFF before running the restore script. Maybe we can check with the MySQL team for a better solution.

PR #4065, opened 2022-09-08
Pmackin/vz 6996 uninstall

pfmackin:
MySQL pods have finalizers that the MySQL operator has to remove after it has done cleanup.  If the MySQL operator is uninstalled before it has a chance to cleanup the MySQL pods, then the finalizers never get removed and the MySQL pods don't get deleted.  This change will cause MySQL operator PreUninstall to wait until MySQL has been uninstalled.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/mysqloperator/mysqloperator_component.go
+++ b/platform-operator/controllers/verrazzano/component/mysqloperator/mysqloperator_component.go
@@ -6,6 +6,7 @@ package mysqloperator
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/mysql"
 	"path/filepath"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


mgianatagh (approved):

platform-operator/controllers/verrazzano/component/mysqloperator/mysqloperator_component.go
```diff
@@ -119,6 +120,38 @@ func (c mysqlOperatorComponent) PreUpgrade(compContext spi.ComponentContext) err
...
+	spiCtx, err := spi.NewContext(compContext.Log(), compContext.Client(), compContext.ActualCR(), compContext.ActualCRV1Beta1(), compContext.IsDryRun())
+	if err != nil {
+		spiCtx.Log().Errorf("Failed to create component context: %v", err)
+		return err
+	}
+	mySQLContext := spiCtx.Init(mysql.ComponentName).Operation(vpocons.UninstallOperation)
+
+	// Check if MySQL is installed
+	mySQLComp := mysql.NewComponent()
+	installed, err := mySQLComp.IsInstalled(mySQLContext)
```
michael-cico, line 141: Do we need a new context for it?

pfmackin: I thought we did.  Each component has a context.

michael-cico: I think since we're just calling `IsInstalled` it may not be necessary to create a whole new context. the configuration won't change.

michael-cico: The context is really an invocation context, we just tie it to a component during call invocations for logging/tracking purposes mainly.  You could argue that the `mysqloperator` context passed in is more relevant than a `mysql` context for the purposes of this call to `IsInstalled`.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/mysqloperator/mysqloperator_component.go
pfmackin: All of the code that i have seen has a context per component.   If we don't need one that we should fix it is general, but make sure it is immutable

PR #4069, opened 2022-09-09
VZ-6928: Profile generator tool

adiforluls:
PR for a utility tool to generate standard Verrazzano profiles.
VZ-6928


ci/scripts/generate_vz_distribution.sh
```diff
@@ -233,11 +233,10 @@ generateVZFullDistribution() {
   tar xzf ${VZ_DISTRIBUTION_COMMON}/${VZ_CLI_DARWIN_AMD64_TARGZ} -C ${distDir}/bin/darwin-amd64
   tar xzf ${VZ_DISTRIBUTION_COMMON}/${VZ_CLI_DARWIN_ARM64_TARGZ} -C ${distDir}/bin/darwin-arm64
 
-  captureBundleContents ${rootDir} ${generatedDir} ${FULL_BUNDLE_CONTENTS}
-
   # Create and upload the final distribution zip file and upload
   echo "Create ${generatedDir}/${VZ_FULL_RELEASE_BUNDLE} and upload ..."
   cp ${VZ_REPO_ROOT}/release/docs/README_FULL.md ${distDir}/README.md
+  captureBundleContents ${rootDir} ${generatedDir} ${FULL_BUNDLE_CONTENTS}
```
adiforluls, line 239: Ignore changes made to this script

pkg/profiles/merge.go
```diff
@@ -0,0 +1,436 @@
...
+
+	actualVelero := actual.Spec.Components.Velero
+	profileVelero := profile.Spec.Components.Velero
+	if actualVelero != nil && profileVelero != nil {
+		actualVelero.ValueOverrides = mergeOverridesV1beta1(actualVelero.ValueOverrides, profileVelero.ValueOverrides)
+	}
+}
+
+////mergeOverrides merges the various profiles overrides of v1beta1.Verrazzano into the actual overrides
+func mergeOverrides(actual, profile []v1alpha1.Overrides) []v1alpha1.Overrides {
```
lake-of-dreams, line 429: Should these one line functions be just omitted and instead call `append` directly instead of calling these functions?

pkg/profiles/merge.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	profileStrings = append(profileStrings, string(data))
+	return profileStrings, nil
+}
+
+//AppendComponentOverrides copies the profile overrides of v1alpha1.Verrazzano over to the actual overrides. Any component that has overrides should be included here.
+// Because overrides lacks a proper merge key, a strategic merge will replace the array instead of merging it. This function stops that replacement from occurring.
+// The profile CR overrides must be appended to the actual CR overrides to preserve the precedence order in the way HelmComponent consumes them.
+func AppendComponentOverrides(actual, profile *v1alpha1.Verrazzano) {
+	actualKubeStateMetrics := actual.Spec.Components.KubeStateMetrics
```
lake-of-dreams, line 156: This seems a bit off to me that we have to do this kind of a thing by actually refer to each named component and then call an append function. Can't we have some sort of an `Interface` defined for the Component and the the `Components` is actually an array or map of those (as the name suggests) and just loop ..

pkg/profiles/merge.go
adiforluls: Well, anything in this file has just been moved from `platform-operator/controllers/transform/merge.go` without any additional changes, I think this merge func was introduced during v1beta1 work, so I wouldn't wanna touch this.

pkg/profiles/merge.go
adiforluls: This falls outside the scope of this PR (also this is moving existing piece of code around). I agree though that this seems bad. The spi package provides a level of polymorphism with component context, but that's controller specific I think and might not fit our cause.
But again, if this should be handled then there should be a separate task.

michael-cico: I think when the Overrides merging was fixed it was done this way because the components array can't be treated polymorphically (if that's a word).  There's no base `Component` interface in the CR as Bhat points out.

We could probably achieve it through reflection though, but at the time it seemed like overkill.  we can take a look at it in the future.  Maybe when we rev the API we can abstract the components behind the interface in the CRD.

michael-cico (approved):

pkg/profiles/merge.go
michael-cico: Actually, if we introduce an interface it shouldn't affect the API usage, right?  Might be something we can do transparently. (not for this PR of course)

kminder (approved):
Approved.

platform-operator/controllers/verrazzano/transform/merge.go
```diff
@@ -4,163 +4,20 @@
 package transform
 
 import (
-	"os"
```
pfmackin, line 7: Why isn't all the merge code moved to pkg?

adiforluls: Because of this dep in other two funcs `platform-operator/internal/config`, it's used to get root dir in image I think. Since it's internal to pl-operator would've had to move this to pkg as well, but the package itself is for operator config, so decided to only move the methods that were generic.

pkg/profiles/merge.go
adiforluls: We have a good chunk of components that share some common characteristics. Therefore a Component interface in the CR sounds like a good idea. Don't think that should affect our current usage of the API as well.

platform-operator/controllers/verrazzano/transform/merge_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
```
anders-swanson, line 9: Why was test file not moved to accompany the merge.go?

pkg/profiles/merge.go
anders-swanson: I am responsible for this mess, but it is also kind of annoying any way you cut it. Consider:
- we would need to introduce some way to look up and merge a component, since we are after all merging two components from two separate CRs here.
- Each component would need to implement the component interface (With something very similar to what we have here)
- some special care would need to be taken with keycloak component (it has a nested component)
- not all components have overrides, but they can still implement the interface. Note also that we already have a Component interface but it means something different (in the context of a component controller)

Now consider the root of the problem, which requires an API change:
The Overrides object has no merge key. This means any `strategicpatchmerge` replaces the Overrides list, which is _not_ what we want. Replacement merges clobber the profile values when merging user values. 
The Overrides object as it exists has no sensible merge key. To fix this, we would need to add a key (and have logic to backfill it if the key doesn't exist). The key would also have to be `omitempty`, since it is a new field to an existing struct.
This pattern is weird too, since you end up mutating the spec of the object to possibly add merge keys for Overrides.

So, rather long route, but that is why it exists as it currently does. I agree it looks terrible and is error-prone.

pkg/profiles/merge.go
lake-of-dreams: @adiforluls still this looks odd to me to have a function doing nothing extra and just calling another function, should have been refactored.

pkg/profiles/merge.go
lake-of-dreams: @anders-swanson do we have a jira to fix this at design level itself?

PR #4079, opened 2022-09-09
VZ 6969: Only clean up VAO resources if CRD present

anders-swanson:
Prevents uninstall hanging when VAO uninstall is fired, but VAO CRDs are not present on the cluster.

### DIFF
--- a/platform-operator/controllers/verrazzano/uninstall.go
+++ b/platform-operator/controllers/verrazzano/uninstall.go
@@ -5,7 +5,6 @@ package verrazzano
 
 import (
 	"context"
-
 	vzappclusters "github.com/verrazzano/verrazzano/application-operator/apis/clusters/v1alpha1"
 	"github.com/verrazzano/verrazzano/pkg/constants"
 	"github.com/verrazzano/verrazzano/pkg/k8s/resource"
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -189,7 +189,7 @@ func (r *Reconciler) deleteMCResources(ctx spi.ComponentContext) error {
 	}
 
 	projects := vzappclusters.VerrazzanoProjectList{}
-	if err := r.List(context.TODO(), &projects, &client.ListOptions{Namespace: vzconst.VerrazzanoMultiClusterNamespace}); err != nil {
+	if err := r.List(context.TODO(), &projects, &client.ListOptions{Namespace: vzconst.VerrazzanoMultiClusterNamespace}); err != nil && !meta.IsNoMatchError(err) {
```
pfmackin, line 192: Does this mean that MC resources might get left around?

anders-swanson: In this case the CRD itself has already been deleted (kind/version not found). If a CRD is deleted, the API server deletes all the CRs, and puts a finalizer on the CRD until the CRs are cleaned up.

anders-swanson: So there should not be any VAO CRs hanging around.

pfmackin: We are not supposed to be deleting CRDs.  This case can happen if VPO install gets deleted before VOA is install, but other than that it should never happen.

anders-swanson: Well, it'll either be deleted or not created yet. Either way, there is no VMC resources to delete.

pfmackin (approved):

PR #4093, opened 2022-09-12
Adding support to include readme.html in the release bundle

saidu941:


kminder (approved):
Approved.

michael-cico (dismissed):

kminder (approved):
Approved.

PR #4098, opened 2022-09-12
Ksnelson/vz 6870 - certmanager component validator

knelson7:


platform-operator/controllers/verrazzano/component/certmanager/certmanager_component_test.go
```diff
@@ -851,6 +852,7 @@ func validationTests(t *testing.T, isUpdate bool) {
 			c := NewComponent()
 			getClientFunc = getTestClient(tt)
 			runValidationTest(t, tt, isUpdate, c)
+			//runValidatonBetaTest(t, tt, isUpdate, c)
```
anders-swanson, line 855: why commented?

knelson7: Oops, old edit. Removing now

anders-swanson (approved):
LGTM

PR #4101, opened 2022-09-13
VZ-6928: Add profiles to release docs

adiforluls:


MelwynPaul (approved):
From doc perspective - looks good.

PR #4103, opened 2022-09-13
VZ-7033: Fix intermittent install issue of MySQL statefulset not going ready

mgianatagh:
Fix intermittent install issue of MySQL statefulset not going ready


jmaron99 (approved):
these changes are brilliant ;)

PR #4107, opened 2022-09-13
VZ-6762 Upgrade Istio to 1.14.3 

guoyongzhang:
upgrade Istio in VZ from 1.13 to 1.14.3

### DIFF
--- a/platform-operator/helm_config/overrides/istio-cr.yaml
+++ b/platform-operator/helm_config/overrides/istio-cr.yaml
@@ -15,11 +15,7 @@ spec:
   values:
     global:
       hub: ghcr.io/verrazzano
-      tag: 1.13.2
-      arch:
-        amd64: 2
-        s390x: 2
-        ppc64le: 2
+      tag: 1.14.3
       multiCluster:
         enabled: false
       istioNamespace: istio-system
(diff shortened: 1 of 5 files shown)


platform-operator/helm_config/overrides/istio-cr.yaml
```diff
@@ -15,11 +15,7 @@ spec:
   values:
     global:
       hub: ghcr.io/verrazzano
-      tag: 1.13.2
```
desagar, line 18: I remember you said that these values (under "arch") went into the affinity section of the ingress/egress gateways. Don't we need to populate these values there?

bobdonat (approved):
LGTM

platform-operator/helm_config/overrides/istio-cr.yaml
guoyongzhang: Istio 1.14.3 has changed how it translates "arch" in the deployment affinity.  As a result, the current arch will always result in "node(s) didn't match Pod's node affinity/selector".  Removing "arch" will remove the arch related affinity, which is consistent what we do with all other deployment (i.e., no arch affinity).

desagar (approved):

PR #4114, opened 2022-09-13
Updating readme to include html in the distribution layout

saidu941:
This is a small change from followup PR: https://github.com/verrazzano/verrazzano/pull/4093


bhatpmk:
Update README_FULL.html as well

bhatpmk (approved):

PR #4122, opened 2022-09-14
VZ-6977 Fix for Jaeger tracing properties not getting populated correctly in Istio CR

anajosep:
VZ-6977 When istio component is converted from `v1alpha1` to `v1beta1` some of the configuration parameters are getting overridden and the Jaeger Tracing related configurations are reset. The fix is done to the istio component and the converter logic to process the overrides in the correct order so that the overridden values are preserved and don't get accidentally overridden.


platform-operator/controllers/verrazzano/component/istio/istio_cr.go
```diff
@@ -94,154 +28,31 @@ type ReplicaData struct {
...
-// value replicas and create Istio gateway yaml
-func configureGateways(istioComponent *vzapi.IstioComponent, externalIP string) (string, error) {
-	var data = ReplicaData{}
-
-	data.IngressReplicaCount = istioComponent.Ingress.Kubernetes.Replicas
-	data.EgressReplicaCount = istioComponent.Egress.Kubernetes.Replicas
-
-	if istioComponent.Ingress.Kubernetes.Affinity != nil {
-		yml, err := yaml.Marshal(istioComponent.Ingress.Kubernetes.Affinity)
+	expandedYamls = append(expandedYamls, jaegerTracingYaml)
```
anders-swanson, line 39: I think the value overrides should get appended with the Helm operation. Components implement GetOverrides which returns their overrides for helm

anders-swanson:
some conflicts came up

ayaranga (approved):

tests/e2e/config/scripts/multicluster_edit_vz.sh
```diff
@@ -20,19 +20,15 @@ if [ "${CLUSTER_COUNT}" -gt 1 ] && [ "$CRD_API_VERSION" == "v1alpha1" ]; then
     yq -i eval '.spec.components.istio.istioInstallArgs[0].value = "true"' "${VZ_CR_FILE}"
     yq -i eval '.spec.components.istio.istioInstallArgs[1].name = "meshConfig.defaultConfig.tracing.sampling"' "${VZ_CR_FILE}"
     yq -i eval '.spec.components.istio.istioInstallArgs[1].value = "90.0"' "${VZ_CR_FILE}"
-    yq -i eval '.spec.components.istio.istioInstallArgs[2].name = "meshConfig.defaultConfig.tracing.tlsSettings.mode"' "${VZ_CR_FILE}"
```
mbenmclean, line 23: This should use overrides, not installArgs. We document overrides to configure this here
https://build.verrazzano.io/docs-stage/docs/monitoring/tracing/jaeger-tracing/#jaeger-tracing-in-a-multicluster-verrazzano-environment
We should test the configuration that we have in our documentation

tests/e2e/config/scripts/multicluster_edit_vz.sh
```diff
@@ -20,19 +20,15 @@ if [ "${CLUSTER_COUNT}" -gt 1 ] && [ "$CRD_API_VERSION" == "v1alpha1" ]; then
...
-    yq -i eval '.spec.components.istio.istioInstallArgs[2].value = "ISTIO_MUTUAL"' "${VZ_CR_FILE}"
-    yq -i eval '.spec.components.istio.istioInstallArgs[3].name = "meshConfig.defaultConfig.tracing.zipkin.address"' "${VZ_CR_FILE}"
-    yq -i eval '.spec.components.istio.istioInstallArgs[3].value = "jaeger-verrazzano-managed-cluster-collector.verrazzano-monitoring.svc.cluster.local.:9411"' "${VZ_CR_FILE}"
-    yq -i eval '.spec.components.istio.istioInstallArgs[4].name = "global.proxy.holdApplicationUntilProxyStarts"' ${VZ_CR_FILE}
-    yq -i eval '.spec.components.istio.istioInstallArgs[4].value = "true"' ${VZ_CR_FILE}
+    yq -i eval '.spec.components.istio.istioInstallArgs[2].name = "meshConfig.defaultConfig.tracing.zipkin.address"' "${VZ_CR_FILE}"
+    yq -i eval '.spec.components.istio.istioInstallArgs[2].value = "jaeger-verrazzano-managed-cluster-collector.verrazzano-monitoring.svc.cluster.local.:9411"' "${VZ_CR_FILE}"
 elif [ "${CLUSTER_COUNT}" -gt 1  ] && [ $CRD_API_VERSION == "v1beta1" ]; then
   yq -i eval '.spec.components.istio.overrides.[0].values.apiVersion = "install.istio.io/v1alpha1"' ${VZ_CR_FILE}
```
mbenmclean, line 26: Same comment about using overrides not install args

platform-operator/controllers/verrazzano/component/istio/istio_cr_test.go
```diff
@@ -681,65 +707,175 @@ spec:
...
+        service:
+          ports:
+          - name: port1
+            nodePort: 32443
+            port: 8000
+            protocol: TCP
+            targetPort: 2000
+          type: NodePort
+      name: istio-ingressgateway
+  values:
```
mbenmclean, line 815: Should this set spec.meshConfig, not spec.values.meshConfig? Did you investigate using that?

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -4,34 +4,56 @@
...
 )
 
 const collectorZipkinPort = 9411
 
-//configureJaeger configures Jaeger for Istio integration and returns install args for the Istio install.
+const istioTracingTemplate = `
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  values:
```
mbenmclean, line 21: Same comment about using spec.meshConfig, not spec.values.meshConfig? Did you investigate using that?

platform-operator/controllers/verrazzano/component/istio/istio_cr_test.go
mbenmclean: See the meshConfig field in https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#IstioOperatorSpec

anajosep: As per the documenation, `spec.meshConfig` is the right way to set the tracing params. `Anything that is available in IstioOperatorSpec should be set above rather than using the passthrough.` (from the documentation). But when `InstallArgs` was implemented in a generic way, where any parameter can be overridden, it was designed to use `spec.values.meshConfig` otherwise it would have required special handling for `meshConfig`. I have kept the same approach. I will test the behavior in a separate branch and see if `spec.meshConfig` can be used.

anajosep: I have incorporated your review comment to use `spec.meshConfig` instead of `spec.values.meshConfig` to override the tracing configurations. When user uses install args, `spec.values.meshConfig` gets used, and that one has the highest precedence. If user is using install overrides, they have the freedom to use either of `spec.meshConfig` or `spec.values.meshConfig` and both works as expected with the later having higher precedence.

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -4,34 +4,101 @@
...
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  meshConfig:
+    defaultConfig:
+      tracing:
+        tlsSettings:
+          mode: "{{.TracingTLSMode}}"
+        zipkin:
+          address: "{{.JaegerCollectorURL}}"
```
mbenmclean, line 31: Should you add something for sampling?

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -4,34 +4,101 @@
...
-		},
-	}, nil
+	t, err := template.New("tracing_template").Parse(istioTracingTemplate)
+	if err != nil {
+		return "", err
+	}
+	var b bytes.Buffer
+	var data = istioTracingTemplateData{}
+	data.JaegerCollectorURL = collectorURL
+	data.TracingTLSMode = "ISTIO_MUTUAL"
```
mbenmclean, line 97: Same question about adding support for sampling

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -4,34 +4,101 @@
...
+// configurations merged with any user provided overrides.
+func buildJaegerTracingYaml(comp *v1beta1.IstioComponent) (string, error) {
+	// Build a list of YAML strings from the istioComponent initargs, one for each arg.
+	// get istio overrides for Jaeger tracing
+	jaegerTracingYaml, err := configureJaegerTracing()
+	if err != nil {
+		return "", err
+	}
+	expandedYamls := []string{jaegerTracingYaml}
+	for _, arg := range comp.ValueOverrides {
```
mbenmclean, line 49: So you are only getting the Values field for overrides. We have support for overrides in secrets and configmaps too. you need to use common.GetInstallOverridesYAML. That will generate a list of strings of all the overrides.
Also, we have a GetOverrides istioComponent method that will return a list of overrides. You should use that to generate the overrides array to pass into common.GetInstallOverridesYAML.

mbenmclean: Ping me if you have questions about how to do this.

tests/e2e/config/scripts/multicluster_edit_vz.sh
```diff
@@ -15,24 +15,13 @@ yq -i eval '.spec.components.kubeStateMetrics.enabled = true' "${VZ_CR_FILE}"
...
   yq -i eval '.spec.components.istio.overrides.[0].values.kind = "IstioOperator"' ${VZ_CR_FILE}
-  yq -i eval '.spec.components.istio.overrides.[0].values.spec.values.meshConfig.defaultConfig.tracing.sampling = 90' ${VZ_CR_FILE}
-  yq -i eval '.spec.components.istio.overrides.[0].values.spec.values.meshConfig.defaultConfig.tracing.tlsSettings.mode = "ISTIO_MUTUAL"' ${VZ_CR_FILE}
-  yq -i eval '.spec.components.istio.overrides.[0].values.spec.values.meshConfig.defaultConfig.tracing.zipkin.address = "jaeger-verrazzano-managed-cluster-collector.verrazzano-monitoring.svc.cluster.local.:9411"' ${VZ_CR_FILE}
-  yq -i eval '.spec.components.istio.overrides.[0].values.spec.values.meshConfig.enableTracing = true' ${VZ_CR_FILE}
+  yq -i eval '.spec.components.istio.overrides.[0].values.spec.meshConfig.defaultConfig.tracing.sampling = 90' ${VZ_CR_FILE}
+  yq -i eval '.spec.components.istio.overrides.[0].values.spec.meshConfig.defaultConfig.tracing.zipkin.address = "jaeger-verrazzano-managed-cluster-collector.verrazzano-monitoring.svc.cluster.local.:9411"' ${VZ_CR_FILE}
```
mbenmclean, line 22: Are we testing jaeger on the admin cluster with the default collector too? I would expect to see one for the admin cluster and one for the managed cluster.

platform-operator/controllers/verrazzano/component/istio/jaeger.go
anajosep: sampling is an optional parameter, if nothing is specified, the default sample rate of 1% would be used.

platform-operator/controllers/verrazzano/component/istio/jaeger.go
anajosep: sampling is an optional parameter, if nothing is specified, the default sample rate of 1% would be used.

tests/e2e/config/scripts/multicluster_edit_vz.sh
anajosep: Yes, the MC tests are verifying system traces in both admin and managed cluster. For managed cluster, the overrides are mandatory as we don't automatically set the Jaeger collector URL in the istio configuration. For admin cluster we automatically fill the Jaeger collector url and the TLS settings due to which special handling is not required in this script.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -274,29 +275,32 @@ func (i istioComponent) createIstioTempFiles(compContext spi.ComponentContext) (
...
-		userFileCR, err := createTempFile(log, istioOperatorYaml)
+		jaegerTracingFile, err := createTempFile(log, jaegerTracingYaml)
 		if err != nil {
 			return files, err
 		}
-		files = append(files, userFileCR)
-
 		// get the install overrides from the VZ CR, write it to a temp file and append it
-		files, err = appendOverrideFilesInOrder(compContext, files)
+		files, err = appendOverrideFilesInOrder(compContext, convertedVZ, files)
```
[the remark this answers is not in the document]
anajosep: No switching of order here. I have added an additional parameter `convertedVZ` here which is the `v1beta1` version of the effective CR.

platform-operator/controllers/verrazzano/component/istio/jaeger.go
anajosep: Ok, I will work on it.

platform-operator/controllers/verrazzano/component/istio/istio_install.go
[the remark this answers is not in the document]
mbenmclean: sg

platform-operator/controllers/verrazzano/component/istio/jaeger.go
mbenmclean: LGTM

tests/e2e/config/scripts/multicluster_edit_vz.sh
mbenmclean: LGTM

mbenmclean (approved):
LGTM

PR #4125, opened 2022-09-14
VZ-6507 guard existing components

shihchang:
webhook validation reports error if any of following components exists
- Istio
- Grafana
- Prometheus

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -18,11 +18,13 @@ import (
 	istioClient "istio.io/client-go/pkg/clientset/versioned"
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	kerrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	appsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	restclient "k8s.io/client-go/rest"
@@ -153,13 +155,28 @@ func GetKubernetesClientsetWithConfig(config *rest.Config) (*kubernetes.Clientse
 	return clientset, err
 }
 
-//GetCoreV1Client Returns the CoreV1Interface
+// GetCoreV1Func is the function to return the CoreV1Interface
+var GetCoreV1Func = GetCoreV1Client
+
+// GetCoreV1Client returns the CoreV1Interface
 func GetCoreV1Client(log ...vzlog.VerrazzanoLogger) (corev1.CoreV1Interface, error) {
 	goClient, err := GetGoClient(log...)
 	if err != nil {
 		return nil, err
 	}
 	return goClient.CoreV1(), nil
+}
+
+// GetAppsV1Func is the function the AppsV1Interface
+var GetAppsV1Func = GetAppsV1Client
+
+// GetAppsV1Client returns the AppsV1Interface
+func GetAppsV1Client(log ...vzlog.VerrazzanoLogger) (appsv1.AppsV1Interface, error) {
+	goClient, err := GetGoClient(log...)
+	if err != nil {
+		return nil, err
+	}
+	return goClient.AppsV1(), nil
 }
 
 // GetIstioClientset returns the clientset object for Istio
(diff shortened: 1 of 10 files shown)


pkg/k8sutil/k8sutil.go
```diff
@@ -153,7 +155,10 @@ func GetKubernetesClientsetWithConfig(config *rest.Config) (*kubernetes.Clientse
 	return clientset, err
 }
 
-//GetCoreV1Client Returns the CoreV1Interface
+//GetCoreV1Func Returns the CoreV1Interface
+var GetCoreV1Func = GetCoreV1Client
+
+//GetCoreV1Cli Returns the CoreV1Interface
```
desagar, line 161: minor nit - comment is different from function name

pkg/k8sutil/k8sutil.go
```diff
@@ -162,6 +167,18 @@ func GetCoreV1Client(log ...vzlog.VerrazzanoLogger) (corev1.CoreV1Interface, err
 	return goClient.CoreV1(), nil
 }
 
+//GetAppsV1Func Returns the AppsV1Interface
+var GetAppsV1Func = GetAppsV1Client
+
+//GetAppsV1Cli Returns the AppsV1Interface
+func GetAppsV1Client(log ...vzlog.VerrazzanoLogger) (appsv1.AppsV1Interface, error) {
```
desagar, line 174: same thing - comment different from func name. I thought one of the linters complains about this.

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
```diff
@@ -234,3 +235,18 @@ func (g grafanaComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzano,
...
 	return nil
 }
+
+// checkExistingGrafana checks if Grafana is already installed
+// OLCNE Istio module may have Grafana installed in istio-system namespace
+func checkExistingGrafana(vz runtime.Object) error {
+	if !vzconfig.IsGrafanaEnabled(vz) {
+		return nil
+	}
+	if err := k8sutil.ErrorIfDeploymentExists(constants.IstioSystemNamespace, ComponentName); err != nil {
```
desagar, line 245: shouldn't this be the verrazzano-system namespace, not istio-system?

desagar: Also, in my VZ install, the Grafana deployment is called vmi-system-grafana. There's a constant in the grafana package already called "grafanaDeployment" in grafana.go that you can use.

platform-operator/controllers/verrazzano/component/grafana/grafana_component_test.go
```diff
@@ -379,3 +380,37 @@ func TestValidateUpdateV1beta1(t *testing.T) {
 	newVz.Spec.Components.Grafana.Enabled = &trueValue
 	assert.NoError(t, NewComponent().ValidateUpdateV1Beta1(oldVz, newVz))
 }
+
+func TestValidateInstall(t *testing.T) {
+	svc := common.MkSvc(constants.IstioSystemNamespace, ComponentName)
```
desagar, line 385: once you fix the code to look in verrazzano system namespace, you should fix these to be in that NS too.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -291,6 +295,9 @@ func (i istioComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzano, ne
 
 // ValidateInstallV1Beta1 checks if the specified new Verrazzano CR is valid for this component to be updated
 func (i istioComponent) ValidateInstallV1Beta1(vz *installv1beta1.Verrazzano) error {
+	if err := checkExistingIstio(vz); err != nil {
```
pfmackin, line 298: You need to make sure this is not the Istio that Verrazzano installed since this code is idempotent and gets called post-install.  Same for all the other checks.

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
shihchang: OLCNE install grafana and prometheus in istio-system namespace. Though that does not cause collision, we still want to report the error of existing grafana when installing.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
```diff
@@ -192,3 +200,18 @@ func (c prometheusComponent) GetCertificateNames(ctx spi.ComponentContext) []typ
 
 	return certificateNames
 }
+
+// checkExistingPrometheus checks if Prometheus is already installed
+// OLCNE Istio module may have Prometheus installed in istio-system namespace
+func checkExistingPrometheus(vz runtime.Object) error {
```
desagar, line 206: suggest renaming func to checkExistingCNEPrometheus

platform-operator/controllers/verrazzano/component/istio/istio_component.go
desagar: agree about Istio - for the other checks, this might not be relevant, because for Grafana and prom operator, the CNE install is in Istio System

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
desagar: yeah I understood that later - hence my suggestion of renaming that checkExisting function so that it is obvious that it's the CNE one it's looking for, not just any Grafana install.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
desagar: For Istio, there's a verrazzano.io/namespace label on the istio-system namespace - can this be safely used to distinguish between VZ installed and not, @pfmackin ?

pkg/k8sutil/k8sutil.go
shihchang: fixed

pkg/k8sutil/k8sutil.go
shihchang: fixed

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
shihchang: fixed

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
shihchang: changed to checkExistingCNEGrafana

platform-operator/controllers/verrazzano/component/istio/istio_component.go
shihchang: changed checkExistingIstio to look for istio-system namespace and "verrazzano.io/namespace" label  which verrazzano labels during installation. If the istio-system namespace is found without  the "verrazzano.io/namespace" label, install validation reports an error.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -562,3 +571,35 @@ func getImageOverrides() ([]bom.KeyValue, error) {
...
+
+func checkNS(ns []v1.Namespace) error {
+	for _, n := range ns {
+		if n.Namespace == IstioNamespace || n.Name == IstioNamespace {
+			for l := range n.Labels {
+				if l == vzNsLabel {
+					return nil
+				}
+			}
+			return fmt.Errorf("existing namespace %s without %s label", IstioNamespace, vzNsLabel)
```
desagar, line 601: IIRC, this error is going to be displayed to the end user. I think this message should be clearer - i.e. something like "found existing Istio installation not created by Verrazzano."

Follow up question - you are no longer checking for pods here - just the existence of the namespace. Is that intentional?

desagar (approved):

PR #4126, opened 2022-09-14
VZ- 7060 don't look at version to enable or disable Keycloak AuthConfig

mbenmclean:
This change removes the logic to look at versions or whether it's an upgrade when deciding to set the kecloakoidc authconfig enabled to true or false. The keycloakoidc authconfig will be enabled unless it is explicitly disabled in the Rancher spec.


kminder (approved):
Approved.

kminder (approved):
Approved.

PR #4136, opened 2022-09-15
VZ-7101: Obfuscate password in mysql db load job spec

fredtibbitts:
Change the command to use the environment variable for the MySQL password instead of putting it directly in the command args.

I tested manually and it works, and I also created a test job that dumps out the command and it correctly substitutes the env var.


stevshap (approved):
lgtm

PR #4143, opened 2022-09-16
VZ-7117 CLI should not fail on Kubernetes API call failure

pfmackin:
VZ-7117 CLI should not fail if Kuberenets API server failure.   Put in retry in case of error.


mgianatagh (approved):

anders-swanson:
seems like we need generic retryable command wrapper.

PR #4144, opened 2022-09-16
VZ-7122: Disable mysql tests

theabmitra:


kminder (approved):
Approved.

mgianatagh (approved):

PR #4146, opened 2022-09-16
VZ-6064:Remove old upgrade test job from the periodic pipeline

Abhijithang:
This PR removes the old upgrade tests job from the periodic pipeline.

### DIFF
--- a/ci/JenkinsfilePeriodicTests
+++ b/ci/JenkinsfilePeriodicTests
@@ -338,21 +338,6 @@ pipeline {
                         }
                     }
                 }
-                stage('Upgrade tests') {
-                    steps {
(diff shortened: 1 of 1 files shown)


michael-cico (approved):

ci/JenkinsfilePeriodicTests
```diff
@@ -338,21 +338,6 @@ pipeline {
                         }
                     }
                 }
-                stage('Upgrade tests') {
```
michael-cico, line 341: should we remove the old job in Jenkins also after this?  and get rid of the jenkinsfile?  Or is the Job still used in older lines?

agv426 (approved):

PR #4152, opened 2022-09-16
VZ 7124: Use v1alpha1 for upgrade if v1beta1 not available 

anders-swanson:


platform-operator/internal/k8s/status/deployment_ready.go
```diff
@@ -78,6 +81,32 @@ func DoDeploymentsExist(log vzlog.VerrazzanoLogger, client clipkg.Client, namesp
...
+	}
+	if deployment.Status.AvailableReplicas < expectedReplicas {
+		log.Progressf("%s is waiting for deployment %s replicas to be %v. Current available replicas is %v", prefix, namespacedName,
+			expectedReplicas, deployment.Status.AvailableReplicas)
+		return false
+	}
+
+	// Velero install deploys a daemonset and deployment with common labels. The labels need to be adjusted so the pod fetch logic works
+	// as expected
+	podSelector := deployment.Spec.Selector
```
pfmackin, line 98: Velero specific code should not be here.  I think this has been discussed before and we gave it a pass.

pfmackin (approved):

platform-operator/internal/k8s/status/deployment_ready.go
anders-swanson: Yeah, I didn't know what to do with it, or even really why it was there.

PR #4161, opened 2022-09-19
VZ-6968: Allow vz CLI install and upgrade to run after VPO has been deployed

bobdonat:
This pull request allows the vz CLI install and upgrade commands to be run after the VPO has been deployed.  

1. install and upgrade can be rerun after hitting Ctrl-C during an install or upgrade
2. running install or upgrade after VPO is deployed for then same version will not timeout and fail
3. Additional checks are in place for install
4. Additional checks are in place for upgrade

As part of this change, moved code to pkg/k8s/status from platform-operator/internal/k8s/status since it is now shared between the install/upgrade code and the cli code.


tools/vz/test/helpers/objects.go
```diff
@@ -76,3 +72,16 @@ func CreateVPOPod(name string) client.Object {
 		},
 	}
 }
+
+func CreateVPOReplicset(name string) client.Object {
```
michael-cico, line 76: typo, Replicaset is misspelled

michael-cico (dismissed):

tools/vz/test/helpers/objects.go
bobdonat: fixed

michael-cico (approved):

PR #4163, opened 2022-09-19
VZ-5768 return progress message for pending IP address

davisgomes:
Progress message update for pending IPs

### DIFF
--- a/platform-operator/controllers/verrazzano/component/istio/istio_component.go
+++ b/platform-operator/controllers/verrazzano/component/istio/istio_component.go
@@ -397,7 +397,7 @@ func (i istioComponent) IsReady(context spi.ComponentContext) bool {
 	if err != nil {
 		// Only log for the Istio component context
 		if context.GetComponent() == ComponentName {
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -397,7 +397,7 @@ func (i istioComponent) IsReady(context spi.ComponentContext) bool {
 	if err != nil {
 		// Only log for the Istio component context
 		if context.GetComponent() == ComponentName {
-			context.Log().Errorf("Ingress external IP pending for component %s: %v", ComponentName, err)
+			context.Log().ErrorfThrottled("Ingress external IP pending for component %s: %v", ComponentName, err)
```
michael-cico, line 400: I would use `Progressf`, otherwise this will generate a stack trace every time I think.

It's not really an error, this is a known situation during install/update.

davisgomes: Okay, I couldn't decide between the two. I'll change if back to progress

michael-cico (approved):

stevshap (approved):

PR #4189, opened 2022-09-20
Fix storage location

agv426:
Set the OCI_OS_LOCATION to get from ephemeral location.

Fixing in release-1.4 will get it into master after


agv426:
This wasn't causing the failure, we don't need this in 1.4 I'm just going to open a new PR for master only

PR #4193, opened 2022-09-21
VZ-7041: Add client role view-users to verrazzano user

bhatpmk:
Add client role view-users to verrazzano user, to allow verrazzano user to view users in the realm verrazzano-system.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -1563,7 +1563,6 @@ func addClientRoleToUser(ctx spi.ComponentContext, cfg *restclient.Config, cli k
 	kcPod := keycloakPod()
 	addRoleCmd := "/opt/jboss/keycloak/bin/kcadm.sh add-roles -r " + targetRealm + " --uusername " + userName + " --cclientid " + clientID + " --rolename " + roleName
 	ctx.Log().Debugf("addRoleCmd: add role %s to the user %s, command = %s", roleName, userName, addRoleCmd)
-	ctx.Log().Infof("addRoleCmd: %s", addRoleCmd)
 	stdout, stderr, err := k8sutil.ExecPod(cli, cfg, kcPod, ComponentName, bashCMD(addRoleCmd))
 	if err != nil {
 		ctx.Log().Errorf("Adding role %s to the user %s failed: stdout = %s, stderr = %s", roleName, userName, stdout, stderr)
--- a/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
+++ b/platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
@@ -54,6 +54,8 @@ const (
 	vzInternalPromUser      = "verrazzano-prom-internal"
 	vzInternalEsUser        = "verrazzano-es-internal"
 	keycloakPodName         = "keycloak-0"
+	realmManagement         = "realm-management"
+	viewUsersRole           = "view-users"
 )
 
 // Define the Keycloak Key:Value pair for init container.
@@ -746,6 +748,11 @@ func configureKeycloakRealms(ctx spi.ComponentContext) error {
 			return err
 		}
 
+		// Add view-users role to verrazzano user
+		err = addClientRoleToUser(ctx, cfg, cli, vzUserName, realmManagement, vzSysRealm, viewUsersRole)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Setting password policy for master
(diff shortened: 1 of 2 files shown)


michael-cico (approved):

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1550,3 +1557,17 @@ func updateRancherClientSecretForKeycloakAuthConfig(ctx spi.ComponentContext) er
 	authConfig[common.AuthConfigKeycloakAttributeClientSecret] = clientSecret
 	return common.UpdateKeycloakOIDCAuthConfig(ctx, authConfig)
 }
+
+// addClientRoleToUser adds client role to the given user in the target realm
+func addClientRoleToUser(ctx spi.ComponentContext, cfg *restclient.Config, cli kubernetes.Interface, userName, clientID, targetRealm, roleName string) error {
+	kcPod := keycloakPod()
+	addRoleCmd := "/opt/jboss/keycloak/bin/kcadm.sh add-roles -r " + targetRealm + " --uusername " + userName + " --cclientid " + clientID + " --rolename " + roleName
```
mbenmclean, line 1564: Is --uusername and --cclientID right? Those look like they might be typos

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
```diff
@@ -1583,3 +1583,19 @@ func TestGetVerrazzanoUserFromKeycloak(t *testing.T) {
...
+func TestAddClientRoleToUser(t *testing.T) {
+	k8sutil.ClientConfig = fakeRESTConfig
+	k8sutil.NewPodExecutor = k8sutilfake.NewPodExecutor
+	cfg, cli, _ := fakeRESTConfig()
+
+	c := fake.NewClientBuilder().WithScheme(k8scheme.Scheme).Build()
+	ctx := spi.NewFakeContext(c, testVZ, nil, false)
+
+	err := addClientRoleToUser(ctx, cfg, cli, "testuser", "test-client", "test-realm", "test-role")
+	assert.NoError(t, err)
```
kminder, line 1600: Shouldn't block merge as this is time sensitive but this isn't much of a test.  What do the fakes even do when k8sutil.ExecPod is invoked.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1550,3 +1557,17 @@ func updateRancherClientSecretForKeycloakAuthConfig(ctx spi.ComponentContext) er
...
+	ctx.Log().Debugf("addRoleCmd: add role %s to the user %s, command = %s", roleName, userName, addRoleCmd)
+	stdout, stderr, err := k8sutil.ExecPod(cli, cfg, kcPod, ComponentName, bashCMD(addRoleCmd))
+	if err != nil {
```
kminder, line 1567: Is there any sort of error that should be retried?  Are there meaningful exit codes from kcadm?

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
bhatpmk: I agree that, I don't have good test coverage to validate the fix. I am exploring a way to validate the change with a functional test
Regarding your question on fakes, I don't know the answer. I have followed the existing pattern

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
bhatpmk: I have seen HTTP 401 and another error while I was exploring the options. Here are the 2 errors from my notes:
1. null [HTTP 401 Unauthorized]
2. Group not found for name: verrazzano-monitors

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
mbenmclean: OK just making sure

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
mbenmclean: Also we're logging the stdout and stderr, but are we logging the error anywhere. In my experience, the error that's produced sometimes has the useful information in it. I think we should be logging that too. While you're at it, perhaps you could also update the rest of the kcadm commands to log the err as well. I think usually for kcadm the err has more useful information that the stderr.

kminder (approved):
Approved

PR #4199, opened 2022-09-21
VZ-7149 refactor network policies so they have their own helm chart

pfmackin:
Currently, NetworkPolicies are installed as part of Verrazzano.  This is a race condition because Verrazzano component depends on several other components.  We want network policies to be installed before anything else. Change the code as follows:

1. Remove the NetworkPolicies from Verrazzano chart
2. Create a new chart for NetworkPolicies
3. Create a networkpolicies component that is run first during install and upgrade
4. Make components depend on the networkpolicies component (so they get installed first)
5. Create all namespaces needed by network policies before applying the chart
6. Disassociate NetworkPolicy resources from verrazzano helm release and associate with verrazzano-network-policies helm release


michael-cico (approved):

platform-operator/controllers/verrazzano/component/registry/registry_test.go
```diff
@@ -67,36 +68,67 @@ func TestGetComponents(t *testing.T) {
...
-	a.Equal(comps[22].Name(), pushgateway.ComponentName)
-	a.Equal(comps[23].Name(), promnodeexporter.ComponentName)
-	a.Equal(comps[24].Name(), jaegeroperator.ComponentName)
-	a.Equal(comps[25].Name(), console.ComponentName)
-	a.Equal(comps[26].Name(), fluentd.ComponentName)
-	a.Equal(comps[27].Name(), velero.ComponentName)
-	a.Equal(comps[28].Name(), rancherbackup.ComponentName)
+	var i int
+	a.Len(comps, 30, "Wrong number of components")
+	a.Equal(comps[i].Name(), networkpolicies.ComponentName)
```
mgianatagh, line 73: Are these checks really necessary?  Is the length check enough?  And maybe the check that network policies is first?

platform-operator/helm_config/charts/verrazzano-network-policies/values.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+name: verrazzano
+
+elasticSearch:
```
mgianatagh, line 5: name openSearch instead?  Same comment would apply for anywhere else that we are not using v1beta1 names.

platform-operator/helm_config/charts/verrazzano-network-policies/values.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
mgianatagh, line 1: copyright on a new file just needs to be 2022,

platform-operator/helm_config/charts/verrazzano-network-policies/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: v1
+description: A Helm chart for Verrazzano Network Policies
+name: verrazzano-network-policies
+version: 1.4.1
```
mgianatagh, line 6: for the master branch should this be 1.5.0?

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
```diff
@@ -0,0 +1,146 @@
...
+	constants.VerrazzanoMonitoringNamespace,
+	constants.VerrazzanoSystemNamespace,
+}
+
+var (
+	// For Unit test purposes
+	writeFileFunc = ioutil.WriteFile
+)
+
+//func resetWriteFileFunc() {
```
mgianatagh, line 51: should this be removed?  Or should it be being used by the unit tests?

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", err
+	}
+	ctx.Log().Debugf("Verrazzano install overrides file %s contents: %s", overridesFileName, string(bytes))
+	return overridesFileName, nil
+}
+
+func appendVerrazzanoValues(ctx spi.ComponentContext, overrides *chartValues) error {
+	effectiveCR := ctx.EffectiveCR()
+
+	overrides.ElasticSearch = &elasticsearchValues{Enabled: vzconfig.IsOpenSearchEnabled(effectiveCR)}
```
mgianatagh, line 103: should be overrides.OpenSearch

platform-operator/controllers/verrazzano/component/registry/registry_test.go
pfmackin: I don't want to remove this unit test.  I guess it prevents someone from inadvertently changing the registry.

platform-operator/helm_config/charts/verrazzano-network-policies/values.yaml
pfmackin: This is the same as the old third party chart so i didn't want to change that.

platform-operator/helm_config/charts/verrazzano-network-policies/Chart.yaml
pfmackin: i think you are right

platform-operator/helm_config/charts/verrazzano-network-policies/values.yaml
pfmackin: ok

platform-operator/controllers/verrazzano/component/registry/registry_test.go
mgianatagh: ok

platform-operator/helm_config/charts/verrazzano-network-policies/values.yaml
mgianatagh: ok

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
pfmackin: I left it for Fred, he needs it for unit tests

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
pfmackin: this is v1apha1, and it is exactly the same as the Verrazzano component.  When we switch to beta1 internally this will all get changed.

mgianatagh (approved):

fredtibbitts (approved):

platform-operator/controllers/verrazzano/component/networkpolicies/netpol_component.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// IsEnabled always returns true since network policies are always installed
+func (c networkPoliciesComponent) IsEnabled(effectiveCR runtime.Object) bool {
+	return true
+}
+
+// PreInstall performs pre-install actions
+func (c networkPoliciesComponent) PreInstall(ctx spi.ComponentContext) error {
+	// Create all namespaces needed by network policies
```
michael-cico, line 78: since the pre upgrade/pre-install both call CreateAndLabelNamespaces, and associateNewtorkPolicies, maybe that block can be consolidated and called from both

platform-operator/helm_config/charts/verrazzano-network-policies/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: v1
+description: A Helm chart for Verrazzano Network Policies
+name: verrazzano-network-policies
+version: 1.5.0
```
michael-cico, line 6: I think the `verrazzano-helper` might also need to account for this new chart, since I think it now updates the version in the charts we own.

michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/networkpolicies/netpol_component.go
pfmackin: it is only 2 calls from the component code, not worth it IMO

platform-operator/helm_config/charts/verrazzano-network-policies/Chart.yaml
pfmackin: I will add a subtask to my JIRA

michael-cico (approved):

stevshap (approved):

PR #4201, opened 2022-09-22
Change timeout to 12 hours, have upgrade testing run 6 in parallel

agv426:
Periodic tests in master and release-1.4 are timing out, increase the timeout and run 6 in parallel for upgrade testing. That should help for now to reduce the overall time, and extend the timeout as well to wait longer. The time issue needs to be investigated separately (will create a task for that).

### DIFF
--- a/ci/JenkinsfilePeriodicTests
+++ b/ci/JenkinsfilePeriodicTests
@@ -47,7 +47,7 @@ def verrazzanoDistributionsFile = "verrazzano_distributions.html"
 
 pipeline {
     options {
-        timeout(time: 9, unit: 'HOURS')
+        timeout(time: 12, unit: 'HOURS')
         skipDefaultCheckout true
         disableConcurrentBuilds()
         timestamps ()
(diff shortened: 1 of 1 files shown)


Abhijithang (approved):

ci/JenkinsfilePeriodicTests
```diff
@@ -343,6 +343,7 @@ pipeline {
                             script {
                                 build job: "/verrazzano-push-triggered-upgrade-minor-release-tests/${CLEAN_BRANCH_NAME}",
                                     parameters: [
+                                        string(name: 'N_JOBS_FOR_EACH_BATCH', value: '6'),
```
Abhijithang, line 346: Or we could directly update it in the upgrade minor release tests [jenkinsfile](https://github.com/verrazzano/verrazzano/blob/bc5a45044ed72b3d9f7182c8b7b05b78333f1c56/ci/upgrade-paths/JenkinsfileTestTriggerMinorRelease#L37)?

agv426: We could do either way but I think for this the issue is happening because the overall pipeline has a timeout that is biting us, so adjusting it only when run in this pipeline to pull the times in

PR #4204, opened 2022-09-22
VZ-7041: Add client role view-users to verrazzano user

bhatpmk:
Add client role view-users to verrazzano user, in the realm verrazzano-system.


kminder (dismissed):
Approved.

michael-cico (approved):

PR #4219, opened 2022-09-22
Fix issues found by go-vet

ericrrath:
Go vet found two issues:

	$ go vet ./...
	# github.com/verrazzano/verrazzano/tests/e2e/multicluster/examples
	tests/e2e/multicluster/examples/example_utils.go:134:3: unreachable code
	# github.com/verrazzano/verrazzano/tools/generate-profiles
	tools/generate-profiles/generate.go:58:3: fmt.Println arg list ends with redundant newline

Both had trivial fixes: simplify the return statements for the first, and change from Println() to Print() for the second.


ericrrath:
@gilbode fixed one of these in a4b799a110, and the other one is [skipped by default](https://github.com/golangci/golangci-lint/blob/v1.47.3/pkg/packages/skip.go#L22), so closing this.

PR #4222, opened 2022-09-23
VZ-7041: Validate the view-user role for verrazzano user

bhatpmk:
Adds a new test case to validate the verrazzano user has view-user client role. Also updated the debug messages in keycloak.go


tests/e2e/verify-install/keycloak/keycloak_test.go
```diff
@@ -21,9 +21,13 @@ import (
 )
 
 const (
-	waitTimeout              = 10 * time.Minute
-	pollingInterval          = 30 * time.Second
-	keycloakNamespace string = "keycloak"
+	waitTimeout       = 10 * time.Minute
```
lake-of-dreams, line 24: aren't we able to use these constants from keycloak package?

tests/e2e/verify-install/keycloak/keycloak_test.go
```diff
@@ -32,6 +36,12 @@ type KeycloakClients []struct {
 	ClientID string `json:"clientId"`
 }
 
+// KeycloakRoles represents an array of roles configured in Keycloak
+type KeycloakRoles []struct {
```
lake-of-dreams, line 40: is it not importable from keycloak package

tests/e2e/verify-install/keycloak/keycloak_test.go
```diff
@@ -100,12 +110,25 @@ var volumeClaims map[string]*corev1.PersistentVolumeClaim
...
 		volumeClaims, err = pkg.GetPersistentVolumeClaims(keycloakNamespace)
 		return volumeClaims, err
 	}, waitTimeout, pollingInterval).ShouldNot(BeNil())
+
+	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
+	if err != nil {
+		Fail(fmt.Sprintf("Failed to get default kubeconfig path: %s", err.Error()))
+	}
+	isKeycloakEnabled = pkg.IsKeycloakEnabled(kubeconfigPath)
+	isMinVersion140, err = pkg.IsVerrazzanoMinVersion("1.4.0", kubeconfigPath)
```
lake-of-dreams, line 128: we are again checking it in line 337 below, should be changed there as well to just use this global value

tests/e2e/verify-install/keycloak/keycloak_test.go
```diff
@@ -500,3 +520,66 @@ func verifyRancherClientURIs(keycloakClient *Client, env string) bool {
...
+func verifyUserClientRole(user, userRole string) bool {
+	var keycloakRoles KeycloakRoles
+
+	// Login to Keycloak
+	if !loginKeycloak() {
+		return false
+	}
+
+	// Get the roles for the verrazzano user
+	cmd := exec.Command("kubectl", "exec", "keycloak-0", "-n", "keycloak", "-c", "keycloak", "--", "/opt/jboss/keycloak/bin/kcadm.sh", "get-roles", "-r", vzSysRealm, "--uusername", user, "--cclientid", realmMgmt, "--effective", "--fields", "id,name")
```
lake-of-dreams, line 560: do we really need the id here, cant't it just be `--fields name`

tests/e2e/verify-install/keycloak/keycloak_test.go
bhatpmk: The constants in keycloak package are not exported, prefer to do that later as a follow up change

tests/e2e/verify-install/keycloak/keycloak_test.go
bhatpmk: Done

tests/e2e/verify-install/keycloak/keycloak_test.go
bhatpmk: Updated to just get the name

tests/e2e/verify-install/keycloak/keycloak_test.go
bhatpmk: The constants in keycloak package are not exported, prefer to do the refactoring later.

lake-of-dreams (approved):
LGTM

PR #4224, opened 2022-09-23
VZ-6932: Validated length of environment name and dns suffix based on user's spec with a webhook

adiforluls:


platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -213,3 +214,55 @@ func GetOverrides(object runtime.Object) interface{} {
...
+func validateLongestHostName(effectiveCR runtime.Object) error {
+	envName := getEnvironmentName(effectiveCR)
+	dnsSuffix, wildcard := getDNSSuffix(effectiveCR)
+	longestHostName := fmt.Sprintf("%s.%s.%s", "elasticsearch.vmi.system", envName, dnsSuffix)
+	if len(longestHostName) > 64 {
+		if wildcard {
+			return fmt.Errorf("The specified environmentName %s is too long", envName)
+		}
+
+		return fmt.Errorf("The specified environmentName %s and DNS suffix %s will result in host names for system endpoints longer than 64 characters", envName, dnsSuffix)
```
adiforluls, line 229: Check if the message can be better articulated.

mbenmclean: Might we worth having the message include something about a commonName and it have the 64 character limit

adiforluls: I originally included that but the error message didn't sound very concise. But we can include that though.

How about we mention how much space is left? Something like `The specified environmentName and the dns suffix are too long for host names of system endpoints. In combination, they must not be longer than x characters for certificates to be generated based on the provided configuration.`. This way they'll know by how much exactly they need to reduce these names.

adiforluls: Made the error log a more helpful message

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -213,3 +215,57 @@ func GetOverrides(object runtime.Object) interface{} {
...
+
+func getEnvironmentName(effectiveCR runtime.Object) string {
+	if cr, ok := effectiveCR.(*vzapi.Verrazzano); ok {
+		return cr.Spec.EnvironmentName
+	}
+	cr := effectiveCR.(*installv1beta1.Verrazzano)
+	return cr.Spec.EnvironmentName
+}
+
+func getDNSSuffix(effectiveCR runtime.Object) (string, bool) {
```
michael-cico, line 245: we already have code to compute the DNS suffix, although that computes it based on the actual domain in the wildcard case.  It may return an error if the ingress IP isn't yet known during install.

michael-cico: See https://github.com/verrazzano/verrazzano/blob/master/platform-operator/internal/vzconfig/ingress.go#L51

michael-cico: Although during initial install/creation, this can't be called by a validator, so you may not have that option.

adiforluls: Yeah I did look at this but decided not to use it because of the reason you pointed.

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -213,3 +215,57 @@ func GetOverrides(object runtime.Object) interface{} {
...
+	dnsSuffix, wildcard := "0.0.0.0", true
+	if cr, ok := effectiveCR.(*vzapi.Verrazzano); ok {
+		if cr.Spec.Components.DNS == nil || cr.Spec.Components.DNS.Wildcard != nil {
+			return fmt.Sprintf("%s.%s", dnsSuffix, vzconfig.GetWildcardDomain(cr.Spec.Components.DNS)), wildcard
```
michael-cico, line 249: The effective config will always have a Wildcard setting right?

adiforluls: Yes

michael-cico (approved):

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -213,3 +215,57 @@ func GetOverrides(object runtime.Object) interface{} {
...
+			dnsSuffix = cr.Spec.Components.DNS.OCI.DNSZoneName
+		} else if cr.Spec.Components.DNS.External != nil {
+			wildcard = false
+			dnsSuffix = cr.Spec.Components.DNS.External.Suffix
+		}
+		return dnsSuffix, wildcard
+	}
+
+	cr := effectiveCR.(*installv1beta1.Verrazzano)
+	if cr.Spec.Components.DNS == nil || cr.Spec.Components.DNS.Wildcard != nil {
```
mbenmclean, line 261: This looks very similar to the code above. Could the logic be combined?

mbenmclean (approved):

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
adiforluls: I had this thought as well while implementing but that's not possible right? Since the two different API version objects are essentially two different types and there's no common interface for them besides `runtime.Object`.

mbenmclean: Yes I had a feeling that might be the case.

PR #4240, opened 2022-09-26
VZ-6159 - Validate supported Kubernetes versions in VPO

lake-of-dreams:
- Maintain list of supported k8s versions in bom
- Validate if the cluster version is either same or a patch. version of the ones supported


lake-of-dreams:
So after this change, the `webhook-init` pod will _ultimately_ be failed and VPO main pod will be in pending status. Since it is a deployment, it is not [supported](https://github.com/kubernetes/kubernetes/issues/24725) to have a `restartPolicy` of `Never` and therefore the `webhook-init` pod restarts few times. Reviewers may please suggest if there is a better way.

michael-cico (approved):

PR #4244, opened 2022-09-26
Fix keycloak-mysql network policy after upgrade

fredtibbitts:
When upgrading to 1.4, we update network policies, including the keycloak-mysql policy. The old policy has a podSelector that matches on the `app=mysql` label. The new policy matches on `tier=mysql` (for the new mysql-operator managed mysql). When the new network policy is installed via helm, the podSelector matchLabels are combined, and since the new mysql pod does not have an `app` label, the network policy is not applied correctly and the upgrade hangs.

Since we cannot add labels to the new mysql pods, this fix is implemented in the verrazzano-network-policies install component, in the `PostUpgrade` step. In `PostUpgrade`, we remove the `app` podSelector label matcher so that only the `tier=mysql` label is matched.

I tested this manually in a kind cluster with Calico by installing 1.3.5 and upgrading to the version on this branch. The podSelector gets updated properly and the upgrade completes as expected.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
+++ b/platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
@@ -5,13 +5,14 @@ package networkpolicies
 
 import (
 	"context"
-	vzos "github.com/verrazzano/verrazzano/pkg/os"
 	"io/fs"
 	"io/ioutil"
+	"os"
+
+	vzos "github.com/verrazzano/verrazzano/pkg/os"
 	netv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	"os"
 	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
@@ -28,10 +29,13 @@ const (
 	tmpSuffix            = "yaml"
 	tmpFileCreatePattern = tmpFilePrefix + "*." + tmpSuffix
 	tmpFileCleanPattern  = tmpFilePrefix + ".*\\." + tmpSuffix
+
+	keycloakMySQLNetPolicyName = "keycloak-mysql"
+	podSelectorAppLabelName    = "app"
 )
 
 // netpolNamespaceNames specifies the NSNs that have network policies
-var netpolNamespaceNames []types.NamespacedName = []types.NamespacedName{
+var netpolNamespaceNames = []types.NamespacedName{
 	{Namespace: vzconst.CertManagerNamespace, Name: "cert-manager"},
 	{Namespace: vzconst.CertManagerNamespace, Name: "external-dns"},
 	{Namespace: constants.IngressNginxNamespace, Name: "ingress-nginx-controller"},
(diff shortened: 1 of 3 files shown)


michael-cico (approved):

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
```diff
@@ -208,3 +212,26 @@ func cleanTempFiles(ctx spi.ComponentContext) {
...
+func fixKeycloakMySQLNetPolicy(ctx spi.ComponentContext) error {
+	netpol := &netv1.NetworkPolicy{}
+	nsn := types.NamespacedName{Namespace: constants.KeycloakNamespace, Name: keycloakMySQLNetPolicyName}
+	if err := ctx.Client().Get(context.TODO(), nsn, netpol); err != nil {
+		ctx.Log().Errorf("Error retrieving network policy %s/%s: %v", constants.KeycloakNamespace, keycloakMySQLNetPolicyName, err)
+		return err
+	}
+
+	// if the podSelector has an "app" label matcher, remove it
+	if _, exists := netpol.Spec.PodSelector.MatchLabels[podSelectorAppLabelName]; exists {
```
pfmackin, line 228: This assumes labels exist but i guess that is ok

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
```diff
@@ -208,3 +212,26 @@ func cleanTempFiles(ctx spi.ComponentContext) {
...
 	}
 }
+
+// fixKeycloakMySQLNetPolicy fixes the keycloak-mysql network policy after an upgrade. When we apply the new version
+// of the network policy, the podSelector match labels are merged with the existing policy match labels, but the
+// old label no longer exists on the new MySQL pod. This function removes that label matcher if it exists.
+func fixKeycloakMySQLNetPolicy(ctx spi.ComponentContext) error {
+	netpol := &netv1.NetworkPolicy{}
+	nsn := types.NamespacedName{Namespace: constants.KeycloakNamespace, Name: keycloakMySQLNetPolicyName}
+	if err := ctx.Client().Get(context.TODO(), nsn, netpol); err != nil {
```
pfmackin, line 222: Wondering if there is ever case where not found is returned

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
fredtibbitts: I think the only way MatchLabels would be nil is if we are upgrading from 1.4+ and we change the network policy to use MatchExpressions instead of MatchLabels. It's probably a good idea to add the nil check though, how about I submit a subsequent PR with that change just so we can get this into the shared branch and unblock upgrade testing?

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
fredtibbitts: I think that would only happen if we completely remove the network policy in a future release, but since we're installing it now I think it makes sense to return an error if we don't find it.

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
fredtibbitts: Merging this and will create another PR to add the nil check.

PR #4249, opened 2022-09-26
VZ-6767 store Kiali signing key as a secret

mbenmclean:
This change generates a random string to use as the Kiali signing key and stores it in the secret. This will prevent the Kiali pod from bouncing after every upgrade. It also modifies the GetValues function to include -o yaml as an argument. This fixes a bug where helm CLI output was getting included in the values file.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/kiali/kiali.go
+++ b/platform-operator/controllers/verrazzano/component/kiali/kiali.go
@@ -6,17 +6,14 @@ package kiali
 import (
 	"context"
 	"fmt"
+
+	"github.com/verrazzano/verrazzano/pkg/bom"
 	globalconst "github.com/verrazzano/verrazzano/pkg/constants"
+	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/kiali/kiali.go
```diff
@@ -192,6 +208,46 @@ func buildKialiHostnameForDomain(dnsDomain string) string {
...
+	}
+	err := ctx.Client().Get(context.TODO(), clipkg.ObjectKey{
+		Namespace: constants.VerrazzanoSystemNamespace,
+		Name:      kialiSigningKeySecret,
+	}, &secret)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return "", ctx.Log().ErrorfThrottledNewErr("Unexpected error getting secret %s, : %V", kialiSigningKeySecret, err)
+		}
+		pw, err := password.GeneratePassword(16)
```
anders-swanson, line 228: generating password twice?

mbenmclean: Oh that's a mistake fixed it

PR #4250, opened 2022-09-27
VZ-7140 Fixed VPO startup probe to check if the webhooks for both v1alpha1 and v1beta1 are ready

anajosep:
Fix for Bug VZ-7140 
The bug fixes addresses 2 issues.
- a minor typo where `%` sign was missed to be prefixed in the format pattern.
- fix the `startupProbe` logic to make the calls to VPO to wait till the webhooks for both `v1alpha1` and `v1beta1` version is available.


platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -44,13 +44,12 @@ spec:
...
-              port: webhook
-              scheme: HTTPS
-              httpHeaders:
-                - name: Content-Type
-                  value: application/json
+            exec:
+              command:
+                - sh
+                - -c
+                - curl -kfs https://localhost:9443/validate-install-verrazzano-io-v1alpha1-verrazzano
```
bobdonat, line 51: The ticket was also to check the convert webhook.  Is there a reason that is not being checked?

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -44,13 +44,12 @@ spec:
...
+                  && curl -kfs https://localhost:9443/validate-install-verrazzano-io-v1beta1-verrazzano
```
bobdonat, line 52: Can you reliable assume the port is always 9443

anajosep: The port is hard coded in the same template yaml file for webhook. So if we are changing the port, we need to do it in 2 places.

bobdonat (approved):

anders-swanson:
Seems incomplete if we are not asserting for all webhooks - it's missing the VMC webhook as well.
To get a complete check, it's necessary to do something like this:
Create a HTTP endpoint (/health/ready or somesuch) that probes all the webhooks so we get full coverage. Ideally /health/ready would return something like this, along with appropriate status code:
```
{
  webhooks: {
    validate-v1-beta1-verrazzano: UP,
    validate-v1-alpha1-verrazzano: DOWN,
    validate-v1-alpha1-vmc: UP,
    convert: UP,
    
  },
  status: NOT_READY
}
```

of course this is just me thinking out loud, but would like to get this more comprehensive.

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
anajosep: I have handled the health check for `/convert` as well by using a `POST` request with custom data.

anders-swanson (dismissed):

anders-swanson:
you may want to wrap the command in a script

anders-swanson (approved):

PR #4258, opened 2022-09-27
VZ-7266 log progress instead of error for Keycloak OIDC config

pfmackin:
Use a log progress messages instead of logging an error when waiting for the Keycloak AuthConfig.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/rancher.go
+++ b/platform-operator/controllers/verrazzano/component/common/rancher.go
@@ -6,6 +6,7 @@ package common
 import (
 	"context"
 	"crypto/x509"
+	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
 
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -6,6 +6,8 @@ package common
 import (
 	"context"
 	"crypto/x509"
+	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
+	"k8s.io/apimachinery/pkg/api/errors"
```
stevshap, line 11: these imports could be cleaned up

stevshap (dismissed):

platform-operator/controllers/verrazzano/component/common/rancher.go
pfmackin: done

stevshap (approved):

PR #4262, opened 2022-09-28
Prevent or Warn for Kiali installation when Istio disabled

neeraj-n-singh:

Summary: kiali deployment  will be enabled whenever istio is enabled or vice versa.

```
kubectl logs verrazzano-platform-operator-78fd9fcd49-kdqn9 -n verrazzano-install  | grep kiali
{"level":"info","@timestamp":"2022-09-27T12:00:26.183Z","caller":"verrazzano/install.go:104","message":"Component kiali-server is disabled, skipping install","resource_namespace":"default","resource_name":"admin","controller":"verrazzano","component":"kiali-server","operation":"install"}
```

### DIFF
--- a/platform-operator/controllers/verrazzano/component/kiali/kiali_component.go
+++ b/platform-operator/controllers/verrazzano/component/kiali/kiali_component.go
@@ -146,7 +146,7 @@ func (c kialiComponent) IsReady(context spi.ComponentContext) bool {
 
 // IsEnabled Kiali-specific enabled check for installation
 func (c kialiComponent) IsEnabled(effectiveCR runtime.Object) bool {
-	return vzconfig.IsKialiEnabled(effectiveCR)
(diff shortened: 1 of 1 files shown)


adiforluls (approved):

platform-operator/controllers/verrazzano/component/kiali/kiali_component.go
```diff
@@ -146,7 +146,7 @@ func (c kialiComponent) IsReady(context spi.ComponentContext) bool {
 
 // IsEnabled Kiali-specific enabled check for installation
 func (c kialiComponent) IsEnabled(effectiveCR runtime.Object) bool {
-	return vzconfig.IsKialiEnabled(effectiveCR)
+	return vzconfig.IsIstioEnabled(effectiveCR) && vzconfig.IsKialiEnabled(effectiveCR)
```
lake-of-dreams, line 149: This does not seem correct that in CR it is still enabled as true but behind the scene we are disabling it in the code based on some other component, it should be webhook based

PR #4269, opened 2022-09-28
VZ-7189: Update Coherence Operator to 3.2.8

dhirupandey:
Update the Coherence Operator to 3.2.8


tests/e2e/multicluster/workloads/mccoherence/coherence_workload_mc_test.go
```diff
@@ -317,7 +318,7 @@ var _ = t.Describe("In Multi-cluster, verify Coherence application", Label("f:mu
 							m := make(map[string]string)
 							m[labelApp] = appName
 							m[clusterNameMetricsLabel] = clusterName
-							return pkg.MetricsExistInCluster(vendoRequestsCount, m, adminKubeconfig)
+							return pkg.MetricsExistInCluster(vendorRequestsCount, m, adminKubeconfig)
```
pfmackin, line 321: Doesn't this have to be fixed in non-multicluster metrics test?

pfmackin (dismissed):

platform-operator/thirdparty/charts/coherence-operator/values.yaml
```diff
@@ -1,17 +1,17 @@
...
 # defaultCoherenceImage is the default application image that will be used if a Coherence
 # resource does not specify an image name.
-defaultCoherenceImage: "ghcr.io/oracle/coherence-ce:21.12.1"
+defaultCoherenceImage: "ghcr.io/oracle/coherence-ce:22.06.1"
 
-# defaultCoherenceUtilsImage is the Coherence Operator utils image that will be used when running
+# defaultCoherenceUtilsImage is the Coherence Operator image that will be used when running
 # Coherence Pods. This image version should typically match the Operator version.
-defaultCoherenceUtilsImage: "ghcr.io/oracle/coherence-operator:3.2.6-utils"
+defaultCoherenceUtilsImage: "ghcr.io/oracle/coherence-operator:3.2.8"
```
kminder, line 14: The utils image isn't a separate image anymore?

platform-operator/thirdparty/charts/coherence-operator/values.yaml
```diff
@@ -21,6 +21,10 @@ watchNamespaces: ""
 # If not set the pull policy is "IfNotPresent".
 imagePullPolicy:
 
+# replicas is the number of Operator replicas to run.
+# The operator runs HA by default, with one of the replicas assuming leadership.
+replicas: 3
```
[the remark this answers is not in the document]
kminder: It can be 3 here I supposed but should be in base.yaml to be 1.
Also the ha.yaml should be updated with the recommended HA value which appears to be 3.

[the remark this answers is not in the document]
kminder: https://github.com/verrazzano/verrazzano/blob/master/examples/ha/ha.yaml

platform-operator/thirdparty/charts/coherence-operator/templates/deployment.yaml
```diff
@@ -168,6 +185,28 @@ spec:
 {{- if .Values.affinity }}
       affinity:
 {{ toYaml .Values.affinity | indent 8 }}
+{{- else }}
```
kminder, line 188: I'm wondering if this should be overridden in base.yaml to include hostname based anti-affinity like all the other compnents.  e.g. https://github.com/verrazzano/verrazzano/blob/master/platform-operator/manifests/profiles/v1beta1/base.yaml#L12
After seeing this I'm also wondering if all our anti-affinity rules should include both hostname and zone based anti-affinity.

tests/e2e/multicluster/workloads/mccoherence/coherence_workload_mc_test.go
dhirupandey: non-multicluster test did not use cpuCFSPeriods, so it did not need fixing

platform-operator/thirdparty/charts/coherence-operator/values.yaml
dhirupandey: With version 3.2.8 of  the Coherence Operator we only have a single image

platform-operator/thirdparty/charts/coherence-operator/values.yaml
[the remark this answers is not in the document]
dhirupandey: Do you mean that I should add something like this to https://github.com/verrazzano/verrazzano/blob/master/examples/ha/ha.yaml
`    coherenceOperator:
         overrides:
             - values:
                 replicas: 3
  `

platform-operator/thirdparty/charts/coherence-operator/templates/deployment.yaml
dhirupandey: I am not sure what needs to be done here. I'll sync up with you before I proceed with anything

platform-operator/thirdparty/charts/coherence-operator/values.yaml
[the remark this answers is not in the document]
kminder: Yes this is what I mean.  This would be done in conjunction with updates to the base.yaml commented on below.

platform-operator/thirdparty/charts/coherence-operator/templates/deployment.yaml
kminder: Either the helm chart or these would need to be updated to include hostname based anti-infinity 
https://github.com/verrazzano/verrazzano/blob/master/platform-operator/manifests/profiles/v1beta1/base.yaml#L67
https://github.com/verrazzano/verrazzano/blob/master/platform-operator/manifests/profiles/v1alpha1/base.yaml#L64
The reason is that we want to ensure that a single node going out of service doesn't cause downtime.  This for example is a very common use case during Kubernetes cluster upgrade.  I'm honestly not sure the best way to combine the topology.kubernetes.io/zone, oci.oraclecloud.com/fault-domain and kubernetes.io/hostname anti-affinity rules.

platform-operator/thirdparty/charts/coherence-operator/values.yaml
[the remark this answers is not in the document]
dhirupandey: ok thanks

platform-operator/thirdparty/charts/coherence-operator/templates/deployment.yaml
dhirupandey: I am also not sure how this can be done. Will need to research this and possibly confer with others in Coherence team about this.

Does this change necessarily need to be done as part of this PR ? or can I do these changes in a separate PR ?

platform-operator/thirdparty/charts/coherence-operator/values.yaml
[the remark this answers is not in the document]
kminder: I don't think this is actually resolved.  By default there should be a single replica if the Cohernece operator when Verrazzano is installed.  I think these changes as it will still result in 3 replicas as neither platform-operator/manifests/profiles/v1alpha1/base.yaml, platform-operator/manifests/profiles/v1beta1/base.yaml or platform-operator/helm_config/overrides/coherence-values.yaml are changed.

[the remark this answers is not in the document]
dhirupandey: I set this conversation to resolved since my question was answered by you. I do understand that this fix needs to go along with inclusion of the anti affinity stuff being discussed below. 

Let me know if you want me to mark this as resolved

platform-operator/thirdparty/charts/coherence-operator/templates/deployment.yaml
thegridman: It should be pretty straight forward to add an additional rule to the anti-affinity spec. We should probably do this in the Operator too rather than in Vz. The current rules are fine for things like OCI with three zones, but thinking about it someone could run in a k8s cluster with a single zone (or none configured) in which case we should fall back to Node anti-affinity. If you want to do this as a one off for this update in Vz that would work, then we can fix the Operator for 3.2.9

platform-operator/helm_config/overrides/coherence-values.yaml
```diff
@@ -1,6 +1,21 @@
-# Copyright (c) 2021, Oracle and/or its affiliates.
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
kminder, line 1: If this isn't a new file it should be...
# Copyright (c) 2021, 2022, Oracle and/or its affiliates.

kminder:
Can you please update CHANGES.md in the root with 
### v1.4.0
Component version updates:

- Coherence Operator v3.2.8

platform-operator/helm_config/overrides/coherence-values.yaml
dhirupandey: Fixed this issue

kminder (approved):
Approved.

PR #4270, opened 2022-09-29
Update refs to deprecated functions; add linter check

ericrrath:
This change updates all references to deprecated functions.  All the io.ioutils funcs had straightforward like-for-like replacements.  The k8s fake client changes are the same kind as those described in fc0045cd4.  There were some uses of strings.Title() that required use of the newer text.cases.Title() func.  And the one reference to net.Error.Temporary() is replaced with net.Error.Timeout() (per the docs).

Finally, the golangci-lint configuration is updated to have staticcheck enforce a check for deprecated references.


platform-operator/controllers/clusters/rancher.go
```diff
@@ -456,15 +456,15 @@ func doRequest(req *http.Request, rc *rancherConfig, log vzlog.VerrazzanoLogger)
...
 
 	retry(defaultRetry, log, func() (bool, error) {
 		// update the body with the saved data to prevent the "zero length body" error
-		req.Body = ioutil.NopCloser(bytes.NewBuffer(buffer))
+		req.Body = io.NopCloser(bytes.NewBuffer(buffer))
 		resp, err = rancherHTTPClient.Do(client, req)
 
 		// check for a network error and retry
-		if nerr, ok := err.(net.Error); ok && nerr.Temporary() {
+		if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
```
adiforluls, line 467: Why did this have to change?

ericrrath: It's [deprecated in golang 1.18.0](https://pkg.go.dev/net@go1.18#Error), and I think there are plans to update this repo to build with even-newer golang 1.19.

anders-swanson: this repo now uses go 1.19

ericrrath: @anders-swanson 👍

anders-swanson (approved):
LGTM

PR #4289, opened 2022-10-03
VZ-7190: Add test to verify the effect of changing heap memory settings for OpenSearch pods

lake-of-dreams:
- Verifies that the container command line contains the same heap settings as set by VMO in OPENSEARCH_JAVA_OPTS

**PS:** The PR will be updated for the VMO image after https://github.com/verrazzano/verrazzano-monitoring-operator/pull/259 is merged

PR #4290, opened 2022-10-03
VZ-7190: Add test to verify the effect of changing heap memory settings for OpenSearch pods(port)

lake-of-dreams:
- Verifies that the container command line contains the same heap settings as set by VMO in OPENSEARCH_JAVA_OPTS

**PS:** The PR will be updated for the VMO image after https://github.com/verrazzano/verrazzano-monitoring-operator/pull/259 is merged

port of https://github.com/verrazzano/verrazzano/pull/4289

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -227,8 +227,8 @@
           "name": "verrazzano-monitoring-operator",
           "images": [
             {
-              "image": "verrazzano-monitoring-operator",
-              "tag": "v1.4.0-20220919132427-9043384",
+              "image": "verrazzano-monitoring-operator-jenkins",
+              "tag": "v1.4.0-20220930101127-50cfb85",
               "helmFullImageKey": "monitoringOperator.imageName",
               "helmTagKey": "monitoringOperator.imageVersion"
             },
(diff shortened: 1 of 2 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -227,8 +227,8 @@
           "name": "verrazzano-monitoring-operator",
           "images": [
             {
-              "image": "verrazzano-monitoring-operator",
-              "tag": "v1.4.0-20220919132427-9043384",
+              "image": "verrazzano-monitoring-operator-jenkins",
+              "tag": "v1.4.0-20220930101127-50cfb85",
```
michael-cico, line 231: It looks like the VMO repo has a release-1.4 branch.  Should we be backporting the fix from the VMO repo master to that branch there, and then use that operator here?  

I'm not sure if we follow the same branching scheme for the other repos as we do in the main VZ repo.

lake-of-dreams: updated with image created from `release-1.4` branch of vmo(with fix)

michael-cico (approved):

lake-of-dreams:
seems to be already merged in https://github.com/verrazzano/verrazzano/pull/4315

PR #4321, opened 2022-10-06
Add option to run Kind tests using ha.yaml

mgianatagh:
Add an option to the Kind test pipeline to install using ha.yaml.

Modify the Periodics pipeline to run the Grafana DB tests with HA enabled.


ci/scripts/prepare_ha_test_at_environment.sh
```diff
@@ -0,0 +1,35 @@
...
+  exit 1
+fi
+INSTALL_CONFIG_FILE="$1"
+
+if [ ! -f "$INSTALL_CONFIG_FILE" ]; then
+  echo "The Verrazzano install file $INSTALL_CONFIG_FILE does not exist"
+  exit 1
+fi
+
+# Update the install configuration to include what the end-to-end tests require
```
pfmackin, line 25: Why don't we have a verrazzano YAML with these settings

pfmackin (dismissed):

ci/scripts/prepare_ha_test_at_environment.sh
mgianatagh: We could, but my concern is that anytime the ha.yaml changes, we would need to remember to change it in two places.

PR #4338, opened 2022-10-10
0 replica case fix for grafana component : t0

absBitm:
Fix : Grafana component never be ready and install/upgrade will never finish during replica count is 0 .


pkg/k8s/status/deployment_ready.go
```diff
@@ -82,6 +82,9 @@ func DoDeploymentsExist(log vzlog.VerrazzanoLogger, client clipkg.Client, namesp
 }
 
 func deploymentFullyReady(log vzlog.VerrazzanoLogger, client clipkg.Client, deployment *appsv1.Deployment, namespacedName types.NamespacedName, expectedReplicas int32, prefix string) bool {
+	if namespacedName.Namespace == constants.VerrazzanoSystemNamespace && namespacedName.Name == constants.GrafanaIngress && *deployment.Spec.Replicas == 0 {
```
anajosep, line 85: Do we need this check ? since you have already added a check in `isGrafanaReady` method.

anders-swanson, line 85: Should not be changed here

platform-operator/controllers/verrazzano/component/grafana/grafana.go
```diff
@@ -24,6 +24,9 @@ func isGrafanaInstalled(ctx spi.ComponentContext) bool {
 // isGrafanaReady checks that the deployment has the minimum number of replicas available and
 // that the admin secret is ready
 func isGrafanaReady(ctx spi.ComponentContext) bool {
+	if *ctx.EffectiveCR().Spec.Components.Grafana.Replicas == 0 {
```
anders-swanson, line 27: nil dereference

anders-swanson (changes requested):
see comments, also consider we must fix this generically. if 0 replicas are requested for the deployment, it does not make sense to wait for it to be ready. 
We can check the spec of the deployment to find out how many replicas there should be, and wait for at most 1 replica.

pkg/k8s/status/deployment_ready.go
absBitm: this method could be invoked from multiple components, I guess right now we should focus on Grafana fix and make this change generic later on

absBitm:
@anders-swanson 

Let Say, N no. of replicas is requested from the vz grafana CR

- case-1 : if N<1 ; return true from isGrafanaReady (skip waiting for pods=expected_replica to be ready) 
- case-2 : if N>=1 ; go ahead to check deployments have the minimum number of specified replicas (expectedReplicas) ready and available

Here in case-2, 
A) should we check deployment has expectedReplicas = N no of replicas followed by its number of pods must match N
B) or should we check deployment has expectedReplicas >= 1 replica followed by its number of pods must match 1 atleast

platform-operator/controllers/verrazzano/component/grafana/grafana.go
absBitm: corrected

platform-operator/controllers/verrazzano/component/grafana/grafana.go
```diff
@@ -24,9 +24,17 @@ func isGrafanaInstalled(ctx spi.ComponentContext) bool {
 // isGrafanaReady checks that the deployment has the minimum number of replicas available and
 // that the admin secret is ready
 func isGrafanaReady(ctx spi.ComponentContext) bool {
+	var expectedReplicas int32 = 1
+	if ctx.EffectiveCR().Spec.Components.Grafana != nil && ctx.EffectiveCR().Spec.Components.Grafana.Replicas != nil {
+		if *ctx.EffectiveCR().Spec.Components.Grafana.Replicas < 1 {
+			return true
+		}
+		expectedReplicas = *ctx.EffectiveCR().Spec.Components.Grafana.Replicas
```
anders-swanson, line 32: Should use 1 as expected replicas if not nil

absBitm: please check now

anders-swanson (approved):
lgtm

PR #4348, opened 2022-10-11
VZ-7349 fix dashboards

mbenmclean:
This change updates the dashboard json files to fix some of the Helidon and Weblogic panels. It also does so me reformatting of the Weblogic dashboard to make it easier to read.


david-cabelus-oracle (commented):
Looks good to me.

PR #4355, opened 2022-10-11
VZ-7267 make install and reconcile a state machine

mbenmclean:
This change converts the reconcileComponents function and the install and update process into an in-memory state machine, similar to how upgrade and uninstall function. This consists of global reconcile state machine, and then a component install state machine. Because of the nature of how reconciling and install works, these state machines are dependent on the VZ State and component state values, but decreases the coupling between the install process and the VZ CR.


platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -45,7 +46,7 @@ func (d opensearchDashboardsComponent) ShouldInstallBeforeUpgrade() bool {
 
 // GetDependencies returns the dependencies of the OpenSearch-Dashbaords component
 func (d opensearchDashboardsComponent) GetDependencies() []string {
-	return []string{vmo.ComponentName}
+	return []string{vmo.ComponentName, opensearch.ComponentName}
```
michael-cico, line 49: Does adding this fix something?  It does kind of make sense, but if OSD is coming up OK without this I'd say leave it, rather than slow it down by waiting for OS which can take a long time.

@pfmackin what do you think?

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,168 @@
...
+type installTracker struct {
+	vzState reconcileState
+	gen     int64
+	compMap map[string]*componentInstallContext
+}
+
+// installTrackerMap has a map of InstallTrackers, one entry per Verrazzano CR resource generation
+var installTrackerMap = make(map[string]*installTracker)
+
+// getUpgradeTrackerKey gets the tracker key for the Verrazzano resource
```
michael-cico, line 47: comment has wrong func name

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,168 @@
...
-				compLog.Oncef("Skipping update for component %s, monitorChanges set to false", comp.Name())
-			} else {
-				oldState := componentStatus.State
-				oldGen := componentStatus.ReconcilingGeneration
-				componentStatus.ReconcilingGeneration = 0
-				if err := r.updateComponentStatus(compContext, "PreInstall started", vzapi.CondPreInstall); err != nil {
+		// reconcileWatchedComponents reconciles first to fix up any broken components
+		case reconcileWatchedComponents:
+			if spiCtx.ActualCR().Status.State != vzapi.VzStateUpgrading {
+				if err := r.reconcileWatchedComponents(spiCtx); err != nil {
```
michael-cico, line 98: Is `err` always an error we want to propagate?   In some cases we were using the error as a requeue signal, but not always propagating it (so for normal/requeue types of errors we don't get a stack trace).

Just wondering if we might have to do something to avoid emitting stack traces for known errors/situations.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,168 @@
...
+func checkGenerationUpdated(spiCtx spi.ComponentContext) bool {
+	for _, comp := range registry.GetComponents() {
+		componentStatus, ok := spiCtx.ActualCR().Status.Components[comp.Name()]
+		if !ok {
+			spiCtx.Log().Debugf("Did not find status details in map for component %s", comp.Name())
+			// if we can't find the component status, enter install loop to try to fix it
+			return true
+		}
+		if checkConfigUpdated(spiCtx, componentStatus, comp.Name()) &&
+			comp.IsEnabled(spiCtx.EffectiveCR()) &&
```
michael-cico, line 147: Nitpick, maybe have `IsEnabled` as the first check which will short-circuit the other checks.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: I saw a failure in install because of OSD. I think I then realized it was an OS issue. I think I can actually remove that but I'll test it

platform-operator/controllers/verrazzano/install.go
mbenmclean: I could log the error here and return nil and a requeue? Would that be better. I'm not sure how we do it in other places.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
mbenmclean: just did an install without that dependency and it looks good.

platform-operator/controllers/verrazzano/install.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/install.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
+type installTracker struct {
+	vzState reconcileState
+	gen     int64
+	compMap map[string]*componentInstallContext
+}
+
+// installTrackerMap has a map of InstallTrackers, one entry per Verrazzano CR resource generation
+var installTrackerMap = make(map[string]*installTracker)
+
+// getInstallTrackerKey gets the tracker key for the Verrazzano resource
```
michael-cico, line 47: So the comments indicate that we key the map based on the generation, but it looks instead like it's keyed on the UID of the CR, which will not change with respect to the generation right?  

Not suggesting we change anything about the impl, but it doesn't seem to line up with the comments.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
+func getInstallTrackerKey(cr *vzapi.Verrazzano) string {
+	return fmt.Sprintf("%s-%s-%s", cr.Namespace, cr.Name, string(cr.UID))
```
michael-cico, line 49: So, by keying off this combo,  we're saying that we only have active generation being reconciled right?  If we get a new update while one was in progress, we will start over right?  I believe that was the behavior before this?

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+func (r *Reconciler) installComponents(spiCtx spi.ComponentContext, tracker *installTracker, preUpgrade bool) (ctrl.Result, error) {
+	spiCtx.Log().Progress("Reconciling components for Verrazzano installation")
+
+	var requeue bool
+
+	// Loop through all of the Verrazzano components and install each one
+	for _, comp := range registry.GetComponents() {
+		installContext := tracker.getComponentInstallContext(comp.Name())
+		result, err := r.installSingleComponent(spiCtx, installContext, comp, preUpgrade)
+		if err != nil || result.Requeue {
```
michael-cico, line 66: OK, this might answer my question from earlier... so, if we get an error back or a requeue indicator, we just requeue.  So it looks like we just depend on the `installSingleComponent` or whatever it calls to log any error and just return a requeue here.  I don't see any path back from this method that will return a non-nil error. 

That's OK, but maybe just means that we should only return a `Result` from this method and not an error.  The contract of the method is that calling it will result in either a requeue or no requeue, meaning it's done reconciling components.

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+
+		case compStatePostInstall:
+			compLog.Oncef("Component %s post-install running", compName)
+			if err := comp.PostInstall(compContext); err != nil {
+				return ctrl.Result{}, err
+			}
+			installContext.state = compStateInstallComplete
+
+		case compStateInstallComplete:
+			if err := r.updateComponentStatus(compContext, "Install complete", vzapi.CondInstallComplete); err != nil {
```
michael-cico, line 200: Not for this pass, but I was thinking we should have `updateComponentStatus` perhaps just record the latest status in the CR, but we only write it once before returning from the reconcile.  That might avoid/minimize the status update errors we see.

The worst case is that we repeat some steps, but I think that's ok in the context of this JIRA.  If do that we should probably do it in a second pass so as not to disrupt this change.

platform-operator/controllers/verrazzano/uninstall.go
```diff
@@ -156,7 +156,7 @@ func (r *Reconciler) reconcileUninstall(log vzlog.VerrazzanoLogger, cr *installv
 
 // getUninstallTracker gets the Uninstall tracker for Verrazzano
 func getUninstallTracker(cr *installv1alpha1.Verrazzano) *UninstallTracker {
-	key := getTrackerKey(cr)
+	key := getInstallTrackerKey(cr)
```
michael-cico, line 159: Nitpick, maybe all the tracker/func terminology should be "Reconcile", e.g., "getReconcileTrackerKey", etc.  Since we're using it for uninstall/updates... from the controller point of view it's just reconciling changes, so I think that's consistent.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
-// reconcileComponents reconciles each component using the following rules:
-//  1. Always requeue until all enabled components have completed installation
-//  2. Don't update the component state until all the work in that state is done, since
-//     that update will cause a state transition
-//  3. Loop through all components before returning, except for the case
-//     where update status fails, in which case we exit the function and requeue
-//     immediately.
+const (
+	// reconcileWatchedComponents is the state where watched components are reconciled
+	reconcileWatchedComponents reconcileState = "watchedComponents"
```
pfmackin, line 17: use the naming convention we have for upgrade and uninstall, with vzState prefix.  Or change it just to state prefix, but it needs to be the same.  Get rid of the reconcile prefixes

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
+	// reconcileVZState is the state where we examine the VZ State and the component generation values and determine what to do
+	reconcileVZState reconcileState = "reconcileVZState"
+
+	// reconcileStartInstall is the state where the VZ Install Started status is written
+	reconcileStartInstall reconcileState = "startInstall"
+
+	// reconcileInstallComponents is the state where the components are being installed
+	reconcileInstallComponents reconcileState = "installComponents"
+
+	// reconcileEnd is the terminal state
```
pfmackin, line 28: Need a state to update status to installed (or reconciled)

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
+	reconcileEnd reconcileState = "reconcileEnd"
+)
+
+// reconcileState identifies the state of a VZ reconcile
+type reconcileState string
+
+// installTracker has the Install context for the Verrazzano Install
+// This tracker keeps an in-memory Install state for Verrazzano and the components that
+// are being Install.
+type installTracker struct {
```
pfmackin, line 38: we can probably make a generic one for all 3 operations to use but that would be a follow up

platform-operator/controllers/verrazzano/install.go
pfmackin: This is the correct key.  The question is when to delete the map entry for the key, if we need to restart, assuming that is what we want to do.

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
+// installTrackerMap has a map of InstallTrackers, one entry per Verrazzano CR resource generation
+var installTrackerMap = make(map[string]*installTracker)
+
+// getInstallTrackerKey gets the tracker key for the Verrazzano resource
+func getInstallTrackerKey(cr *vzapi.Verrazzano) string {
+	return fmt.Sprintf("%s-%s-%s", cr.Namespace, cr.Name, string(cr.UID))
+}
+
+// getInstallTracker gets the install tracker for Verrazzano
+func getInstallTracker(cr *vzapi.Verrazzano) *installTracker {
```
pfmackin, line 53: Again we can probably consolidate this tracker stuff into tracker.go, but not this PR

platform-operator/controllers/verrazzano/install.go
```diff
@@ -4,213 +4,167 @@
...
-//     that update will cause a state transition
-//  3. Loop through all components before returning, except for the case
-//     where update status fails, in which case we exit the function and requeue
-//     immediately.
+const (
+	// reconcileWatchedComponents is the state where watched components are reconciled
+	reconcileWatchedComponents reconcileState = "watchedComponents"
+
+	// reconcileVZState is the state where we examine the VZ State and the component generation values and determine what to do
+	reconcileVZState reconcileState = "reconcileVZState"
```
pfmackin, line 20: Can you use a more descriptive name for the state, this is really generic

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+			if !preUpgrade && !isVersionOk(compLog, comp.GetMinVerrazzanoVersion(), spiCtx.ActualCR().Status.Version) {
+				// User needs to do upgrade before this component can be installed
+				compLog.Progressf("Component %s cannot be installed until Verrazzano is upgraded to at least version %s",
+					comp.Name(), comp.GetMinVerrazzanoVersion())
+				installContext.state = compStateInstallEnd
+				continue
+			}
+			installContext.state = compStateInstallStarted
+
+		case compStateInstallReady:
```
pfmackin, line 137: please move this to the next state after compStateWaitReady

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+			if preUpgrade && !comp.ShouldInstallBeforeUpgrade() {
+				installContext.state = compStateInstallEnd
+				continue
+			}
+
+			// Determine the state of the component to know what to do with it
+			switch componentStatus.State {
+			case vzapi.CompStateDisabled:
+				installContext.state = compStateInstallDisabled
+			case vzapi.CompStatePreInstalling:
```
pfmackin, line 113: Why are we looking at the component state here?  Shouldn't we be following the in-memory state machine?  Is this a VPO restart optimization?

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+				if !comp.MonitorOverrides(compContext) && comp.IsEnabled(spiCtx.EffectiveCR()) {
+					compLog.Oncef("Skipping update for component %s, monitorChanges set to false", comp.Name())
+				} else {
+					installContext.state = compStateInstallStarted
+					continue
+				}
+			}
+			installContext.state = compStateInstallEnd
+
+		case compStateInstallStarted:
```
pfmackin, line 154: This should be installStarting

pfmackin, line 154: Wondering if installStarting is clearer.  When i saw this name, i thought that we had done the pre-install and the install (helm install)

platform-operator/controllers/verrazzano/install_component.go
pfmackin: You bring up an overall problem that we have with the 2 return params.  Sometimes we use and error to indicate requeue and sometimes we return nil error and but result. requeue = true.  We have this in other places, maybe we need a JIRA just for cleaning that up.

platform-operator/controllers/verrazzano/install_component.go
```diff
@@ -0,0 +1,256 @@
...
+		installContext.state = compStateInstallEnd
+	}
+
+	for installContext.state != compStateInstallEnd {
+		switch installContext.state {
+		case compStateInstallInit:
+
+			compLog.Debugf("Component %s is being reconciled", compName)
+
+			if !comp.IsOperatorInstallSupported() {
```
pfmackin, line 97: should we get rid if this since we only do operator based install?

platform-operator/controllers/verrazzano/install.go
mbenmclean: I think once we reach the global install end we delete it. Otherwise we won't be able to reconcile watched components

platform-operator/controllers/verrazzano/install_component.go
michael-cico: I think that's a problem at the component level to be sure, but in these top-level methods in the controller they can implement a convention between them.

Normalizing error/requeue conditions across the components is a bigger deal than what I'm talking about.  I just meant for this particular method, called from the top-level `reconcileComponents`, we don't need to return both.  Just whether or not to requeue, but we can discuss that.

platform-operator/controllers/verrazzano/install.go
mbenmclean: VzStateReconcile

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: Yes probably. There was talk we might repurpose that spi method though @michael-cico did we end up doing that?

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: Follow on work.

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: This is correct where it is. I updated the name though to make it more clear what it is

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: Changed the name to show that this is about writing status

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: in follow-on work

platform-operator/controllers/verrazzano/uninstall.go
mbenmclean: This is a typo, this is in uninstall

mbenmclean: combined all the getTrackerKey functions into one.

platform-operator/controllers/verrazzano/install.go
mbenmclean: fixed comment

platform-operator/controllers/verrazzano/install_component.go
mbenmclean: tracking this as follow-on work

platform-operator/controllers/verrazzano/install.go
michael-cico: I think the main thing would be to look at the logs to see if we have any regressions in dumping stack traces for errors that are "normal" or expected, i.e., ingress IP not yet available, etc.  Not critical stuff to be sure, but it will jump out at you.  But you can fix any of those on a follow-up PR.

michael-cico (approved):

pfmackin (approved):

PR #4359, opened 2022-10-12
VZ-7330: Misspell linter

SukruthBhat:
Added misspell linter to the list of enabled linters.
Ran make check command locally and corrected the misspelled words.

Jenkins build was failing failing in the System component metrics stage. After probing I figured out that the failure was because vmoTimestampMetric, which is one of the constants for sample metrics of system components had it's value changed from "vmo_configmap_last_succesful_timestamp" to "vmo_configmap_last_successful_timestamp" as successful was reported as a spelling error by misspell linter.

In order to solve this, I have commented out misspell linter in golangci yaml file and also kept the constant value as it is.


.golangci.yaml
```diff
@@ -20,6 +20,7 @@ linters:
     - unused
     - staticcheck
     - stylecheck
+#    - misspell
```
adiforluls, line 23: Why is this commented out? You aren’t running the misspell checker if you're doing that.

michael-cico, line 23: do we not intend to run this on every linter run?

michael-cico: Never mind, just re-read the PR description

michael-cico: Do we intend to fix the metrics naming and the corresponding test in another pass, or just leave it?

adiforluls: If this is an internal metric then this should get fixed in this pass only, the scope of this task is to enable the linter so we can't leave that out.

theabmitra, line 23: read the entire description , i think we need to uncomment this settings like this 
misspell:
     ignore-words:
      - someword

SukruthBhat: I have changed the linter settings of misspell, now succesful misspell will be ignored.

adiforluls (approved):
Approved.

PR #4368, opened 2022-10-13
VZ-7360: Mutating webhook for mysql backup jobs

theabmitra:
Develop a mutating webhook which adds istio-annotations to backup job when it is invoked by mysql-operator so that the job can communicate to etcd. This also applies to cronjobs


jmaron99:
I feel this probably belongs in the VPO rather than VAO.  This sort of platform related processing seems more appropriate there.

theabmitra:
Closing this PR as this needs to be moved in VPO

PR #4374, opened 2022-10-13
VZ-7127 Separate VPO Webhooks into separate deployment

bryantidd:
Separate VPO webhooks into their own deployment so that they can be scaled and resiliency provided


platform-operator/main.go
```diff
@@ -173,19 +173,65 @@ func main() {
...
+				os.Exit(1)
+			}
+			if err = (&installv1beta1.Verrazzano{}).SetupWebhookWithManager(mgr, log); err != nil {
+				log.Errorf("Failed to setup install.v1beta1.Verrazzano webhook with manager: %v", err)
+				os.Exit(1)
+			}
+			mgr.GetWebhookServer().CertDir = config.CertDir
+		}
+		// Set up the validation webhook for VMC
+		if config.WebhooksEnabled {
```
bobdonat, line 218: No need for this if.  It can be included in previous if statement

platform-operator/main.go
```diff
@@ -5,6 +5,7 @@ package main
 
 import (
 	"flag"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/validator"
```
bobdonat, line 8: Is this same main.go being used in both the VPO deployment and webhook deployment

bobdonat: If that is the case, it appears that a bunch of code is run that is not needed.

bryantidd: yes, I would like to improve this in a followup PR.

platform-operator/internal/k8s/certificate/certificate.go
```diff
@@ -162,8 +163,18 @@ func writeFile(filepath string, pem *bytes.Buffer) error {
 
 // UpdateValidatingnWebhookConfiguration sets the CABundle
 func UpdateValidatingnWebhookConfiguration(kubeClient kubernetes.Interface, caCert *bytes.Buffer) error {
+	var oldValidatingWebhook *adminv1.ValidatingWebhookConfiguration
+	oldValidatingWebhook, err := kubeClient.AdmissionregistrationV1().ValidatingWebhookConfigurations().Get(context.TODO(), OldOperatorName, metav1.GetOptions{})
+	if err == nil {
```
bobdonat, line 168: This is type of thing we normally do in the upgrade code. Seems to be following a different pattern here.

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -22,37 +22,20 @@ spec:
...
-              httpHeaders:
-                - name: Content-Type
-                  value: application/json
-            periodSeconds: 5
-            timeoutSeconds: 3
+            exec:
+              command:
+                - /verrazzano/platform-operator/scripts/wait4webhook.sh
+            periodSeconds: 10
+            timeoutSeconds: 30
```
pfmackin, line 38: We want the probe to retry frequently, maybe every 5 secs.  Also, the webhook might take up to 2 min to start so this has to handle that.

platform-operator/internal/k8s/netpolicy/netpolicy.go
```diff
@@ -32,25 +33,37 @@ const (
...
+	netPolicies := newNetworkPolicies(ip, port)
+	for _, netPolicy := range netPolicies {
+		objKey := &netv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: netPolicy.ObjectMeta.Name, Namespace: netPolicy.ObjectMeta.Namespace}}
 
-	opResult, err := controllerutil.CreateOrUpdate(context.TODO(), client, objKey, func() error {
-		netPolicy.Spec.DeepCopyInto(&objKey.Spec)
-		return nil
-	})
+		opResult, err := controllerutil.CreateOrUpdate(context.TODO(), client, objKey, func() error {
+			netPolicy.Spec.DeepCopyInto(&objKey.Spec)
```
pfmackin, line 56: This is not the pattern that we normally use for CreateOrUpdate.  Usually we just provide the object key and set the object data in the callback vs setting the object data before the CreateOrUpdate and doing a deep copy.  This seems like it is ok though.

platform-operator/main.go
```diff
@@ -173,19 +173,65 @@ func main() {
...
+		if err != nil {
+			log.Errorf("Failed to create a controller-runtime manager: %v", err)
+			os.Exit(1)
+		}
+
+		installv1alpha1.SetComponentValidator(validator.ComponentValidatorImpl{})
+		installv1beta1.SetComponentValidator(validator.ComponentValidatorImpl{})
+
+		// Setup the validation webhook
+		if config.WebhooksEnabled {
```
pfmackin, line 205: All the webhook logic should be put in its own function.  Please do this as a followup

pfmackin (approved):

PR #4379, opened 2022-10-13
VZ-7127 Secondary cleanup PR

bryantidd:
Cleanup code and add clarity to concerns

### DIFF
--- a/platform-operator/main.go
+++ b/platform-operator/main.go
@@ -142,8 +142,7 @@ func main() {
 }
 
 func startWebhookServers(config internalconfig.OperatorConfig, log *zap.SugaredLogger) {
-	// initWebhooks flag is set when called from an initContainer.  This allows the certs to be setup for the
-	// validatingWebhookConfiguration resource before the operator container runs.
+	// initWebhooks flag is set when called from an webhook deployment.  This allows separation of webhooks and operator
 
 	log.Debug("Creating certificates used by webhooks")
 	caCert, err := certificate.CreateWebhookCertificates(config.CertDir)
--- a/platform-operator/main.go
+++ b/platform-operator/main.go
@@ -88,8 +88,6 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&config.CertDir, "cert-dir", config.CertDir, "The directory containing tls.crt and tls.key.")
-	flag.BoolVar(&config.WebhooksEnabled, "enable-webhooks", config.WebhooksEnabled,
-		"Enable webhooks for the operator")
 	flag.BoolVar(&config.DryRun, "dry-run", config.DryRun, "Run operator in dry run mode.")
 	flag.BoolVar(&config.WebhookValidationEnabled, "enable-webhook-validation", config.WebhookValidationEnabled,
 		"Enable webhooks validation for the operator")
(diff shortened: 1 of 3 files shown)


platform-operator/main.go
```diff
@@ -135,105 +135,111 @@ func main() {
 	// initWebhooks flag is set when called from an initContainer.  This allows the certs to be setup for the
```
jmaron99, line 135: comment appears to be out of date

platform-operator/main.go
```diff
@@ -135,105 +135,111 @@ func main() {
...
 	// validatingWebhookConfiguration resource before the operator container runs.
 	if config.InitWebhooks {
```
jmaron99, line 137: also looks like config.WebhooksEnabled is no longer used?

platform-operator/main.go
```diff
@@ -135,105 +135,110 @@ func main() {
...
-		log.Debug("Creating certificates used by webhooks")
-		caCert, err := certificate.CreateWebhookCertificates(config.CertDir)
```
jmaron99, line 139: I'll need to modify this method to allow for the update of my webhook with the cert config.  I'll add "webhookName" as an argument.

jmaron99 (approved):

platform-operator/main.go
```diff
@@ -135,105 +135,110 @@ func main() {
...
-			os.Exit(1)
-		}
+		startWebhookServers(config, log)
+	} else {
+		reconcilePlatformOperator(config, log)
+	}
+}
+
+func startWebhookServers(config internalconfig.OperatorConfig, log *zap.SugaredLogger) {
+	// initWebhooks flag is set when called from an webhook deployment.  This allows separation of webhooks and operator
```
bobdonat, line 145: comment looks to out of place

bryantidd: I'll move it up, I fixed the comment content, but didn't move it to the right place now

platform-operator/main.go
bobdonat: I noticed the same thing.  Prior to this we allowed the webhooks to be initialized but not be enabled.  Not sure why that was needed. If we no longer want that then the flag should be removed.

platform-operator/main.go
```diff
@@ -132,108 +132,111 @@ func main() {
...
-			os.Exit(1)
-		}
+		startWebhookServers(config, log)
+	} else {
+		reconcilePlatformOperator(config, log)
+	}
+}
+
+func startWebhookServers(config internalconfig.OperatorConfig, log *zap.SugaredLogger) {
+
```
pfmackin, line 144: Can you remove the blank line

platform-operator/main.go
```diff
@@ -132,108 +132,111 @@ func main() {
...
-		if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-			log.Errorf("Failed starting controller-runtime manager: %v", err)
-			os.Exit(1)
-		}
+		startWebhookServers(config, log)
+	} else {
+		reconcilePlatformOperator(config, log)
+	}
+}
+
```
pfmackin, line 142: Can you add a function description explaining that this code only runs in the VPO-Webhook, not the VPO.  Might be confusing to someone looking at this.

platform-operator/main.go
bryantidd: Will do

platform-operator/main.go
bryantidd: yes

platform-operator/main.go
bryantidd: Yes

pfmackin (approved):

PR #4389, opened 2022-10-14
VZ-6932 Webhook to validate DNS config: part deux

adiforluls:
PR to add a webhook to validate whether the specified dns config or environment name will result in internal hostnames longer than 64 characters. Hostnames longer than 64 characters cause a problem in cert generation.
As compared to previous change, the webhook has been moved to certmanager component code along with UTs. 

Additionally, after discussing with Mike G and Kevin, we aren't going to block an update/upgrade with the webhook if the dns config did not change.


platform-operator/controllers/verrazzano/component/certmanager/certmanager.go
```diff
@@ -946,3 +950,57 @@ func GetOverrides(object runtime.Object) interface{} {
...
+		return cr.Spec.EnvironmentName
+	}
+	cr := effectiveCR.(*v1beta1.Verrazzano)
+	return cr.Spec.EnvironmentName
+}
+
+func getDNSSuffix(effectiveCR runtime.Object) (string, bool) {
+	dnsSuffix, wildcard := "0.0.0.0", true
+	if cr, ok := effectiveCR.(*vzapi.Verrazzano); ok {
+		if cr.Spec.Components.DNS == nil || cr.Spec.Components.DNS.Wildcard != nil {
```
bobdonat, line 983: This code can be moved to fa unction since it repeated below

bobdonat: Maybe not.  Looks like they are using different versions of the cr

bobdonat (approved):

mbenmclean (approved):

PR #4392, opened 2022-10-14
Reduce the number of times hack/update-codegen.sh is executed

bobdonat:
This pull request reduces the number of times hack/update-codegen.sh is executed when using the make target manfests.  The end result is that the make target runs faster and the clientsets have a different package name.  Updated the tests to use the new package name.


tests/e2e/pkg/metrics.go
```diff
@@ -10,7 +10,7 @@ import (
 	"net/http"
 
 	promoperapi "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
-	vaoClient "github.com/verrazzano/verrazzano/application-operator/clients/app/clientset/versioned"
+	vaoClient "github.com/verrazzano/verrazzano/application-operator/clientset/versioned"
```
pfmackin, line 13: How do we distinguish between the client sets of the multiple apis?  Are they all merged into a single client set?   Same question for VPO

pfmackin:
Does this require release note?

bobdonat:
> Does this require release note?

Possibly, we don't document these but I guess someone could find them and use them.

tests/e2e/pkg/metrics.go
bobdonat: Yes, all VAO apis are merged into one clientset.  Same with VPO.

pfmackin (approved):

PR #4394, opened 2022-10-14
VZ 4481

anders-swanson:


platform-operator/controllers/availability/controller.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Start starts the Controller if it is not already running.
+// It is safe to call Start multiple times, additional goroutines will not be created
+func (c *Controller) Start() {
+	if c.shutdown != nil {
+		// already running, so nothing to do
+		return
+	}
+	c.shutdown = make(chan int)
```
pfmackin, line 40: need comments on why you need this go routine and what it is doing.

anders-swanson: Updated comments in the availability controller, is it clearer now?

PR #4396, opened 2022-10-14
VZ-7053: Workaround Rancher upgrade issues

fredtibbitts:
We are seeing multiple issues when upgrading Rancher in Verrazzano. One issue manifests with errors such as this in the Rancher pod logs:
```
available chart version (100.0.3+up0.3.9) for fleet is less than the min version (100.0.5+up0.3.11)
```
Another issue appears when upgrading from 1.2.2 where the old Fleet Helm release is deleted after the new Fleet release, resulting in cluster resources (cluster roles for example) being deleted which causes RBAC failures.

The cause of at least some of the problems is that the ClusterRepo resources for the system charts does not get updated properly on upgrade. As a result the new Rancher pods use stale chart data.

The work around is to do the following in Rancher PreUpgrade:
1. Scale down the Rancher deployment
2. Delete ClusterRepo resources

Then when the Rancher upgrade proceeds, it will recreate the ClusterRepo objects with the new chart data. Scaling down the deployment is unfortunately necessary, otherwise the ClusterRepo may get sync'ed with an old Rancher pod and it will contain the commit for the old release. Then when the new Rancher pods come up, they sync in the opposite direction and set the bundled chart commit to match the old commit from the ClusterRepo.


platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -269,161 +269,112 @@ func isRancherReady(ctx spi.ComponentContext) bool {
...
 	}
 
-	// If any pods were restarted, return an error so that the IsReady check will not continue
-	// any further.  Checks will resume again after the pod is ready again.
-	if podsRestarted {
-		return fmt.Errorf("Rancher IsReady: pods were restarted, waiting for them to be ready again")
+	// get the deployment and check the status to see if all replicas are down, if not we'll reconcile again
+	// until all replicas are down
+	if err := c.Get(context.TODO(), namespacedName, &deployment); err != nil {
+		return client.IgnoreNotFound(err)
```
davisgomes, line 308: Could this be accomplished by inducing another reconcile? In your testing, has this ever succeeded or have we had to reconcile regardless. Maybe we could return a retryable error as the default and only return nil when the first `deployment.Status.AvailableReplicas == 0` returns true (as you have in the beginning)

fredtibbitts: Yes I think that would work. I'd also have to add a check to only update the deployment if the replicas is not already zero, but what you're suggesting should work.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -208,7 +208,15 @@ func (r rancherComponent) PreInstall(ctx spi.ComponentContext) error {
 	return nil
 }
 
-//Install
+// PreUpgrade
+/* Runs pre-upgrade steps
+- Scales down Rancher pods and deletes the ClusterRepo resources to work around Rancher upgrade issues (VZ-7053)
+*/
+func (r rancherComponent) PreUpgrade(ctx spi.ComponentContext) error {
+	return chartsNotUpdatedWorkaround(ctx)
```
davisgomes, line 216: Does this now happen for all upgrades to 1.5.X, regardless of the from version? I know it'd be ugly to check, but we might want to phase this out at some point.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -208,7 +208,15 @@ func (r rancherComponent) PreInstall(ctx spi.ComponentContext) error {
 	return nil
 }
 
-//Install
+// PreUpgrade
+/* Runs pre-upgrade steps
+- Scales down Rancher pods and deletes the ClusterRepo resources to work around Rancher upgrade issues (VZ-7053)
```
davisgomes, line 213: I assume including VZ-7053 is okay in this case.

platform-operator/controllers/verrazzano/component/rancher/rancher_test.go
```diff
@@ -303,3 +309,93 @@ func TestGetRancherHostnameNotFound(t *testing.T) {
 	_, err := getRancherHostname(fake.NewClientBuilder().WithScheme(getScheme()).Build(), &vzapi.Verrazzano{})
 	assert.NotNil(t, err)
 }
+
+// TestChartsNotUpdatedWorkaround tests the chartsNotUpdatedWorkaround function
+// GIVEN an existing Rancher installation
+//
+//	WHEN chartsNotUpdatedWorkaround is called
+//	THEN the Rancher deployment has been scaled down and the ClusterRepo resources for system charts are deleted
```
davisgomes, line 317: Nitpicky, but is this comment formatting correct?

platform-operator/controllers/verrazzano/upgrade_component.go
```diff
@@ -96,7 +96,6 @@ func (r *Reconciler) upgradeSingleComponent(spiCtx spi.ComponentContext, upgrade
 		case compStatePreUpgrade:
 			compLog.Oncef("Component %s pre-upgrade running", compName)
 			if err := comp.PreUpgrade(compContext); err != nil {
-				compLog.Errorf("Failed pre-upgrading component %s: %v", compName, err)
```
davisgomes, line 99: Why is this comment removed?

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
fredtibbitts: I don't know, I haven't tested upgrading from all possible versions. Also this fix will need to get backported to the 1.4 line.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
fredtibbitts: That was already included in other comments, so I didn't really think about it, but I think it's ok to include it. There are other places in the code doing the same thing so there's precedent at least.

platform-operator/controllers/verrazzano/component/rancher/rancher_test.go
fredtibbitts: I followed the same pattern that was used for all of the other test comments. It does look weird, but it's consistent and it seems to have something to do with gofmt in 1.19.

platform-operator/controllers/verrazzano/upgrade_component.go
fredtibbitts: It's not a comment, but I removed it because when I returned a `RetryableError` from `PreUpgrade` due to waiting for the deployment to scale down, it dumped a big ugly stack trace. Also other similar functions (`PreInstall`, etc.) do not have this log statement; they just return the error.

platform-operator/controllers/verrazzano/component/rancher/rancher.go
fredtibbitts: OK made that update.

davisgomes (dismissed):

platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -269,161 +269,107 @@ func isRancherReady(ctx spi.ComponentContext) bool {
...
-	err := c.List(ctx, podList, client.InNamespace(ComponentNamespace), client.MatchingLabels{"app": "rancher"})
-	if err != nil {
+// chartsNotUpdatedWorkaround - workaround for VZ-7053, where some of the Helm charts are not
+// getting updated after Rancher upgrade. This workaround will scale the Rancher deployment down
+// and then delete the ClusterRepo resources. This must be done in PreUpgrade. Scaling down is
+// necessary to prevent an old pod from updating the ClusterRepo with an old commit. When Rancher
+// is upgraded, new Rancher pods will start and create new ClusterRepo resources with the correct
+// chart data.
+func chartsNotUpdatedWorkaround(ctx spi.ComponentContext) error {
+	if err := scaleDownRancherDeployment(ctx.Client(), ctx.Log()); err != nil {
```
pfmackin, line 279: We should only do this if the ClusterRepo resource exists (i think) since this affects HA (Rancher will have downtime).

fredtibbitts: That's why this is called from PreUpgrade. Doesn't PreUpgrade imply an existing installation exists, or can PreUpgrade be called without a previous installation? Rancher can't function without ClusterRepo resources for the system charts so if there's an existing Rancher installation, the ClusterRepos must exist.

davisgomes (approved):

davisgomes (approved):

PR #4400, opened 2022-10-14
VZ-7360: Mutating webhook for mysql backup jobs in VPO

theabmitra:
Develop a mutating webhook which adds istio-annotations to backup job when it is invoked by mysql-operator so that the job can communicate to etcd. This also applies to cronjobs


platform-operator/apis/verrazzano/mysql-webhook/mysql_backup_webhook.go
```diff
@@ -0,0 +1,222 @@
...
+	Defaulters    []MySQLDefaulter
+}
+type MySQLDefaulter interface {
+}
+
+// ConvertAPIVersionToGroupAndVersion splits APIVersion into API and version parts.
+// An APIVersion takes the form api/version (e.g. networking.k8s.io/v1)
+// If the input does not contain a / the group is defaulted to the empty string.
+// apiVersion - The combined api and version to split
+func ConvertAPIVersionToGroupAndVersion(apiVersion string) (string, string) {
```
michael-cico, line 58: Looks like this is duplicated from the VAO code?  Maybe refactor it to a shared location, like pkg/k8sutil

platform-operator/helm_config/charts/verrazzano-application-operator/templates/verrazzano-application-operator.yaml
```diff
@@ -307,6 +308,27 @@ rules:
       - get
       - list
       - update
+  - apiGroups:
```
michael-cico, line 311: Is this supposed to be in the VAO chart?

theabmitra: yes. the pod request is received by that as well.

platform-operator/apis/verrazzano/mysql-webhook/mysql_backup_webhook.go
theabmitra: this can be done in a separate PR since these are being referenced by app operator code. Making that change will cause more diff which wont be specific to this change

michael-cico (approved):

michael-cico (approved):

PR #4418, opened 2022-10-17
Updating development version from 1.4.1 to 1.4.2

stevshap:
Updating development version from 1.4.1 to 1.4.2


fredtibbitts (commented):
I notice the BOM has component versions for some of the components, and those are at "1.4.0" for the Verrazzano components, I'm not even sure where those versions are used but are we supposed to be updating those as well?

stevshap:
The next step after these images are published is to update the BOM (from my interpretation at least)

adiforluls (approved):

PR #4423, opened 2022-10-18
VZ-5880: Refactor tests/e2e/pkg/resource.go to pkg/k8s/resource/resource_util.go

saketmht:
This is a PR for the first step towards implementing the workaround for VZ-5880. The file was moved from tests/e2e/pkg/ to pkg/k8s/resource/ so that it can referenced from the mcagent.go. The file was renamed to resource_util.go and unit tests were added for the file as well.


pkg/k8s/resource/testdata/secret.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
desagar, line 1: Where were these testdata files before? If they were somewhere under the tests/e2e directory, they should not be moved out into the pkg area - this is only for main code and unit tests.

It would be better to put these testdata files back to where they were previously, and have the calling code call FindTestDataFile and pass the correct file.

desagar: Oh, never mind about these files. I see they are only used by unit tests.

But I still do think the file path should be passed in fully to the resource_util functions, instead of computed using FindTestDataFile in those functions.

saketmht: These testdata files are actually for the unit tests that I wrote. I created them. They were not moved from anywhere

tests/e2e/examples/bobsbooks/bobs_books_test.go
```diff
@@ -124,12 +125,12 @@ func deployBobsBooksExample(namespace string) {
 	// Note: creating the app config first to verify that default metrics traits are created properly if the app config exists before the components
 	t.Logs.Info("Create application resources")
 	Eventually(func() error {
-		return pkg.CreateOrUpdateResourceFromFileInGeneratedNamespace("examples/bobs-books/bobs-books-app.yaml", namespace)
+		return resource.CreateOrUpdateResourceFromFileInGeneratedNamespace("examples/bobs-books/bobs-books-app.yaml", namespace)
```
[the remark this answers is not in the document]
saketmht: Okay, I will change all the references accordingly and remove the FindTestDataFile function from the resource_util.go

fredtibbitts (commented):
A general comment about your imports. Imports are divided into at least two sections. The first section always contains the go built-in package imports. Then there is a newline, and then there are one or more sections for all other imports. You have a bunch of files where the newline has been removed and non-built-in packages are included in the first section. Those should be cleaned up. 

Also you can configure your IDE to automatically format imports using `goimports` or `gofmt` so that it will do all of the organizing for you.

pkg/k8s/resource/resource_util.go
```diff
@@ -39,9 +41,10 @@ var nsGvr = schema.GroupVersionResource{
 // This is intended to be equivalent to `kubectl apply`
 // The cluster used is the one set by default in the environment
 func CreateOrUpdateResourceFromFile(file string) error {
+	var logger = vzlog.DefaultLogger()
```
fredtibbitts, line 44: I'm wondering if the logger should instead be passed into these functions. The comment on `vzlog.DefaultLogger` says it's normally used for testing, but now these functions will be used in the product as well as from test code. Do we know that the logger returned from `DefaultLogger` is consistent with the loggers used by product code that will be calling these functions?

pkg/k8s/resource/resource_util_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// Test using environment variable
+	err = os.Setenv(k8sutil.EnvVarTestKubeConfig, kubeConfigPath)
+	asserts.NoError(err)
+
+	logger := vzlog.DefaultLogger().GetZapLogger()
+	err = CreateOrUpdateResourceFromFile(file, logger)
+	asserts.NoError(err)
+
+	// Reset env variable
+	err = os.Setenv(k8sutil.EnvVarTestKubeConfig, prevEnvVar)
```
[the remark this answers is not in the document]
saketmht: I have made the changes accordingly.

fredtibbitts (approved):

PR #4440, opened 2022-10-19
NoJIRA Fix tests to properly use VZ cli for Upgrade

bryantidd:


michael-cico (approved):

ci/chaos/Jenkinsfile
```diff
@@ -323,7 +326,8 @@ pipeline {
 
                     else
                         echo "Upgrading VPO to this commit version"
-                        kubectl apply -f ${WORKSPACE}/upgrade-test-operator.yaml
+                        ${GO_REPO_PATH}/vz upgrade --operator-file ${WORKSPACE}/upgrade-test-operator.yaml --version ${VERRAZZANO_DEV_VERSION}
+                         kubectl wait --timeout=45m --for=condition=UpgradeComplete verrazzano/my-verrazzano
```
bobdonat, line 330: This kubectl wait command is not needed.  You should pass the --timeout on the vz upgrade command.

bobdonat: If the upgrade fails it looks like there will be an additional wait of 45m before timing out

PR #4442, opened 2022-10-19
VZ 7476: Synchronize availability channels

anders-swanson:
- Use bi-directional communication to prevent status channel from blocking
- Ensure updates are made to the VZ CR availability when main Verrazzano reconciler is idle


platform-operator/controllers/verrazzano/health/availability.go
```diff
@@ -75,6 +80,42 @@ func (p *PlatformHealth) newStatus(log vzlog.VerrazzanoLogger, vz *vzapi.Verrazz
...
+	p.status = status
+	// if cluster Verrazzano has identical status, don't send an update
+	if status != nil && !status.needsUpdate() {
+		return nil
+	}
+	select {
+	case <-p.ackChannel:
+		// if the last status has been ack'd, send a new status
+		p.statusChannel <- p.status
+	default:
```
pfmackin, line 93: So you have this thread updating the status if you didn't get an ack from the other thread?

PR #4444, opened 2022-10-19
VZ-7132: Clean Up Old Code in OpenSearch Post-Upgrade

marcogelle:
Deletes old, unused code, namely `updateElasticsearchResources` in opensearch_component.go and all functions that are exclusively called by this. Unit tests that called the deleted functions are accordingly deleted as well.

Note that VZ-7132 was originally to modify `waitForPodsWithReadyContainer`, but the function is never called for more recent versions of Verrazzano. Instead, this PR attempts to remove dead code.


theabmitra (approved):
Approved

PR #4446, opened 2022-10-19
Check Effective CR for Rancher Keycloak integration

mbenmclean:
In the Rancher code, there is a check if Keycloak is enabled before configuration the Keycloak Rancher integration. However, that only looks at the Actual CR. If the install were configured so that Rancher were enabled but Keycloak wasn't this would cause an error. This change that check to look at the effective CR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 Fixes:
 
 - Rancher upgrade intermittently fails with errors stating that the available chart version is less than the minimum chart version for Rancher system charts.
-- Fixed a bug related to when to integate Rancher with Keycloak auth
+- Fixed a bug related to Rancher/Keycloak auth integration for managed cluster profile
 
 ### v1.4.1
 Fixes:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 Fixes:
 
 - Rancher upgrade intermittently fails with errors stating that the available chart version is less than the minimum chart version for Rancher system charts.
+- Fixed a bug related to when to integate Rancher with Keycloak auth
 
 ### v1.4.1
 Fixes:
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -364,7 +364,7 @@ func activateDrivers(log vzlog.VerrazzanoLogger, c client.Client) error {
 // +enables or disables Keycloak Auth provider.
 func configureAuthProviders(ctx spi.ComponentContext) error {
 	log := ctx.Log()
-	if vzconfig.IsKeycloakEnabled(ctx.ActualCR()) && isKeycloakAuthEnabled(ctx.ActualCR()) {
+	if vzconfig.IsKeycloakEnabled(ctx.EffectiveCR()) && isKeycloakAuthEnabled(ctx.EffectiveCR()) {
```
lake-of-dreams, line 367: @mbenmclean may be I don't understand in which case will the existing solution fail, will it fail when someone installs with
```
- rancher
      enabled: true
- keycloak
        enabled: false
```
won't it will work in this case since it reads actual CR values?

mbenmclean: It will fail when someone installs managed-cluster profile with rancher enabled. This is because using Actual CR will think Keycloak is enabled because it is nil but in fact it's disabled because of the managed-cluster profile. I verified this is broken.

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 Fixes:
 
 - Rancher upgrade intermittently fails with errors stating that the available chart version is less than the minimum chart version for Rancher system charts.
+- Fixed a bug related to when to integate Rancher with Keycloak auth
```
lake-of-dreams, line 5: please change to `Fixed a bug related to Rancher/Keycloak auth integration for managed cluster profile` as thats when the bug occurs

lake-of-dreams (approved):

PR #4452, opened 2022-10-20
fallback to tier=mysql lookup

agv426:
fallback to the tier lookup when waiting for mysql pod to restart


fredtibbitts (dismissed):

kminder (approved):
Approved.

PR #4455, opened 2022-10-20
VZ-7481: Add undocumented install/upgrade option to override timeout for VPO

mgianatagh:
Add a VZ command line option to override the hard coded value of 5 minutes that the CLI waits for VPO to start.

Updated the HA test pipeline to use a longer wait timeout.

### DIFF
--- a/tools/vz/cmd/helpers/vpo.go
+++ b/tools/vz/cmd/helpers/vpo.go
@@ -7,6 +7,13 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
 	"github.com/spf13/cobra"
 	vzconstants "github.com/verrazzano/verrazzano/pkg/constants"
 	"github.com/verrazzano/verrazzano/pkg/k8s/ready"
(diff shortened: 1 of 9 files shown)


michael-cico (approved):

tools/vz/cmd/helpers/vpo.go
```diff
@@ -180,14 +172,15 @@ func WaitForPlatformOperator(client clipkg.Client, vzHelper helpers.VZHelper, co
 	// Wait for the verrazzano-platform-operator pod to be found
 	seconds := 0
 	retryCount := 0
+	retryMax := int(timeout.Seconds())
```
pfmackin, line 175: are you missing the number of seconds?

pfmackin, line 175: Is retryMax a retry count or a time value?    It is called timeout.Seconds but it is compared with a retryCount

mgianatagh: The existing logic used a max retry value (in seconds) and performed retries in a loop using an increment value (which is 1 second).  I didn't want to change that logic, so I took the timeout value (which can be in any time unit) and converted it to seconds (on line 175).  

I am going to merge as is for now, because I need this for testing.  But let me know if you want me to rework anything here internally
 in a follow-up PR.

PR #4456, opened 2022-10-20
Callign helm component pre-install/upgrade/uninstall from components …

saidu941:
…even when overrided

### DIFF
--- a/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
+++ b/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
@@ -83,7 +83,10 @@ func (c applicationOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error
 	if err != nil {
 		return err
 	}
-	return labelAnnotateWorkloadDefinitions(ctx.Client())
+	if err := labelAnnotateWorkloadDefinitions(ctx.Client()); err != nil {
+		return err
+	}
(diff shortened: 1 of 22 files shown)


platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
```diff
@@ -85,7 +85,10 @@ func (c applicationOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error
 	if err != nil {
 		return err
 	}
-	return labelAnnotateWorkloadDefinitions(ctx.Client())
+	if err := labelAnnotateWorkloadDefinitions(ctx.Client()); err != nil {
+		return err
+	}
+	return c.HelmComponent.PreUpgrade(ctx)
```
michael-cico, line 91: Looks like the code that does the check/fix in `PreInstall`/`PreUpgrade` in the `HelmComponent` code is largely duplicated.  Looks like the same logic implemented slightly differently.  We should probably consolidate that behind a single func that is called from both places, if that is indeed the case.

It’s not part of these changes, must have been merged before.  that can be investigated/changes as a follow up if you prefer.

michael-cico (approved):

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
davisgomes: Will address as a follow-up.

PR #4459, opened 2022-10-20
VZ-7326: Mysql backup automated tests

theabmitra:
Re-introducing mysql backup tests with KInd


tests/e2e/backup/helpers/models.go
```diff
@@ -429,6 +446,164 @@ type VeleroPodVolumeRestores struct {
 	} `json:"status"`
 }
 
+type MySQLBackupModel struct {
```
michael-cico, line 449: not a blocker, but why do we need these `Model` helpers?

michael-cico (approved):

tests/e2e/backup/helpers/models.go
theabmitra: To parse the o/p. We dont need the entire model, just a few fields, but have kept them in case we want to expand on something later on

michael-cico (approved):

PR #4460, opened 2022-10-20
VZ-7421 loggen

pfmackin:
Added worker manager and setup structure to collect metrics. 
Created common worker runner that will do work in a loop.


tools/psr/backend/config/config.go
```diff
@@ -42,8 +42,8 @@ type EnvVarDesc struct {
 }
 
 type CommonConfig struct {
-	WorkerType           string
-	IterationSleepMillis time.Duration
+	WorkerType          string
+	IterationSleepNanos time.Duration
```
stevshap, line 46: Why switch to nano? ms seems more standard

stevshap (approved):

tools/psr/backend/config/config.go
pfmackin: Because duration is in nanos.  Millis was actually incorrect

stevshap: got it

PR #4476, opened 2022-10-24
VZ-6464 Push Registration Secrets to managed cluster in VMC Reconcile

davisgomes:
This change should automatically push the Multicluster agent and registration secrets to a managed cluster if it has been registered in the Rancher UI.


pkg/test/mockmatchers/matchers.go
```diff
@@ -29,3 +30,25 @@ func (u *URIMatcher) Matches(x interface{}) bool {
...
+type URIMethodMatcher struct {
+	method, path string
+}
+
+func MatchesURIMethod(method, path string) gomock.Matcher {
+	return &URIMethodMatcher{method, path}
+}
+
+func (u *URIMethodMatcher) Matches(x interface{}) bool {
+	if !gomock.AssignableToTypeOf(&http.Request{}).Matches(x) {
```
desagar, line 45: would be nice to be able to write this like:
return MatchesURI(u.path).Matches(req) &&  req.Method == u.Method

But I know the casts would happen multiple times and it's not ideal :-(

platform-operator/apis/clusters/v1alpha1/verrazzanomanagedcluster_types.go
```diff
@@ -39,6 +39,10 @@ const (
 	// ManagedCARetrieved = true means that the managed cluster CA cert has been retrieved and
 	// populated. This is done by the VMC controller via the Rancher API proxy for the managed cluster.
 	ConditionManagedCARetrieved ConditionType = "ManagedCARetrieved"
+
```
desagar, line 42: does this not require a make manifests?

platform-operator/controllers/clusters/push_manifest_objects.go
```diff
@@ -0,0 +1,74 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// pushManifestObjects applies the Verrazzano manifest objects to the managed cluster.
+// To access the managed cluster, we are taking advantage of the Rancher proxy
+func (r *VerrazzanoManagedClusterReconciler) pushManifestObjects(vmc *clusterapi.VerrazzanoManagedCluster) (bool, error) {
+	for _, condition := range vmc.Status.Conditions {
+		if condition.Type == clusterapi.ConditionManifestPushed && condition.Status == corev1.ConditionTrue {
+			// Do nothing if the condition is met in the status
```
desagar, line 18: maybe a Oncef log here would help.

platform-operator/controllers/clusters/push_manifest_objects.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Create or Update the agent and registration secrets
+	agentSecret := corev1.Secret{}
+	agentSecret.Namespace = constants.VerrazzanoSystemNamespace
+	agentSecret.Name = constants.MCAgentSecret
+	regSecret := corev1.Secret{}
+	regSecret.Namespace = constants.VerrazzanoSystemNamespace
+	regSecret.Name = constants.MCRegistrationSecret
+	err = createOrUpdateSecretRancherProxy(&agentSecret, rc, clusterID, func() error {
+		agentSecret, err = r.getSecret(vmc.Namespace, GetAgentSecretName(vmc.Name), true)
```
desagar, line 44: I'm not sure how this works. You are reassigning the pointer variable, so whatever the argument agentSecret is pointing to won't be changed here, right?

platform-operator/controllers/clusters/push_manifest_objects.go
davisgomes: I had a log meesage here, but I moved the location it to when the status is updated. I think the log fits better there, and it should convey the same message with similar timing.

platform-operator/controllers/clusters/push_manifest_objects.go
```diff
@@ -0,0 +1,74 @@
...
+			Namespace: constants.VerrazzanoSystemNamespace,
+			Name:      constants.MCAgentSecret,
+		}
+		return nil
+	}, r.log)
+	if err != nil {
+		return false, err
+	}
+	err = createOrUpdateSecretRancherProxy(&regSecret, rc, clusterID, func() error {
+		regSecret, err = r.getSecret(vmc.Namespace, GetRegistrationSecretName(vmc.Name), true)
```
desagar, line 59: Same pointer q. as above for regSecret here.

platform-operator/apis/clusters/v1alpha1/verrazzanomanagedcluster_types.go
davisgomes: It does not because it is the same object (Condition) with differently populated fields. Nothing about the API for the VMC is changing.

pkg/test/mockmatchers/matchers.go
davisgomes: I think that would be interesting, but since it's a very simple string comparison, I think it's more understandable to keep it separate.

platform-operator/controllers/clusters/push_manifest_objects.go
desagar: looks like the passed in vars are not used in the caller. We should just get rid of the agentSecret parameter I think.

davisgomes: `r.GetSecret` actually returns a `v1.Secret` variable not a `*v1.Secret`. I am reassigning the value of the pointer and not the pointer itself.

platform-operator/controllers/clusters/rancher.go
```diff
@@ -612,3 +620,127 @@ func getProxyURL() string {
...
+		return err
+	}
+	if newKey := client.ObjectKeyFromObject(secret); key != newKey {
+		return log.ErrorfNewErr("MutateFn cannot mutate secret name and/or secret namespace")
+	}
+	return nil
+}
+
+// rancherSecretGet simulates a client get request through the Rancher proxy for secrets
+func rancherSecretGet(secret *corev1.Secret, rc *rancherConfig, clusterID string, log vzlog.VerrazzanoLogger) error {
```
fredtibbitts, line 665: Should these Get, Create, Update functions do a `nil` check on the secret?

platform-operator/controllers/clusters/vmc_controller_test.go
```diff
@@ -1707,6 +1731,20 @@ func expectSyncManifest(t *testing.T, mock *mocks.MockClient, mockStatus *mocks.
 		})
 }
 
+func expectPushManifest(mockRequestSender *mocks.MockRequestSender) {
+	mockRequestSender.EXPECT().
+		Do(gomock.Not(gomock.Nil()), mockmatchers.MatchesURI(clustersPath+"/"+unitTestRancherClusterID)).
+		DoAndReturn(func(httpClient *http.Client, req *http.Request) (*http.Response, error) {
+			var resp *http.Response
+			r := io.NopCloser(bytes.NewReader([]byte(`{"state":"inactive", "agentImage":"test"}`)))
```
fredtibbitts, line 1739: Is this really what comes back in the response when creating/updating resource via the proxy? If it was just copy/paste then you can probably remove it and just return an empty body.

platform-operator/controllers/clusters/rancher.go
davisgomes: They probably should, even if we supply the pointers

PR #4488, opened 2022-10-25
VZ-7467.  Update VPO and webhook deployments to allow webhook to scale

michael-cico:
Update the VPO/Webhook code to allow webhook scaling
- Change the creation of the webhook CA/TLS certs to be stored in secrets in the install namespace using a VPO initContainer
- Have the webhooks use the certs from those secrets when creating/updating their webhook configurations
- Add a `webhookswait` initContainer for the VPO to wait for the webhooks to be alive before starting the other containers
- Remove the startup probe from the VPO
- Add default anti-affinity settings for the VPO and webhook deployments
- Add a. `set -x` in `waitforwebhook.sh` to allow us to see what's going on in the `webhookswait` initContainer


platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -23,11 +23,22 @@ spec:
     spec:
       terminationGracePeriodSeconds: 0
       initContainers:
-        - name: "{{ .Values.name }}-init"
+        - name: initwebhooks
           imagePullPolicy: {{ .Values.imagePullPolicy }}
           image: {{ .Values.image }}
-          command:
-            - /verrazzano/platform-operator/scripts/wait4webhook.sh
+          args:
```
pfmackin, line 29: Does this mean the VPO is doing webhook initialization?  I thought that was all done in the Webhook pod

michael-cico: No generally this PR means that the VPO does the cert creation and storage into secrets in the first initContainer, and the second one waits for the webhooks to be up and running before allowing the VPO pod to continue to come up.

michael-cico: The webhook pod(s) do the work of creating/updating the webhook configurations and deleting the old one.

mgianatagh (approved):

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
michael-cico: talked with @mgianatagh , going to merge this and keep the ticket open to see if it improves the HA test scenarios, and look into using the leader-election approach to make the webhook pods self-contained and not rely on the VPO to manage the certificates.

PR #4502, opened 2022-10-26
VZ-6458: End-to-end test for Rancher cluster and VMC auto sync

fredtibbitts:
This PR adds end-to-end testing of the Rancher cluster auto sync. When a cluster is imported using the Rancher UI or API, we now automatically create a VMC. When a cluster is deleted using the Rancher UI or API, we delete the VMC.

The changes include refactoring to allow code to be re-used for testing.


michael-cico (approved):

tests/e2e/pkg/rancher.go
```diff
@@ -206,3 +220,42 @@ func verifyAuthConfigAttribute(name string, actual interface{}, expected interfa
...
+
+func EventuallyGetRancherHost(log *zap.SugaredLogger, api *APIEndpoint) (string, error) {
+	rancherHost := EventuallyGetIngressHost(log, api, rancher.ComponentNamespace, common.RancherName)
+	if rancherHost == "" {
+		return "", fmt.Errorf("got empty Rancher ingress host")
+	}
+	return rancherHost, nil
+}
+
+func CreateNewRancherConfig(log *zap.SugaredLogger, kubeconfigPath string) (*clusters.RancherConfig, error) {
```
davisgomes, line 232: Could this be done by making the newRancherConfig method public or are they too different?

fredtibbitts: @desagar can comment better than I can, but I think the main issue is the `newRancherConfig` function configures the Rancher client for running inside the cluster, and the tests run outside the cluster. Seems ok to have the separate function for testing even though there is a little duplication there.

davisgomes (approved):

PR #4505, opened 2022-10-26
VZ-7506: Make CLI resilient to streaming console output during install/upgrade

mgianatagh:
During install/upgrade if the connection is lost to streaming console output, the CLI use to just exit.  The install or upgrade continues, but environments like our pipelines think the install has failed because the CLI has exited with an error.

This PR has changes to have the CLI make up to 5 attempts to reconnect.  For example, if the VPO pod recycles during install, the CLI will automatically re-establish the connection to the new VPO pod and continue streaming the console output.


tools/vz/cmd/helpers/vpo.go
```diff
@@ -207,18 +203,42 @@ func WaitForOperationToComplete(client clipkg.Client, kubeClient kubernetes.Inte
...
+					time.Sleep(5 * time.Second)
+					continue
+				}
+				connectRetryCount = 0
+				sc.Split(bufio.ScanLines)
+			}
+
+			scannedOk := sc.Scan()
+			if !scannedOk {
+				fmt.Fprintf(outputStream, "Lost connection to the console output, attempting to reconnect\n")
```
bobdonat, line 230: Should the error be shown here too? I think sc.Err() with return the error.

bobdonat:
Is his same change needed for uninstall

tools/vz/cmd/helpers/vpo.go
mgianatagh: sc.Err was always blank for me.  I guess I could check if it was non-nil and tailor the message based on that.

tools/vz/cmd/helpers/vpo.go
```diff
@@ -207,18 +203,42 @@ func WaitForOperationToComplete(client clipkg.Client, kubeClient kubernetes.Inte
...
+		const maxConnectRetries = 5
+
 		for {
-			sc.Scan()
+			if sc == nil {
+				sc, err = getScanner(client, kubeClient)
+				if err != nil {
+					connectRetryCount++
+					fmt.Fprintf(outputStream, fmt.Sprintf("Failed to connect to the console output, retry %d of %d: %v\n", connectRetryCount, maxConnectRetries, err))
+					if connectRetryCount >= 5 {
```
michael-cico, line 218: the retry count/interval has a max of 25 secs cumulatively, which may not be enough if a pod's moving etc.

tools/vz/cmd/helpers/vpo.go
michael-cico: I think the problem I saw during uninstall also was that the `Scan` call hung until the operation finished or some other timeout kicked in, and there was no error.  So this impl is able to pick back up when the pod is whacked?

tools/vz/cmd/install/install.go
```diff
@@ -220,7 +214,7 @@ func runCmdInstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelper)
 	}
 
 	// Wait for the Verrazzano install to complete
-	return waitForInstallToComplete(client, kubeClient, vzHelper, vpoPodName, types.NamespacedName{Namespace: vzNamespace, Name: vzName}, timeout, logFormat)
+	return waitForInstallToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vzNamespace, Name: vzName}, timeout, logFormat)
```
michael-cico, line 217: Do we need to update the `uninstall` command code also?

stevshap: +1

mgianatagh: I can address uninstall in a follow-up PR

tools/vz/cmd/helpers/vpo.go
mgianatagh: how long do you feel we should wait?  1 or 2 minutes?

michael-cico (approved):

PR #4510, opened 2022-10-26
VZ 7470: Component Availability helpers

anders-swanson:
Implement component availability helpers, if component owned deploy, sts, ds, ing objects are 'available', the component is available. Availability attempts to make use of existing pod readiness gates to keep things simple - ideally a pod's readiness gate should correspond with its availability.

Will follow up with implementations for remaining components, so where will be some minor changes. App operator is implemented as an example.

### DIFF
--- a/pkg/k8s/ready/available.go
+++ b/pkg/k8s/ready/available.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package ready
+
+import (
+	"context"
+	"fmt"
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	appsv1 "k8s.io/api/apps/v1"
+	networkingv1 "k8s.io/api/networking/v1"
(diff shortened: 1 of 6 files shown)


pkg/k8s/ready/available.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return nil
+}
+
+func isDeploymentAvailable(client clipkg.Client, nsn types.NamespacedName) error {
+	deploy := &appsv1.Deployment{}
+	if err := client.Get(context.TODO(), nsn, deploy); err != nil {
+		return handleGetError(err, nsn, typeDeployment)
+	}
+	return handleReplicasNotReady(deploy.Status.ReadyReplicas, deploy.Status.Replicas, nsn, typeDeployment)
```
pfmackin, line 78: so all replicas have to be ready for a component to be available?  Seems like there should partially available state.

michael-cico (dismissed):

pkg/k8s/ready/available.go
anders-swanson: I think you're right, but 'partially' available is a much more nebulous thing to determine.

pfmackin: can you create a followup task to look into this.

pfmackin (approved):

PR #4515, opened 2022-10-27
Change logging for isInnoDBClusterOnline to be less verbose

mgianatagh:
Change logging for isInnoDBClusterOnline to be less verbose


platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -197,7 +197,7 @@ func isMySQLReady(ctx spi.ComponentContext) bool {
 
 // isInnoDBClusterOnline returns true if the InnoDBCluster resource cluster status is online
 func isInnoDBClusterOnline(ctx spi.ComponentContext) bool {
-	ctx.Log().Progress("Waiting for InnoDBCluster to be online")
```
pfmackin, line 200: This should only log once a minute, is that too much?

bobdonat (approved):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
pfmackin, line 200: We don't want to lose this in the normal install/upgrade flow

mgianatagh: this message seems like more of a lower level debug info.  Does the user really want a repeating message that the VPO is checking the mysql db status?  Even if I change this to progress it will come out once an hour.  The rest of the function is only outputting debug info too.

pfmackin (approved):

PR #4517, opened 2022-10-27
Fix Rancher cluster sync logger

fredtibbitts:
The logger in the Rancher cluster sync code was not being set correctly. As a result the log level configured on the VPO (via command line switch) was not honored. I tested by setting the log level to debug on the VPO and confirmed that debug log message appear.

I also changed a Progress message to a Debug message.

### DIFF
--- a/platform-operator/controllers/clusters/sync_rancher_clusters.go
+++ b/platform-operator/controllers/clusters/sync_rancher_clusters.go
@@ -39,7 +39,7 @@ type RancherClusterSyncer struct {
 func (r *RancherClusterSyncer) StartSyncing() {
 	// initialize logger - the AddCallerSkip is needed otherwise the caller in the log message always shows "vzlog.go"
 	l := r.Log.Desugar().WithOptions(zap.AddCallerSkip(2))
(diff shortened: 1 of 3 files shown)


davisgomes (approved):

platform-operator/controllers/clusters/sync_rancher_clusters.go
```diff
@@ -33,11 +32,15 @@ var clustersResponseHash []byte
...
 type RancherClusterSyncer struct {
 	client.Client
+	Log *zap.SugaredLogger
 }
 
 // StartSyncing starts the Rancher cluster synchronization loop
 func (r *RancherClusterSyncer) StartSyncing() {
-	log := r.initLogger()
+	// initialize logger - the AddCallerSkip is needed otherwise the caller in the log message always shows "vzlog.go"
+	l := r.Log.Desugar().WithOptions(zap.AddCallerSkip(2))
```
pfmackin, line 41: why is this needed for Rancher?

desagar (approved):

platform-operator/controllers/clusters/sync_rancher_clusters.go
fredtibbitts: Per conversation, I don't understand why this is needed. Will file a ticket for additional investigation.

PR #4521, opened 2022-10-27
VZ-7431 PSR GetLogs Metrics

mbenmclean:
This change reworks the GetLogs worker and adds metrics to it.


tools/psr/backend/workmanager/runner.go
```diff
@@ -125,7 +125,7 @@ func (r runner) RunWorker(conf config.CommonConfig, log vzlog.VerrazzanoLogger)
 		startIteration := time.Now().UnixNano()
 		err := r.Worker.DoWork(conf, log)
 		if err != nil {
-			log.Errorf("Failed calling %s to do work: %v", r.Worker.GetWorkerDesc().EnvName, err)
```
pfmackin, line 128: I think EnvName makes more sense since this is what the user sees in the override YAML file.  They don't have directly visibility to the MetricsName.  Either way, I don't feel strongly about this the names should be very close.

tools/psr/manifests/charts/worker/templates/ingestauthorizationpolicy.yaml
```diff
@@ -5,7 +5,7 @@
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
-  name: {{ .Release.Namespace }}-ingest-authzpol
+  name: {{ .Release.Name }}-{{ .Release.Namespace }}-ingest-authzpol
```
pfmackin, line 8: you don't need the namespace since it is namespaced scoped.

tools/psr/manifests/charts/worker/templates/masterauthorizationpolicy.yaml
```diff
@@ -5,7 +5,7 @@
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
-  name: {{ .Release.Namespace }}-master-authzpol
+  name: {{ .Release.Name }}-{{ .Release.Namespace }}-master-authzpol
```
pfmackin, line 8: remove the Release.Namespace

pfmackin (dismissed):

tools/psr/backend/workmanager/runner.go
mbenmclean: I'll leave it for now and we can adjust later perhaps

mbenmclean: Need to repush so I changed

stevshap (approved):

PR #4522, opened 2022-10-28
Check external secret in case of external ES/OS 

IshaGirdhar:
Fluentd preinstall/pre-upgrade always waits for es-internal-secret which is not used in the case of external ES/OS. This PR adds the code to check if it has an external secret, it will wait for that. If no external secret is provided, will wait for the system secret (as is the current behavior).


anders-swanson (approved):
LGTM

PR #4525, opened 2022-10-28
VZ-6465 create e2e tests to verify the pushed Manifest 

davisgomes:

### DIFF
--- a/tests/e2e/config/scripts/register_managed_cluster.sh
+++ b/tests/e2e/config/scripts/register_managed_cluster.sh
@@ -116,16 +116,20 @@ fi
 
 # get the admin user token from the Rancher API
 RANCHER_URL=$(kubectl --kubeconfig ${ADMIN_KUBECONFIG} get vz -o jsonpath='{.items[0].status.instance.rancherUrl}')
+echo "RANCHER_URL: ${RANCHER_URL}"
 RANCHER_ADMIN_PASS=$(kubectl --kubeconfig ${ADMIN_KUBECONFIG} get secret -n cattle-system rancher-admin-secret -o jsonpath={.data.password} | base64 --decode)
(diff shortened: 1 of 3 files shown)


tests/e2e/config/scripts/register_managed_cluster.sh
```diff
@@ -114,11 +114,25 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
-# export manifest on admin
-kubectl --kubeconfig ${ADMIN_KUBECONFIG} get secret verrazzano-cluster-${MANAGED_CLUSTER_NAME}-manifest -n verrazzano-mc -o jsonpath={.data.yaml} | base64 --decode > register-${MANAGED_CLUSTER_NAME}.yaml
+# get the admin user token from the Rancher API
+RANCHER_URL=$(kubectl --kubeconfig ${ADMIN_KUBECONFIG} get vz -o jsonpath='{.items[0].status.instance.rancherUrl}')
+echo "RANCHER_URL: ${RANCHER_URL}"
+RANCHER_ADMIN_PASS=$(kubectl --kubeconfig ${ADMIN_KUBECONFIG} get secret -n cattle-system rancher-admin-secret -o jsonpath={.data.password} | base64 --decode)
+echo "RANCHER_ADMIN_PASS: ${RANCHER_ADMIN_PASS}"
```
fredtibbitts, line 121: I'm wondering if we should not be dumping the passwords and tokens to the console? The cluster is short-lived so it's probably not a big deal but it always makes me uncomfortable when passwords/tokens are visible like this.

fredtibbitts (approved):
Just the one question about the passwords, otherwise LGTM.

tests/e2e/config/scripts/register_managed_cluster.sh
davisgomes: Yea I was going back and forth because it gives us some nice diagnostic help, but it might not be good to throw these out.

PR #4532, opened 2022-10-30
VZ-7506: Make CLI resilient to loosing stream output of uninstall log

mgianatagh:
Make CLI resilient to loosing stream output of uninstall log

### DIFF
--- a/tools/vz/cmd/uninstall/uninstall.go
+++ b/tools/vz/cmd/uninstall/uninstall.go
@@ -297,8 +297,8 @@ func waitForUninstallToComplete(client client.Client, kubeClient kubernetes.Inte
 		return err
 	}
 
-	re := regexp.MustCompile(cmdhelpers.VpoSimpleLogFormatRegexp)
-	go func(outputStream io.Writer, sc *bufio.Scanner) {
+	go func(outputStream io.Writer, sc *bufio.Scanner, useUninstallJob bool) {
+		re := regexp.MustCompile(cmdhelpers.VpoSimpleLogFormatRegexp)
 		var err error
 		secondsWaited := 0
 		maxSecondsToWait := int(vpoTimeout.Seconds())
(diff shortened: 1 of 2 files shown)


tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -359,6 +378,35 @@ func waitForUninstallToComplete(client client.Client, kubeClient kubernetes.Inte
 	return timeoutErr
 }
 
+// getScanner - get scanner for uninstall console output
+func getScanner(useUninstallJob bool, client client.Client, kubeClient kubernetes.Interface, vzHelper helpers.VZHelper, namespacedName types.NamespacedName) (*bufio.Scanner, error) {
```
michael-cico, line 382: I wonder if there's enough overlap here to consolidate the `getScanner` funcs, but not for this PR I think.  They share the same logic I think (get log from a job, or from the VPO), at least for install/uninstall maybe.  Might be tough since upgrade never used a job.

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -289,36 +285,59 @@ func getUninstallJobPodName(c client.Client, vzHelper helpers.VZHelper, jobName
...
+						return
+					}
+					time.Sleep(secondsPerRetry * time.Second)
+					continue
+				}
+				secondsWaited = 0
+				sc.Split(bufio.ScanLines)
+			}
+
+			scannedOk := sc.Scan()
```
michael-cico, line 323: Does `Scan` return when the VPO goes down?  In my own testing `Scan` just hung until the operation finished or timed out and got closed, IIRC.

michael-cico (approved):

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -79,6 +79,9 @@ func NewCmdUninstall(vzHelper helpers.VZHelper) *cobra.Command {
 	cmd.PersistentFlags().Bool(constants.DryRunFlag, false, "Simulate an uninstall.")
 	_ = cmd.PersistentFlags().MarkHidden(constants.DryRunFlag)
 
+	// Hide the flag for overriding the default wait timeout for the platform-operator
```
bobdonat, line 82: Is this flag needed for uninstall?

stevshap (approved):

tools/vz/cmd/uninstall/uninstall.go
mgianatagh: I wanted to obtain the timeout for re-connecting to the VPO from a common location.  If we expose these flags they will need further thought on how to present them to a user.

PR #4533, opened 2022-10-31
VZ-7543: E2E tests for OpenSearch metrics and grafana dashboards

adiforluls:
PR to add E2E tests for OpenSearch metrics and grafana dashboards.


tests/e2e/grafana-db/grafana_test.go
```diff
@@ -240,6 +269,29 @@ var _ = t.Describe("Test Grafana Dashboard Persistence", Label("f:observability.
...
+					pkg.Log(pkg.Error, err.Error())
+					return false
+				}
+				if resp.StatusCode != http.StatusOK {
+					pkg.Log(pkg.Error, fmt.Sprintf(grafanaErrMsgFmt, resp.StatusCode, string(resp.Body)))
+					return false
+				}
+				body := make(map[string]map[string]string)
+				json.Unmarshal(resp.Body, &body)
+				return strings.Contains(body["dashboard"]["title"], openSearchMetricsDbTitle)
```
anajosep, line 291: Can we check for some metrics, like cluster state, shard count etc are having valid values ? Is it possible to check that there are no errors in the dashboard UI through these tests ?

tests/e2e/grafana-db/grafana_test.go
```diff
@@ -240,6 +269,29 @@ var _ = t.Describe("Test Grafana Dashboard Persistence", Label("f:observability.
 			return strings.Contains(body["dashboard"]["title"], systemDashboardTitle)
 		}).WithPolling(pollingInterval).WithTimeout(waitTimeout).Should(BeTrue())
 	})
+
+	// GIVEN a running grafana instance
+	// WHEN a call is made to Grafana Dashboard with UID corresponding to OpenSearch Summary Dashboard
+	// THEN the dashboard metadata of the corresponding dashboard is returned
+	if ok, _ := pkg.IsVerrazzanoMinVersion("1.3.0", kubeconfigPath); ok {
```
anajosep, line 276: Same test seems to be duplicated in 2 places ? is that needed ?

anajosep: Can we have the testing method in a common place, say `pkg/grafana.go` and call the method from different test suites ?

adiforluls: Yeah in this suite we delete the grafana pod and test again to see whether out DB backup works. Agree though we should have one method in pkg to reduce duplication everywhere.

tests/e2e/grafana-db/grafana_test.go
adiforluls: I don't think we can test UI interactions directly via ginkgo or an API. I'll review whether we can verify metrics like shard count returning valid values. So far for metrics the only thing I could see we test is that the metrics exist with correct labels.

anajosep (approved):

PR #4534, opened 2022-10-31
Add test cases for log pkg

rahulsrivastava71:

### DIFF
--- a/pkg/log/vzlog/vzlog_test.go
+++ b/pkg/log/vzlog/vzlog_test.go
@@ -137,6 +137,40 @@ func TestErrorfThrottledNewErr(t *testing.T) {
 	DeleteLogContext(rKey)
 }
 
+// TestErrorfNewErr tests the error msg
+// GIVEN a ProgressLogger with a frequency of 30 seconds
+func TestErrorfNewErr(t *testing.T) {
+	const messageTemplate = "Mymessage %s"
+	const messageParameter = "test2"
+	msg := fmt.Sprintf(messageTemplate, messageParameter)
+	logger := fakeLogger{expectedMsg: msg}
+	const rKey = "testns/errorsNew"
+	rl := EnsureContext(rKey)
+	l := rl.EnsureLogger("comp1", &logger, zap.S()).SetFrequency(30)
+
+	assert.Error(t, l.ErrorfNewErr(messageTemplate, messageParameter))
+	assert.Equal(t, 1, logger.count)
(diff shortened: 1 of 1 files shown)


pkg/log/vzlog/vzlog_test.go
```diff
@@ -264,6 +298,79 @@ func TestZap(t *testing.T) {
...
+	rl := DefaultLogger()
+	rl.Oncef("comp1", &logger, zap.S())
+	assert.NotNil(t, rl)
+}
+
+// TestSetZapLogger is testing the call for SetZapLogger and it's not null
+func TestSetZapLogger(t *testing.T) {
+	rl := DefaultLogger()
+	rl.SetZapLogger(zap.S())
+	assert.NotNil(t, rl)
```
pfmackin, line 323: you need to call rl.GetZapLogger and make sure it matches what was just set

pkg/log/vzlog/vzlog_test.go
```diff
@@ -264,6 +298,79 @@ func TestZap(t *testing.T) {
...
+}
+
+// TestDebug is testing the call for all Debug , it's not null
+func TestDebug(t *testing.T) {
+	rl := DefaultLogger()
+	rl.Debug("comp1", zap.S())
```
pfmackin, line 329: All of the logging with zap.S() doesn't really test anything.  You need to use the fake logger to make sure what you logged worked.

pkg/log/vzlog/vzlog_test.go
rahulsrivastava71: this test case is only checking for setzaplogger as an individual component so to check other method on basis of this method is already done in other test case

pkg/log/vzlog/vzlog_test.go
rahulsrivastava71: i can remove that as this , but we can keep this also as this method takes multiple arguements so i kept string and one zap object

pfmackin: Look at TestLogFormat how we check the message in the fake logger.  This is what you need to do for all of these tests, you have to log, then make sure the expected string got logged.

rahulsrivastava71: TestLogFormat is testing the EnsureLogger method which is again a different component and this is unit test cases , so the above test case is again checking the single component.

pfmackin (approved):

PR #4539, opened 2022-10-31
VZ-7533 PSR add base chart for worker

pfmackin:
Add base chart for worker so that worker specific charts are isolated. 
Added logging for runner so that we know a worker had a successful DoWork call.
Make some name changes to resources so that they are consistently named.

### DIFF
--- a/tools/psr/manifests/charts/worker/charts/getlogs/templates/ingestauthorizationpolicy.yaml
+++ b/tools/psr/manifests/charts/worker/charts/getlogs/templates/ingestauthorizationpolicy.yaml
@@ -1,11 +1,10 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-{{- if .Values.istio.enabled }}
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
-  name: {{ .Release.Name }}-{{ .Release.Namespace }}-ingest-authzpol
+  name: {{ template "worker.fullName" .}}-ingest-authzpol
   namespace: verrazzano-system
 spec:
   selector:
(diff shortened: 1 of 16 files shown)


tools/psr/manifests/charts/worker/charts/getlogs/templates/masterauthorizationpolicy.yaml
```diff
@@ -1,11 +1,10 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-{{- if .Values.istio.enabled }}
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
-  name: {{ .Release.Name }}-{{ .Release.Namespace }}-master-authzpol
```
mbenmclean, line 8: This does not guarantee  uniqueness. You could have the same worker type deployed with the same chart name in different namespaces. This would cause things to break if one of the releases was uninstalled, but not the other. You need to include the namespace as part of the name for this.

tools/psr/manifests/charts/worker/charts/getlogs/templates/ingestauthorizationpolicy.yaml
```diff
@@ -1,11 +1,10 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-{{- if .Values.istio.enabled }}
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
-  name: {{ .Release.Name }}-{{ .Release.Namespace }}-ingest-authzpol
+  name: {{ template "worker.fullName" .}}-ingest-authzpol
```
mbenmclean, line 7: This does not guarantee uniqueness. You could have the same worker type deployed with the same chart name in different namespaces. This would cause things to break if one of the releases was uninstalled, but not the other. You need to include the namespace as part of the name for this.

tools/psr/manifests/charts/worker/charts/getlogs/templates/masterauthorizationpolicy.yaml
pfmackin: fixed

tools/psr/manifests/charts/worker/charts/getlogs/templates/ingestauthorizationpolicy.yaml
pfmackin: fixed

mbenmclean (approved):

PR #4546, opened 2022-10-31
VZ-7565 add worker threads (goroutines)

pfmackin:
Run a worker in multiple goroutines so that we can scale a pod vertically (scale-up)

### DIFF
--- a/tools/psr/backend/config/config.go
+++ b/tools/psr/backend/config/config.go
@@ -6,6 +6,7 @@ package config
 import (
 	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
 	"github.com/verrazzano/verrazzano/tools/psr/backend/osenv"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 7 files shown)


tools/psr/backend/config/config.go
```diff
@@ -58,8 +65,18 @@ func GetCommonConfig(log vzlog.VerrazzanoLogger) (CommonConfig, error) {
 		sleepDuration = 10 * time.Nanosecond
 	}
 
+	threadCount, err := strconv.Atoi(PsrEnv.GetEnv(PsrWorkerThreadCount))
+	if err != nil {
+		return CommonConfig{}, log.ErrorfNewErr("Error parsing worker thread count: %v", err)
+	}
+	// Max thread count is 100
+	if threadCount > 100 {
```
stevshap, line 73: Why 100?

pfmackin: no special reason other than prevent the pod from maxing cpu.  I can tweek it higher if need be.

stevshap: ok works for now

stevshap (approved):

tools/psr/backend/workmanager/runner.go
```diff
@@ -46,6 +49,7 @@ var _ WorkerRunner = runner{}
 // runnerMetrics holds the metrics produced by the runner. Metrics must be thread safe.
 type runnerMetrics struct {
 	loopCount                  metrics.MetricItem
+	workerThreadCount          metrics.MetricItem
```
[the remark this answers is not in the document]
mbenmclean: Or actually looks like it might be the number of active threads?

tools/psr/backend/workmanager/runner.go
```diff
@@ -17,6 +17,9 @@ const (
 	LoopCount     = "loop_count"
 	LoopCountHelp = "The number of loop iterations executed"
 
+	WorkerThreadCount     = "worker_thread_count"
```
mbenmclean, line 20: I think active_getlogs_thread_count might be better. I think including the name of the worker and the word active would be good.

tools/psr/backend/workmanager/runner.go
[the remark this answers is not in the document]
pfmackin: I should total in the name let me fix that

tools/psr/backend/workmanager/runner.go
pfmackin: Every metric already has psr-<worker-type>- prefix

PR #4548, opened 2022-10-31
VZ-7499: use mysql-operator from BOM when creating innodbcluster

mgianatagh:
use mysql-operator from BOM when creating innodbcluster


platform-operator/thirdparty/charts/mysql/values.yaml
```diff
@@ -6,6 +6,8 @@ image:
     enabled: false
     secretName:
 
+mysqlOperator:
+  image:
 
 credentials:
   root:
```
jmaron99, line 13: Can you do me a favor and remove the commented out legacyUpgrade values at the bottom of the file?  They are no longer really used by the helm process but are rather manually created to support the creation of the db restore job so they really aren't configurable by the user.  I had that thought on a walk last night post merging my commit and could do that in a separate commit, but since this one is already up... ;)

michael-cico (approved):

platform-operator/thirdparty/charts/mysql/values.yaml
mgianatagh: I will handle that in a subsequent PR, this one already has a passing test suite, so I rather get it merged now.

jmaron99: Yeah - that was me.  Forgot :)

PR #4550, opened 2022-11-01
VZ-7535 update PSR OpenSearch query 

mbenmclean:
This change updates the OpenSearch query to 10 random OR conditionals.


tools/psr/backend/workers/opensearch/getlogs/getlogs.go
```diff
@@ -208,3 +208,81 @@ func (w getLogs) GetMetricList() []prometheus.Metric {
...
+                {
+          "match": {
+            "message": "%s"
+          }
+        }
+      ]
+    }
+  }
+}`,
+		GetRandomLowerAlpha(),
```
pfmackin, line 271: ok for now but better to have a function that loops and returns a string

pfmackin (dismissed):

stevshap (dismissed):

tools/psr/backend/workers/opensearch/getlogs/getlogs.go
mbenmclean: Sound good

PR #4556, opened 2022-11-01
VZ-4767.  VPO webhook scaling

michael-cico:
Re-introduces VZ webhook scaling capability.  This approach attempts to get rid of the somewhat-circular dependency between the VPO and the webhook deployments in the first attempt.

In this implementation, we move the initContainer that creates the certs and secrets into the webhook deployment; in addition, if the secrets already exist, we re-use them (i.e., if the `Create` call fails if the secret already exists, we read the data from the secrets and use that).  The certs are no longer directly mounted from the secrets.

Also, the code implementing the different run modes (webhook-init, webhooks, operator) is refactored into an internal package to allow for additional unit testing (in a new PR to come).

There is still some work to consider around cycling the certs; this change also bumps the expiration to 3 years.  In a follow-on PR we will look to rotate these during upgrade by deleting the secrets and bouncing the webhook deployment.
 
Change details:
- Move init container that initializes webhook certs/secrets to the webhook deployment
- If secrets already exist, the code uses the existing secret data to populate the webhook certs and update the webhook configurations
- refactor init code for containers into separate package for readability and (eventual) unit testing
- consolidate error handling to single return/exit point from main
- Re-add default anti-affinity rules for the VPO to the chart
- Updates the VPO switches to reflect the new execution modes appropriately
- Reintroduces change to scale up the webhook for the HA tests
- Adds a couple of Makefile targets to simplify/automate building the operator manifest using local docker images and loading them into a local kind cluster

### DIFF
--- a/ci/ha/JenkinsfileRollingUpgrade
+++ b/ci/ha/JenkinsfileRollingUpgrade
@@ -304,7 +304,7 @@ pipeline {
                             echo "Installing Verrazzano"
                             ${GO_REPO_PATH}/vz install --filename ${INSTALL_CONFIG_FILE} --operator-file ${WORKSPACE}/platform-operator.yaml --timeout 45m --platform-operator-timeout 10m
 
-                            # For now, manually scale the VPO webhook to 2 replicas until we have hooks for it in the product
+                            # For now, manually scale the VPO webhook to 3 replicas until we have hooks for it in the product
                             kubectl  scale deployment -n verrazzano-install verrazzano-platform-operator-webhook --replicas 3
                             kubectl  rollout status deployment -n verrazzano-install verrazzano-platform-operator-webhook -w --timeout 15m
                        """
--- a/ci/ha/JenkinsfileRollingUpgrade
+++ b/ci/ha/JenkinsfileRollingUpgrade
@@ -303,6 +303,10 @@ pipeline {
 
                             echo "Installing Verrazzano"
                             ${GO_REPO_PATH}/vz install --filename ${INSTALL_CONFIG_FILE} --operator-file ${WORKSPACE}/platform-operator.yaml --timeout 45m --platform-operator-timeout 10m
+
+                            # For now, manually scale the VPO webhook to 2 replicas until we have hooks for it in the product
+                            kubectl  scale deployment -n verrazzano-install verrazzano-platform-operator-webhook --replicas 3
+                            kubectl  rollout status deployment -n verrazzano-install verrazzano-platform-operator-webhook -w --timeout 15m
                        """
                     }
                     post {
--- a/platform-operator/Makefile
+++ b/platform-operator/Makefile
@@ -166,7 +166,7 @@ endif
 unit-test: go-install
 	$(GO) test -v  ./internal/... ./controllers/... ./apis/...
 
-CLUSTER_NAME = verrazzano
+CLUSTER_NAME ?= verrazzano
 BUILD_DEPLOY = build/deploy
 OPERATOR_YAML ?= ${BUILD_DEPLOY}/operator.yaml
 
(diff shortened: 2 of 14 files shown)


ci/ha/JenkinsfileRollingUpgrade
```diff
@@ -303,6 +303,10 @@ pipeline {
 
                             echo "Installing Verrazzano"
                             ${GO_REPO_PATH}/vz install --filename ${INSTALL_CONFIG_FILE} --operator-file ${WORKSPACE}/platform-operator.yaml --timeout 45m --platform-operator-timeout 10m
+
+                            # For now, manually scale the VPO webhook to 2 replicas until we have hooks for it in the product
```
mgianatagh, line 307: 3 replicas

platform-operator/Makefile
```diff
@@ -176,6 +176,21 @@ generate-operator-yaml:
 	mkdir -p $${OPERATOR_DIR} ; \
 	env DOCKER_IMAGE=${DOCKER_IMAGE_FULLNAME}:${DOCKER_IMAGE_TAG} IMAGE_PULL_SECRETS=${IMAGE_PULL_SECRETS} APP_OPERATOR_IMAGE=${VERRAZZANO_APPLICATION_OPERATOR_IMAGE} ../tools/scripts/generate_operator_yaml.sh > ${OPERATOR_YAML}
 
+.PHONY: generate-local-operator-yaml
+generate-local-operator-yaml:
+	OPERATOR_DIR=$$(dirname ${OPERATOR_YAML}) ; \
+	mkdir -p $${OPERATOR_DIR} ; \
+	env DOCKER_IMAGE=${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG} IMAGE_PULL_SECRETS=${IMAGE_PULL_SECRETS} APP_OPERATOR_IMAGE=${VERRAZZANO_APPLICATION_OPERATOR_IMAGE} ../tools/scripts/generate_operator_yaml.sh > ${OPERATOR_YAML}
+
+.PHONY: load-local-operators
```
mgianatagh, line 185: I am surprised this works.  I have found that I had to tag the application-operator image to have the tags expected in the BOM, because that is that will be used when VPO deploys it.

michael-cico: I have to have a local build of the application-operator at the same tag, generally, for this to work.  I can back this out if we want.

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -91,7 +85,10 @@ metadata:
 spec:
   replicas: 1
   strategy:
-    type: Recreate
+    type: RollingUpdate
```
pfmackin, line 88: I think we want recreate since we had that problem way back where the old VPO was doing work while the new one was coming up.  We made this change to fix that.

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -102,13 +99,29 @@ spec:
...
+      initContainers:
+        - name: initwebhooks
+          imagePullPolicy: {{ .Values.imagePullPolicy }}
+          image: {{ .Values.image }}
+          args:
+            - --zap-log-level=info
+            - --run-webhook-init=true
+          resources:
+            requests:
+              memory: 72Mi
```
pfmackin, line 114: where did the 72Mi come from?

platform-operator/internal/config/config.go
```diff
@@ -55,29 +52,36 @@ type OperatorConfig struct {
 	// VersionCheckEnabled enables/disables version checking for upgrade.
 	VersionCheckEnabled bool
 
-	// WebhooksEnabled enables/disables Webhooks for the operator
-	WebhooksEnabled bool
+	// RunWebhooks Inidicates the executable the operator code is executing the webhooks path
+	RunWebhooks bool
```
pfmackin, line 56: Can you say that this is only used by the WebHook pod

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
michael-cico: this is for the webhook, not the VPO.  The operator deployment still specifies `Recreate`

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
michael-cico: It was there in one of the earlier impls.  Seems oddly specific to me.

pfmackin (approved):

platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
bryantidd: I think everything should have some value for requests. Note it's not a limit. This is needed if you want to do autoscaling I believe.

michael-cico: OK

bryantidd (approved):

PR #4557, opened 2022-11-01
JIRA: VZ-7475 create and run database load job prior to invoking helm upgrade for mysql

jmaron99:
The job responsible for re-loading the keycloak db and user is now started prior to the helm upgrade of mysql to ensure that the correct PV is bound to the job rather than the new mysql pods.


platform-operator/controllers/verrazzano/component/mysql/legacydb.go
```diff
@@ -191,15 +261,129 @@ func handleLegacyDatabasePreUpgrade(ctx spi.ComponentContext) error {
...
+			return err
+		}
+	}
+
+	if err := createLegacyUpgradeJob(ctx); err != nil {
+		return err
+	}
+
+	// wait till the pod shows up so that it is bound to the PV
+	if err := waitForJobPodRunning(ctx, time.Duration(60)*time.Second); err != nil {
```
michael-cico, line 282: Not sure if we should wait for 60 seconds here; this will return an error on timeout right?  Maybe like 10s or something so we don't block the reconciler for too long.

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
```diff
@@ -191,15 +261,129 @@ func handleLegacyDatabasePreUpgrade(ctx spi.ComponentContext) error {
...
+				return err
+			}
+			var kvs []bom.KeyValue
+			if kvs, err = appendLegacyUpgradePersistenceValues(&bomFile, []bom.KeyValue{}); err != nil {
+				return err
+			}
+
+			values := loadJobValues{}
+			for _, kv := range kvs {
+				if kv.Key == "legacyUpgrade.container.image" {
```
michael-cico, line 340: be good to make sure we run the private registry tests with this, although not sure the job will be around long enough for the tests to pick up any bugs in pulling the wrong images.

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
jmaron99: 60 seconds is the time out period.  if the pod comes up sooner the wait terminates and the processing continues.

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
jmaron99: Not sure what to do with this comment - run the private registry tests on this branch as a condition for approval/commit?

michael-cico: Just thought it might save a revert.  I guess it's half comment/half-question, just wanted to point it out.  My question part is it whether or not you think we'll capture anything by running it.

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
michael-cico: I was thinking of slow pull times, etc.  Could be 6 of one, 1/2 dozen the other.  If there are slow pull times everything will be affected.

michael-cico: @pfmackin what do you think?

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
jmaron99: Not entirely sure.  Only thing I imagine I can do is try to run it.  It doesn't a whole day does it?

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
jmaron99: I'm not sure I follow.  If we need to run the job we will be stuck at this point in the reconciliation loop whether we come back and retry after 10 seconds or wait, or am I wrong?

michael-cico (approved):

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
michael-cico: we will, but other things can proceed.  This will block all things for 60s right now.

jmaron99: Only in the case where the pod fails to come up, which should be an edge case.  Otherwise it will just block till the pod transitions to running (+/- 1 sec)

PR #4559, opened 2022-11-01
Added log message around the distribution failure

saidu941:
Added more logging around the failure while comparing the two slices


pkg/string/slice_test.go
```diff
@@ -207,8 +208,8 @@ func TestCompareTwoSlices(t *testing.T) {
 	slice3 := []string{"s1", "s2", "s4"}
 	slice4 := []string{"s1", "s2"}
 
-	assert.True(CompareTwoSlices(slice1, slice2))
-	assert.False(CompareTwoSlices(slice1, slice3))
-	assert.False(CompareTwoSlices(slice2, slice3))
-	assert.False(CompareTwoSlices(slice1, slice4))
+	assert.True(CompareTwoSlices(vzlog.DefaultLogger(), slice1, slice2))
```
anders-swanson, line 211: you can create logger once, if you like. `logger := vzlog.DefaultLogger()` then use `logger` in each call.
Please also note that this pattern you are using can be a parameterized test.

pkg/string/slice.go
```diff
@@ -67,15 +69,19 @@ func SliceAddString(slice []string, s string) ([]string, bool) {
 }
 
 // compareSlices compares 2 string slices after sorting
-func CompareTwoSlices(slice1 []string, slice2 []string) bool {
+func CompareTwoSlices(log vzlog.VerrazzanoLogger, slice1 []string, slice2 []string) bool {
 	sort.Strings(slice1)
```
anders-swanson, line 73: You should be copying the slices before sorting them. This is going to mutate the input parameters, perhaps unexpectedly.

pkg/string/slice.go
```diff
@@ -67,15 +69,19 @@ func SliceAddString(slice []string, s string) ([]string, bool) {
 }
 
 // compareSlices compares 2 string slices after sorting
-func CompareTwoSlices(slice1 []string, slice2 []string) bool {
+func CompareTwoSlices(log vzlog.VerrazzanoLogger, slice1 []string, slice2 []string) bool {
```
anders-swanson, line 72: Can you call this function AreSlicesEqualWithoutOrder or something similar. I don't know what `compare` means just looking at the function signature.

pkg/string/slice_test.go
saidu941: Addressed and updated PR to use log from test framework

pkg/string/slice.go
saidu941: Addressed and updated PR

pkg/string/slice.go
```diff
@@ -67,15 +69,19 @@ func SliceAddString(slice []string, s string) ([]string, bool) {
...
 	sort.Strings(slice1)
 	sort.Strings(slice2)
 
 	if len(slice1) != len(slice2) {
+		log.Errorf("%s doesn't match with %s", slice1, slice2)
```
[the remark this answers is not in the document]
saidu941: Addressed and updated PR to use log from test framework

pkg/string/slice.go
saidu941: Addressed and updated PR

pkg/string/slice.go
```diff
@@ -67,15 +67,19 @@ func SliceAddString(slice []string, s string) ([]string, bool) {
...
+func AreSlicesEqualWithoutOrder(slice1 []string, slice2 []string) bool {
+	s1 := make([]string, len(slice1))
+	s2 := make([]string, len(slice2))
+	copy(s1, slice1)
+	copy(s2, slice2)
+	sort.Strings(s1)
+	sort.Strings(s2)
 
-	if len(slice1) != len(slice2) {
+	if len(s1) != len(s2) {
```
anders-swanson, line 78: you could move this check before the copying/sorting as an optimization.

anders-swanson (dismissed):

pkg/string/slice.go
saidu941: Addressed and updated PR

anders-swanson (approved):

PR #4562, opened 2022-11-01
VZ-7391: Fix race condition for verify registry test

Abhijithang:
This PR fixes the race condition for verify registry tests in upgrade paths pipeline.


desagar (dismissed):
just a question - does the build have to be done inside the loop?

desagar (approved):

PR #4563, opened 2022-11-01
VZ 7578: Work around omitempty replicas for node groups

anders-swanson:
Omit empty replicas for node groups was causing user defined `0` values to be ignored when creating the EffectiveCR.


theabmitra (approved):
Approved

theabmitra (approved):
approved

PR #4566, opened 2022-11-01
Adding http get worker to PSR app

vanajamukkara:
http get worker which accesses the given endpoint. 
Writes the metrics total_get_requests, total_get_requests_succeeded, total_get_requests_failed


tools/psr/backend/workmanager/manager.go
```diff
@@ -75,6 +76,8 @@ func getWorker(wt string) (spi.Worker, error) {
 	switch wt {
 	case config.WorkerTypeExample:
 		return example.NewExampleWorker()
+    case config.WorkerTypeHttpGet:
```
pfmackin, line 79: Looks like formatting is messed up

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (w httpGetWorker) WantIterationInfoLogged() bool {
+	return false
+}
+
+func (w httpGetWorker) DoWork(conf config.CommonConfig, log vzlog.VerrazzanoLogger) error {
+    var lc, ls, lf int64
+    var u string = config.PsrEnv.GetEnv(config.PsrEndpoint)
+    log.Infof("Endpoint %s", u)
```
pfmackin, line 134: Change to log.Oncef

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,167 @@
...
+    log.Infof("HttpGet Worker doing work")
```
pfmackin, line 135: Change to log.Oncef

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,167 @@
...
+    //increment totalGetRequestsCount
+    lc = atomic.AddInt64(&w.httpgetMetrics.totalGetRequestsCount.Val, 1)
+    resp, err := http.Get(u)
+    if err != nil {
+        return err
+    }
+    if resp == nil {
+        return fmt.Errorf("GET request to endpoint %s received a nil response", u)
+    }
+    fmt.Println("The status code of the get request is: ", resp.StatusCode)
```
pfmackin, line 146: Remove the Println.  Just generator metrics like getlogs.go worker.

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,167 @@
...
+        }
+        //Convert the body to type string
+        sb := string(body)
+        fmt.Println("The response body: ", sb)
+    }
+
+    logMsg := fmt.Sprintf("HttpGet worker total requests %v, " +
+                " successful requests %v, failed requests %v",
+                lc, ls,lf)
+    log.Infof(logMsg)
```
pfmackin, line 164: Don't log.  If you really need this then use log.Debugf

pfmackin:
In general, the worker should not log, use metrics to convey information.

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	"github.com/verrazzano/verrazzano/tools/psr/backend/config"
+	"github.com/verrazzano/verrazzano/tools/psr/backend/metrics"
+	"github.com/verrazzano/verrazzano/tools/psr/backend/osenv"
+	"github.com/verrazzano/verrazzano/tools/psr/backend/spi"
+	"sync/atomic"
+)
+
+const (
+	totalGetRequestsCount     = "total_get_requests"
```
pfmackin, line 23: total needs to be at the end like get_requests_total see https://prometheus.io/docs/practices/naming/

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,168 @@
...
+	totalGetRequestsCountHelp = "The total number of get requests logged"
+
+	totalGetRequestsSucceededCount     = "total_get_requests_succeeded"
```
pfmackin, line 26: name should have _count_

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package http
+
+import (
+	"io"
+	"net/http"
+)
+
```
stevshap, line 10: This should be one import block, and the github imports should be separated from the go packages.

Here is an example of how we try to format imports:
https://github.com/verrazzano/verrazzano/blob/master/platform-operator/main.go

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("GET request to endpoint received a nil response")
+	}
+	if resp.StatusCode == 200 {
+		ls = atomic.AddInt64(&w.workerMetrics.getRequestsSucceededCountTotal.Val, 1)
+	} else {
+		lf = atomic.AddInt64(&w.workerMetrics.getRequestsFailedCountTotal.Val, 1)
+		//Read the response body on the line below.
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			log.Errorf("Error reading response body: %v", err)
```
pfmackin, line 136: We should not log in this test since it affects the OpenSearch logging load.  If this test is doing a millions gets per second and there is error then it will inadvertently generate a ton of logs.  At least use log.ErrorfThrottled

tools/psr/manifests/usecases/http/get.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+
+global:
+  envVars:
+    PSR_WORKER_TYPE: httpget
+    # SERVICE_NAME, SERVICE_NAMESPACE, SERVICE_PORT and PATH are required to form HTTP endpoint URL to access. HTTPS is not supported.
+    SERVICE_NAME:
```
mbenmclean, line 9: Why is this empty? Does this ever get set?

tools/psr/backend/workers/http/get/get.go
```diff
@@ -0,0 +1,147 @@
...
+func (w get) WantIterationInfoLogged() bool {
+	return false
+}
+
+func (w get) DoWork(conf config.CommonConfig, log vzlog.VerrazzanoLogger) error {
+	var lc, ls, lf int64
+	//increment getRequestsCountTotal
+	lc = atomic.AddInt64(&w.workerMetrics.getRequestsCountTotal.Val, 1)
+	resp, err := http.Get("http://" + config.PsrEnv.GetEnv(ServiceName) +
+		"." + config.PsrEnv.GetEnv(ServiceNamespace) +
```
mbenmclean, line 119: This will always be empty with how you have it implemented. You never get the namespace.

tools/psr/manifests/usecases/http/get.yaml
vanajamukkara: These are passed as overrides from helm with --set

tools/psr/backend/workers/http/get/get.go
vanajamukkara: They are required environment variables. pod log will have errors if these are not set.

tools/psr/manifests/usecases/http/get.yaml
pfmackin: you can also use -f with these defined in a file

pfmackin (dismissed):

stevshap (approved):

pfmackin (approved):

PR #4574, opened 2022-11-02
Enhance UT coverage

sk2701:
UT coverage for pkg/bom increased from 67.96% to 82.52%

### DIFF
--- a/pkg/bom/bom_test.go
+++ b/pkg/bom/bom_test.go
@@ -302,14 +302,16 @@ func TestBomComponentVersion(t *testing.T) {
 // WHEN I ask for the version of that component
 // THEN the correct version is returned if found, or an error if not found
 func TestBomGetComponentVersion(t *testing.T) {
-	bom, err := NewBom(realBomFilePath)
+	bom, err := NewBom(testBomFilePath)
 	assert.NoError(t, err)
 
-	_, err = bom.GetComponentVersion("cert-manager")
+	ver, err := bom.GetComponentVersion("cert-manager")
 	assert.NoError(t, err)
+	assert.Equal(t, ver, "v1.7.1")
 
-	_, err = bom.GetComponentVersion("foo")
+	ver, err = bom.GetComponentVersion("foo")
 	assert.Error(t, err)
+	assert.Equal(t, ver, "")
 }
 
 // TestGetSubcomponentImages tests the GetSubcomponentImages method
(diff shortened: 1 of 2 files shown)


pkg/bom/bom_test.go
```diff
@@ -297,3 +296,41 @@ func TestBomComponentVersion(t *testing.T) {
...
+	bom, err := NewBom(testBomFilePath)
+	assert.NoError(t, err)
+
+	imageNum := bom.GetSubcomponentImageCount("foo")
+	if imageNum != 0 {
+		t.Errorf("The number of counted images is incorrect")
+	}
+
+	imageNum = bom.GetSubcomponentImageCount(ingressControllerComponent)
+	if imageNum != 2 {
```
adiforluls, line 332: Use asserts to verify the outcome. Example: if you're expecting 2 images then it should be something like `assert.Equal(imageNum, 2)`.

pkg/bom/bom_test.go
```diff
@@ -297,3 +296,41 @@ func TestBomComponentVersion(t *testing.T) {
 	assert.NotNil(t, c)
 	assert.NotNil(t, c.Version)
 }
+
+func TestBomGetComponentVersion(t *testing.T) {
```
adiforluls, line 300: Add func comment to unit tests explaining what the test is about and what scenario it's testing.
`TestXYZ tests the func XYZ for the following scenario
GIVEN some resource
WHEN <scenario>
THEN something is returned`

adiforluls (dismissed):

pkg/bom/bom_test.go
```diff
@@ -297,3 +296,49 @@ func TestBomComponentVersion(t *testing.T) {
...
+
+// TestBomGetComponentVersion tests the GetComponentVersion method
+// GIVEN a call to GetComponentVersion for a valid component
+// WHEN I ask for the version of that component
+// THEN the correct version is returned if found, or an error if not found
+func TestBomGetComponentVersion(t *testing.T) {
+	bom, err := NewBom(realBomFilePath)
+	assert.NoError(t, err)
+
+	_, err = bom.GetComponentVersion("cert-manager")
```
adiforluls, line 308: Add assertion for version as well. Instead of real bom use a fake bom since the version will keep updating in the real bom.

pkg/bom/bom_test.go
```diff
@@ -297,3 +296,49 @@ func TestBomComponentVersion(t *testing.T) {
...
+
+// TestGetSubcomponentImages tests the GetSubcomponentImages method
+// GIVEN a call to GetSubcomponentImages for a valid subcomponent
+// WHEN I ask for the images of that component
+// THEN the correct images are returned if found, or an error if not found
+func TestGetSubcomponentImages(t *testing.T) {
+	bom, err := NewBom(testBomFilePath)
+	assert.NoError(t, err)
+
+	_, err = bom.GetSubcomponentImages(ingressControllerComponent)
```
adiforluls, line 323: Add assertion for sub component images as well.

adiforluls (approved):
LGTM

PR #4582, opened 2022-11-02
VZ-6602 - Sync agent secret from admin cluster and always use Rancher URL in managed cluster kubeconfig when Rancher is enabled (remove rancherBasedKubeconfig feature flag)

desagar:

### DIFF
--- a/application-operator/mcagent/mcagent_cluster_secrets.go
+++ b/application-operator/mcagent/mcagent_cluster_secrets.go
@@ -49,27 +49,18 @@ func (s *Syncer) syncAgentSecretFromAdminCluster() (controllerutil.OperationResu
 		return opResult, err
 	}
 
-	// Get the local cluster agent secret
 	agentSecret := corev1.Secret{}
-	err = s.LocalClient.Get(s.Context, client.ObjectKey{
-		Namespace: constants.VerrazzanoSystemNamespace,
-		Name:      constants.MCAgentSecret,
-	}, &agentSecret)
-	if err != nil {
-		return opResult, err
-	}
-
-	// Update the local cluster agent secret if the admin-kubeconfig or managed-cluster-name have changed
-	if !agentSecretsEqual(agentSecret, adminAgentSecret) {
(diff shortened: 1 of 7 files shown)


application-operator/mcagent/mcagent_cluster_secrets.go
```diff
@@ -33,6 +33,45 @@ func (s *Syncer) syncClusterCAs() (controllerutil.OperationResult, error) {
...
+		Namespace: constants.VerrazzanoMultiClusterNamespace,
+		Name:      getAgentSecretName(s.ManagedClusterName),
+	}, &adminAgentSecret)
+	if err != nil {
+		return opResult, err
+	}
+
+	// Get the local cluster agent secret
+	agentSecret := corev1.Secret{}
+	err = s.LocalClient.Get(s.Context, client.ObjectKey{
```
fredtibbitts, line 54: Why not remove this get and the if == check below and just let CreateOrUpdate do it?

desagar: that's a good point. Will rewrite it that way.

application-operator/mcagent/mcagent_cluster_secrets.go
```diff
@@ -33,6 +33,36 @@ func (s *Syncer) syncClusterCAs() (controllerutil.OperationResult, error) {
...
+	if err != nil {
+		return opResult, err
+	}
+
+	agentSecret := corev1.Secret{}
+	agentSecret.Name = constants.MCAgentSecret
+	agentSecret.Namespace = constants.VerrazzanoSystemNamespace
+	return controllerutil.CreateOrUpdate(s.Context, s.LocalClient, &agentSecret, func() error {
+		// Update the local cluster agent secret if the admin-kubeconfig or managed-cluster-name have changed
+		if !agentSecretsEqual(agentSecret, adminAgentSecret) {
```
fredtibbitts, line 57: I don't think this is necessary, CreateOrUpdate does a deep compare already, so it will only update if the data changes.

fredtibbitts (approved):

fredtibbitts (approved):

PR #4583, opened 2022-11-02
VMO PVC Resize

anders-swanson:
Support dynamic PVC resize with the VMO if the underlying storage class allows it.


stevshap (approved):

theabmitra (approved):
Approved

PR #4586, opened 2022-11-02
WIP: Refresh module POC

michael-cico:


platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -133,3 +136,8 @@ type Component interface {
 
 	Reconcile(ctx ComponentContext) error
 }
+
```
michael-cico, line 139: Need to align this with the new `ComponentUninstaller` interface; possibly no longer need `ModuleComponent`?

PR #4593, opened 2022-11-02
 Get the existing VMC before status updates to avoid conflicts

davisgomes:
This should help to avoid most conflicts seen updating the status of the VMC in the reconcile process.

### DIFF
--- a/platform-operator/controllers/clusters/sync_manifest_secret.go
+++ b/platform-operator/controllers/clusters/sync_manifest_secret.go
@@ -176,7 +176,16 @@ func (r *VerrazzanoManagedClusterReconciler) updateRancherStatus(ctx context.Con
 	vmc.Status.RancherRegistration.Status = status
 	vmc.Status.RancherRegistration.ClusterID = rancherClusterID
 	vmc.Status.RancherRegistration.Message = message
-	err := r.Status().Update(ctx, vmc)
+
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -176,7 +176,16 @@ func (r *VerrazzanoManagedClusterReconciler) updateRancherStatus(ctx context.Con
 	vmc.Status.RancherRegistration.Status = status
 	vmc.Status.RancherRegistration.ClusterID = rancherClusterID
 	vmc.Status.RancherRegistration.Message = message
-	err := r.Status().Update(ctx, vmc)
+
+	// Fetch the existing VMC to avoid conflicts in the status update
+	existingVMC := &clusterapi.VerrazzanoManagedCluster{}
```
desagar, line 181: couldn't we just copy the values directly into the existingVMC instead of copying to the vmc variable and then to existingVMC?

That way you don't even have to pass the vmc variable into this function

desagar (approved):

platform-operator/controllers/clusters/sync_manifest_secret.go
davisgomes: Actually, this will avoid a get request if no updates are required, so I think I'm going to keep it as is

fredtibbitts (approved):

PR #4594, opened 2022-11-02
VZ-7432 PSR first pass at psrctl CLI tool

pfmackin:
Add psrctl explain do describe available scenarios
Add psrctl start to run the example scenario
Add code to extract embedded manifests file to temp dir


tools/psr/psrctl/cmd/explain/explain.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/verrazzano/verrazzano/tools/psr/psrctl/pkg/embedded"
+	"github.com/verrazzano/verrazzano/tools/psr/psrctl/pkg/scenario"
+	cmdhelpers "github.com/verrazzano/verrazzano/tools/vz/cmd/helpers"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/helpers"
+)
+
+const (
+	CommandName = "explain"
+	helpShort   = "Explain a PSR test scenario"
+	helpLong    = `The command 'explain' explains a available scenario that can be started`
```
mbenmclean, line 18: an available

tools/psr/psrctl/cmd/explain/explain.go
```diff
@@ -0,0 +1,50 @@
...
+		return explainCmdExplain(cmd, vzHelper)
+	}
+	cmd.Example = helpExample
+
+	return cmd
+}
+
+// explainCmdExplain - explain the "psrctl explain" command
+func explainCmdExplain(cmd *cobra.Command, vzHelper helpers.VZHelper) error {
+	fmt.Println("Listing available scenarios ...")
```
mbenmclean, line 35: does this command list the help for one scenario or explain all scenarios. This makes it sound like the latter, but the command help above makes it sound like the former. Please clarify the help if this just lists all the scenario. Maybe change to List scenarios?

mbenmclean: Or perhaps have it take an argument that's the scenario name and list only the scenario that matches, but list all of them if no argument is provided?

tools/psr/psrctl/cmd/start/start.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package start
+
```
mbenmclean, line 5: I prefer run as the command name honestly. Any reason why you changed it?

tools/psr/psrctl/cmd/explain/explain.go
pfmackin: All for now, but a single one is coming in future PR
Explain lines up with kubectl explain.  List will show running scenarios.

tools/psr/psrctl/cmd/start/start.go
pfmackin: start / stop are symmetric and i think it is more accurate.  
Run is more of a steady state description (rstart a car then it is running, pod is running, startup probe, etc)

stevshap: I had it as run orginally

tools/psr/psrctl/cmd/explain/explain.go
pfmackin: fixed

tools/psr/psrctl/cmd/explain/explain.go
pfmackin: i fixed the command help

stevshap (approved):

PR #4596, opened 2022-11-02
JIRA:VZ-7591 update rancher user lists as users are deleted

jmaron99:
If there is an issue while deleting the rancher users during the test the user deletion will be re-attempted.  There is a need to remove the deleted users from the list being provided to the deletion method to avoid 404 responses for deletion of users already deleted and the build ultimately failing.


theabmitra (approved):
Approved

PR #4597, opened 2022-11-03
VZ-7541: Increase coverage by adding more UTs for pkg/k8utils

IshaGirdhar:
UT cases added for pkg/k8utils
Previously, Coverage for this package was 43.89 %
Now, With this MR, coverage is around 81 %

### DIFF
--- a/pkg/k8sutil/k8sutil_test.go
+++ b/pkg/k8sutil/k8sutil_test.go
@@ -4,6 +4,7 @@ package k8sutil_test
 
 import (
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common"
 	"net/url"
 	"os"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/k8sutil/k8sutil_test.go
```diff
@@ -308,7 +310,27 @@ func TestExecPod(t *testing.T) {
...
+	spdyfake.PodExecResult = func(url *url.URL) (string, string, error) { return resultString, "", nil }
+	cfg, client := spdyfake.NewClientsetConfig()
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "ns",
+			Name:      "name",
+		},
+	}
+	stdout, _, err := k8sutil.ExecPodNoTty(client, cfg, pod, "container", []string{"run", "some", "command"})
+	assert.Nil(t, err)
```
adiforluls, line 332: Can you add a negative case as well? (i.e. one that will generate an error, eg: execing into a pod that doesn't exist). See if you can add negative tests for all the other tests as well.

anajosep (approved):

pkg/k8sutil/k8sutil_test.go
IshaGirdhar: Added

anajosep (approved):

PR #4612, opened 2022-11-03
VZ 7602: Fluentd MC Configuration moved to component chart

anders-swanson:
The Managed Cluster configuration for fluentd is moved to component chart.


platform-operator/controllers/verrazzano/component/common/managed_cluster.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetManagedClusterRegistrationSecret fetches the managed cluster registraiton secret if it exists.
```
fredtibbitts, line 15: Minor nit, spelling: registraiton.

fredtibbitts (commented):
Just the one very minor comment. I think it would be good to have @desagar and @davis-gomes review this too since I'm not deeply familiar with this code.

platform-operator/controllers/verrazzano/component/fluentd/override.go
```diff
@@ -103,19 +109,38 @@ func appendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
 	return kvs, nil
 }
 
-func appendFluentdOverrides(effectiveCR *vzapi.Verrazzano, overrides *fluentdComponentValues) {
+func appendFluentdLogging(client clipkg.Client, fluentd *vzapi.FluentdComponent, overrides *fluentdComponentValues) error {
+	overrides.Logging = &loggingValues{}
+	registrationSecret, err := common.GetManagedClusterRegistrationSecret(client)
```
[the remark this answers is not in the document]
michael-cico: We try to avoid keying off the profile name.  Is there something else that can be used as an indicator?

platform-operator/controllers/verrazzano/controller.go
```diff
@@ -842,12 +843,41 @@ func (r *Reconciler) watchResources(namespace string, name string, log vzlog.Ver
...
+	err = r.Controller.Watch(
 		&source.Kind{Type: &batchv1.Job{}},
 		createReconcileEventHandler(namespace, name),
 		createPredicate(func(e event.CreateEvent) bool {
 			return r.isMysqlOperatorJob(e, log)
 		}))
+	if err != nil {
+		return err
+	}
+	log.Debugf("Watching for the registration secret to reconcile Verrzzano CR %s/%s", namespace, name)
```
davisgomes, line 855: Similar note here. Do you think it would be beneficial to only add the watch for managed profiles if it's feasible to check?

platform-operator/helm_config/charts/verrazzano-fluentd/values.yaml
```diff
@@ -10,6 +10,11 @@ image:
   terminationGracePeriodSeconds: 60
 
 logging:
+  credentialsSecret: verrazzano-es-internal
+  usernameKey: username
+  passwordKey: password
+  clusterName: local
+  osURL: 'http://verrazzano-authproxy-elasticsearch:8775'
```
davisgomes, line 17: Why is the chart modification necessary? Should these be applied as overrides?

platform-operator/controllers/verrazzano/component/fluentd/override.go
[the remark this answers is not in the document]
anders-swanson: This is the same indicator used by the MCAgent. If there is any other besides the profile name, I am not aware.

platform-operator/helm_config/charts/verrazzano-fluentd/values.yaml
anders-swanson: The chart wasn't flexible enough as written, so I added some extra parameters. Check the YAMLs to see why it was necessary.

platform-operator/controllers/verrazzano/component/fluentd/override.go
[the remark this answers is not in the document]
desagar: The presence of the registration secret is what we usually use to determine what the fluentd information should be. We should not use managed-cluster profile. Anders added a private isManaged function earlier that checks if the reg secret has data, and that's probably a more reasonable approach.

platform-operator/controllers/verrazzano/component/fluentd/override.go
```diff
@@ -103,19 +109,38 @@ func appendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
...
+	if err != nil {
+		return err
+	}
+	if registrationSecret == nil {
```
desagar, line 118: It would be good to have a test case in fluentd_test.go which fakes the presence/absence of the registration secret.

platform-operator/controllers/verrazzano/controller.go
desagar: Same answer as above. We should not restrict by profile. Managed clusters can theoretically use any profile.

platform-operator/helm_config/charts/verrazzano-fluentd/values.yaml
desagar: I have the same q.

desagar (approved):

PR #4617, opened 2022-11-04
VZ-7544: Verify VMI OpenSearch plugins are installed correctly

adiforluls:


anajosep:
Can you reduce the code duplication warning sonar finding ?

adiforluls:
> Can the code duplication warning from sonarcloud be addressed ?

This is odd as duplication reported was 0% last time around. The main logic is already in a separate method which is being called inside Eventually. So further putting util code in a func would be useless but can do that if you think so.

anajosep (approved):

PR #4618, opened 2022-11-04
VZ-5880: Sync cattle-cluster-agent on the managed cluster after DNS update

saketmht:
After a DNS update, Rancher updates the deployment yaml for the cattle-cluster-agent(with the correct url) in the registration manifest for the managed cluster. But this update is not synced to cattle-agent on managed cluster which results in the following error:

`time="2022-06-03T13:45:31Z" level=error msg="Remotedialer proxy error" error="x509: certificate is valid for rancher.default.sslip.io, not rancher.default.172.18.0.231.nip.io"`
`time="2022-06-03T13:45:32Z" level=error msg="Failed to dial steve aggregation server: x509: certificate is valid for rancher.default.sslip.io, not rancher.default.172.18.0.231.nip.io"`

This PR is to patch the cattle-cluster-agent when the deployment changes in the registration manifest on the managed cluster.


application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+
+	config, err := k8sutil.BuildKubeConfig(kubeconfigPath)
+	if err != nil {
+		log.Errorf("failed to create incluster config: %v", err)
+		return err
+	}
+	log.Debugf("Built kubeconfig: %s, now applying manifest", config.Host)
+
+	// Data[10] contains the yaml for the cattle-cluster-agent
+	//err := resource.CreateOrUpdateResourceFromBytes(data[10], log)
```
fredtibbitts, line 78: Remove commented out code.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+	patch, err := yaml.ToJSON(data[10])
```
fredtibbitts, line 79: Relying on the order/position of the resources in the YAML is very fragile. If the order changes then all off this code could break. It would be much better to not assume the order and look for the specific resources you're interested in.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+
+	gvr := schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
+	err = resource.PatchResourceFromBytes(gvr, types.StrategicMergePatchType, "cattle-system", "cattle-cluster-agent", patch, config)
+	if err != nil {
+		log.Errorf("failed to patch cattle-cluster-agent: %v", err)
+		return err
+	}
+
+	// Data[8] contains the yaml for the cattle-credential used by the cattle-cluster-agent
+	err = resource.CreateOrUpdateResourceFromBytesUsingConfig(data[8], config)
```
fredtibbitts, line 93: Should the secret be updated before the deployment is patched?

application-operator/mcagent/mcagent_cattle_agent_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// TestSyncer_syncCattleClusterAgent tests the syncCattleClusterAgent of Syncer
+// WHEN trying to sync the cattle-cluster-agent
```
fredtibbitts, line 26: These comments should use GIVEN/WHEN/THEN.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to fetch manifest secret for %s cluster: %v", s.ManagedClusterName, err)
+	}
+	s.Log.Debugf(fmt.Sprintf("Found manifest secret for %s cluster: %s", s.ManagedClusterName, manifestSecret.Name))
+
+	manifestData := manifestSecret.Data["yaml"]
+	yamlSlices := bytes.Split(manifestData, []byte("---\n"))
+
+	if len(yamlSlices) < 12 {
```
fredtibbitts, line 38: See other comment, this seems too fragile.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/verrazzano/verrazzano/pkg/k8s/resource"
+	"github.com/verrazzano/verrazzano/pkg/k8sutil"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (s *Syncer) syncCattleClusterAgent(kubeconfigPath string) error {
```
fredtibbitts, line 23: Add a function comment that describes what is being sync'ed.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+		return fmt.Errorf("failed to update the cattle-cluster-agent on %s cluster: %v", s.ManagedClusterName, err)
+	}
+
+	s.CattleAgentHash = cattleAgentHash
+	s.Log.Infof("Successfully synched cattle-cluster-agent")
+
+	return nil
+}
+
+func updateCattleAgent(data [][]byte, log *zap.SugaredLogger, kubeconfigPath string) error {
```
fredtibbitts, line 68: Add function comment.

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+	"crypto/sha256"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
+
+	"github.com/verrazzano/verrazzano/pkg/k8s/resource"
+	"github.com/verrazzano/verrazzano/pkg/k8sutil"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
```
desagar, line 17: The VerrazzanoMultiClusterNamespace constant is available in the application-operator directory as well as from pkg/constants at the top level. Better to use one of those than the one from platform-operator. 

There is a top level pkg/constants where the same constant is present as well. At some point we should refactor and remove the shared constant from platform operator and application operator, but that will really balloon the scope of this PR. For now I'd say just use the top level one.

application-operator/mcagent/mcagent_cattle_agent.go
desagar: completely agree with this. You can consider using the [gabs](https://github.com/Jeffail/gabs) parser to parse the individual yaml sections and query the gabs.Container object to check whether it is the cattle-cluster-agent deployment. 
We use the gabs parser elsewhere in the code - examples are [here](https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/clusters/vmc_controller_test.go#L2136) and [here](https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/clusters/rancher.go#L256)

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+
+	manifestData := manifestSecret.Data["yaml"]
+	yamlSlices := bytes.Split(manifestData, []byte("---\n"))
+
+	if len(yamlSlices) < 12 {
+		s.Log.Debugf("The registration manifest doesn't have all the resources. Will try to update the cattle-cluster-agent in the next iteration")
+		return nil
+	}
+
+	cattleAgentSlice := yamlSlices[10]
```
desagar, line 43: minor point - the word "slice" in Golang means roughly an array. I would suggest using a different name for these things ("Section" instead of "Slice" for example)

application-operator/mcagent/mcagent_cattle_agent.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Debugf("Successfully patched cattle-cluster-agent and created a new cattle-credential secret")
+
+	return nil
+}
+
+func createHash(data []byte) string {
+	// The cattle-credential-xxxxx contains a short-lived token
+	// It changes every few pulls of the rancher manifest even if there is no change in the cattle-agent
+	// So remove the secret name before creating the hash
+	split := bytes.SplitAfter(data, []byte("secretName: "))
```
desagar, line 107: why split? Why not hash the whole thing?

application-operator/mcagent/mcagent_cattle_agent.go
saketmht: Manually it didn't make a difference which was done first. It worked both ways. The cattle-cluster-agent deployment stays in ContainerCreating state while the secret is not created and then starts successfully when the secret is created. So I think its better to create the secret after the deployment is patched

application-operator/mcagent/mcagent_cattle_agent.go
saketmht: The hash is being created from the cattle-cluster-agent deployment instead of the cattle secret. The deployment also contains the secret name(which keeps changing because of the token) so removing that part before creating the hash.

PR #4623, opened 2022-11-04
Re-introduce authproxy strategy to set maxUnavailable to 0 to force create-before-delete on RollingUpdates

michael-cico:
Re-introduce change to set authproxy strategy maxUnavailable to 0 to force create-before-delete semantics on RollingUpdate.

This also changes the authproxy update tests to account for maxSurge during rolling update; the tests were not filtering out unschedulable nodes, and when the behavior changed to spin up new pods before deleting any existing ones it was throwing off the counts.

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
@@ -40,7 +40,7 @@ spec:
   strategy:
     rollingUpdate:
       maxSurge: 1
-      maxUnavailable: 1
+      maxUnavailable: 0
     type: RollingUpdate
   template:
     metadata:
--- a/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+# Copyright (c) 2021, 2023, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/tests/e2e/update/authproxy/authproxy_update_test.go
+++ b/tests/e2e/update/authproxy/authproxy_update_test.go
@@ -5,10 +5,10 @@ package authproxy
 
 import (
 	"fmt"
+	"sigs.k8s.io/yaml"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
-	"github.com/onsi/gomega"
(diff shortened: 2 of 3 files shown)


tests/e2e/update/authproxy/authproxy_update_test.go
```diff
@@ -175,12 +180,8 @@ var _ = t.Describe("Update authProxy", Label("f:platform-lcm.update"), func() {
 			m := AuthProxyPodPerNodeAffintyModifier{}
 			update.UpdateCRWithRetries(m, pollingInterval, waitTimeout)
 
-			expectedRunning := nodeCount - 1
+			expectedRunning := nodeCount
```
guoyongzhang, line 183: did something change when there is more than one node in the cluster?  auth-proxy is created in control plane node too?

tests/e2e/update/authproxy/authproxy_update_test.go
```diff
@@ -175,12 +180,8 @@ var _ = t.Describe("Update authProxy", Label("f:platform-lcm.update"), func() {
...
 			expectedPending := true
-			if nodeCount == 1 {
-				expectedRunning = nodeCount
-				expectedPending = false
```
guoyongzhang, line 182: There is auth-proxy pod pending even in single-node cluster?

tests/e2e/update/authproxy/authproxy_update_test.go
michael-cico: I didn't change anything really in the tests, but the new strategy forces the new pods to come up before deleting the old ones.

Maybe we should sync up on this, I may not fully understand what the test configuration here.

tests/e2e/update/authproxy/authproxy_update_test.go
michael-cico: See previous comment

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
```diff
@@ -40,7 +40,7 @@ spec:
   strategy:
     rollingUpdate:
       maxSurge: 1
-      maxUnavailable: 1
+      maxUnavailable: 0
```
shihchang, line 43: During update, this will allow the authproxy to be temporarily unavailable, is that the expected effect?

michael-cico: This actually should prevent authproxy from becoming unavailable.

michael-cico: I should say, the combination of `maxSurge: 1` and `maxUnavailable: 0` is what prevents downtime.  If there's only 1 replica available with `maxUnavailable: 1` the existing pod can be deleted before the new one becomes available.

michael-cico: https://www.bluematador.com/blog/kubernetes-deployments-rolling-update-configuration

mgianatagh (approved):

shihchang (approved):

PR #4626, opened 2022-11-04
Refactor updater test API to test common pkg

michael-cico:
Moves the e2e test update interfaces to a common package to share with tests outside the `tests/e2e/update` package.


guoyongzhang (approved):
LGTM

PR #4633, opened 2022-11-04
Remove WebLogic types and replace with Unstructured

rjeberhard:
The WebLogic workload controller isn't using these types and they aren't up to date. As a first step toward updating V8o to use WKO 4.0, this PR removes these pre-existing v8 types and instead treats the Domain resource content as Unstructured.


pfmackin (approved):

bobdonat (approved):
LGTM

PR #4640, opened 2022-11-04
Fix Mysql Values validating webhook when  there are multiple overrides, Spec.Version field isn't set

michael-cico:
Addresses the case for the Mysql values validating webhook when the `spec.version` field isn't set, also the case where the `podSpec` override is not the first one in the array.
- Need to use the VZ status field if the new spec.version is empty, and fall back to the internal BOM version if neither is set
- fix the MySQL values validators to loop through the overrides looking for podspec changes,
instead of just in the first item in the array.
- Add MySQL warning validator e2e tests
- Re-introduce component validators sanity check test, wrap in `Eventually` assertions to account for spurious errors


mgianatagh (approved):

stevshap (approved):

platform-operator/apis/verrazzano/webhooks/mysql_values_webhook_v1alpha1_test.go
```diff
@@ -111,19 +156,26 @@ func TestNoValidationWarningWithoutServerPodSpecV1alpha1(t *testing.T) {
...
 				},
 			},
 		},
 	}
 
-	req := newAdmissionRequest(admissionv1.Update, newVz)
+	req := newAdmissionRequest(admissionv1.Update, newVz, &v1alpha1.Verrazzano{})
 	res := m.Handle(context.TODO(), req)
 	asrt.True(res.Allowed, "Expected request to be allowed with warnings")
 	asrt.Len(res.Warnings, 0, "Expected there to be one warning")
```
ayaranga, line 181: nit: Assert messages need to be corrected. Not related to your change though.

michael-cico: Fixed and pushed.

michael-cico: nice catch

mgianatagh (approved):

PR #4645, opened 2022-11-07
VZ-7559: Add unit test cases for rancher backup component

gaurav-05:
Added unit tests for Rancher Backup.
Changed from 60.5% -> 88.2%


anajosep (dismissed):

platform-operator/controllers/verrazzano/component/rancherbackup/rancher_backup_component_test.go
```diff
@@ -166,6 +166,140 @@ func TestInstallUpgrade(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestValidateUpdateMethods(t *testing.T) {
```
adiforluls, line 169: Add func comments like this everywhere
```
// TestXYZ tests the <func-name> method for the following scenario(s)
// GIVEN ....
// WHEN ....
// THEN ....
```

gaurav-05: I did check this but as checked few other functions this did not appear to me a standard way so ignored. I can add if this is a standard practice.

adiforluls: Yeah this is a standard practice I believe. Probably those funcs were missed during review.

adiforluls (approved):

PR #4652, opened 2022-11-07
VZ-6469 update the cluster syncronization e2e tests and fix some multicluster registration bugs

davisgomes:


ci/multicluster/Jenkinsfile
```diff
@@ -949,10 +954,11 @@ def verifyDeregisterManagedClusters() {
...
                     export MANAGED_CLUSTER_NAME="managed${count-1}"
                     export MANAGED_KUBECONFIG="${KUBECONFIG_DIR}/${count}/kube_config"
-                    export KUBECONFIG="${KUBECONFIG_DIR}/${count}/kube_config"
-                    kubectl -n verrazzano-system delete secret verrazzano-cluster-registration
+                    export ADMIN_KUBECONFIG="${KUBECONFIG_DIR}/1/kube_config"
+                    ./tests/e2e/config/scripts/deregister_managed_cluster.sh
                     cd ${GO_REPO_PATH}/verrazzano/tests/e2e
-                    ginkgo -p --randomize-all -v --keep-going --no-color ${GINKGO_REPORT_ARGS} -tags="${params.TAGGED_TESTS}" --focus-file="${params.INCLUDED_TESTS}" --skip-file="${params.EXCLUDED_TESTS}" multicluster/verify-deregister/...
+                    ginkgo build multicluster/verify-deregister/
+                    ginkgo -p --randomize-all -v --keep-going --no-color ${GINKGO_REPORT_ARGS} -tags="${params.TAGGED_TESTS}" --focus-file="${params.INCLUDED_TESTS}" --skip-file="${params.EXCLUDED_TESTS}" multicluster/verify-deregister/*.test
```
desagar, line 961: What's the difference between using verify-deregister/... versus using *.test?
Also, we should be able to do the ginkgo build once outside the loop instead of each time. Abi made a similar change recently and reported that it worked.

platform-operator/controllers/clusters/sync_manifest_secret.go
```diff
@@ -154,7 +154,10 @@ func (r *VerrazzanoManagedClusterReconciler) syncCACertSecret(vmc *clusterapi.Ve
 			vmc.Spec.CASecret = caSecretName
 			// update the VMC with ca secret name
 			r.log.Infof("Updating VMC %s with managed cluster CA secret %s", vmc.Name, caSecretName)
-			err = r.Update(context.TODO(), vmc)
+			// Replace the update call with a new pointer
```
desagar, line 157: this comment is a bit misleading. I think you meant something like "Replace the VMC in the update call with a copy. That way the existing VMC status updates do not get overwritten by the update.

platform-operator/controllers/clusters/vmc_controller.go
```diff
@@ -445,7 +445,9 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
 	}
 
 	// Replace the existing status conditions and state with the conditions generated from this reconcile
-	existingVMC.Status.Conditions = vmc.Status.Conditions
+	for _, genCondition := range vmc.Status.Conditions {
+		r.setStatusCondition(existingVMC, genCondition, genCondition.Type == clustersv1alpha1.ConditionManifestPushed)
```
desagar, line 449: why is manifest pushed the only condition for which we need to update on time change? The Ready condition time could also change, couldn't it, based on updates?

davisgomes: This was the original implementation. Since we set the ready time every reconcile, we only want to update the time if the status has changed. The time reflects the most recent time the condition changed to ready.

tests/e2e/multicluster/verify-cluster-sync/cluster_sync_test.go
```diff
@@ -0,0 +1,257 @@
...
+	deleted, err := clusters.DeleteClusterFromRancher(rc, clusterID, vzlog.DefaultLogger())
+	Expect(err).ShouldNot(HaveOccurred())
+	Expect(deleted).To(BeTrue())
+
+	// Eventually, a VMC with that cluster name should be deleted
+	Eventually(func() bool {
+		pkg.Log(pkg.Info, "Waiting for VMC to be deleted")
+		_, err := client.ClustersV1alpha1().VerrazzanoManagedClusters(constants.VerrazzanoMultiClusterNamespace).Get(context.TODO(), clusterName, metav1.GetOptions{})
+		return errors.IsNotFound(err)
+	}).WithPolling(pollingInterval).WithTimeout(waitTimeout).Should(BeTrue())
```
desagar, line 177: For the deletion cases, do you think we should add a Consistently at the end to verify that it stays deleted for some time?

platform-operator/controllers/clusters/sync_manifest_secret.go
davisgomes: Thanks. this is much clearer

tests/e2e/multicluster/verify-cluster-sync/cluster_sync_test.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// testRancherClusterDeletion tests a cluster deleted in Rancher
+func testRancherClusterDeletion(rc *clusters.RancherConfig, client *versioned.Clientset, clusterName, clusterID string) {
+	// GIVEN a Rancher cluster is deleted using Rancher API/UI
+	// WHEN the Rancher cluster is appropriately labeled
+	// THEN the VMC for the cluster is deleted
+
+	// The VMC should have the clusterID field set before we attempt to delete
+	Eventually(func() bool {
```
desagar, line 163: I know you added this wait for the cluster id to be populated in the VMC for a good reason, but users might not wait. What happens if we don't wait? Is there anything we can do to improve the user experience?

desagar (dismissed):

tests/e2e/multicluster/verify-cluster-sync/cluster_sync_test.go
davisgomes: If they don't wait, the cluster gets recreated. The likelihood that the customer will continuously run into this issue is low. At worst, they might have to delete a cluster twice.

tests/e2e/multicluster/verify-cluster-sync/cluster_sync_test.go
davisgomes: Yea, I think this is a good idea. I will add this on

ci/multicluster/Jenkinsfile
davisgomes: I don't see any examples of this outside the for loop. I'm going to copy the upgrade-paths implementation for consistency.

desagar (approved):

fredtibbitts (approved):

PR #4656, opened 2022-11-07
VZ 7636: Refactor Jaeger Operator out of MCAgent

anders-swanson:
Any Jaeger Operator functionality in the MCAgent is moved to the component, so only the VPO (the owner of Jaeger Operator) is configured those resources.


anajosep:
Old unit test is removed, can that be modified, and moved to the new package for jaeger ?

ayaranga (approved):

ayaranga (dismissed):

PR #4671, opened 2022-11-08
VZ-7554: Stop logging large stack traces while components are waiting

marcogelle:
Fluentd logs a large stack trace while waiting for the creation of the `verrazzano-es-internal` secret during preinstall. Jaeger Operator and Istio also log large stack traces while waiting, but they are during postinstall and install, respectively. This PR aims to prevent these stack traces from filling up the main log.

Fluentd and Jaeger Operator already have progress logs when waiting in these scenarios. Istio did not, so this PR also adds a progress log for when Istio returns a `RetryableError` due to needing to wait.

### DIFF
--- a/platform-operator/controllers/verrazzano/reconcile/install_component.go
+++ b/platform-operator/controllers/verrazzano/reconcile/install_component.go
@@ -150,7 +150,11 @@ func (r *Reconciler) installSingleComponent(spiCtx spi.ComponentContext, install
 			// If component is not installed,install it
 			compLog.Oncef("Component %s install started ", compName)
 			if err := comp.Install(compContext); err != nil {
-				compLog.ErrorfThrottled("Error running Install for component %s: %v", compName, err)
+				// Only log this error if this is not a RetryableError
+				if _, ok := err.(spi2.RetryableError); !ok {
+					compLog.ErrorfThrottled("Error running Install for component %s: %v", compName, err)
+				}
+
 				return ctrl.Result{Requeue: true}
 			}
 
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/reconcile/install_component.go
```diff
@@ -4,6 +4,7 @@
 package reconcile
 
 import (
+	spi2 "github.com/verrazzano/verrazzano/pkg/controller/errors"
```
anders-swanson, line 7: why is it called `spi2`?

marcogelle: Actually, I didn't pay attention to that; it was just what my IDE imported it as. Looking at other places where this is imported, I think I'll change the name to `ctrlerrors`.

platform-operator/controllers/verrazzano/reconcile/install_component.go
```diff
@@ -135,7 +136,11 @@ func (r *Reconciler) installSingleComponent(spiCtx spi.ComponentContext, install
 			}
 			compLog.Progressf("Component %s pre-install is running ", compName)
 			if err := comp.PreInstall(compContext); err != nil {
-				compLog.ErrorfThrottled("Error running PreInstall for component %s: %v", compName, err)
+				// Only log this error if this is not a RetryableError
+				if _, ok := err.(ctrlerrors.RetryableError); !ok {
```
anders-swanson, line 140: suggest to create function `IsRetryableError(err error) bool` in the ctrlerrors package? It would do same thing as this conditional.

marcogelle: I think that's a good idea. I'll make that change now.

anders-swanson (approved):
LGTM

PR #4685, opened 2022-11-09
VZ-7662 initialize the Helm chart for the Verrazzano Cluster Operator

davisgomes:
Template helm chart for the Verrazzano Cluster Operator

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-cluster-operator/.helmignore
+++ b/platform-operator/helm_config/charts/verrazzano-cluster-operator/.helmignore
@@ -1,0 +1,22 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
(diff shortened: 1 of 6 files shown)


platform-operator/helm_config/charts/verrazzano-cluster-operator/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: v2
```
desagar, line 3: the app operator and platform operator have apiVersion as v1 in their charts. Should we use the same here?

platform-operator/helm_config/charts/verrazzano-cluster-operator/.helmignore
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
desagar, line 1: The helmignore files in the VPO and VAO charts are much smaller. Any reason why we need all these entries?

davisgomes: This was the default from helm create. I don't think it hurts to include these.

platform-operator/helm_config/charts/verrazzano-cluster-operator/Chart.yaml
davisgomes: Yes we should. I'm not sure why the helm chart generated this way.

desagar (approved):

PR #4690, opened 2022-11-09
VZ 7552: Availability is non-binary

anders-swanson:
Make component status availability a non-binary value.

### DIFF
--- a/platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
@@ -158,8 +158,8 @@ type ComponentStatusMap map[string]*ComponentStatusDetails
 
 // ComponentStatusDetails defines the observed state of a component.
 type ComponentStatusDetails struct {
-	// Whether or not a component is available for use.
-	Available *bool `json:"available,omitempty"`
+	// Whether a component is available for use.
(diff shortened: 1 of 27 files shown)


platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
```diff
@@ -224,6 +224,16 @@ type Condition struct {
 	Type ConditionType `json:"type"`
 }
 
+// ComponentAvailability identifies the availability of a Verrazzano Component.
+type ComponentAvailability string
+
+const (
+	//ComponentAvailable signifies that a Verrazzano Component is ready for use.
+	ComponentAvailable = "Available"
+	//ComponentUnavailable signifies that a Verrazzano Component is not ready for use.
```
michael-cico, line 233: Does this mean "unavailable needs reconcile", like we discussed in the meeting?  So, we may have another value for "unavailable, do nothing"?

michael-cico (approved):

platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
anders-swanson: Unavailable means health check is failing, and the Verrazzano Component is not ready for use.

rosemarymarano (approved):
LGTM

PR #4698, opened 2022-11-10
VZ-7478 Add prerequisite validator webhook

ayaranga:
A webhook has been added to check the Verrazzano prerequisites and present an appropriate warning to the user while applying the Verrazzano custom resource.

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -154,24 +154,6 @@ func GetKubernetesClientsetWithConfig(config *rest.Config) (*kubernetes.Clientse
 	return clientset, err
 }
 
-// GetKubernetesNodeList returns a list of Kubernetes nodes
-func GetKubernetesNodeList(k8sClient client.Client) (*v1.NodeList, error) {
-	nodeList := &v1.NodeList{}
-	err := k8sClient.List(
-		context.TODO(),
-		nodeList,
-		&client.ListOptions{},
-	)
-	if err != nil {
-		return nil, err
-	}
-	nodes := map[string]bool{}
-	for _, node := range nodeList.Items {
-		nodes[node.GetName()] = true
-	}
-	return nodeList, nil
-}
-
 // GetCoreV1Func is the function to return the CoreV1Interface
 var GetCoreV1Func = GetCoreV1Client
 
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -154,6 +154,24 @@ func GetKubernetesClientsetWithConfig(config *rest.Config) (*kubernetes.Clientse
 	return clientset, err
 }
 
+// GetKubernetesNodeList returns a list of Kubernetes nodes
+func GetKubernetesNodeList(k8sClient client.Client) (*v1.NodeList, error) {
+	nodeList := &v1.NodeList{}
+	err := k8sClient.List(
+		context.TODO(),
+		nodeList,
+		&client.ListOptions{},
+	)
+	if err != nil {
+		return nil, err
+	}
+	nodes := map[string]bool{}
+	for _, node := range nodeList.Items {
+		nodes[node.GetName()] = true
+	}
+	return nodeList, nil
+}
+
 // GetCoreV1Func is the function to return the CoreV1Interface
 var GetCoreV1Func = GetCoreV1Client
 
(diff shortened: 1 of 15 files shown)


pkg/k8sutil/k8sutil.go
```diff
@@ -154,6 +154,24 @@ func GetKubernetesClientsetWithConfig(config *rest.Config) (*kubernetes.Clientse
 	return clientset, err
 }
 
+// GetKubernetesNodeList returns a list of Kubernetes nodes
```
pfmackin, line 157: put this code in a new package/file   verrazzzano/pkg/k8s/node/node.go

pkg/vzchecks/requirement.go
```diff
@@ -0,0 +1,82 @@
...
+		ephemeralStorage: &resourceInfo{resourceType: EphemeralStorage, allocatable: resource.MustParse("100G")},
+	}
+	ProdProfile = VZRequirement{
+		nodeCount:        3,
+		cpu:              &resourceInfo{resourceType: CPU, allocatable: resource.MustParse("4")},
+		memory:           &resourceInfo{resourceType: Memory, allocatable: resource.MustParse("32G")},
+		ephemeralStorage: &resourceInfo{resourceType: EphemeralStorage, allocatable: resource.MustParse("100G")},
+	}
+	ManagedProfile = VZRequirement{
+		nodeCount:        3,
```
pfmackin, line 64: I think this is too high for managed cluster.  Customers might have edge clusters that are very lightweight single node.  Count should be 1, but double check with Kevin.  Also, memory seems high but maybe it is ok.

pkg/k8sutil/k8sutil.go
ayaranga: Done. Thanks.

pkg/vzchecks/requirement.go
ayaranga: I have changed it to 1 and checking with Kevin as well.

pkg/vzchecks/requirement.go
```diff
@@ -0,0 +1,82 @@
...
+		cpu:              &resourceInfo{resourceType: CPU, allocatable: resource.MustParse("4")},
+		memory:           &resourceInfo{resourceType: Memory, allocatable: resource.MustParse("32G")},
+		ephemeralStorage: &resourceInfo{resourceType: EphemeralStorage, allocatable: resource.MustParse("100G")},
+	}
+	profileMap = map[ProfileType]VZRequirement{
+		Dev:            DevProfile,
+		Prod:           ProdProfile,
+		ManagedCluster: ManagedProfile,
+	}
+)
```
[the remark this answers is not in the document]
ayaranga: I had defaulted to prod in precheck file. Made a change to check in this file as well.

pkg/vzchecks/requirement.go
kminder: Should be 1.

pkg/vzchecks/precheck.go
```diff
@@ -0,0 +1,59 @@
...
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
+)
+
+// PrerequisiteCheck checks the prerequisites before applying the Verrazzano CR
+func PrerequisiteCheck(client clipkg.Client, profile ProfileType) []error {
+	return preCheck(client, profile)
+}
+
+func preCheck(client clipkg.Client, profile ProfileType) []error {
```
kminder, line 20: This probably shouldn't be using profile as the primary means to determine what the resource requirements are going to be.  We should discuss.

pkg/vzchecks/requirement.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	CPU              = ResourceType(k8score.ResourceCPU)
+	Memory           = ResourceType(k8score.ResourceMemory)
+	EphemeralStorage = ResourceType(k8score.ResourceEphemeralStorage)
+)
+
+const (
+	nodeCountReqMsg = "minimum required number of worker nodes for %s profile is %d but the available number of worker nodes is %d"
```
kminder, line 44: At a minimum take the profile statement out of these messages.

ayaranga: Done

pkg/vzchecks/precheck.go
ayaranga: As discussed, we will take this up as a separate incremental PR

kminder (approved):
Approved as an intermediate solution.  This needs to be followed up such that the requirements are derived from the enabled components and their configuration rather than relying on the base profile name alone.

PR #4701, opened 2022-11-10
VZ-7524: VPO workaround for MySQL pods stuck waiting for readiness gates

mgianatagh:
Temporary workaround for issue with MySQL where pods can be stuck waiting for readiness gates during install (and possibly upgrade too).


michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -184,13 +187,78 @@ func (c mysqlComponent) isMySQLReady(ctx spi.ComponentContext) bool {
...
+	// Initiate repair only if time to wait period has been exceeded
+	expiredTime := c.LastTimeReadinessGateRepairStarted.Add(5 * time.Minute)
+	if time.Now().After(expiredTime) {
+		// Check if the current not ready state is due to readiness gates not met
+		ctx.Log().Debug("Checking if MySQL not ready due to pods waiting for readiness gates")
+
+		selector := metav1.LabelSelectorRequirement{Key: mySQLComponentLabel, Operator: metav1.LabelSelectorOpIn, Values: []string{mySQLDComponentName}}
+		podList := k8sready.GetPodsList(ctx.Log(), ctx.Client(), types.NamespacedName{Namespace: ComponentNamespace}, &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{selector}})
+		if podList == nil || len(podList.Items) == 0 {
+			return fmt.Errorf("no pods found matching selector %s", selector.String())
```
pfmackin, line 222: should be "Failed, no pods
Also might be good to have more context, like Failed checking MySQL readiness because no pods...

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -184,13 +187,78 @@ func (c mysqlComponent) isMySQLReady(ctx spi.ComponentContext) bool {
...
+
+			// All readiness gates must be true
+			if len(pod.Spec.ReadinessGates) != readyCount {
+				// Restart the mysql-operator to see if it will finish setting the readiness gates
+				ctx.Log().Info("Restarting the mysql-operator to see if it will repair MySQL pods stuck waiting for readiness gates")
+
+				operSelector := metav1.LabelSelectorRequirement{Key: "name", Operator: metav1.LabelSelectorOpIn, Values: []string{mysqloperator.ComponentName}}
+				operPodList := k8sready.GetPodsList(ctx.Log(), ctx.Client(), types.NamespacedName{Namespace: mysqloperator.ComponentNamespace}, &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{operSelector}})
+				if operPodList == nil || len(podList.Items) != 1 {
+					return fmt.Errorf("no pods found matching selector %s", selector.String())
```
pfmackin, line 250: same as above

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -184,13 +187,78 @@ func (c mysqlComponent) isMySQLReady(ctx spi.ComponentContext) bool {
...
+					}
+				}
+			}
+
+			// All readiness gates must be true
+			if len(pod.Spec.ReadinessGates) != readyCount {
+				// Restart the mysql-operator to see if it will finish setting the readiness gates
+				ctx.Log().Info("Restarting the mysql-operator to see if it will repair MySQL pods stuck waiting for readiness gates")
+
+				operSelector := metav1.LabelSelectorRequirement{Key: "name", Operator: metav1.LabelSelectorOpIn, Values: []string{mysqloperator.ComponentName}}
```
pfmackin, line 247: looks like common code as above, maybe move to func

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -184,13 +187,78 @@ func (c mysqlComponent) isMySQLReady(ctx spi.ComponentContext) bool {
...
 
+// repairMySQLPodsWaitingReadinessGates - temporary workaround to repair issue were a MySQL pod
+// can be stuck waiting for its readiness gates to be met.
+func (c mysqlComponent) repairMySQLPodsWaitingReadinessGates(ctx spi.ComponentContext) error {
+	if c.LastTimeReadinessGateRepairStarted.IsZero() {
+		*c.LastTimeReadinessGateRepairStarted = time.Now()
+		return nil
+	}
+
+	// Initiate repair only if time to wait period has been exceeded
```
pfmackin, line 213: Would it make more sense to measure the time from MySQL being ready to when the readiness state is set.  In other words, it doesn't hurt to check for readiness immediately since controller-runtime has a cache and it is very cheap.  Once MySQL is really ready, then the clock starts ticking before you restart the mysql operator.

mgianatagh: I don't feel that approach would work.  The problem is MySQL is not going ready because the readiness gates are not being met.

platform-operator/controllers/verrazzano/component/mysql/mysql.go
mgianatagh: A common function is already being used here (GetPodList).  The code around it is to setup the input parameters, and then check for the desired result for the context the call is being used.

pfmackin (dismissed):

michael-cico (approved):

PR #4716, opened 2022-11-14
VZ-7732: Increase unit test coverage for application-operator.controllers.navigation

saketmht:
Increased coverage from 68.05% to 94.6%


saketmht:
> Can you address the 65.9% code duplication reported by sonar

Most of the duplication is coming from the similar EXPECT() calls for the mock client. Most of them are needed to test different scenarios in the code.

anajosep (approved):

PR #4724, opened 2022-11-14
VZ-7666 part1 - Refactoring - move enabled checks in VZ CR to a top level 'vz' package

desagar:


mbenmclean (approved):

pkg/vz/enabled.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2021, 2022, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-package vzconfig
+package vz
```
pfmackin, line 4: vz is too obscure.  Can it be vzcr or something?  I have some vzcr code for PSR that i can move here.

pfmackin (approved):

PR #4729, opened 2022-11-14
VZ-7658 added unit tests for file.go

alyssavlev:

### DIFF
--- a/pkg/os/file_test.go
+++ b/pkg/os/file_test.go
@@ -3,11 +3,41 @@
 package os
 
 import (
-	"github.com/stretchr/testify/assert"
-	"go.uber.org/zap"
+	"fmt"
 	"os"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


pkg/os/file_test.go
```diff
@@ -3,12 +3,42 @@
...
+	}{
+		{name: "WithPrefixAndSuffix", createPattern: "test1-*.yaml"},
+		{name: "WithPrefixOnly", createPattern: "test2-*"},
+		{name: "WithSuffixOnly", createPattern: "*-test3"},
+		{name: "NoPrefixNorSuffix", createPattern: "test4"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			assert := assert.New(t)
```
desagar, line 32: I would suggest adding a deletion of the tempfiles as follows: defer RemoveTempfile(temp.Name()) 
This line can be right after calling CreateTempFile

desagar (approved):

PR #4730, opened 2022-11-15
Enhance UT coverage for opensearch component

sk2701:
Coverage increased from 73.89% to 93.81%


adiforluls (commented):
Can you get rid of the 5 code smells? Also review whether the duplication can be brought down.

sk2701:
> Can you get rid of the 5 code smells? Also review whether the duplication can be brought down.

Removed the code smells, don't think that the duplication can be reduced further because all the code is needed

adiforluls (approved):

PR #4735, opened 2022-11-15
VZ-7759 Add helm configmap to transfer install overrides to the VCO

davisgomes:
The install overrides for the label selector need to be communicated to the VCO. Through this configmap, we can determine the state of the install overrides and correctly apply the label selector

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/configmap.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/configmap.yaml
@@ -13,4 +13,3 @@ data:
 {{ toYaml .Values.syncRancherClusters.clusterSelector | indent 4 }}
   {{- end -}}
 {{- end -}}
-
--- a/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/configmap.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/configmap.yaml
@@ -1,0 +1,14 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+{{- if .Values.syncRancherClusters.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.name }}-selector
+  namespace: {{ .Values.namespace }}
+data:
+  {{- if .Values.syncRancherClusters.clusterSelector -}}
+  selector.yaml: {{ .Values.syncRancherClusters.clusterSelector }}
+  {{- end -}}
+{{- end -}}
+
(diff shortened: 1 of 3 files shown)


platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/configmap.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+{{- if .Values.syncRancherClusters.enabled -}}
```
desagar, line 3: isn't this always going to be defined as false unless overridden, since it's in values.yaml? Do we need this if statement?

desagar: oh, I see. If enabled is false we don't write the CM

platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/deployment.yaml
```diff
@@ -21,4 +21,21 @@ spec:
...
         imagePullPolicy: {{ .Values.imagePullPolicy }}
+        env:
+        - name: CLUSTER_SYNC_ENABLED
+          valueFrom:
+            configMapKeyRef:
+              name: {{ .Values.name }}-selector
+              key: enabled
+        volumeMounts:
+          - name: cluster-selector
+            mountPath: /var/selector
```
desagar, line 32: I'd like to have the word "cluster" and/or "rancher" somewhere in either the filename or the mount path. Maybe /var/syncRancherClusters for the path and selector.yaml for the file name? 
That way when we reference it in the code it is clear what it is.

platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/deployment.yaml
```diff
@@ -21,4 +21,21 @@ spec:
       - name: {{ .Values.name }}
         image: {{ .Values.image }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
+        env:
+        - name: CLUSTER_SYNC_ENABLED
```
desagar, line 25: RANCHER_CLUSTER_SYNC_ENABLED?

platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/deployment.yaml
davisgomes: Updated

platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/deployment.yaml
davisgomes: Updated this as well

desagar (approved):

PR #4737, opened 2022-11-15
Update metrics location to fileName:lineNumber

anders-swanson:


tests/e2e/pkg/test/framework/metrics/ginkgo_metrics.go
```diff
@@ -104,38 +103,44 @@ func NewLogger(pkg string, ind string, paths ...string) (*zap.SugaredLogger, err
...
-		log = log.With(JenkinsJob, jobPipeline)
-	}
-	gitCommit := os.Getenv("GIT_COMMIT")
-	//Tagging commit with the branch.
-	if gitCommit != "" {
-		gitCommitAndBranch := branchName + "/" + gitCommit
-		log = log.With(CommitHash, gitCommitAndBranch)
-	}
+	log, _ = withEnvVarMutate(log, BuildURL, "BUILD_URL", func(buildURL string) string {
+		return strings.Replace(buildURL, urlReplace, separator, 1)
```
fredtibbitts, line 113: Why not use https://pkg.go.dev/net/url#PathEscape since it will handle escaping for all chars?

anders-swanson: This code gets rid of the URL escaped term. I am not sure why it did a replace instead of a decode though.

fredtibbitts: Oh I see, I read it the other way around.

fredtibbitts (dismissed):

tests/e2e/pkg/test/framework/metrics/ginkgo_metrics.go
anders-swanson: I changed it to a query un-escape anyway, since that's what this code was originally trying to do

fredtibbitts (approved):

PR #4738, opened 2022-11-15
VZ-7423.  Fix use of image pull secret with OAM component, add psrctl hooks for secret

michael-cico:
Fix use of image pull secret with OAM component, add psrctl hooks for secret
- make `verrazzano-container-registry` the default pull secret, since the PSR repos are all private for now
- also, make the communication of additional HelmOverrides more general in the Manager impl

Example:

```
psrctl start -s ops-s1 -p mysecret -w "ghcr.io/verrazzano/psr-backend-jenkins:xxxxx"
```

assuming that `mysecret` is an existing pull secret and the credentials are valid.

### DIFF
--- a/tools/psr/manifests/charts/worker/values.yaml
+++ b/tools/psr/manifests/charts/worker/values.yaml
@@ -1,9 +1,9 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-imagePullSecrets: [ ]
-
-#imageName: ghcr.io/verrazzano/psr-backend:local-c10faf9
+#imageName:
+imagePullSecrets:
+  - name: verrazzano-container-registry
 imagePullPolicy: IfNotPresent
 
 replicas: 1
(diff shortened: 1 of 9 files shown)


pfmackin (approved):

stevshap (approved):

tools/psr/manifests/charts/worker/values.yaml
```diff
@@ -1,9 +1,9 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-imagePullSecrets: [ ]
-
-#imageName: ghcr.io/verrazzano/psr-backend:local-c10faf9
+#imageName:
+imagePullSecrets:
+  - name: verrazzano-container-registry
```
mbenmclean, line 6: If secret doesn't exist what will happen? What do we want to happen?

mbenmclean (approved):

tools/psr/manifests/charts/worker/values.yaml
michael-cico: nothing, but if the image repo is private then it won't pull.

PR #4740, opened 2022-11-15
VZ-7654 PSR Readme

stevshap:
Updating PSR Readme to add new use cases and some refactoring

### DIFF
--- a/tools/psr/README.md
+++ b/tools/psr/README.md
@@ -178,7 +178,6 @@ global:
     
     PSR_WORKER_THREAD_COUNT - threads per worker
     default: 1
-    
 ```
 
 ### Example
@@ -224,8 +223,11 @@ The PostLogs worker makes batch http post requests to OpenSearch, in-cluster. Th
 
 #### Configuration
 ```
-LOG_ENTRIES: 1
-LOG_LENGTH: 1
+LOG_ENTRIES - number of log messages per post request
+default: 1
+
+LOG_LENGTH - number of characters per log message
+defaultL 1
 ```
 
 #### Run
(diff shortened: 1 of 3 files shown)


tools/psr/README.md
```diff
@@ -54,37 +54,34 @@ OpenSearch out and in, repeatedly.
 
 To run a scenario, you install two Helm releases as follows (note some use cases
```
mbenmclean, line 55: Why are we telling people to run a scenario by running two helm commands. Isn't the whole point of the psrcli to have it be one command?

stevshap: I haven't added any CLI-specific docs yet so this will be resolved in a follow up PR

michael-cico (approved):

tools/psr/README.md
pfmackin: This is the use case section, not scenario.  A developer would most likely be running use cases with Helm when developing a new worker, at least that is what we have been doing.

tools/psr/README.md
```diff
@@ -54,37 +54,34 @@ OpenSearch out and in, repeatedly.
...
-  PSR_MIN_PODS: 3  
-  PSR_INTERATION_DELAY: 5s
-  PSR_DURATION : 1h
+  envVars:
+    PSR_WORKER_TYPE: scale
+    OPEN_SEARCH_TIER: master
+    SCALE_DELAY_PER_TIER: 5s
+    MIN_REPLICA_COUNT: 3
+    MAX_REPLICA_COUNT: 5
 ```
```
pfmackin, line 70: missing global as in

global:
  envVars:
    PSR_WORKER_TYPE: scale

tools/psr/README.md
```diff
@@ -217,3 +216,69 @@ no configuration overrides
...
+### Scale
+#### Description
+The Scale worker continuously scales one tier of OpenSearch out and in. This worker must run in the mesh.
+
+#### Configuration
+```
+OPEN_SEARCH_TIER - OpenSearch tier
+default: master
+
+SCALE_DELAY_PER_TIER - scale delay
```
pfmackin, line 247: SCALE_DELAY_PER_TIER is obsolete obsolete

PR #4741, opened 2022-11-15
VZ-7676: Cluster operator Rancher cluster controller

fredtibbitts:
This PR adds the implementation of the Rancher cluster controller to the Verrazzano Cluster Operator. The controller watches Rancher cluster resources and creates and deletes VMCs accordingly.

This PR also fixes problems related to overriding the cluster operator image when installing from a private registry and also adds clusterrole and clusterrolebinding to the VCO helm chart.

Notes:

1. The controller is set up to take a label selector that will be used to filter Rancher clusters. It's not currently being set and will sync all clusters. A separate PR will wire up the label selector that the user provides in the VZ CR.
2. The code that was doing this in the VPO has been disabled but not removed. It will be removed in a PR that @desagar is working on.

### DIFF
--- a/cluster-operator/controllers/rancher/rancher_cluster_controller.go
+++ b/cluster-operator/controllers/rancher/rancher_cluster_controller.go
@@ -19,12 +19,15 @@ import (
 
 	clustersv1alpha1 "github.com/verrazzano/verrazzano/cluster-operator/apis/v1alpha1"
 	vzconst "github.com/verrazzano/verrazzano/pkg/constants"
+	vzstring "github.com/verrazzano/verrazzano/pkg/string"
 )
 
 const (
 	createdByLabel      = "app.kubernetes.io/created-by"
 	createdByVerrazzano = "verrazzano"
 	localClusterName    = "local"
+
+	finalizerName = "verrazzano.io/rancher-cluster"
 )
 
 type RancherClusterReconciler struct {
@@ -58,7 +61,7 @@ func (r *RancherClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	cluster := &unstructured.Unstructured{}
 	cluster.SetGroupVersionKind(gvk)
-	err := r.Client.Get(context.TODO(), req.NamespacedName, cluster)
+	err := r.Get(context.TODO(), req.NamespacedName, cluster)
 	if err != nil && !errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -69,11 +72,18 @@ func (r *RancherClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	// if the deletion timestamp is set, delete the corresponding VMC
-	if cluster.GetDeletionTimestamp() != nil {
-		if err := r.deleteVMC(cluster); err != nil {
-			return ctrl.Result{}, err
+	if !cluster.GetDeletionTimestamp().IsZero() {
+		if vzstring.SliceContainsString(cluster.GetFinalizers(), finalizerName) {
+			if err := r.deleteVMC(cluster); err != nil {
+				return ctrl.Result{}, err
+			}
 		}
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, r.removeFinalizer(cluster)
+	}
+
+	// add a finalizer to the Rancher cluster if it doesn't already exist
+	if err := r.ensureFinalizer(cluster); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	// ensure the VMC exists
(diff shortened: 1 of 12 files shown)


cluster-operator/controllers/rancher/rancher_cluster_controller.go
```diff
@@ -0,0 +1,187 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// deleteVMC deletes the VMC associated with a Rancher cluster
+func (r *RancherClusterReconciler) deleteVMC(cluster *unstructured.Unstructured) error {
+	displayName, err := r.getClusterDisplayName(cluster)
```
davisgomes, line 144: Seems like cluster is only used to get the display name. Could you pass in just the display name string?

cluster-operator/main.go
```diff
@@ -41,27 +47,62 @@ func main() {
...
 		MetricsBindAddress:     metricsAddr,
 		Port:                   9443,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "42d5ea87.verrazzano.io",
-	})
+	}
+
+	// if the user has specified a label selector to filter Rancher clusters, create a custom cache that applies the selector
+	// TODO populate this from the selector provided by the user
```
davisgomes, line 73: i think I remember hearing TODOs should be left out of PRs, but if were going to update this very soon, I don't think it's a huge issue.

platform-operator/internal/operatorinit/run_operator.go
```diff
@@ -61,11 +61,11 @@ func StartPlatformOperator(config config.OperatorConfig, log *zap.SugaredLogger,
...
-	rancherClusterSyncer := &clusters.RancherClusterSyncer{
-		Client: mgr.GetClient(),
-		Log:    log,
-	}
-	go rancherClusterSyncer.StartSyncing()
+	//rancherClusterSyncer := &clusters.RancherClusterSyncer{
+	//	Client: mgr.GetClient(),
+	//	Log:    log,
+	//}
+	//go rancherClusterSyncer.StartSyncing()
```
davisgomes, line 68: Can this be removed?

cluster-operator/main.go
fredtibbitts: Removed "TODO".

platform-operator/internal/operatorinit/run_operator.go
fredtibbitts: Removed the code.

cluster-operator/controllers/rancher/rancher_cluster_controller.go
fredtibbitts: I think I'm going to leave it as-is. It makes the calls from `Reconcile` consistent (that is, the calls to ensure and delete pass the cluster object). If `Reconcile` already had the display name then I would just pass that, but it doesn't so I'd just be moving the `getClusterDisplayName` call to `Reconcile`.

desagar (approved):

cluster-operator/controllers/rancher/rancher_cluster_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	if errors.IsNotFound(err) {
+		r.Log.Debugf("Rancher cluster %v not found, nothing to do", req.NamespacedName)
+		return ctrl.Result{}, nil
+	}
+
+	// if the deletion timestamp is set, delete the corresponding VMC
+	if !cluster.GetDeletionTimestamp().IsZero() {
+		if vzstring.SliceContainsString(cluster.GetFinalizers(), finalizerName) {
```
davisgomes, line 76: Do we still want to delete even if the finalizer is not set?

cluster-operator/controllers/rancher/rancher_cluster_controller.go
```diff
@@ -69,11 +72,18 @@ func (r *RancherClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	// if the deletion timestamp is set, delete the corresponding VMC
-	if cluster.GetDeletionTimestamp() != nil {
-		if err := r.deleteVMC(cluster); err != nil {
-			return ctrl.Result{}, err
+	if !cluster.GetDeletionTimestamp().IsZero() {
+		if vzstring.SliceContainsString(cluster.GetFinalizers(), finalizerName) {
+			if err := r.deleteVMC(cluster); err != nil {
+				return ctrl.Result{}, err
```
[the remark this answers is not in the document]
fredtibbitts: I think all errors are logged by the controller itself, but we could add another log message here. I'm not sure what you mean by "show up on the resource being deleted".

cluster-operator/controllers/rancher/rancher_cluster_controller.go
fredtibbitts: I think the general pattern is to only delete if the finalizer is set, since we always add the finalizer and the finalizer is only removed after a successful VMC deletion. That's what we seem to do in other controllers. That said, I don't see harm in always deleting it.

cluster-operator/controllers/rancher/rancher_cluster_controller.go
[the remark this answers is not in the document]
desagar: I meant if I described the resource, somewhere on its status.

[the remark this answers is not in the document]
desagar: but if the error is logged, I think that's fine.

desagar (approved):

PR #4744, opened 2022-11-16
Add test cases for fluentd component

rahulsrivastava71:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/fluentd/fluentd_component_test.go
+++ b/platform-operator/controllers/verrazzano/component/fluentd/fluentd_component_test.go
@@ -647,3 +647,86 @@ func getFakeComponentContext(c client.WithWatch) spi.ComponentContext {
 	}, nil, false)
 	return ctx
 }
+
+// Test isReady when it's called with component context
+func TestIsReady(t *testing.T) {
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/fluentd/fluentd_component_test.go
```diff
@@ -647,3 +647,86 @@ func getFakeComponentContext(c client.WithWatch) spi.ComponentContext {
 	}, nil, false)
 	return ctx
 }
+
+// Test isReady when it's called with component context
+func TestIsReady(t *testing.T) {
+	c := fake.NewClientBuilder().Build()
+	ctx := spi.NewFakeContext(c, &v1alpha1.Verrazzano{}, nil, false)
+	assert.False(t, NewComponent().IsReady(ctx))
```
adiforluls, line 655: Add good case as well where the method will return true (happy path)

rahulsrivastava71: that is tested in other scenario, in other test case

adiforluls (approved):
Approved. Remember to add the ticket no. in your PR title and merge commit message before merging.

PR #4746, opened 2022-11-16
VZ 7757: Increase UT coverage for platform-operator/internal/config

neeraj-n-singh:
VZ 7757: Increase UT coverage to 97.5% from 47.5%

### DIFF
--- a/platform-operator/internal/config/config_test.go
+++ b/platform-operator/internal/config/config_test.go
@@ -76,6 +76,7 @@ func TestSetConfig(t *testing.T) {
 
 	conf := Get()
 	SetDefaultBomFilePath("/etc/bom.json")
+	defer SetDefaultBomFilePath("")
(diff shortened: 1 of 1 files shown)


platform-operator/internal/config/config_test.go
```diff
@@ -58,24 +71,35 @@ func TestSetConfig(t *testing.T) {
 		WebhookValidationEnabled: false,
 		VerrazzanoRootDir:        "/root",
 		HealthCheckPeriodSeconds: int64(0),
+		DryRun:                   true,
 	})
 
 	conf := Get()
-
+	SetDefaultBomFilePath("/etc/bom.json")
```
adiforluls, line 78: Add a defer statement to reset the default bom file path `defer SetDefaultBomFilePath("")`

neeraj-n-singh: Added as per feedback

adiforluls (approved):

PR #4747, opened 2022-11-16
Fix diagnostic URL for bug-report

saidu941:
This PR have changes
> To All jenkins files to include environment variable


Jenkinsfile
```diff
@@ -162,6 +161,11 @@ pipeline {
                         env.VERRAZZANO_VERSION = "${env.VERRAZZANO_VERSION}-${env.BUILD_NUMBER}+${SHORT_COMMIT_HASH}"
                     }
                     DOCKER_IMAGE_TAG = "v${VERRAZZANO_DEV_VERSION}-${TIMESTAMP}-${SHORT_COMMIT_HASH}"
+                    def analysisurl="https://verrazzano.io/" + env.VERRAZZANO_VERSION + "/docs/troubleshooting/diagnostictools/analysisadvice/"
```
bhatpmk, line 164: You need to consider only major and minor version with prefix "v" and without dot (.) between major and minor version. The version in .verrazzano-development-version includes the patch as well.

Links to versioned doc is available here https://verrazzano.io/archive/docs/

Jenkinsfile
```diff
@@ -844,3 +848,11 @@ def metricBuildDuration() {
         }
     }
 }
+
+def setEffectiveDocsVersion() {
```
bhatpmk, line 852: Move the logic to a shell script, and return true / false from there to set the environment variable in Jenkinsfile.

saidu941: Updated PR with the changes

Jenkinsfile
saidu941: Updated the PR

Jenkinsfile
```diff
@@ -171,6 +170,9 @@ pipeline {
                         SUSPECT_LIST = getSuspectList(commitList, userMappings)
                         echo "Suspect list: ${SUSPECT_LIST}"
                     }
+                    // derive the USE_V8O_DOC_STAGE for the docs
+                    sh "chmod +x ${WORKSPACE}/ci/scripts/derive_v8o_doc_stage.sh"
```
agv426, line 174: set the script as executable and push it into source control that way, then remove this chmod from all the spots you are doing it at runtime

ci/scripts/derive_v8o_doc_stage.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+#
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+set -o pipefail
+
+if [ -z "$1" ]; then
```
agv426, line 9: this could be callable outside of Jenkins as well for user's local builds.

You could have it explicitly check for VERRAZZANO_DEV_VERSION being set and give a usage message that it needs to be set to use this.

Thinking that we should be able to do this for local builds as well

saidu941: Addressed and updated the PR

Jenkinsfile
saidu941: Addressed and updated the PR

ci/kind/Jenkinsfile
```diff
@@ -199,12 +199,16 @@ pipeline {
                 script {
                     def props = readProperties file: '.verrazzano-development-version'
                     VERRAZZANO_DEV_VERSION = props['verrazzano-development-version']
+                    env.VERRAZZANO_DEV_VERSION = "${VERRAZZANO_DEV_VERSION}"
                     TIMESTAMP = sh(returnStdout: true, script: "date +%Y%m%d%H%M%S").trim()
                     SHORT_COMMIT_HASH = sh(returnStdout: true, script: "git rev-parse --short=8 HEAD").trim()
                     DOCKER_IMAGE_TAG = "${VERRAZZANO_DEV_VERSION}-${TIMESTAMP}-${SHORT_COMMIT_HASH}"
+                    env.VERRAZZANO_VERSION = "${VERRAZZANO_DEV_VERSION}"
```
agv426, line 206: Is env.VERRAZZANO_VERSION needed?

agv426 (approved):

agv426 (dismissed):

lake-of-dreams (changes requested):
Maybe this might have been discussed before but I find it weird and counterintuitive to pass the VERRAZZANO_DEV_VERSION via Jenkins to some shell script which then is executed to  further creates a new environment variable USE_V8O_DOC_STAGE  which are being used in code for tools that should function as standalone cli's? 
I mean why can't the code in cli directly
1. make a http connection to https://raw.githubusercontent.com/verrazzano/verrazzano/master/.verrazzano-development-version and figure out erstwhile VERRAZZANO_DEV_VERSION, and deduce `docsVersion` and then
2. make a http connection to https://verrazzano.io/$docsVersion/docs/troubleshooting/diagnostictools/analysisadvice and see if it returns 404
Do we expect user to set `USE_V8O_DOC_STAGE` before running these cli's ?

tools/vz/cmd/version/version.go
```diff
@@ -64,12 +63,7 @@ func runCmdVersion(vzHelper helpers.VZHelper) error {
 }
 
 func GetEffectiveDocsVersion() string {
-	if os.Getenv("USE_V8O_DOC_STAGE") == "true" || len(cliVersion) == 0 {
```
agv426, line 67: When did this get changed, we must NOT rely on the USE_V8O_DOC_STAGE for normal cases, it only exists as an optional override in developer environments on the main line where there is no versioned doc released for it yet so we can form "deve" links

The normal case is that the environment variable does not exist and the CLI here will know the version it was built with and will use the versioned doc links that will exist for the released version

PR #4751, opened 2022-11-16
VZ-7654 psrctl README

stevshap:
First pass at README file for the psrctl CLI

### DIFF
--- a/tools/psr/psrctl/README.md
+++ b/tools/psr/psrctl/README.md
@@ -1,0 +1,47 @@
+# psrctl
+
+The PSR CLI is a tool to run and modify PSR test scenarios consisting of one or more `use cases`.
+
+You can run a PSR scenario against a live cluster using existing scenarios such as `opensearch-s1`:
+```
+psrctl start -s ops-s1
+```
+
+To create new scenarios, create a `scenario.yaml` file under `manifests/scenarios` using the following convention:
+```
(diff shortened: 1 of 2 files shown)


tools/psr/psrctl/README.md
```diff
@@ -0,0 +1,47 @@
+# psrctl
+
+The PSR CLI is a tool to run and modify PSR test scenarios consisting of one or more `use cases`.
+
+You can run a PSR scenario against a live cluster using existing scenarios such as `opensearch-s1`:
+```
+psrctl start -s ops-s1
+```
+
+To create new scenarios, create a `scenario.yaml` file under `manifests/scenarios` using the following convention:
```
mbenmclean, line 10: Maybe mention that you can then run that scenario using -d manifests/scenarios/myscenario -s ops-myscenario or something like that

mbenmclean (dismissed):

tools/psr/psrctl/README.md
pfmackin: Yes that is a good idea (-d mention)
Also, you should mention that the usecase-overrides dir is needed with the overrides for the scenario.  You can refer to a built-in scenario as an example

stevshap: ok I will add this

PR #4759, opened 2022-11-17
VZ-7396: Remove deprecated ES/kibana access URL on a new install and preserve it for upgrade.

Abhijithang:
This PR does the following:
    - New install - Removes Elasticsearch/Kibana references from ingresses.
    - Upgrade - preserves the deprecated Elasticsearch/Kibana ingress URL access by creating a redirect ingress for permanent redirection.


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1415,6 +1423,15 @@ func upgradeStatefulSet(ctx spi.ComponentContext) error {
 
 func populateSubdomainInTemplate(ctx spi.ComponentContext, tmpl string) (string, error) {
 	data := templateData{}
+
+	// Update verrazzano-pkce client redirect and web origin uris if deprecated host exists in the ingress
+	osHostExists, err := DoesIngressHostExist(ctx, constants.VerrazzanoSystemNamespace, constants.OpensearchIngress)
```
anders-swanson, line 1428: I still think this variable name is confusing. It's not whether the OS host exists, but whether the deprecated host exists.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1415,6 +1423,15 @@ func upgradeStatefulSet(ctx spi.ComponentContext) error {
 
 func populateSubdomainInTemplate(ctx spi.ComponentContext, tmpl string) (string, error) {
```
anders-swanson, line 1424: Is there any chance keycloak update will complete before the opensearch ingress is updated during a verrazzano upgrade? Seems like if keycloak goes first, it will miss the deprecated host because the new ingress doesn't exist yet (requires VMO opensearch reconciliation)

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1584,3 +1601,20 @@ func addClientRoleToUser(ctx spi.ComponentContext, cfg *restclient.Config, cli k
 	ctx.Log().Oncef("Added client role %s to the user %s", roleName, userName)
 	return nil
 }
+
+// DoesIngressHostExist returns true if ingress host exists
+func DoesIngressHostExist(ctx spi.ComponentContext, namespace string, ingressName string) (bool, error) {
+	ingress := &networkv1.Ingress{}
+	err := ctx.Client().Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: ingressName}, ingress)
+	if err != nil {
+		return false, nil
```
anders-swanson, line 1610: why is err nil in the return value?

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -510,8 +515,12 @@ data:
             serverUrl = me.makeMonitoringComponentBackendUrl('prometheus-operator-kube-p-prometheus', '9090')
         elseif backend == 'kibana' then
             serverUrl = me.makeVmiBackendUrl(backend, '5601')
+        elseif backend == 'osd' then
+            serverUrl = me.makeVmiBackendUrl('kibana', '5601')
```
anders-swanson, line 519: does osd map to kibana?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
Abhijithang: Makes sense, will change it. Thanks.

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -510,8 +515,12 @@ data:
             serverUrl = me.makeMonitoringComponentBackendUrl('prometheus-operator-kube-p-prometheus', '9090')
         elseif backend == 'kibana' then
             serverUrl = me.makeVmiBackendUrl(backend, '5601')
+        elseif backend == 'osd' then
+            serverUrl = me.makeVmiBackendUrl('kibana', '5601')    
         elseif backend == 'elasticsearch' then
-            serverUrl = me.makeVmiBackendUrl('es-ingest', '9200')
+            serverUrl = me.makeVmiBackendUrl('os-ingest', '9200')
```
anders-swanson, line 521: just curious why the backend is hardcoded here but the other entries use the `backend` var. think I am missing something

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
Abhijithang: Yes, kibana deployment and services need to be refactored, that's a different PR.

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -106,6 +107,14 @@ var (
 var _ = t.BeforeSuite(func() {
 	var err error
 	httpClient = pkg.EventuallyVerrazzanoRetryableHTTPClient()
+	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
+	if err != nil {
+		Fail(fmt.Sprintf("Failed to get default kubeconfig path: %s", err.Error()))
```
anders-swanson, line 112: can you use t.Fail

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -106,6 +107,14 @@ var (
...
+	}
+	isMinVersion150, err = pkg.IsVerrazzanoMinVersion("1.5.0", kubeconfigPath)
+	if err != nil {
+		Fail(err.Error())
```
anders-swanson, line 116: "

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
Abhijithang: I am just looking for a false return if the ingress doesn't exists. Error case does not apply for my requirement. So if it's misleading, I can remove it and make it  just underscore? What do you suggest?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
Abhijithang: I think it is according to the registry order from this file: platform-operator/controllers/verrazzano/component/registry/registry.go. So it won't upgrade before VMO. Also, I have tested with several iterations and didn't encounter the scenario where keycloak upgrades first.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
anders-swanson: You could ignore not found error, but what about some other client error?

Abhijithang: If it's not found it returns error. So I had to remove it.  But you are right, it has to be handled. Sorry for missing on that. 
I am thinking of using errors.IsNotFound(err) and return false incase of not found, else throw error.  Any better idea?

anders-swanson: Yes, use IsNotFound

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
```diff
@@ -57,8 +57,8 @@ var netpolNamespaceNames = []types.NamespacedName{
 	{Namespace: vzconst.VerrazzanoSystemNamespace, Name: "oam-kubernetes-runtime"},
 	{Namespace: vzconst.VerrazzanoSystemNamespace, Name: "vmi-system-es-master"},
 	{Namespace: vzconst.VerrazzanoSystemNamespace, Name: "vmi-system-es-data"},
-	{Namespace: vzconst.VerrazzanoSystemNamespace, Name: "vmi-system-es-ingest"},
-	{Namespace: vzconst.VerrazzanoSystemNamespace, Name: "vmi-system-kibana"},
```
bobdonat, line 61: will the old netpol references be cleaned up during upgrade

Abhijithang: Yes, VMO will delete them after creating new ingresses.

bobdonat: I think so since they are changed in the helm chart

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
bobdonat: i don't think there is any guarantee when VMO reconcillation will happen

platform-operator/controllers/verrazzano/component/networkpolicies/netpol.go
Abhijithang: You mean you are not sure of them cleaned up by VMO? Sorry, just wanted to verify if my understanding is right.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
Abhijithang: Good catch Anders :) Let me know if this is okay.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1584,3 +1601,18 @@ func addClientRoleToUser(ctx spi.ComponentContext, cfg *restclient.Config, cli k
 	ctx.Log().Oncef("Added client role %s to the user %s", roleName, userName)
 	return nil
 }
+
+// DoesDeprecatedIngressHostExist returns true if ingress host exists
+func DoesDeprecatedIngressHostExist(ctx spi.ComponentContext) (bool, error) {
+	ingressList := &networkv1.IngressList{}
```
anders-swanson, line 1607: You probably want to list by a selector or at least the VMI namespace?

anders-swanson (approved):

anders-swanson (approved):

PR #4771, opened 2022-11-17
 VZ-7797: Custom fail method for test framework

anders-swanson:
Implement a custom fail method that emits failure metrics. This will ensure that only one "failed" metric is pushed per test. Main changes are in ginkgo_wrapper.go and ginkgo_metrics.go. The rest is boilerplate changes to use the new fail method.


bobdonat (approved):
LGTM

PR #4776, opened 2022-11-18
VZ-7466.  Prototype "none" profile

michael-cico:


platform-operator/manifests/profiles/v1alpha1/none.yaml
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
kminder, line 1: Just 2022?

platform-operator/manifests/profiles/v1beta1/none.yaml
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
kminder, line 1: Just 2022?

PR #4782, opened 2022-11-18
Register all test panics as failed tests

anders-swanson:
Adds failure metrics for tests that panic outside of gomega fail handlers. Previously we would miss these metrics.


bobdonat (approved):
LGTM

PR #4786, opened 2022-11-18
JIRA: VZ-4480 generate and register comp health metrics

jmaron99:
Cumulative metrics for enabled and available components, as well as per component availability metrics, are created, updated and registered.


platform-operator/controllers/verrazzano/status/availability.go
```diff
@@ -69,11 +70,34 @@ func (p *HealthChecker) newStatus(log vzlog.VerrazzanoLogger, vz *vzapi.Verrazza
 			a := p.getComponentAvailability(component, componentStatus.State, ctx)
 			if a.available == vzapi.ComponentAvailable {
 				countAvailable++
+				// update the component availability metric
```
anders-swanson, line 73: Can be simplified like `metricsexporter.SetComponentAvailabilityMetric(component.Name(), a.available)` ?

jmaron99: Only if you want to go with the more obvious, maintainable approach ;)

Modified.

anders-swanson (approved):

PR #4810, opened 2022-11-21
VZ-7283 consolidate PreInstall/PreUpgrade and fix unit tests

davisgomes:


platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -673,6 +654,19 @@ func (h HelmComponent) resolveNamespace(ctx spi.ComponentContext) string {
 	return namespace
 }
 
+// PreInstallUpgrade handles the helm release status and secret cleanup for the helm upgrade or install to avoid conflicts
+func (h HelmComponent) PreInstallUpgrade(ctx spi.ComponentContext) error {
```
michael-cico, line 658: This doesn't need to be exported right?

davisgomes: No good point. I'll make it private

michael-cico (approved):

PR #4815, opened 2022-11-22
VZ-7793-oke deletion enhancement

theabmitra:

### DIFF
--- a/tests/e2e/config/scripts/delete_oke_cluster.sh
+++ b/tests/e2e/config/scripts/delete_oke_cluster.sh
@@ -59,7 +59,10 @@ do
     export KUBECONFIG=${KUBECONFIG_DIR}/$i/kube_config
     export VERRAZZANO_KUBECONFIG=$KUBECONFIG
   fi
-  cleanup_vz_resources
+  #cleanup_vz_resources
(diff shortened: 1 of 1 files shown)


tests/e2e/config/scripts/delete_oke_cluster.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
bhatpmk, line 3: You need to update the copyright year as "2020, 2022"

agv426 (approved):

bhatpmk (approved):

tests/e2e/config/scripts/delete_oke_cluster.sh
```diff
@@ -59,6 +59,9 @@ do
     export KUBECONFIG=${KUBECONFIG_DIR}/$i/kube_config
     export VERRAZZANO_KUBECONFIG=$KUBECONFIG
   fi
+  resName=$(kubectl get vz -o jsonpath='{.items[*].metadata.name}')
```
michael-cico, line 62: should we be doing this before the forced deletions above?

PR #4819, opened 2022-11-22
VZ-7478 Add prerequisite validator webhook (#4698)

ayaranga:
A webhook has been added to check the Verrazzano prerequisites and
present an appropriate warning to the user while applying the
Verrazzano custom resource. 

This PR is same as #4698 with one additional commit to cleanup the newly added webhook.


kminder (approved):
Approved.

PR #4823, opened 2022-11-22
VZ-4480 edit component health metric to add a value for "disabled"

knelson7:


platform-operator/controllers/verrazzano/status/availability.go
```diff
@@ -72,7 +72,7 @@ func (p *HealthChecker) newStatus(log vzlog.VerrazzanoLogger, vz *vzapi.Verrazza
 				countAvailable++
 			}
 			// update the component availability metric
-			err := metricsexporter.SetComponentAvailabilityMetric(component.Name(), a.available)
+			err := metricsexporter.SetComponentAvailabilityMetric(component.Name(), a.available, component.IsEnabled(ctx.EffectiveCR()))
```
anders-swanson, line 75: Can you calculate IsEnabled once, and reuse the value? it's already used in the conditional block above this line.

platform-operator/metricsexporter/metricsexporter.go
```diff
@@ -106,16 +106,19 @@ type ComponentHealth struct {
 }
 
 // This member function returns the simpleGaugeMetric that holds the upgrade time for a component
-func (c *ComponentHealth) SetComponentHealth(name string, availability bool) (prometheus.Gauge, error) {
+func (c *ComponentHealth) SetComponentHealth(name string, availability bool, isEnabled bool) (prometheus.Gauge, error) {
 	setting := 0
-	if availability {
-		setting = 1
+	if isEnabled {
```
anders-swanson, line 111: so, available = 1, enabled but not available = -1, and disabled = 0? Does it make sense to have 1 => available, 0 => unavailable, -1 => disabled? Then you could do an addition instead of these nested if statements?

anders-swanson: isEnabled : true => 0, isEnabled : false => -1
availability : true => 1, availability : false => 0
setting := availability + isEnabled

just a thought.

anders-swanson (approved):

PR #4825, opened 2022-11-22
VZ-7765: Wire up cluster label selector in cluster operator

fredtibbitts:
This PR wires up the cluster label selector in the cluster operator. We now auto create VMCs from Rancher clusters when:

1. Rancher cluster sync is enabled
2. Rancher cluster labels match the optional label selector configured in the Verrazzano CR. If sync is enabled and no cluster selector is specified, all Rancher clusters will be synchronized.

This PR also adds end-to-end tests for cluster selectors.

### DIFF
--- a/cluster-operator/controllers/rancher/rancher_cluster_controller_test.go
+++ b/cluster-operator/controllers/rancher/rancher_cluster_controller_test.go
@@ -48,6 +48,104 @@ func TestReconcileCreateVMC(t *testing.T) {
 	asserts.Equal(clusterName, vmc.Status.RancherRegistration.ClusterID)
 }
 
+// GIVEN a Rancher cluster resource is created with labels
+// AND   the user has specified a cluster selector that matches the labels
+// WHEN  the reconciler runs
+// THEN  a VMC is created and the cluster id is set in the status
+func TestReconcileWithClusterSelector(t *testing.T) {
+	asserts := assert.New(t)
+
+	const (
+		labelName  = "test-label"
+		labelValue = "test-label-value"
+	)
+
+	cluster := newCattleCluster(clusterName, displayName)
+	cluster.SetLabels(map[string]string{labelName: labelValue})
+	fakeClient := fake.NewClientBuilder().WithScheme(newScheme()).WithObjects(cluster).Build()
+
+	reconciler := newRancherClusterReconciler(fakeClient)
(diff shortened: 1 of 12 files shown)


cluster-operator/controllers/vmc/rancher.go
```diff
@@ -195,6 +192,44 @@ func ImportClusterToRancher(rc *RancherConfig, clusterName string, log vzlog.Ver
...
+		commaIfNeeded := ","
+		if i == 0 {
+			commaIfNeeded = ""
+		}
+		i++
+		labelsJSONString = fmt.Sprintf(`%s%s "%s": "%s"`, labelsJSONString, commaIfNeeded, label, val)
+	}
+	return labelsJSONString
+}
+
```
davisgomes, line 232: Is there a reason to use this instead of JSON marshalling the map? It seems like it should accomplish the same thing.

fredtibbitts: Agreed, I think you can replace this entire function with `json.Marshal(labels)`.

cluster-operator/controllers/rancher/rancher_cluster_controller_test.go
```diff
@@ -81,7 +179,10 @@ func TestReconcileDeleteVMC(t *testing.T) {
 	vmc := newVMC(displayName)
 	vmc.Status.RancherRegistration.ClusterID = clusterName
 	fakeClient := fake.NewClientBuilder().WithScheme(newScheme()).WithObjects(cluster, vmc).Build()
+
 	reconciler := newRancherClusterReconciler(fakeClient)
+	// disable cluster sync, the VMC (if it exists) should be deleted even if this flag is false
+	reconciler.ClusterSyncEnabled = false
```
desagar, line 185: is there something that tests this case with ClusterSyncEnabled = true?

cluster-operator/controllers/vmc/rancher.go
desagar: yeah probably. I wrote this code.

cluster-operator/controllers/rancher/rancher_cluster_controller_test.go
fredtibbitts: Added that.

desagar (approved):

davisgomes (approved):

PR #4827, opened 2022-11-22
Abehern/vz 7560 enforce ut branch coverage gt or eq master

AbeHern:
Updated coverage-number-comparison.sh to use remote UT coverage numbers to pass of fail a feature build


build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+
+#Get latest line-rate from master/release for comparison
+LOCAL_BRANCH_LINE_RATE=$(grep -i '<coverage' coverage.xml | awk -F' ' '{print $2}' | sed -E 's/line-rate=\"(.*)\"/\1/')
+LOCAL_BRANCH_VERSION=$(grep -i 'version=' .verrazzano-development-version | awk -F '=' '{print $2}' | head -c3)
+LOCAL_BRANCH_NAME=$(git branch  --no-color  | grep -E '^\*' | sed 's/\*[^a-z]*//g')
```
anders-swanson, line 10: Should be already present via the Jenkinsfile setting GIT_BRANCH? https://github.com/verrazzano/verrazzano/blob/master/Jenkinsfile#L130

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
...
+
+#Defaulted to 0 and only writes when enabled to unit-test-coverage-number.txt if coverage passes and if on release-* or master branch
+WRITE_ON_ENABLED=0
+
+#Returns 1 if Local-line-rate is gte Remote-line-rate, otherwise returns 0
+compareCoverageNumbers(){
+  REMOTE_LINE_RATE=$(cat unit-test-coverage-number.txt)
+  RATE="$LOCAL_BRANCH_LINE_RATE >= $REMOTE_LINE_RATE"
+  RESULT=$(echo "$RATE" | bc)
+    if [[ "$RESULT" -eq 1 ]]
```
anders-swanson, line 23: use consistent indentation, 2 spaces on the line above, 4 spaces here.

build/coverage-number-comparison.sh
AbeHern: Calling it in my file would just be $GIT_BRANCH then? Also I realized that I don't use LOCAL_BRANCH_NAME anywhere. It's defined but never used, I'll clean that up. I was using $CLEAN_BRANCH_NAME also in the Jenkinsfile https://github.com/verrazzano/verrazzano/blob/master/Jenkinsfile#L59 Should I switch to $GIT_BRANCH?

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
...
+        then
+          echo "PASS."
+          echo "Branch-line-rate: $LOCAL_BRANCH_LINE_RATE is gte to Remote-line-rate: $REMOTE_LINE_RATE"
+          writeToObjectStorage
+        else
+          echo "WARNING: Unit Test coverage(line-rate) does NOT pass"
```
anders-swanson, line 29: Should we only say WARNING if actually is just a warning? E.g., FAIL_ON_ENABLED is 0?

AbeHern: I was following the logic of the pseduo we wrote yesterday, I thought the only time we fail would be on a feature branch?

Check if git branch to regex (release-* or master)
        if coverage passes
	       put the coverage number in obj storage
Else if on dev branch 
	check if release coverage num exists in obj storage
		try to download it
	if it doesn’t exist download master 
		compare and fail if not gte

build/coverage-number-comparison.sh
anders-swanson: Either CLEAN_BRANCH_NAME or GIT_BRANCH should be fine.

build/coverage-number-comparison.sh
anders-swanson: I specifically mean the echo warning message

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "$LOCAL_BRANCH_LINE_RATE" > unit-test-coverage-number.txt
+    oci --region us-phoenix-1 os object put --force --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name "$CLEAN_BRANCH_NAME"/unit-test-coverage-number.txt --file unit-test-coverage-number.txt
+  fi
+}
+
+#Determine if on a release or master branch and download unit-test-coverage-number
+if [[ ! "$CLEAN_BRANCH_NAME" =~ ^release-[0-9]+\.[0-9]+$|^master$ ]]
+then
+    FAIL_ON_ENABLED=1
+    echo "Trying to download unit-test-coverage-number.txt from release-$LOCAL_BRANCH_VERSION..."
```
anders-swanson, line 51: can you define a new var: `UNIT_TEST_TXT_FILE=unit-test-coverage-number.txt`, and use that anywhere you have `unit-test-coverage-number.txt`?

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
...
+      echo "Trying to download unit-test-coverage-number.txt from master..."
+      oci --region us-phoenix-1 os object get --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name master/unit-test-coverage-number.txt --file unit-test-coverage-number.txt
+    fi
+
+    #Runs when we are on a feature branch and determines if line coverage passes
+    compareCoverageNumbers
+
+else
+  echo "Is a release-* or master branch..."
+  WRITE_ON_ENABLED=1
```
anders-swanson, line 65: I think Jenkins should set this. On local I do not want to upload the coverage number. So, `WRITE_ON_ENABLED` should be defined as a Jenkinsfile environment variable. 
Could you also rename this variable to `UPLOAD_UNIT_TEST_COVERAGE` so it is more clear what it's for?

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "Writing new coverage number to unit-test-coverage-number.txt ..."
+    echo "$LOCAL_BRANCH_LINE_RATE" > unit-test-coverage-number.txt
+    oci --region us-phoenix-1 os object put --force --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name "$CLEAN_BRANCH_NAME"/unit-test-coverage-number.txt --file unit-test-coverage-number.txt
+  fi
+}
+
+#Determine if on a release or master branch and download unit-test-coverage-number
+if [[ ! "$CLEAN_BRANCH_NAME" =~ ^release-[0-9]+\.[0-9]+$|^master$ ]]
+then
+    FAIL_ON_ENABLED=1
```
anders-swanson, line 50: 1. Can you rename this to `FAIL_IF_COVERAGE_DECREASED` or something similarly descriptive
2. Can you define this in the Jenkinsfile as a boolean parameter, and then assign it as an environment variable?

build/coverage-number-comparison.sh
AbeHern: Confused a bit, the Warning, will always display for a build that does not pass coverage, but the job will ONLY fail if we're working on a feature branch. Should release-* and master branches also fail and not just warn?

anders-swanson: I'm probably being too nitpicky, you can leave the echo statements as is.

Jenkinsfile
```diff
@@ -115,6 +117,10 @@ pipeline {
         OCIR_SCAN_REGISTRY = credentials('ocir-scan-registry')
         OCIR_SCAN_REPOSITORY_PATH = credentials('ocir-scan-repository-path')
         DOCKER_SCAN_CREDS = credentials('v8odev-ocir')
+
+        // used to write to object storage, or fail build if UT coverage does not pass
+        FAIL_BUILD_COVERAGE = 'params.FAIL_IF_COVERAGE_DECREASED'
+        UPLOAD_UT_COVERAGE = 'params.UPLOAD_UNIT_TEST_COVERAGE'
```
anders-swanson, line 123: I think you have to define like `= "${params.UPLOAD_UNIT_TEST_COVERAGE}"`

AbeHern: Just changed

Jenkinsfile
```diff
@@ -39,6 +39,8 @@ pipeline {
         booleanParam (description: 'Whether to fail the Integration Tests to test failure handling', name: 'SIMULATE_FAILURE', defaultValue: false)
         booleanParam (description: 'Whether to perform a scan of the built images', name: 'PERFORM_SCAN', defaultValue: false)
         booleanParam (description: 'Whether to wait for triggered tests or not. This defaults to false, this setting is useful for things like release automation that require everything to complete successfully', name: 'WAIT_FOR_TRIGGERED', defaultValue: false)
+        booleanParam (description: 'Whether to fail build if UT coverage number decreases lower than its release-* coverage from object storage. This defaults to false, meaning we are assuming the build will pass.', name: 'FAIL_IF_COVERAGE_DECREASED', defaultValue: false)
```
agv426, line 42: The parameter description should be clarified I think.

The behaviour of FAIL_IF_COVERAGE_DECREASED only applies to non release-*/master branches, for master/release it will always fail. The setting is helpful for users testing their branches to tell it to not fail.

Maybe add a "ignored by master/release builds, useful for user testing" or similar note

Jenkinsfile
```diff
@@ -39,6 +39,8 @@ pipeline {
...
+        booleanParam (description: 'Whether to write new passing UT coverage number to object storage. This only occurs for release-* or master branches with gte coverage. Defaults to false, only time to upload is if we are on release-* or master.', name: 'UPLOAD_UNIT_TEST_COVERAGE', defaultValue: false)
```
agv426, line 43: I'm not sure what purpose the UPLOAD_UNIT_TEST_COVERAGE is serving currently here.

release-*/master branches will override it and upload it.

I think this mainly could be useful for someone debugging the coverage ratcheting in a branch, however if doing that it would maybe need to also download from the user branch as well (ie: if testing it out in a user branch).

For this PR I don't think it hurts to leave it as-is. But maybe we can extend this in the future to be more of a UNIT_TEST_COVERAGE_TEST_MODE setting that will download and upload from the user branch to test things out against the coverage numbers only in the branch. And just default the UPLOAD_UT_COVERAGE variable below to false, doesn't have to come from a parameter value

Jenkinsfile
anders-swanson: default value should be true?

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Writing new coverage number to $UNIT_TEST_TXT_FILE ..."
+    echo "$LOCAL_BRANCH_LINE_RATE" > "$UNIT_TEST_TXT_FILE"
+    echo "Putting in object storage at $CLEAN_BRANCH_NAME/$UNIT_TEST_TXT_FILE"
+    oci --region us-phoenix-1 os object put --force --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name "$CLEAN_BRANCH_NAME"/"$UNIT_TEST_TXT_FILE" --file "$UNIT_TEST_TXT_FILE"
+  fi
+}
+
+#Determine if on a release or master branch and download unit-test-coverage-number
+if [[ ! "$CLEAN_BRANCH_NAME" =~ ^release-[0-9]+\.[0-9]+$|^master$ ]]; then
+  FAIL_BUILD_COVERAGE=true
```
anders-swanson, line 43: if we set to true it overwrites the jenkinsfile value - so it will always compare coverage on a feature branch

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Trying to download $UNIT_TEST_TXT_FILE from master..."
+    oci --region us-phoenix-1 os object get --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name master/"$UNIT_TEST_TXT_FILE" --file "$UNIT_TEST_TXT_FILE"
+  fi
+
+  #Runs when we are on a feature branch and determines if line coverage passes
+  compareCoverageNumbers
+
+else
+  echo "Is a release-* or master branch..."
+  UPLOAD_UT_COVERAGE=true
```
anders-swanson, line 57: if we set to true it overrides the jenkinsfile value - so it will always upload the coverage amount

AbeHern: It will only upload the coverage number if it is gte. The call to upLoadCoverage() only happens at [https://github.com/verrazzano/verrazzano/blob/abehern/vz-7560-Enforce-UT-branch-coverage-gt-or-eq-master/build/coverage-number-comparison.sh#L20](url) which is when the coverage is gte its remote comparison So although UPLOAD_UT_COVERAGE is set to true in the script, it is only set to true when on a release-* or master branch but doesn't upload if the coverage has decreased

build/coverage-number-comparison.sh
AbeHern: It's always comparing coverage

Jenkinsfile
AbeHern: I created it a trigger for only uploading release and master coverages. It gets set to true when on a release-* or master branch but the call to uploadToObjectStorage only happens at  [https://github.com/verrazzano/verrazzano/blob/abehern/vz-7560-Enforce-UT-branch-coverage-gt-or-eq-master/build/coverage-number-comparison.sh#L20](url) So it only actually uploads if it passes AND if it's on a release-* or master branch. At least that's the logic I was going for.

Jenkinsfile
AbeHern: Maybe I've made things much more confusing than they need to be, but I created both bools to be flags to be accepted in conditions that will fail a build. So on NON release-* or master branches the bool is set to true in [https://github.com/verrazzano/verrazzano/blob/abehern/vz-7560-Enforce-UT-branch-coverage-gt-or-eq-master/build/coverage-number-comparison.sh#L43](url) but its value is never read unless the coverage number decreases. So if a non release*/master branch fails coverage it will fail, otherwise it will pass BUT NOT upload its coverage

build/coverage-number-comparison.sh
```diff
@@ -0,0 +1,56 @@
...
+    echo "Trying to download $UNIT_TEST_TXT_FILE from master..."
+    oci --region us-phoenix-1 os object get --namespace "$OCI_OS_NAMESPACE" -bn "$OCI_OS_BUCKET" --name master/"$UNIT_TEST_TXT_FILE" --file "$UNIT_TEST_TXT_FILE"
+  fi
+
+  #Runs when we are on a feature branch and determines if line coverage passes
+  compareCoverageNumbers
+
+else
+  echo "Is a release-* or master branch..."
+   if [[ "$UPLOAD_UT_COVERAGE" = true ]]; then
```
anders-swanson, line 53: fix indentation here

AbeHern: Fixed

anders-swanson (approved):

PR #4835, opened 2022-11-23
VZ-7502 Create a Rancher user and Role for multicluster interactions

davisgomes:
To provide some security for VPO operations on managed clusters, we can create this new user with limited multicluster permissions.

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/clusterrole.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/clusterrole.yaml
@@ -74,3 +74,19 @@ rules:
       - get
       - list
       - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: verrazzano-cluster-rancher-user
+rules:
+  - apiGroups:
+      - ""
+    resources:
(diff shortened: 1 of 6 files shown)


platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/clusterrole.yaml
```diff
@@ -73,4 +73,20 @@ rules:
       - update
       - get
       - list
-      - watch
\ No newline at end of file
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
fredtibbitts, line 79: Don't you also need a ClusterRoleBinding somewhere?

fredtibbitts:
One other question, when we use the Rancher proxy to talk to the managed cluster, don't we now need to use this new user?

platform-operator/helm_config/charts/verrazzano-cluster-operator/templates/clusterrole.yaml
davisgomes: The Clusterrolebinding is created from the Rancher template resources, so we don't need a traditional kubernetes binding if we only plan to use this for the Rancher proxy

PR #4839, opened 2022-11-23
VZ-7503 Use the Verrazzano Cluster User for Multicluster Interaction

davisgomes:
To tighten security, we now use the Verrazzano Cluster User to make requests to managed clusters.


cluster-operator/controllers/vmc/update_rancher_rolebinding.go
```diff
@@ -0,0 +1,95 @@
...
+// to grant the Verrazzano cluster user the correct permissions on the managed cluster
+func (r *VerrazzanoManagedClusterReconciler) updateRancherClusterRoleBindingTemplate(vmc *v1alpha1.VerrazzanoManagedCluster) error {
+	if vmc == nil {
+		r.log.Debugf("Empty VMC, no ClusterRoleBindingTemplate created")
+		return nil
+	}
+
+	clusterID := vmc.Status.RancherRegistration.ClusterID
+	if len(clusterID) == 0 {
+		r.log.Debugf("Waiting to create ClusterRoleBindingTemplate for cluster %s, Rancher ClusterID not found in the VMC status", vmc.GetName())
```
fredtibbitts, line 43: Should this be a Progress message?

cluster-operator/controllers/vmc/update_rancher_rolebinding.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil
```
fredtibbitts, line 44: Does this need to return an error so that it gets retried?

pkg/rancher/rancher_config.go
```diff
@@ -0,0 +1,329 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package rancherutil
```
fredtibbitts, line 4: Generally the package name is the same as the name of the enclosing folder. If we add more files to `pkg/rancher` they will all need to use the same package name, `rancherutil` in this case. I'm not sure how big a deal that is. We can refactor in the future if needed.

pkg/rancher/rancher_config.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		log.Errorf("Failed to get Rancher TLS root CA: %v", err)
+		return nil, err
+	}
+	rc.CertificateAuthorityData = caCert
+
+	log.Debugf("Checking for Rancher additional CA in secret %s", cons.AdditionalTLS)
+	rc.AdditionalCA = common.GetAdditionalCA(rdr)
+
+	if admin {
```
fredtibbitts, line 99: It might be better to refactor this function so the token is set separately, it would make it more flexible so it can be used for any user.

fredtibbitts (approved):

cluster-operator/controllers/vmc/update_rancher_rolebinding.go
davisgomes: I don't believe so, because the cluster ID getting updated should re-trigger a reconcile

cluster-operator/controllers/vmc/update_rancher_rolebinding.go
davisgomes: This is set to debug for a few of the VMC processes. If I update this one, I should update them all.

PR #4842, opened 2022-11-24
VZ-7993: Update Rancher tests to handle user deletion appropriately

theabmitra:
Rancher delete users is failing consistently after a backup is done. This PR comments out that part since restore operation also does that automatically and brings back the users created. This is also verified once restore is complete. 

This will also help in periodics

### DIFF
--- a/tests/e2e/backup/rancher/rancher_backup_test.go
+++ b/tests/e2e/backup/rancher/rancher_backup_test.go
@@ -376,7 +376,7 @@ var _ = t.Describe("Rancher Backup and Restore,", Label("f:platform-verrazzano.r
 		})
 
 	})
-	
+
 	t.Context("Rancher restore", func() {
 		WhenRancherBackupInstalledIt("Start restore after rancher backup is completed", func() {
 			Eventually(func() error {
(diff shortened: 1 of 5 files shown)


tests/e2e/backup/rancher/rancher_backup_test.go
```diff
@@ -377,13 +377,13 @@ var _ = t.Describe("Rancher Backup and Restore,", Label("f:platform-verrazzano.r
...
-			}, waitTimeout, pollingInterval).Should(BeTrue(), "Delete rancher user")
-		})
-	})
+	//t.Context("Disaster simulation", func() {
+	//	WhenRancherBackupInstalledIt("Delete all users that were created as part of pre-suite", func() {
+	//		Eventually(func() bool {
+	//			return DeleteRancherUsers(common.RancherURL)
+	//		}, waitTimeout, pollingInterval).Should(BeTrue(), "Delete rancher user")
+	//	})
+	//})
```
ayaranga, line 386: nit: is it required to leave it commented or remove this code altogether? Also, is it better to make DeleteRancherUsers true if the user doesn't exist?

ayaranga (dismissed):

ayaranga (approved):

PR #4846, opened 2022-11-28
VZ-7502 PR to reintroduce the Cluster Operator User and Role

davisgomes:
The cluster role is now created in the Rancher component to fix a dependency issue


platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -529,13 +529,13 @@ func ConfigureAuthProviders(ctx spi.ComponentContext) error {
...
+// createOrUpdateRoleTemplates creates or updates the verrazzano-admin, verrazzano-monitor, and verrazzano-cluster-user RoleTemplates
 func createOrUpdateRoleTemplates(ctx spi.ComponentContext) error {
-	if err := createOrUpdateRoleTemplate(ctx, VerrazzanoAdminRoleName); err != nil {
+	if err := CreateOrUpdateRoleTemplate(ctx, VerrazzanoAdminRoleName); err != nil {
 		return err
 	}
 
-	return createOrUpdateRoleTemplate(ctx, VerrazzanoMonitorRoleName)
+	return CreateOrUpdateRoleTemplate(ctx, VerrazzanoMonitorRoleName)
 }
```
fredtibbitts, line 539: Is there supposed to be another call here for the verrazzano-cluster-user? The comment was updated so I expected to see another call to create the role template for the new user.

davisgomes: The RoleTemplate is now created as part of the Cluster Operator install. I made this function public so that is could be used in the Cluster Operator component. This was done to prevent an upgrade dependency where Rancher Upgrade was waiting for the Cluster Operator install to start.

davisgomes: The comment is a relic of the old version. I can update it to remove the old roletemplate change

fredtibbitts (approved):

PR #4851, opened 2022-11-28
VZ-4480 change metric names to match vz crd component json names

knelson7:


knelson7:
some of these commits have already gone in. Made separate branch and PR with only new changes

PR #4852, opened 2022-11-28
VZ-4480 change metric names to match vz crd component json names

knelson7:


anders-swanson:
can you use the constants from the components instead of string literals?

anders-swanson (approved):

anders-swanson (approved):

PR #4853, opened 2022-11-28
update individual component files so json name matches json name in crd

knelson7:


platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
```diff
@@ -224,7 +224,7 @@ func (g grafanaComponent) PostUpgrade(ctx spi.ComponentContext) error {
 func (g grafanaComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Verrazzano) error {
 	// do not allow disabling active components
 	if vzcr.IsGrafanaEnabled(old) && !vzcr.IsGrafanaEnabled(new) {
-		return fmt.Errorf("Disabling component Grafana not allowed")
+		return fmt.Errorf("Disabling component grafana not allowed")
```
jmaron99, line 227: I think this is OK (changing the printout case) but I have run into sensitivity about case with existing products, although that was mostly in documentation.  I'm not sure who to check this with?  @kminder ?

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -220,7 +220,7 @@ func (o opensearchComponent) Name() string {
 func (o opensearchComponent) isOpenSearchEnabled(old *installv1beta1.Verrazzano, new *installv1beta1.Verrazzano) error {
 	// Do not allow disabling of any component post-install for now
 	if vzcr.IsOpenSearchEnabled(old) && !vzcr.IsOpenSearchEnabled(new) {
-		return fmt.Errorf("Disabling component OpenSearch not allowed")
+		return fmt.Errorf("Disabling component opensearch not allowed")
```
jmaron99, line 223: same question about case...

platform-operator/controllers/verrazzano/component/coherence/coherence_component.go
```diff
@@ -30,7 +30,7 @@ const ComponentName = "coherence-operator"
 // ComponentNamespace is the namespace of the component
 const ComponentNamespace = constants.VerrazzanoSystemNamespace
 
-// ComponentJSONName is the josn name of the verrazzano component in CRD
+// ComponentJSONName is the c name of the verrazzano component in CRD
```
kminder, line 33: This should be JSON.  They should all be JSON not json.

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
kminder: Lower case is fine as long as it refers to something in code/API as this does.  If it refers to the product then it needs to match the trademarked name.  In this case I read this as referring to the component name in the API so grafana is correct.  In fact it should probably be a formatted string using ComponentJSONName instead of being hard coded in the message.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component_test.go
```diff
@@ -100,7 +100,7 @@ func TestGetMinVerrazzanoVersion(t *testing.T) {
 //	THEN a string containing JSONName is returned
 func TestGetJSONName(t *testing.T) {
 	jsonName := NewComponent().GetJSONName()
-	assert.Equal(t, jsonName, "opensearch-dashboards")
+	assert.Equal(t, jsonName, "opensearchDashboards")
```
kminder, line 103: This seems like a test purely to get code coverage.  Just make it future proof and assert against the constant.

platform-operator/controllers/verrazzano/component/grafana/grafana_component.go
```diff
@@ -233,7 +233,7 @@ func (g grafanaComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Verra
 func (g grafanaComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzano, new *installv1beta1.Verrazzano) error {
 	// do not allow disabling active components
 	if vzcr.IsGrafanaEnabled(old) && !vzcr.IsGrafanaEnabled(new) {
-		return fmt.Errorf("Disabling component Grafana not allowed")
+		return fmt.Errorf("Disabling component grafana not allowed")
```
anders-swanson, line 236: use `fmt.Errorf("Disabling component %s not allowed", ComponentJSONName)`

anders-swanson (approved):

PR #4854, opened 2022-11-28
Abehern/vz 7560 enforce ut branch coverage gt or eq master

AbeHern:
Removes UT coverage comparison from "precommit" stage and renames the target name to be more meaningful


fredtibbitts (commented):
If I'm reading this correctly, it means we won't find out if we've lowered the coverage until the code has been pushed and it runs in the pipeline. It would be nice to be able to figure that out locally before pushing code, but maybe that's difficult to do?

AbeHern:
> If I'm reading this correctly, it means we won't find out if we've lowered the coverage until the code has been pushed and it runs in the pipeline. It would be nice to be able to figure that out locally before pushing code, but maybe that's difficult to do?

I agree knowing ones local coverage before building would be better. As far as I know there isn't an obvious solution to that but it is known. This fix is to address people local builds not breaking as a result of the coverage comparison

fredtibbitts (approved):

PR #4855, opened 2022-11-28
VZ-7849: secure pods in verrazzano-install namespace from running as root or escalating privs

mgianatagh:
secure pods in verrazzano-install namespace from running as root or escalating privs

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
@@ -25,6 +25,9 @@ spec:
       initContainers:
         - name: webhookswait
           imagePullPolicy: {{ .Values.imagePullPolicy }}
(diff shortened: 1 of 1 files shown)


platform-operator/helm_config/charts/verrazzano-platform-operator/templates/deployment.yaml
```diff
@@ -143,3 +157,5 @@ spec:
         - name: webhook-certs
           emptyDir: {}
       serviceAccountName: {{ .Values.name }}
+      securityContext:
+        runAsNonRoot: true
```
kminder, line 161: Does this need allowPrivilegeEscalation: false?

mgianatagh: that is not an option for the pod level security context

kminder (approved):
Approved.

PR #4857, opened 2022-11-28
Add verrazzano-cluster-operator to list of pods that do not have an istio sidecar

fredtibbitts:
Prior to this change, when updating the VZ CR the verrazzano-cluster-operator pod would get restarted, but it has no istio sidecar so it should not be restarted.


kminder:
What about a test?

fredtibbitts:
> What about a test?

I did update the unit test, and I did test it locally to make sure the unnecessary restart was no longer happening. What sort of test did you have in mind?

desagar (approved):

kminder (approved):
Approved.

PR #4865, opened 2022-11-29
VZ-7770: Added USE_V8O_DOC_STAGE  for vz tool

neeraj-n-singh:
In ocne CI tests, whenever we have failures and we run vz tools against the same, in that scenario it will show non-release version doc, as a consequence developer can't verify the things.
There was a keyword which is used to determine the same, added that keyword in CI definition will resolve the things.

### DIFF
--- a/ci/olcne/Jenkinsfile
+++ b/ci/olcne/Jenkinsfile
@@ -201,6 +201,7 @@ pipeline {
         // OCI_COMPARTMENT_ID=credentials('oci-tiburon-dev-compartment-ocid')
         // OCI_TELEMETRY_URL=credentials('oci-telemetry-url')
         VZ_CLI_TARGZ="vz-linux-amd64.tar.gz"
+        USE_V8O_DOC_STAGE = "${env.BRANCH_NAME == 'master' ? 'true' : 'false'}"
 
         // used to emit metrics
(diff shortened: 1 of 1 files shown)


ci/olcne/Jenkinsfile
```diff
@@ -201,6 +201,7 @@ pipeline {
         // OCI_COMPARTMENT_ID=credentials('oci-tiburon-dev-compartment-ocid')
         // OCI_TELEMETRY_URL=credentials('oci-telemetry-url')
         VZ_CLI_TARGZ="vz-linux-amd64.tar.gz"
+        USE_V8O_DOC_STAGE = "${env.BRANCH_NAME == 'master' ? 'true' : 'false'}"
```
kminder, line 204: Where does this get used?  I don't see a use in this Jenkinsfile.  Might be worth a comment.

agv426: See Saidu's PR that will update all of the pipelines:
https://github.com/verrazzano/verrazzano/pull/4747

This is used by the CLI to determine if the source that was used to build the CLI is still in development or was released, it will use that to determine whether the URLs for analysis advice it generates point to the staging docs (not released yet), or the versioned docs (released and those exist)

PR #4888, opened 2022-12-01
VZ-7802 Add additional OpenSearch PSR scenarios

stevshap:
Scenarios added are different permutations of:
- WriteLogs
- GetLogs
- PostLogs
- Restart
- Scale


tools/psr/manifests/scenarios/opensearch/combine-all-scenarios/scenario.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: can you change name to all-usecases

tools/psr/manifests/scenarios/opensearch/s6/scenario.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
```
pfmackin, line 1: This is missing usecase-overrides dir.  Do we need this s6 scenario?

pfmackin (approved):

PR #4892, opened 2022-12-02
VZ-7641-test jaeger resources

rahulsrivastava71:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
+++ b/platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
@@ -6,11 +6,10 @@ package operator
 import (
 	"context"
 	"fmt"
-	"path/filepath"
-
 	"github.com/verrazzano/verrazzano/pkg/k8s/ready"
 	"github.com/verrazzano/verrazzano/pkg/vzcr"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/networkpolicies"
+	"path/filepath"
 
 	installv1beta1 "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1beta1"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common"
@@ -223,10 +222,14 @@ func (c jaegerOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error {
 	if !installed && doDefaultJaegerInstanceDeploymentsExists(ctx) {
 		return ctx.Log().ErrorfNewErr("Conflicting Jaeger instance %s/%s exists! Either disable the Verrazzano's default Jaeger instance creation by overriding jaeger.create Helm value for Jaeger Operator component to false or delete and recreate the existing Jaeger deployment in a different namespace: %v", ComponentNamespace, globalconst.JaegerInstanceName, err)
 	}
-	err = removeOldJaegerResources(ctx)
-	if err != nil {
-		return err
+	//if installed is false then component is not installed by helm
+	if !installed {
+		err = removeOldJaegerResources(ctx)
+		if err != nil {
+			return err
+		}
 	}
+
 	createInstance, err := isCreateDefaultJaegerInstance(ctx)
 	if err != nil {
 		return err


platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -223,10 +224,21 @@ func (c jaegerOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error {
...
-		return err
+	//get the source version before upgrade
+	sourceVersion := ctx.EffectiveCR().Spec.Version
+	//convert source verion to float
+	s, err := strconv.ParseFloat(sourceVersion, 64)
+	if err == nil {
+		ctx.Log().Debug(s)
+	}
+	//don't remove resources if version to upgrade is greater than 1.3
+	if s < 1.4 {
```
adiforluls, line 235: See https://github.com/verrazzano/verrazzano/tree/master/pkg/semver
Use this package's methods to get and compare Verrazzano version instead of directly reading and formatting raw version from the CR.

Look at this as an example https://github.com/verrazzano/verrazzano/blob/4ec94ff3b970f87253c4247ac8c08c7b5c35f572/platform-operator/controllers/verrazzano/reconcile/controller.go#L226

pshirodkar-orcl, line 235: use constant https://github.com/verrazzano/verrazzano/blob/master/platform-operator/constants/constants.go#L138 instead of "1.4"

rahulsrivastava71: i need to compare with the float value and semver.NewSemVersion(actualCR.Spec.Version) does not gives a float value because version can be float value .

rahulsrivastava71: added the code for the same

rahulsrivastava71: updated it please check

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -223,10 +223,23 @@ func (c jaegerOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error {
...
 	}
-	err = removeOldJaegerResources(ctx)
-	if err != nil {
-		return err
+	//get the source version before upgrade
+	sourceVersion := ctx.EffectiveCR().Spec.Version
+	//convert source verion to float
+	//s, err := strconv.ParseFloat(sourceVersion, 64)
+	specVersion, err := semver.NewSemVersion(sourceVersion)
+	targetVersion, err := semver.NewSemVersion(ctx.ActualCR().Spec.Version)
```
adiforluls, line 231: `EffectiveCR` = `ActualCR` + a set of YAMLs merged together by the VPO at runtime. What you need to do is compare the version in actualCR with `v1.4.0` . Use the constant to do that. https://github.com/verrazzano/verrazzano/blob/3caaf3f663a5d9d498c9c310b1e55fefd3e36758/platform-operator/constants/constants.go#L138

Also add a unit test to validate your change.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -223,10 +223,23 @@ func (c jaegerOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error {
 	if !installed && doDefaultJaegerInstanceDeploymentsExists(ctx) {
 		return ctx.Log().ErrorfNewErr("Conflicting Jaeger instance %s/%s exists! Either disable the Verrazzano's default Jaeger instance creation by overriding jaeger.create Helm value for Jaeger Operator component to false or delete and recreate the existing Jaeger deployment in a different namespace: %v", ComponentNamespace, globalconst.JaegerInstanceName, err)
 	}
-	err = removeOldJaegerResources(ctx)
-	if err != nil {
-		return err
+	//get the source version before upgrade
+	sourceVersion := ctx.EffectiveCR().Spec.Version
+	//convert source verion to float
```
adiforluls, line 228: Remove comments that are irrelevant

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
rahulsrivastava71: ok let me check this

rahulsrivastava71: done can you please check

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
rahulsrivastava71: removed

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
adiforluls: Can you add a unit test as well for this change? (i.e. the function to clean up resources is only called when the version is less than v1.4.0, otherwise it is not called).

adiforluls: I'm not sure if it's possible to get the version we're upgrading from in PreUpgrade func, the effectiveCR and actualCR would contain the version we're upgrading to, not the version we're upgrading from when called from PreUpgrade I think. @michael-cico would you know?

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -223,10 +223,26 @@ func (c jaegerOperatorComponent) PreUpgrade(ctx spi.ComponentContext) error {
...
+	targetVersion, err := semver.NewSemVersion(constants.VerrazzanoVersion1_4_0)
+	if err == nil {
+		ctx.Log().ErrorfNewErr("targetversion", targetVersion)
+	}
+	//don't remove resources if version to upgrade is greater than 1.3
+	if specVersion.IsLessThan(targetVersion) {
+		err = removeOldJaegerResources(ctx)
+		if err != nil {
+			return err
+		}
```
ayaranga, line 243: I think we can make the code more readable, something like
```
        sourceVersion, _ := semver.NewSemVersion(ctx.EffectiveCR().Spec.Version)
	ver140, _ := semver.NewSemVersion(constants.VerrazzanoVersion1_4_0)
	if sourceVersion.IsLessThan(ver140) {
```
Things to check:
- Do we need a check if ctx.EffectiveCR().Spec.Version is not empty?
- Do we need to catch the error for semver.NewSemVersion(ctx.EffectiveCR().Spec.Version)? For the 1.4.0 case, it's not needed.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
ayaranga: This needs to be verified. @rahulsrivastava71 can you please confirm?

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
ayaranga: If we are not able to get the source VZ version, then better to remove the resources only if Jaeger is not installed by helm component. We already have `installed` variable set above to true if the Jaeger Helm component is already installed. You can remove the resource only if `installed` is set to `false`.

rahulsrivastava71: @ayaranga i have implemented the same can you please check this once

ayaranga (approved):

PR #4898, opened 2022-12-02
VZ-7600 Worker must be able to run for a specific amount of time

stevshap:
- Implemented code and tests for PsrDuration parameter
- Default value is "-1s" which signifies no time limit


tools/psr/backend/workmanager/manager.go
```diff
@@ -51,7 +51,7 @@ func StartWorkerRunners(log vzlog.VerrazzanoLogger) error {
 
 	// start metrics server as go routine
 	log.Info("Starting metrics server")
-	mProviders := []spi.WorkerMetricsProvider{}
+	var mProviders []spi.WorkerMetricsProvider
```
pfmackin, line 54: why did you change this?  Normally we use :=

tools/psr/backend/workmanager/manager.go
```diff
@@ -112,6 +112,6 @@ func getWorker(wt string) (spi.Worker, error) {
 	case config.WorkerTypeReceiveAlerts:
 		return alerts.NewAlertsWorker()
 	default:
-		return nil, fmt.Errorf("Failed, invalid worker type '%s'", wt)
+		return nil, fmt.Errorf("failed, invalid worker type '%s'", wt)
```
pfmackin, line 115: use Failed

tools/psr/backend/workmanager/runner.go
```diff
@@ -127,14 +128,21 @@ func (r workerRunner) RunWorker(conf config.CommonConfig, log vzlog.VerrazzanoLo
...
+		_ = log.GetZapLogger().Sync()
+
 		durationSecondsTotal := time.Now().Unix() - startTimeSecs
 		atomic.StoreInt64(&r.runnerMetrics.workerLoopNanoSeconds.Val, time.Now().UnixNano()-startLoop)
 		atomic.StoreInt64(&r.runnerMetrics.workerDurationTotalSeconds.Val, durationSecondsTotal)
 		if r.Worker.WantLoopInfoLogged() {
 			log.Infof("Loop Count: %v, Total seconds from start of the first worker loop until now: %v", loopCount, durationSecondsTotal)
 		}
+		duration, _ := time.ParseDuration(strconv.FormatInt(durationSecondsTotal, 10) + "s")
+		if duration >= conf.PsrDuration && conf.PsrDuration != config.UnlimitedWorkerDuration {
```
pfmackin, line 140: You should calculate endTime once before looping, same place where startTime is calculated.

tools/psr/backend/workmanager/runner.go
```diff
@@ -127,14 +128,21 @@ func (r workerRunner) RunWorker(conf config.CommonConfig, log vzlog.VerrazzanoLo
...
-		log.GetZapLogger().Sync()
+		_ = log.GetZapLogger().Sync()
+
 		durationSecondsTotal := time.Now().Unix() - startTimeSecs
 		atomic.StoreInt64(&r.runnerMetrics.workerLoopNanoSeconds.Val, time.Now().UnixNano()-startLoop)
 		atomic.StoreInt64(&r.runnerMetrics.workerDurationTotalSeconds.Val, durationSecondsTotal)
 		if r.Worker.WantLoopInfoLogged() {
 			log.Infof("Loop Count: %v, Total seconds from start of the first worker loop until now: %v", loopCount, durationSecondsTotal)
 		}
+		duration, _ := time.ParseDuration(strconv.FormatInt(durationSecondsTotal, 10) + "s")
```
pfmackin, line 139: You don't need to add 's'.  Do this like config.go
	sleepDuration, err := time.ParseDuration(PsrEnv.GetEnv(PsrLoopSleep))

tools/psr/backend/osenv/env.go
```diff
@@ -53,7 +53,7 @@ func (e *envData) addItemConfig(c EnvVarDesc) error {
 	val := GetEnvFunc(c.Key)
 	if len(val) == 0 {
 		if c.Required {
-			return fmt.Errorf("Failed, missing required Env var %s", c.Key)
+			return fmt.Errorf("failed, missing required Env var %s", c.Key)
```
pfmackin, line 56: All log messages start with upper case

tools/psr/backend/workmanager/runner_test.go
```diff
@@ -81,9 +82,10 @@ func TestRunWorker(t *testing.T) {
 		name      string
 		loops     int64
 		expectErr bool
+		duration  time.Duration
 	}{
-		{name: "oneIter", loops: 1, expectErr: false},
-		{name: "tenIter", loops: 10, expectErr: false},
```
pfmackin, line 86: we still want the loop number of iteration tests, just add the new ones and name them something duration

tools/psr/backend/workmanager/runner.go
```diff
@@ -127,14 +128,21 @@ func (r workerRunner) RunWorker(conf config.CommonConfig, log vzlog.VerrazzanoLo
...
+		if duration >= conf.PsrDuration && conf.PsrDuration != config.UnlimitedWorkerDuration {
+			log.Infof("Worker has reached its' run duration of %s", conf.PsrDuration)
```
pfmackin, line 141: change its` to its

tools/psr/backend/workmanager/runner.go
mbenmclean: don't use PsrEnv.GetEnv in the runner or worker code. Use field in the conf parameter

tools/psr/backend/config/config.go
```diff
@@ -86,6 +89,15 @@ func GetCommonConfig(log vzlog.VerrazzanoLogger) (CommonConfig, error) {
 		return CommonConfig{}, err
 	}
 
+	duration, err := time.ParseDuration(PsrEnv.GetEnv(PsrDuration))
+	if err != nil {
+		return CommonConfig{}, log.ErrorfNewErr("Error parsing worker duration: %v", err)
+	}
+	// Negative values are not allowed, -1s represents no duration timeout
+	if duration < (UnlimitedWorkerDuration) {
```
mbenmclean, line 97: if duration < 0
if someone passed duration as -1ns it would not be < 1s but would still be negative, so problematic

tools/psr/backend/workmanager/runner.go
stevshap: I don't understand where that would be used

tools/psr/backend/workmanager/runner.go
stevshap: Duration is the duration that has already taken place, which is an int of seconds. I need it to be converted to type time.Duration

mbenmclean: I don't personally have a problem with how you did it.

tools/psr/backend/workmanager/runner.go
```diff
@@ -128,13 +128,19 @@ func (r workerRunner) RunWorker(conf config.CommonConfig, log vzlog.VerrazzanoLo
...
 		}
 		log.GetZapLogger().Sync()
+
 		durationSecondsTotal := time.Now().Unix() - startTimeSecs
 		atomic.StoreInt64(&r.runnerMetrics.workerLoopNanoSeconds.Val, time.Now().UnixNano()-startLoop)
 		atomic.StoreInt64(&r.runnerMetrics.workerDurationTotalSeconds.Val, durationSecondsTotal)
 		if r.Worker.WantLoopInfoLogged() {
 			log.Infof("Loop Count: %v, Total seconds from start of the first worker loop until now: %v", loopCount, durationSecondsTotal)
 		}
+		if time.Duration(durationSecondsTotal) > conf.PsrDuration*time.Second && conf.PsrDuration != config.UnlimitedWorkerDuration {
```
pfmackin, line 138: I think this needs to be >= conf.PsrDuration, so if the user says run for 10 secs, it will terminate at 10 secs.  Also, can you put spaces between PsrDuration * time.Seconds

pfmackin (approved):

PR #4901, opened 2022-12-02
VZ-7900 update system health check dashboard metric names and ensure dashboard is autogenerated

knelson7:


platform-operator/controllers/verrazzano/healthcheck/availability_test.go
```diff
@@ -38,8 +39,12 @@ func (f fakeComponent) Name() string {
...
 }
 
-func newFakeComponent(name string, available vzapi.ComponentAvailability, enabled bool) fakeComponent {
-	return fakeComponent{name: name, available: available, enabled: enabled}
+func (f fakeComponent) GetJSONName() string {
+	return f.JSONName
+}
+
+func newFakeComponent(name string, JSONname string, available vzapi.ComponentAvailability, enabled bool) fakeComponent {
+	return fakeComponent{name: name, JSONName: name, available: available, enabled: enabled}
```
anders-swanson, line 47: JSONName parameter not used? Probably not needed for this fake object.

knelson7: My mistake, it should be fakeComponent{name: name, JSONName: JSONname ...) 
It didn't cause an error because the components used in our tests have the same Name and JSON name (grafana, opensearch, and rancher), but it seems like a good idea to pass in the variable since some components don't have identical names.

anders-swanson (approved):

PR #4903, opened 2022-12-05
VZ-7371: Check events for addtional support data

saketmht:
In case there are no matches for any of the existing patterns when there is a component failure, the respective events.json for the component is used to find out events with type "Warning" to provide some additional support data (the message and reason related to that event), which might help in figuring out the reason for the issue.

### DIFF
--- a/tools/vz/pkg/analysis/internal/util/cluster/events.go
+++ b/tools/vz/pkg/analysis/internal/util/cluster/events.go
@@ -132,7 +132,7 @@ func CheckEventsForWarnings(log *zap.SugaredLogger, events []corev1.Event, event
 		for index, previousEvent := range finalEventList {
 			// If we already iterated through an event for the same involved object with given eventType
 			// And that event occurred before this current event, replace it accordingly
-			if IsInvolvedObjectSame(event, previousEvent) && event.LastTimestamp.Time.After(previousEvent.LastTimestamp.Time) {
+			if isInvolvedObjectSame(event, previousEvent) && event.LastTimestamp.Time.After(previousEvent.LastTimestamp.Time) {
 				foundInvolvedObject = true
 				// Remove the previous event from the final list
 				finalEventList = append(finalEventList[:index], finalEventList[index+1:]...)
@@ -150,13 +150,16 @@ func CheckEventsForWarnings(log *zap.SugaredLogger, events []corev1.Event, event
 	}
 
 	for _, event := range finalEventList {
+		if len(event.Message) == 0 {
+			continue
+		}
 		message = append(message, fmt.Sprintf("Namespace: %s, %s %s, Message: %s, Reason: %s", event.InvolvedObject.Namespace, event.InvolvedObject.Kind, event.InvolvedObject.Name, event.Message, event.Reason))
 	}
 
 	return message, nil
 }
 
-func IsInvolvedObjectSame(eventA corev1.Event, eventB corev1.Event) bool {
+func isInvolvedObjectSame(eventA corev1.Event, eventB corev1.Event) bool {
 	return eventA.InvolvedObject.Kind == eventB.InvolvedObject.Kind &&
 		(eventA.InvolvedObject.Name == eventB.InvolvedObject.Name) &&
 		eventA.InvolvedObject.Namespace == eventB.InvolvedObject.Namespace
--- a/tools/vz/pkg/analysis/internal/util/cluster/events.go
+++ b/tools/vz/pkg/analysis/internal/util/cluster/events.go
@@ -96,8 +96,8 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
 	return serviceEvents, nil
 }
 
-// GetEventsRelatedToComponentNs gets events related to a component
-func GetEventsRelatedToComponentNs(log *zap.SugaredLogger, clusterRoot string, componentNamespace string, timeRange *files.TimeRange) (componentEvents []corev1.Event, err error) {
+// GetEventsRelatedToComponentNamespace gets events related to a component namespace
+func GetEventsRelatedToComponentNamespace(log *zap.SugaredLogger, clusterRoot string, componentNamespace string, timeRange *files.TimeRange) (componentEvents []corev1.Event, err error) {
 	log.Debugf("GetEventsRelatedToComponentNs called for component in namespace %s", componentNamespace)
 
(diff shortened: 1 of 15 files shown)


tools/vz/pkg/analysis/internal/util/cluster/events.go
```diff
@@ -95,6 +96,72 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
 	return serviceEvents, nil
 }
 
+// GetEventsRelatedToComponentNamespace gets events related to a component namespace
+func GetEventsRelatedToComponentNamespace(log *zap.SugaredLogger, clusterRoot string, componentNamespace string, timeRange *files.TimeRange) (componentEvents []corev1.Event, err error) {
```
bhatpmk, line 100: The bug-report contains <namespace>/events.json already, I am not clear on the need for this new function. 

In my opinion, you can just get the namespace for the component, unmarshal events.json from the directory with the name same as namespace and process the events

tools/vz/pkg/analysis/internal/util/cluster/events.go
```diff
@@ -95,6 +96,72 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
...
+		}
+		componentEvents = append(componentEvents, eventList...)
+	}
+	return componentEvents, nil
+}
+
+// CheckEventsForWarnings goes through the events list for a specific Component/Service/Pod
+// and checks if there exists an event with type Warning and returns the corresponding reason and message
+// as an additional supporting data
+func CheckEventsForWarnings(log *zap.SugaredLogger, events []corev1.Event, timeRange *files.TimeRange) (message []string, err error) {
```
bhatpmk, line 127: Can you please supply type Warning to this function, so that any change in the type can be handled easily in future ?

tools/vz/pkg/analysis/internal/util/cluster/install.go
```diff
@@ -412,3 +425,19 @@ func reportInstallIssue(log *zap.SugaredLogger, clusterRoot string, compsNotRead
 	issueReporter.AddKnownIssueMessagesFiles(report.ComponentsNotReady, clusterRoot, messages, files)
 	return nil
 }
+
+func getUniqueNamespaces(log *zap.SugaredLogger, compsNoMessages []string) map[string]bool {
```
bhatpmk, line 429: vzhelper.go has a function to get all the namespaces - GetNamespacesForAllComponents. Also there is a function to remove the duplicates in a slice in vzcapture.go. You can just make use of those existing fucntions.

tools/vz/pkg/analysis/internal/util/cluster/events.go
saketmht: That is what's being done in this function. The additional thing it does is instead of getting all the events, it first gets the list of pods for the component namespace and returns the list of events associated to those pods. Unmarshalling the events.json directly gives the list of events for not only pods but also for other resources like replicasets, deployements, services which I think would clutter the analysis report a bit too much. So for now only considering the events related to pods for failed components.

tools/vz/pkg/analysis/internal/util/cluster/events.go
saketmht: Sure

tools/vz/pkg/analysis/internal/util/cluster/events.go
```diff
@@ -95,6 +96,72 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
...
+	}
+
+	for _, event := range finalEventList {
+		message = append(message, fmt.Sprintf("Namespace: %s, %s %s, Message: %s, Reason: %s", event.InvolvedObject.Namespace, event.InvolvedObject.Kind, event.InvolvedObject.Name, event.Message, event.Reason))
+	}
+
+	return message, nil
+}
+
+func IsInvolvedObjectSame(eventA corev1.Event, eventB corev1.Event) bool {
```
bhatpmk, line 159: If this function is not useful outside of events.go, you can avoid exporting it.

tools/vz/pkg/analysis/internal/util/cluster/events.go
```diff
@@ -95,6 +96,72 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
...
+		}
+		componentEvents = append(componentEvents, eventList...)
+	}
+	return componentEvents, nil
+}
+
+// CheckEventsForWarnings goes through the events list for a specific Component/Service/Pod
+// and checks if there exists an event with type Warning and returns the corresponding reason and message
+// as an additional supporting data
+func CheckEventsForWarnings(log *zap.SugaredLogger, events []corev1.Event, eventType string, timeRange *files.TimeRange) (message []string, err error) {
```
bhatpmk, line 127: If you are not using time range, not sure why do you pass it to the function

tools/vz/pkg/analysis/internal/util/cluster/events.go
```diff
@@ -95,6 +96,72 @@ func GetEventsRelatedToService(log *zap.SugaredLogger, clusterRoot string, servi
...
+			}
+		}
+		// No previous event for this specific involved object
+		// Add to the finalEventList, if event type matches
+		if !foundInvolvedObject && event.Type == eventType {
+			finalEventList = append(finalEventList, event)
+		}
+	}
+
+	for _, event := range finalEventList {
```
bhatpmk, line 152: I remember seeing warnings without any message. Did you see any such case ? If so, can we consider the events with non-empty message ?

tools/vz/pkg/analysis/internal/util/cluster/events.go
saketmht: Fixed

tools/vz/pkg/analysis/internal/util/cluster/events.go
saketmht: Added a check for empty messages

tools/vz/pkg/analysis/internal/util/cluster/events.go
saketmht: The other functions in the file also have the timerange parameter which is not being used. Since we are planning to implement searching in a certain timerange in the future, I have added it there just like the other funcs.

agv426 (approved):

PR #4904, opened 2022-12-05
VZ-7875: Pipeline to download source files

bhatpmk:
This change adds a new Jenkinsfile to download the source from various repositories defined in an input properties file and creates an archive.

### DIFF
--- a/release/builds/JenkinsfilePRT
+++ b/release/builds/JenkinsfilePRT
@@ -103,14 +103,16 @@ pipeline {
         stage('Download Source') {
             steps {
                 script {
-                    sh """
-                        mkdir -p ${WORKSPACE}/downloaded_source
-                        ${WORKSPACE}/release/scripts/download_source_prt.sh ${WORKSPACE}/verrazzano_${VERRAZZANO_DEV_VERSION}-images.txt ${WORKSPACE}/repo_urls.properties ${WORKSPACE}/downloaded_source
-                        ls ${WORKSPACE}/downloaded_source
-                        mkdir -p ${WORKSPACE}/archive_dir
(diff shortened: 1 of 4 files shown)


release/builds/JenkinsfilePRT
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    parameters {
+        string (description: 'The release branch', name: 'RELEASE_BRANCH', defaultValue: 'NONE', trim: true)
+        string (description: 'The source commit for the release (required for full release)', name: 'COMMIT_TO_USE', defaultValue: 'NONE', trim: true )
+        string (description: 'The release version (major.minor.patch format, e.g. 1.5.0)', name: 'RELEASE_VERSION', defaultValue: 'NONE', trim: true)
+    }
+
+    environment {
+        OCR_CREDS = credentials('ocr-pull-and-push-account')
```
agv426, line 27: Remove these unused creds, looks like none of the ones in this block are used (at least directly). OCR_CREDS, NETRC_FILE, DOCKER_CREDS, DOCKER_REPO

release/builds/JenkinsfilePRT
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
```
agv426, line 1: Thinking we should use a better name for this Jenkinsfile, maybe JenkinsfileSourceArchive or something similar

If we ever automate the PRT process itself that Dave needs to follow, then we may call that JenkinsfilePRT,

bhatpmk: Done

release/builds/JenkinsfilePRT
bhatpmk: Done

agv426 (approved):

PR #4905, opened 2022-12-05
VZ-7876: fix fluentd config to prevent duplication of logs in OS

lake-of-dreams:
port of #4918

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 Fixes:
 
 - Fixed a bug that resulted in user applications restarting during upgrade from a previous Verrazzano release.
-- Fixed fluentd configuration to prevent de-duplication of logs in OpenSearch on fluentd restarts or upgrade.
+- Fixed fluentd configuration to prevent duplication of logs in OpenSearch on fluentd restarts or upgrade.
 
 ### v1.4.2
 Fixes:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 Fixes:
 
 - Fixed a bug that resulted in user applications restarting during upgrade from a previous Verrazzano release.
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 Fixes:
 
 - Fixed a bug that resulted in user applications restarting during upgrade from a previous Verrazzano release.
+- Fixed fluentd configuration to prevent duplication of logs in OpenSearch on fluentd restarts or upgrade.
```
bhatpmk, line 5: Can you please use "Fluentd" instead of "fluentd"

platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
```diff
@@ -164,7 +164,7 @@ data:
       @type tail
       # @id in_tail
       path /var/log/containers/*.log
-      pos_file /tmp/fluentd-containers.log.pos
+      pos_file /var/log/vz-fluentd-containers.log.pos
```
kminder, line 167: Are there any upgrade implications with this?

CHANGES.md
lake-of-dreams: fixed

platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
lake-of-dreams: 1. If upgrade happens from a version that has this fix to a version having this fix, no duplicate logs will be send to OS since the `pos_file` will be on host and contain last offsets.
2. if upgrade happens from a version that does not have the fix, the OS will still have duplicate logs once since `pos_file` will be anew and on host. any further restarts/upgrades should not result in duplicate logs

bhatpmk (approved):

PR #4915, opened 2022-12-05
Exclude PSR from top-level unit-test target

michael-cico:
Exclude tools/psr from the unit-test target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ unit-test-coverage-ratcheting:
 
 .PHONY: unit-test
 unit-test:
-	go test $$(go list ./... | grep -Ev /tests/e2e)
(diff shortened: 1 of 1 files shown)


anders-swanson (approved):

Makefile
```diff
@@ -98,7 +98,7 @@ unit-test-coverage-ratcheting:
 
 .PHONY: unit-test
 unit-test:
-	go test $$(go list ./... | grep -Ev /tests/e2e)
+	go test $$(go list ./... | grep -Ev "/tests/e2e|/tools/psr")
```
bobdonat, line 101: Won't that also exclude the psr unit tests?

michael-cico: Yes, we don't run those as part of the main unit/coverage stuff.  PSR is being intentionally isolated to it's own build.

PR #4917, opened 2022-12-06
VZ-7875: Use helper to get the list of repository URLs

bhatpmk:
This change gets the list of source repositories from the Verrazzano helper in the build pipeline. The script to download the source code is updated to allow additional repositories.

### DIFF
--- a/release/builds/JenkinsfileSourceArchive
+++ b/release/builds/JenkinsfileSourceArchive
@@ -46,6 +46,7 @@ pipeline {
         REPO_URLS_PROPS = "vzsource_repo_urls.properties"
         REPO_URLS1_PROPS = "vzsource_repo_urls_1.properties"
         REPO_URLS2_PROPS = "vzsource_repo_urls_2.properties"
+        REPO_URLS3_PROPS = "vzsource_repo_urls_3.properties
     }
 
     stages {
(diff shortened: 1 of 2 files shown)


release/builds/JenkinsfileSourceArchive
```diff
@@ -21,12 +21,17 @@ pipeline {
         string (description: 'The release branch', name: 'RELEASE_BRANCH', defaultValue: 'NONE', trim: true)
         string (description: 'The source commit for the release (required for full release)', name: 'COMMIT_TO_USE', defaultValue: 'NONE', trim: true )
         string (description: 'The release version (major.minor.patch format, e.g. 1.5.0)', name: 'RELEASE_VERSION', defaultValue: 'NONE', trim: true)
+        string (description: 'verrazzano-helper branch, user branch name is used when testing verrazzano-helper changes', name: 'VERRAZZANO_HELPER_BRANCH',
```
agv426, line 24: We need to create a followup I think for the default helper branch to use (ie: master or release-* based default).

For this change this is fine though

release/builds/JenkinsfileSourceArchive
```diff
@@ -83,48 +89,80 @@ pipeline {
...
+                GITHUB_AUTH_TOKEN = credentials('github-api-token-release-process')
+            }
+            steps {
+                script {
+                    sh """
+                        oci --region ${OCI_REGION} os object get --namespace ${OBJECT_STORAGE_NS} -bn ${OCI_OS_SHARED_BUCKET} --name ${params.VERRAZZANO_HELPER_BRANCH}/verrazzano-helper --file ${WORKSPACE}/verrazzano-helper
+                        chmod uog+x ${WORKSPACE}/verrazzano-helper
+
+                        # Redirect the output of verrazzano-helper get repo-urls to files
+                        ${WORKSPACE}/verrazzano-helper get repo-urls all > ${WORKSPACE}/${REPO_URLS_PROPS}
```
agv426, line 107: it may be useful for reviewing/debugging if we have the repo-url files and the image file it used kept as artifacts in the job.

More thinking for when we are doing releases, if something goes wrong it is easier for folks to eyeball them from a job artifact list rather than trying to piece together what may have gone wrong.

release/builds/JenkinsfileSourceArchive
bhatpmk: Agree

release/builds/JenkinsfileSourceArchive
bhatpmk: I have updated the PR to archive the repo-url files and the image file

agv426 (approved):

PR #4919, opened 2022-12-06
Add OpenSearch rule to PSR

mbenmclean:
This change adds an OpenSearch prometheusrule to the PSR chart. It also adds an istioInject field so that it will put things in and out of the mesh correctly


tools/psr/manifests/charts/worker/charts/opensearch/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  labels:
+    release: prometheus-operator
+  name: psr-{{ template "worker.fullName" .}}-opensearch
+  namespace: {{ .Release.Namespace }}
+spec:
+  groups:
+    - name: psr
+      rules:
+        - alert: LowOpenSearchStorageAvailability
```
stevshap, line 15: maybe LowOpenSearchAvailableStorage instead?

stevshap (dismissed):

tools/psr/manifests/usecases/prometheus/alerts.yaml
```diff
@@ -3,4 +3,4 @@
 
 global:
   envVars:
-    PSR_WORKER_TYPE: http-alerts
\ No newline at end of file
+    PSR_WORKER_TYPE: http-alerts
```
pfmackin, line 6: can you make this prom-alerts worker (follow up PR)

mbenmclean: fixed

stevshap (approved):

pfmackin (dismissed):

PR #4921, opened 2022-12-06
VZ-7637 - Enable use of Rancher proxy URL in kubeconfig used by managed clusters to connect to admin cluster.

desagar:
Before DNS updates, copy the Rancher ingress and TLS secret(s) so that managed clusters can use the old DNS name to access the admin cluster until the update is propagated.

### DIFF
--- a/platform-operator/controllers/verrazzano/reconcile/rancher.go
+++ b/platform-operator/controllers/verrazzano/reconcile/rancher.go
@@ -7,6 +7,7 @@ import (
 	"context"
 
 	"github.com/google/uuid"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/rancher"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/registry"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
@@ -15,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // createRancherIngressAndCertCopies - creates copies of Rancher ingress and cert secret if
@@ -33,6 +35,9 @@ func (r *Reconciler) createRancherIngressAndCertCopies(ctx spi.ComponentContext)
 			if !errors.IsNotFound(err) {
 				ctx.Log().Infof("Cannot make a copy of Rancher ingress %s/%s - get ingress failed: %v", ingressName.Namespace, ingressName.Name, err)
 			}
+			continue
+		}
+		if r.ingressCopyExists(ctx, ing) {
 			continue
 		}
 		newIngressTLSList := r.createIngressCertSecretCopies(ctx, ing)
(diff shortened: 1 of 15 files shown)


platform-operator/controllers/verrazzano/reconcile/rancher.go
```diff
@@ -0,0 +1,139 @@
...
+			ctx.Log().Debugf("Rancher ingress host %s has DNS suffix %s, skipping copy", ing.Spec.TLS[0].Hosts[0], dnsSuffix)
+			continue
+		}
+
+		newIngressTLSList := r.createIngressCertSecretCopies(ctx, ing)
+		newIngressNSN := types.NamespacedName{Name: "vz-" + ing.Name, Namespace: ingressName.Namespace}
+		err = r.createIngressCopy(newIngressNSN, ing, newIngressTLSList)
+		if err != nil {
+			ctx.Log().Infof("Failed to create a copy of Rancher ingress %s/%s - create ingress failed: %v", ingressName.Namespace, ingressName.Name, err)
+			continue
```
davisgomes, line 63: What is the justification for ignoring this error?

davisgomes: And the ones prior

platform-operator/controllers/verrazzano/reconcile/rancher_test.go
```diff
@@ -0,0 +1,104 @@
...
+				asserts.Equal(t, ing.Spec.TLS[i].SecretName, "vz-"+tls.SecretName)
+			}
+		}
+	}
+
+	secretList := corev1.SecretList{}
+	err = c.List(context.TODO(), &secretList)
+	asserts.NoError(t, err)
+	asserts.Equal(t, 4, len(secretList.Items))
+}
```
davisgomes, line 104: Could there be a test here that verifies that everything runs smoothly (or does what you expect) if some of the expected resources are missing?

platform-operator/controllers/verrazzano/reconcile/rancher.go
desagar: I think the idea I had was that this was not something that should fail the Rancher install/update and hence the VZ install/update. This copying is needed only in a subset of cases and there is a workaround i.e. manually reapplying the registration.
Of course, the downside is that diagnosability when multicluster does not work is made a bit more difficult. I am open to opinions on whether we should make this a failing change.

fredtibbitts: I had thought about this too, we'd have to make `beforeInstallComponents` return an error and then requeue and not proceed with installing components until it passes. I've gone back and forth, but I think I'm leaning towards making this a "best effort" and not fail the install. We can always make it more restrictive in the future.

tests/e2e/update/certac/certac_test.go
```diff
@@ -177,7 +217,88 @@ func verifyRegistration() {
...
+
+// reapplyManagedClusterRegManifest reapplies the registration manifest on managed clusters. For
+// self-signed certs, this manual step is expected of users so that the admin kubeconfig used by the
+// managed clusters can be reliably updated to use the new CA cert. Otherwise intermittent timing
+// related failures may occur
+func reapplyManagedClusterRegManifest(newCACert string) {
+	// CAUpdateTime := time.Now()
+	for _, managedCluster := range managedClusters {
+		waitForManifestSecretUpdated(managedCluster.Name, newCACert)
+		// waitForVMCReadyAfterTime(managedCluster.Name, CAUpdateTime)
```
fredtibbitts, line 233: Nit, remove commented code (and the one a few lines above).

tests/e2e/update/certac/certac_test.go
```diff
@@ -177,7 +217,88 @@ func verifyRegistration() {
...
+		gomega.Eventually(func() error {
+			reg, err := adminCluster.GetManifest(managedCluster.Name)
+			if err != nil {
+				pkg.Log(pkg.Error, fmt.Sprintf("could not get manifest for managed cluster %v error: %v", managedCluster.Name, err))
+				return err
+			}
+			pkg.Log(pkg.Info, fmt.Sprintf("Reapplying registration manifest for managed cluster %s", managedCluster.Name))
+			managedCluster.Apply(reg)
+			return nil
+		}).WithTimeout(waitTimeout).WithPolling(pollingInterval).Should(gomega.BeNil(), fmt.Sprintf("Reapply registration manifest failed for cluster %s", managedCluster.Name))
```
fredtibbitts, line 243: `ShouldNot(HaveOccurred())` reads a little better.

desagar: thanks, missed this one. I had updated it elsewhere. Fixed.

tests/e2e/update/certac/certac_test.go
desagar: done

platform-operator/controllers/verrazzano/reconcile/rancher_test.go
desagar: I'm not sure what you mean by missing resources, but I added one for when the copy already exists and made sure it was updated without errors. (e.g. when you do 2 DNS updates in succession).

davisgomes: I wasn't sure what feasible errors you would come across, but that seems reasonable to me.

davisgomes (approved):

fredtibbitts (approved):

davisgomes (approved):

PR #4926, opened 2022-12-06
VZ-7839: remove remaining Rancher custom resources during post uninstall

bobdonat:
This pull request removes any remaining Rancher custom resources during post uninstall.  In addition, an e2e test was added to make sure there are no Rancher resources remaining after uninstalling Rancher.


platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -110,13 +110,18 @@ func postUninstall(ctx spi.ComponentContext) error {
...
 	}
 
+	crds := getCRDList(ctx)
+
+	// Remove any Rancher custom resources that remain
+	removeCRs(ctx, crds)
+
 	// Remove any Rancher CRD finalizers that may be causing CRD deletion to hang
-	removeCRDFinalizers(ctx)
+	removeCRDFinalizers(ctx, crds)
```
michael-cico, line 119: Are some of the CRDs being deleted by Rancher?  We're not deleting any right?  I didn't think we were doing that ourselves.

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -147,12 +152,48 @@ func deleteWebhooks(ctx spi.ComponentContext) error {
...
 	}
+
+	return crds
+}
+
+// removeCRs deletes any remaining Rancher cattle.io cluster scoped custom resources
+func removeCRs(ctx spi.ComponentContext, crds *v1.CustomResourceDefinitionList) {
+	ctx.Log().Oncef("Removing Rancher cluster scoped custom resources")
+	for _, crd := range crds.Items {
+		if strings.HasSuffix(crd.Name, ".cattle.io") && crd.Spec.Scope == v1.ClusterScoped {
```
michael-cico, line 170: are there any namespace-scoped resources we need to clean up?  I think there are possibly Rancher user resources hanging around also.

bobdonat: I did not see any after an uninstall so I limited the removing to cluster-scoped resources.

michael-cico (dismissed):

michael-cico (approved):

PR #4943, opened 2022-12-07
VZ-7794 add validate pod calls to authpolicy_test

knelson7:


tests/e2e/istio/authz/authpolicy_test.go
```diff
@@ -38,8 +42,23 @@ var beforeSuite = t.BeforeSuiteFunc(func() {
 	deployFooApplication()
 	deployBarApplication()
 	deployNoIstioApplication()
-	beforeSuitePassed = true
 	metrics.Emit(t.Metrics.With("deployment_elapsed_time", time.Since(start).Milliseconds()))
+
+	//Resources for application bar
+	update.ValidatePods(sleepWorkloadName, labelPodName, barNamespace, 1, false)
```
kminder, line 48: Doesn't seem to me that these will produce a very meaningful error message it they fail.  Have you tried it?  Specifically the value in the test failure should be meaningful, not just a message in the test output somewhere.

tests/e2e/istio/authz/authpolicy_test.go
```diff
@@ -38,8 +42,23 @@ var beforeSuite = t.BeforeSuiteFunc(func() {
 	deployFooApplication()
 	deployBarApplication()
 	deployNoIstioApplication()
-	beforeSuitePassed = true
 	metrics.Emit(t.Metrics.With("deployment_elapsed_time", time.Since(start).Milliseconds()))
+
+	//Resources for application bar
```
kminder, line 47: Should this validation just be added to the various deploy*Application functions?

tests/e2e/istio/authz/authpolicy_test.go
knelson7: I tested it out by passing in the wrong replica count and the error printed is "expect 3 running pods, but got 1".
I added information to the error and now it prints "Deployment name springboot-frontend: expect 3 running pods, but got 1."
Not sure what additional things you'd want to add, but let me know. Things like deployName, labelName, and nameSpace could be easily added as well.
Validate pods is used in quite a few places so Abi and I were trying to avoid any big changes at first, but if it needs to be reworked I'll look into it.

tests/e2e/istio/authz/authpolicy_test.go
knelson7: I saw that that's how validate pods is normally used, but I checked with Abi and we don't want any of the tests to run until the pods are all up and running.

kminder (approved):
Approved.

tests/e2e/istio/authz/authpolicy_test.go
kminder: Adding the deployment name is sufficient for now.  In the future it might be meaningful to have a little more information about any existing pods that aren't running.  For example are the image pull issues or crash backoff loops happenning?

tests/e2e/istio/authz/authpolicy_test.go
kminder: I think you are missing my point.  I was just thinking that deployFooApplication would be more self contained if it didn't even return until is confirms the app was definitely up probably by calling update.ValudatePods.

Abhijithang (approved):

tests/e2e/istio/authz/authpolicy_test.go
Abhijithang: That's a good idea. I think every deploy functions should have update.ValidatePods so we stop where it fails. Thanks Kevin. Will take that up in the follow up PR.

PR #4945, opened 2022-12-07
VZ 7919: Handle errors in app operator main

anders-swanson:
Handle errors instead of continuing.

### DIFF
--- a/application-operator/internal/operatorinit/run_operator.go
+++ b/application-operator/internal/operatorinit/run_operator.go
@@ -4,7 +4,6 @@
 package operatorinit
 
 import (
-	"fmt"
 	clustersv1alpha1 "github.com/verrazzano/verrazzano/application-operator/apis/clusters/v1alpha1"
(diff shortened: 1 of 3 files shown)


application-operator/internal/operatorinit/run_operator.go
```diff
@@ -40,15 +41,15 @@ func StartApplicationOperator(metricsAddr string, enableLeaderElection bool, def
 		LeaderElectionID:   "5df248b3.verrazzano.io",
 	})
 	if err != nil {
-		log.Errorf("Failed to start manager: %v", err)
+		return fmt.Errorf("Failed to start manager: %v", err)
```
pfmackin, line 44: If a func is passed a logger then it should the error.  Use  return log.ErrorfNewErr(..)
Same for every place where that was changed.

davisgomes: The caller should not log the error in this case AFAIK. We always log the error as close to the source as we can wen the logger is present.

PR #4948, opened 2022-12-07
Update the DNS Cert Update test to give more reliable feedback

davisgomes:

### DIFF
--- a/tests/e2e/update/env-dns-cm/env_dns_cert_update_test.go
+++ b/tests/e2e/update/env-dns-cm/env_dns_cert_update_test.go
@@ -4,6 +4,7 @@
 package envdnscm
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"os/exec"
@@ -19,6 +20,10 @@ import (
 	"github.com/verrazzano/verrazzano/tests/e2e/pkg"
 	"github.com/verrazzano/verrazzano/tests/e2e/pkg/test/framework"
 	"github.com/verrazzano/verrazzano/tests/e2e/pkg/update"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


tests/e2e/update/env-dns-cm/env_dns_cert_update_test.go
```diff
@@ -81,191 +85,239 @@ var (
...
 		validateIngressList(currentEnvironmentName, currentDNSDomain)
 		validateVirtualServiceList(currentDNSDomain)
 	})
 
-	t.It("Update and verify environment name", func() {
+	t.Context("Update and verify environment name", func() {
 		m := EnvironmentNameModifier{testEnvironmentName}
-		update.UpdateCRWithRetries(m, pollingInterval, waitTimeout)
+		updateCR(m)
 		validateIngressList(testEnvironmentName, currentDNSDomain)
```
fredtibbitts, line 118: It's nice to see the actual `It` test specs in the Context, I think it makes it easier to reason about what the test is asserting without having to follow the functions, but it would be a lot of code duplication. I'm torn on this one.

fredtibbitts: If you do leave these functions it might be good to prefix the functions that have test specs with `It`. So `ItValidatesIngressList` for example. That way you can tell which functions have test specs vs which ones do not.

davisgomes: I'll leave these functions with the specified name change. It makes them easier to reuse for each context.

tests/e2e/update/env-dns-cm/env_dns_cert_update_test.go
```diff
@@ -81,191 +85,239 @@ var (
...
+}
+
+func verifyIngressAccess(log *zap.SugaredLogger) {
+	if log == nil {
+		log = zap.S()
+	}
+
+	t.DescribeTable("Access Ingresses",
+		func(access func() error) {
+			Eventually(func() error {
```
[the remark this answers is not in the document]
davisgomes: Good point. I'll update

fredtibbitts (approved):

PR #4949, opened 2022-12-07
Update the name for the Verrazzano Cluster Registrar user

davisgomes:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
+++ b/platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
@@ -7,6 +7,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
```diff
@@ -92,7 +95,7 @@ func createVZClusterUser(ctx spi.ComponentContext) error {
 	}
 
 	// Send a request to see if the user exists
-	reqURL := rc.BaseURL + usersByNamePath + vzconst.VerrazzanoClusterRancherUsername
+	reqURL := rc.BaseURL + usersByNamePath + strings.Replace(clusterRegName, " ", "%20", -1)
```
fredtibbitts, line 98: It would be better to use https://pkg.go.dev/net/url#PathEscape (or QueryEscape if it's a query param).

fredtibbitts (approved):

fredtibbitts (approved):

PR #4950, opened 2022-12-08
VZ-7892: Fix upgrade from 1.2.x to master by removing deprecated ES service

Abhijithang:
This PR removes the deprecated ES service when Verrazzano is upgraded from 1.2.x to master. ES service does not get removed because auth proxy was part of the Verrazzano component that was moved to its own component in 1.3.x. Therefore, the post-upgrade step is not executed, thus leaving the ES service on the cluster.


kminder (approved):
Approved.

fredtibbitts (approved):

PR #4967, opened 2022-12-09
Mask bearer token in log message

fredtibbitts:
When installing the cluster operator and Rancher is enabled, the VPO makes an API call to create a user. The request is logged and the auth bearer token showed up in clear text in the log. This PR masks the token.

Snippet from a log with the fix:
```
..."clusteroperator/cluster_operator.go:137","message":"Sending request message: 
Type: POST, URL: https://ingress-controller-ingress-nginx-controller.ingress-nginx/v3/users,
headers: map[Authorization:Bearer ****** Content-Type:application/json], payload...
```

### DIFF
--- a/platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
+++ b/platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
@@ -134,7 +134,7 @@ func createVZClusterUser(ctx spi.ComponentContext) error {
 	logMsg := fmt.Sprintf("Sending request message: Type: %s, URL: %s, headers: %v, payload %s", http.MethodPost, reqURL, headers, payload)
 	// Mask the bearer token in the log message
 	logMsg = vzpassword.MaskFunction("Authorization:Bearer ")(logMsg)
-	ctx.Log().Infof(logMsg)
+	ctx.Log().Debugf(logMsg)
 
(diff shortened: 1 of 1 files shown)


platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator.go
```diff
@@ -130,7 +130,12 @@ func createVZClusterUser(ctx spi.ComponentContext) error {
 	}
 
 	headers["Content-Type"] = "application/json"
-	ctx.Log().Infof("Sending request message: Type: %s, URL: %s, headers: %v, payload %s", http.MethodPost, reqURL, headers, payload)
+
+	logMsg := fmt.Sprintf("Sending request message: Type: %s, URL: %s, headers: %v, payload %s", http.MethodPost, reqURL, headers, payload)
```
michael-cico, line 134: Do we need to log this at all?  Or just log something like "Sending request to create Verrazzano cluster user in Rancher" ?

davisgomes: I think it's helpful for debugging, so maybe a debug message would be better

michael-cico (approved):

davisgomes (approved):

PR #4970, opened 2022-12-12
VZ-7931: service as a variable in grafana wls dashboards

adiforluls:
PR to make service a variable instead of a hardcoded value in wls grafana dashboards. Service name can be formatted to a different name than expected.


platform-operator/manifests/dashboards/weblogic/weblogic_dashboard.json
```diff
@@ -3809,6 +3809,38 @@
...
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "includeAll": false,
+        "label": "Service",
+        "multi": true,
+        "name": "service",
+        "options": [],
+        "query": {
+          "query": "label_values(wls_jvm_uptime{verrazzano_cluster=~\"$vzcluster\", weblogic_domainUID=\"$domainName\", weblogic_clusterName=\"$clusterName\",weblogic_serverName=\"$serverName\"},service)",
```
lake-of-dreams, line 3831: maybe I don't get it completely, but is the value of `service` variable being deduced from `service` variable which will on metrics `wls_jvm_uptime` ?

adiforluls: Yeah, followed a pattern here. That's how we're getting other values such as domain, cluster and server names.

lake-of-dreams: I find it a bit different, for example in https://github.com/verrazzano/verrazzano/pull/4970/files#diff-976e4655b510a3b2b80d703eac831f1d8da4a8fe72a3ed33e96bf24a7bbe308eL3799 , `serverName` is being deduced from a label `weblogic_serverName` and in https://github.com/verrazzano/verrazzano/pull/4970/files#diff-976e4655b510a3b2b80d703eac831f1d8da4a8fe72a3ed33e96bf24a7bbe308eL3764, the variable `clusterName` is being derived from `weblogic_clusterName`.

In this query however, it is reading the label `service` and from that creating the variable/label `service` which I find odd. Does the label `service` already exist on the metrics? if yes, why do we want to create another?

adiforluls: `service` is a label name in the metric, just like `weblogic_serverName` and `weblogic_clusterName`. With this query we're getting the value of the corresponding label (i.e. `service=<value>`). Currently we were hardcoding this based on domain and cluster name.

lake-of-dreams: maybe what will make it a bit less confusing is change the `name` to `serviceName` here

adiforluls: Renamed the var

lake-of-dreams (approved):

PR #4972, opened 2022-12-12
VZ-4482 fix component metrics names in metricsMap and take out unnecessary values 

knelson7:


platform-operator/controllers/verrazzano/component/registry/registry.go
```diff
@@ -106,6 +106,29 @@ func getComponents() []spi.Component {
 	return componentsRegistry
 }
 
+func GetMetricComponents() []spi.Component {
```
anders-swanson, line 109: Don't really like this, can we contain it to the metrics package? Maybe keep the inverse, a map of

platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -164,36 +138,23 @@ func initSimpleCounterMetricMap() map[metricName]*SimpleCounterMetric {
...
 		authproxyMetricName:            newMetricsComponent(authproxy.ComponentJSONName),
-		oamMetricName:                  newMetricsComponent(oam.ComponentJSONName),
-		appoperMetricName:              newMetricsComponent(appoper.ComponentJSONName),
 		istioMetricName:                newMetricsComponent(istio.ComponentJSONName),
-		weblogicMetricName:             newMetricsComponent(weblogic.ComponentJSONName),
 		nginxMetricName:                newMetricsComponent(nginx.ComponentJSONName),
 		certmanagerMetricName:          newMetricsComponent(certmanager.ComponentJSONName),
-		clusterOperatorMetricName:      newMetricsComponent(clusteroperator.ComponentJSONName),
 		externaldnsMetricName:          newMetricsComponent(externaldns.ComponentJSONName),
-		rancherMetricName:              newMetricsComponent(rancher.ComponentJSONName),
```
anders-swanson, line 175: Some of these components are necessary

platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -389,3 +424,7 @@ func SetComponentAvailabilityMetric(name string, availability vzapi.ComponentAva
 	MetricsExp.internalData.componentHealth.SetComponentHealth(compMetric.metricName, availability == vzapi.ComponentAvailable, isEnabled)
 	return nil
 }
+
+func IsNonMetricComponent(componentName string) bool {
+	return componentName == vmo.ComponentName || componentName == networkpolicies.ComponentName
```
anders-swanson, line 429: You can do like this, instead of using OR statement:
```
var nonMetricComponents = map[string]bool = {
  vmo.ComponentJSONName: true,
  networkpolicies.ComponentJSONName: true,
}
```

``` func IsNonMetricComponent(componentJSONName string) bool {
  return nonMetricComponents(componentJSONName)
```

This may help down the line if we need to add more components. Also, please note the use of the ComponentJSONName - I think that's what we always want to be using in this package for consistency?

platform-operator/controllers/verrazzano/healthcheck/availability.go
```diff
@@ -58,6 +58,10 @@ func (p *HealthChecker) newStatus(log vzlog.VerrazzanoLogger, vz *vzapi.Verrazza
 		Components: map[string]vzapi.ComponentAvailability{},
 	}
 	for _, component := range components {
+		// If component is not in the metricsMap, move on to the next component
+		if metricsexporter.IsNonMetricComponent(component.Name()) {
```
anders-swanson, line 62: I think we only want to skip the metrics logic if it's a non metric compnent? otherwise the rest of the availability logic is lost.

anders-swanson (approved):

platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -348,11 +349,17 @@ func AnalyzeVerrazzanoResourceMetrics(log vzlog.VerrazzanoLogger, cr vzapi.Verra
 				upgradeCompletionTime = status.LastTransitionTime
 			}
 		}
+		found, component := registry.FindComponent(componentName)
```
anders-swanson, line 352: fyi this makes a nested loop, but probably not a big deal

PR #4973, opened 2022-12-12
VZ-7965: Fetch namespaces from component registry

IshaGirdhar:
This PR removes the map that holds the component namespace mapping used by VZ cli tool, now tool will fetch the namespaces directly from the component registry.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -5,8 +5,6 @@ package constants
 
 import (
 	"time"
-
-	platformOperatorConstants "github.com/verrazzano/verrazzano/platform-operator/constants"
 )
 
 // RestartVersionAnnotation - the annotation used by user to tell Verrazzano application to restart its components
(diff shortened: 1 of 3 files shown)


agv426 (approved):

bhatpmk (approved):

pkg/constants/constants.go
```diff
@@ -175,71 +173,19 @@ const VerrazzanoClusterRancherUsername = "vz-cluster-reg"
 
 // Components Names
 const (
-	OamKubernetesRuntime          = "oam-kubernetes-runtime"
```
pfmackin, line 178: Why did you remove several entries?

IshaGirdhar: These entries were used only by the intermediate map that VZ tool was using, which we are not using anymore with this PR, so removed unused constants as well, rest are used in other code in VZ

bhatpmk (approved):

PR #4978, opened 2022-12-12
VZ 7901: PreInstall apply CRDs

anders-swanson:
Apply CRDs as part of PreInstall.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/vmo/vmo_component.go
+++ b/platform-operator/controllers/verrazzano/component/vmo/vmo_component.go
@@ -94,6 +94,14 @@ func (c vmoComponent) IsInstalled(ctx spi.ComponentContext) (bool, error) {
 	return true, nil
 }
 
+// PreInstall applies the VMO CRDs
+func (c vmoComponent) PreInstall(ctx spi.ComponentContext) error {
+	if err := common.ApplyCRDYaml(ctx, config.GetHelmVMOChartsDir()); err != nil {
+		return err
+	}
+	return c.HelmComponent.PreInstall(ctx)
+}
+
 // PreUpgrade VMO pre-upgrade processing
 func (c vmoComponent) PreUpgrade(context spi.ComponentContext) error {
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/vmo/vmo_component.go
```diff
@@ -94,6 +94,14 @@ func (c vmoComponent) IsInstalled(ctx spi.ComponentContext) (bool, error) {
 	return true, nil
 }
 
+// PreInstall applies the VMO CRDs
+func (c vmoComponent) PreInstall(ctx spi.ComponentContext) error {
```
adiforluls, line 98: vmo crd is currently applied by the verrazzano component in PreInstall/PreUpgrade, so this is not needed here I think. Or, the applying of vmo crd can be completely moved here in PreUpgrade as well.

https://github.com/verrazzano/verrazzano/blob/72dea82448eb5feeebd9833a7e13c5e6883123ca/platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_component.go#L77

pfmackin: I think this is for VMO chart CRDs

pfmackin (approved):

platform-operator/controllers/verrazzano/component/vmo/vmo_component.go
adiforluls: Right, if you look at verrazzano component PreInstall and PreUpgrade the VMO chart CRDs are being applied there currently. We may wanna remove it from there.

anders-swanson: Thanks, I will remove it.

anders-swanson: Looking again, `verrazzano` component still creates VMI resources in its Install, so leaving it for now. Once `verrazzano` component is further reduced, it could be removed.

PR #4997, opened 2022-12-15
VZ-7910: Make Rancher uninstall non-blocking

marcogelle:
The post-uninstall process of the Rancher component blocks the uninstallation of the other components due to using the Rancher uninstall system tool. This PR changes Rancher's post-uninstall to run in a background goroutine in order to prevent it from blocking rest of the uninstall process.

The unit tests in `rancher_uninstall_test.go` are modified in order to address the changes to the code.

The changes made are similar to how the Istio component installs in a background goroutine.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
@@ -6,11 +6,13 @@ package rancher
 import (
 	"context"
 	"fmt"
-	"github.com/verrazzano/verrazzano/pkg/os"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	osexec "os/exec"
 	"regexp"
 	"strings"
+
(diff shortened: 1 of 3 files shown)


michael-cico (approved):

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -54,8 +56,131 @@ var rancherSystemNS = []string{
...
+var postUninstallFunc postUninstallFuncSig = invokeRancherSystemToolAndCleanup
+
+// postUninstallRoutineParams - Used to pass args to the postUninstall goroutine
+type postUninstallRoutineParams struct {
+	ctx spi.ComponentContext
+}
+
+// uninstallMonitor - Represents a monitor object used by the component to monitor a background goroutine used for running
+// rancher uninstall operations asynchronously.
+type postUninstallMonitor interface {
```
pfmackin, line 75: This code looks the same as the Istio background check code.  Is it possible to make the code common in a follow up PR.

PR #5001, opened 2022-12-16

IshaGirdhar:

### DIFF
--- a/tests/e2e/examples/bobsbooks/bobs_books_test.go
+++ b/tests/e2e/examples/bobsbooks/bobs_books_test.go
@@ -278,14 +278,10 @@ var _ = t.Describe("Bobs Books test", Label("f:app-lcm.oam",
 		// WHEN the application configuration uses a default metrics trait
 		// THEN confirm that all the scrape targets are healthy
 		t.It("Verify all scrape targets are healthy for the application", func() {
-			pkg.Concurrently(
-				func() {
-					Eventually(func() (bool, error) {
-						var componentNames = []string{"bobby-coh", "bobby-helidon", "bobby-wls", "bobs-mysql-deployment", "bobs-mysql-service", "bobs-orders-wls", robertCoh, "robert-helidon"}
-						return pkg.ScrapeTargetsHealthy(pkg.GetScrapePools(namespace, "bob-books", componentNames))
-					}, shortWaitTimeout, shortPollingInterval).Should(BeTrue())
-				},
-			)
+			Eventually(func() (bool, error) {
+				var componentNames = []string{"bobby-coh", "bobby-helidon", "bobby-wls", "bobs-mysql-deployment", "bobs-mysql-service", "bobs-orders-wls", robertCoh, "robert-helidon"}
+				return pkg.ScrapeTargetsHealthy(pkg.GetScrapePools(namespace, "bob-books", componentNames))
+			}, shortWaitTimeout, shortPollingInterval).Should(BeTrue())
 		})
 		// Verify Istio Prometheus scraped metrics
 		// GIVEN a deployed Bob's Books application
--- a/tests/e2e/examples/bobsbooks/bobs_books_test.go
+++ b/tests/e2e/examples/bobsbooks/bobs_books_test.go
@@ -261,60 +261,15 @@ var _ = t.Describe("Bobs Books test", Label("f:app-lcm.oam",
 		})
 	})
 	t.Context("Metrics.", Label("f:observability.monitoring.prom"), FlakeAttempts(5), func() {
-		// Verify application Prometheus scraped metrics
+		// Verify application Prometheus scraped targets
 		// GIVEN a deployed Bob's Books application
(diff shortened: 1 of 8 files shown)


tests/e2e/pkg/metrics.go
```diff
@@ -178,6 +179,28 @@ func MetricsExist(metricsName, key, value string) bool {
 	return MetricsExistInCluster(metricsName, m, kubeconfigPath)
 }
 
+// ScrapeTargetsHealthy validates the health of the scrape targets for the given namespace
+func ScrapeTargetsHealthy(namespace string) bool {
```
kminder, line 183: Should this return bool, error?

kminder, line 183: I'm not a fan a just using the namespace.  For example if no scrape targets get created for some but should have this will still pass, right?  I think the caller should be passing in the scrape targets that it expects.

IshaGirdhar: Done

IshaGirdhar: Passing the exact list of scrapePools expected for that application now and either any of the scrape targets is not up or the target doesn't exist for the scrapePool, we are failing the test.

tests/e2e/examples/bobsbooks/bobs_books_test.go
```diff
@@ -261,60 +273,16 @@ var _ = t.Describe("Bobs Books test", Label("f:app-lcm.oam",
...
 	t.Context("Metrics.", Label("f:observability.monitoring.prom"), FlakeAttempts(5), func() {
-		// Verify application Prometheus scraped metrics
+		// Verify application Prometheus scraped targets
 		// GIVEN a deployed Bob's Books application
 		// WHEN the application configuration uses a default metrics trait
-		// THEN confirm that metrics are being collected
-		t.It("Retrieve application Prometheus scraped metrics", func() {
+		// THEN confirm that all the scrape targets are healthy
+		t.It("Verify all scrape targets are healthy for the application", func() {
 			pkg.Concurrently(
```
kminder, line 281: Is this concurrently still requried?

tests/e2e/examples/helidonconfig/helidon_config_test.go
```diff
@@ -172,21 +172,18 @@ var _ = t.Describe("Helidon Config OAM App test", Label("f:app-lcm.oam",
...
-	// Verify Prometheus scraped metrics
+	// Verify Prometheus scraped targets
 	// GIVEN OAM helidon-config app is deployed
 	// WHEN the component and appconfig without metrics-trait(using default) are created
-	// THEN the application metrics must be accessible
+	// THEN the application scrape targets must be healthy
 	t.Describe("Metrics.", Label("f:observability.monitoring.prom"), func() {
-		t.It("Retrieve Prometheus scraped metrics", func() {
+		t.It("Verify all scrape targets are healthy for the application", func() {
 			pkg.Concurrently(
```
kminder, line 181: Check all of these Concurrently uses to see if they are still required.

tests/e2e/pkg/metrics.go
```diff
@@ -178,6 +180,38 @@ func MetricsExist(metricsName, key, value string) bool {
...
+					return isHealthy, errors.New("target with scrapePool" + targetScrapePool + "and scrapeURL" + scrapeURL + "is not healthy")
+				}
+			}
+		}
+		// If target with scrapePool not found, then return false and error
+		if !found {
+			Log(Error, fmt.Sprintf("target with scrapePool %s is not found", scrapePool))
+			return isHealthy, errors.New("target with scrapePool" + scrapePool + "not found")
+		}
+		isHealthy = true
```
kminder, line 210: Something seems off here.  Couldn't this result in true being returned if some later target/pool is found to be unhealthy?  Why isn't the return on 212 just return true, nil and every other return fakse, errors.New(...)?

tests/e2e/examples/bobsbooks/bobs_books_test.go
IshaGirdhar: Removed

tests/e2e/examples/helidonconfig/helidon_config_test.go
IshaGirdhar: Done

tests/e2e/pkg/metrics.go
IshaGirdhar: This was an oversight, thanks for pointing it out, we actually didn't need this variable. Fixed.

tests/e2e/workloads/weblogic/weblogic_workload_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
kminder, line 1: This needs to be 2022, 2023,

PR #5003, opened 2022-12-16
VZ-7471: Fix for i/o timeout errors when installing Verrazzano on a RKE2 cluster

bobdonat:
This pull request fixes i/o timeout errors when installing Verrazzano on a RKE2 cluster.  The cause of the timeout was the VPO network policy which only allowed egress to one of three k8s API server IP addresses.  The agreed upon solution was to remove the egress from the VPO network policy.


platform-operator/internal/k8s/netpolicy/netpolicy.go
```diff
@@ -88,32 +88,47 @@ func getAPIServerServiceIPAndPort(clientset kubernetes.Interface) (string, int32
 	return "", 0, fmt.Errorf("unable to find a host and port for the kubernetes API server service")
 }
 
-// getAPIServerIPAndPort returns the IP address and port of the Kubernetes API server.
-func getAPIServerIPAndPort(clientset kubernetes.Interface) (string, int32, error) {
+// getAPIServerIPsAndPort returns an IP address list and port of the Kubernetes API server.
```
pfmackin, line 91: What happens on cluster upgrade where a new Kubernetes node is added?   I think we should just get rid of egress policy completely, we don't have it for any other component as far as I know.

bobdonat: I have pushed new changes to remove the egress policy for VPO.

pfmackin (approved):

PR #5011, opened 2022-12-19
VZ-7910 Part 2: Generalize Monitor code for background Istio install and Rancher uninstall

marcogelle:
This refactors the code that runs the Istio install and Rancher uninstall in the background. Related to [https://github.com/verrazzano/verrazzano/pull/4997](https://github.com/verrazzano/verrazzano/pull/4997). This PR aims to reduce duplicated code.

Pulls out the monitor object and its methods into the `platform-operator/controllers/verrazzano/component/common` package (I'm unsure about this, please advise if you think I should move it).

Also wrote some unit tests for the monitor object in `monitor_test.go`.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/monitor.go
+++ b/platform-operator/controllers/verrazzano/component/common/monitor.go
@@ -61,7 +61,6 @@ func (m *MonitorType) Run(operation BackgroundFunc) {
 	go func(outputChan chan bool) {
 		// The function will execute once, sending true on success, false on failure to the channel reader
 		err := operation()
-		defer func() { m.running = false }()
 		if err != nil {
 			outputChan <- false
 			return
--- a/platform-operator/controllers/verrazzano/component/common/monitor.go
+++ b/platform-operator/controllers/verrazzano/component/common/monitor.go
@@ -61,6 +61,7 @@ func (m *MonitorType) Run(operation BackgroundFunc) {
 	go func(outputChan chan bool) {
 		// The function will execute once, sending true on success, false on failure to the channel reader
 		err := operation()
+		defer func() { m.running = false }()
 		if err != nil {
 			outputChan <- false
 			return
--- a/platform-operator/controllers/verrazzano/component/common/monitor.go
+++ b/platform-operator/controllers/verrazzano/component/common/monitor.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package common
+
+import (
+	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
+)
+
+// Monitor - Represents a monitor object used by the component to monitor a background goroutine used for running
+// install, uninstall, etc. operations asynchronously.
+type Monitor interface {
+	// CheckResult - Checks for a result from the goroutine; returns either the result of the operation, or an error indicating
+	// the operation is still in progress
+	CheckResult() (bool, error)
+	// Reset - Resets the monitor and closes any open channels
+	Reset()
+	// IsRunning - returns true of the monitor/goroutine are active
+	IsRunning() bool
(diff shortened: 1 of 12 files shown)


platform-operator/controllers/verrazzano/component/common/monitor.go
```diff
@@ -0,0 +1,88 @@
...
+
+package common
+
+import (
+	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
+)
+
+// Monitor - Represents a monitor object used by the component to monitor a background goroutine used for running
+// install, uninstall, etc. operations asynchronously.
+type Monitor interface {
```
michael-cico, line 12: Do you think we should rename this to something more meaningful?  when it was buried in the Istio implementation code and not shared I don't think we cared as much.

Maybe `BackgroundProcessMonitor`, or something like that?

marcogelle: I agree that's a good name. So, I will change `Monitor` -> `BackgroundProcessMonitor`.
What do you think about the names of the other structs? Should I similarly change
`MonitorType` -> `BackgroundProcessMonitorType`
and
`FakeMonitorType` -> `FakeBackgroundProcessMonitorType`

michael-cico: Yes, that seems fine.  @pfmackin do you have any preferences or suggestions?

platform-operator/controllers/verrazzano/component/common/monitor.go
```diff
@@ -0,0 +1,88 @@
...
+var _ Monitor = &MonitorType{}
+
+// FakeMonitorType - a fake monitor object, useful for unit testing.
+type FakeMonitorType struct {
+	Result  bool
+	Err     error
+	Running bool
+}
+
+func (f *FakeMonitorType) CheckResult() (bool, error)   { return f.Result, f.Err }
```
anders-swanson, line 82: General practice with Fake objects is to have them in a seprate package, e.g., monitor/fake

marcogelle: Ok, I can make that change. So in the new monitor package I will make, will the file structure look like this?

monitor package directory, containing:
    - monitor.go
    - monitor_test.go
    - fake directory, containing:
        - fake_monitor.go
        - fake_monitor_test.go

anders-swanson: Yes, thank you

marcogelle: Actually nevermind, I think I answered my own question by looking at some existing examples in the repo. I'll work on this change.

platform-operator/internal/monitor/fake/fake_monitor.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package fake
+
+import "github.com/verrazzano/verrazzano/platform-operator/internal/monitor"
+
+// FakeMonitorType - a fake monitor object, useful for unit testing.
+type FakeBackgroundProcessMonitorType struct {
```
anders-swanson, line 9: because it's the `fake` package, you can just call this type BackgroundProcessMonitor or similar. Any user will reference it like `fake.BackgroundProcessMonitor{}`, so it's already visibly a fake.

anders-swanson (approved):

platform-operator/internal/monitor/fake/fake_monitor.go
anders-swanson: It's also a go style practice to not start type names with the package name, for that reason.

marcogelle: Ok, I'll rename it to `BackgroundProcessMonitorType`

anders-swanson (approved):

michael-cico (approved):

PR #5023, opened 2022-12-20
VZ-7963 update VPO component install and upgrade duration metrics 

knelson7:


platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -205,6 +206,24 @@ func initComponentHealthMetrics() *ComponentHealth {
 	}
 }
 
+func initComponentInstallDurationMetrics() *ComponentInstallDuration {
```
anders-swanson, line 209: Are you going to get rid of the old metrics in another PR?

platform-operator/metricsexporter/metricsexporter_utils.go
```diff
@@ -205,6 +206,24 @@ func initComponentHealthMetrics() *ComponentHealth {
...
+	return &ComponentInstallDuration{
+		installDuration: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "vz_platform_operator_component_install_duration_seconds",
+			Help: "The duration of the latest installation of each component in seconds",
+		}, []string{"component"}),
```
anders-swanson, line 214: use a const for "component" since you are using it multiple times

platform-operator/metricsexporter/metricsexporter_utils.go
knelson7: I was going to do it in this one. Is it better to split it up into a separate PR though? 
I haven't figured out the best to get rid of the old ones yet because if I don't need the two metrics defined in newMetricsComponent, then the struct is just a name...

anders-swanson: Yeah... That kind of comes down to how we decided to implement metrics for our controllers in a weird way. If you take a look at any other project, they declare their metrics are package-level variables: https://github.com/etcd-io/etcd/blob/9bb134233a42c0339cd7c5355dd9c2772e580944/server/etcdserver/metrics.go#L28
Which are then referenced from whatever piece of code needs to emit the metric. It's very simple and does away with a lot of the cruft we added.

anders-swanson: `newMetricsComponent` can probably go away... If you have the gaugevec, it wouldn't be useful anymore.

anders-swanson (approved):
LGTM

PR #5029, opened 2022-12-20
VZ-8067 Export metrics from the Cluster Operator

davisgomes:

### DIFF
--- a/cluster-operator/controllers/rancher/rancher_cluster_controller.go
+++ b/cluster-operator/controllers/rancher/rancher_cluster_controller.go
@@ -54,11 +54,11 @@ var (
 		Help: "The duration of the reconcile process for cluster objects",
 	})
 	reconcileErrorCount = promauto.NewCounter(prometheus.CounterOpts{
-		Name: "vz_cluster_operator_reconcile_errors_total",
+		Name: "vz_cluster_operator_reconcile_cluster_error_total",
 		Help: "The amount of errors encountered in the reconcile process",
 	})
 	reconcileSuccessCount = promauto.NewCounter(prometheus.CounterOpts{
-		Name: "vz_cluster_operator_reconcile_success_total",
+		Name: "vz_cluster_operator_reconcile_cluster_success_total",
 		Help: "The number of times the reconcile process succeeded",
 	})
 )
--- a/cluster-operator/controllers/rancher/rancher_cluster_controller.go
+++ b/cluster-operator/controllers/rancher/rancher_cluster_controller.go
@@ -6,10 +6,10 @@ package rancher
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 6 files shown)


cluster-operator/controllers/rancher/rancher_cluster_controller.go
```diff
@@ -45,6 +48,21 @@ var gvk = schema.GroupVersionKind{
...
+	})
+	reconcileErrorCount = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "vz_cluster_operator_reconcile_errors_total",
+		Help: "The amount of errors encountered in the reconcile process",
+	})
+	reconcileSuccessCount = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "vz_cluster_operator_reconcile_success_total",
+		Help: "The number of times the reconcile process succeeded",
+	})
+)
```
fredtibbitts, line 64: This operator has two controllers: this one for Rancher clusters and another one for VMCs. Should we have another similar set of metrics in vmc_controller.go? These names and descriptions would need to reflect they are for the Rancher clusters. Or was that something you planned to do in a follow-on PR?

fredtibbitts (approved):
Looks like a good start. Do we need to add/change any network policies or Istio auth policies to allow Prometheus scraping? We may have already done that, but if you tested in a cluster with Calico enabled and scraping works then we should be good.

cluster-operator/controllers/rancher/rancher_cluster_controller.go
```diff
@@ -45,6 +48,21 @@ var gvk = schema.GroupVersionKind{
 	Kind:    "Cluster",
 }
 
+var (
+	reconcileTimeMetric = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "vz_cluster_operator_reconcile_cluster_duration_seconds",
+		Help: "The duration of the reconcile process for cluster objects",
+	})
+	reconcileErrorCount = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "vz_cluster_operator_reconcile_errors_total",
```
knelson7, line 57: Re metric naming conventions as of now, we're using error singular—"vz_cluster_operator_reconcile_error_total". Forgot to mention it when you asked, my apologies.

cluster-operator/controllers/rancher/rancher_cluster_controller.go
davisgomes: Yes. I can add these for the other reconcile loop.

cluster-operator/controllers/rancher/rancher_cluster_controller.go
davisgomes: Thanks. I will update these names in the following push

knelson7 (approved):

PR #5035, opened 2022-12-21
VZ-7904: update VMO image to add default index patterns

karan56625:
update VMO image to add default metrics.


adiforluls (approved):
LGTM

PR #5043, opened 2022-12-22
VZ-7888 pod security e2e tests

mbenmclean:
This adds a pod security e2e test to confirm that pods are deploy with the correct pod security configuration

### DIFF
--- a/tests/e2e/verify-install/security/pod_security_test.go
+++ b/tests/e2e/verify-install/security/pod_security_test.go
@@ -3,15 +3,17 @@ package security
 import (
 	"context"
 	"fmt"
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	"strings"
+	"time"
+
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
 	"github.com/verrazzano/verrazzano/tests/e2e/pkg/test/framework"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"strings"
-	"time"
 )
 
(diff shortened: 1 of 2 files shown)


michael-cico (approved):

tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -0,0 +1,192 @@
...
+				podList, err = clientset.CoreV1().Pods(ns).List(context.TODO(), v1.ListOptions{})
+				return podList, err
+			}, waitTimeout, pollingInterval).ShouldNot(BeNil())
+
+			pods := podList.Items
+			for _, pod := range pods {
+				if shouldSkipPod(pod.Name, ns) {
+					continue
+				}
+				Expect(expectPodSecurityForNamespace(pod)).To(BeTrue())
```
michael-cico, line 79: Can you modify this so that the `expect` functions return a meaningful error, and expect the error to be nil or something?  That way it will include the error in the test result without having to poke through the logs.

This is to conform to the guidance from the recent e2e tests conventions that were agreed upon.

michael-cico (dismissed):

michael-cico (approved):

tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -0,0 +1,191 @@
...
+var _ = BeforeSuite(beforeSuite)
+
+var _ = t.AfterEach(func() {})
+
+var _ = t.Describe("Ensure pod security", Label("f:security.podsecurity"), func() {
+	kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
+	if err != nil {
+		Fail(fmt.Sprintf("Failed to get default kubeconfig path: %s", err.Error()))
+	}
+	for ns := range skipPods {
```
pfmackin, line 63: calling this skip pods is sort of misleading since it is a map of namespaces to check where you also have comma separated pods to skip.  I guess this is ok but it would be better to have a list of structs {namespace, []skiplist}.

pfmackin (approved):

PR #5045, opened 2022-12-23
VZ-8053.  Update pod security tests to test for non-standard capabilities

michael-cico:
Update pod security e2e tests to check for `Add` capabilities for any non-standard capabilities added

Also, reorganized the tests a bit:
- converted to a test table
- changed tests to accumulate all errors for a namespace and display them on failure, and not short-circuit on the first error

Sample failure report

```

  Expected
      <[]error | len:6, cap:6>: [
          <*errors.errorString | 0xc000760100>{
              s: "SecurityContext not configured correctly for pod verrazzano-application-operator-84f7658754-t5q7t, container verrazzano-application-operator, Privileged != false",
          },
          <*errors.errorString | 0xc000760140>{
              s: "SecurityContext not configured correctly for pod verrazzano-application-operator-84f7658754-t5q7t, container verrazzano-application-operator, AllowPrivilegeEscalation != false",
          },
          <*errors.errorString | 0xc000760190>{
              s: "only NET_BIND_SERVICE capability allowed, found unexpected capabilities added to container verrazzano-application-operator in pod verrazzano-application-operator-84f7658754-t5q7t: [NET_BIND_SERVICE MKNOD]",
          },
          <*errors.errorString | 0xc000760220>{
              s: "SecurityContext not configured correctly for pod verrazzano-authproxy-f9d8767f6-54np9, container verrazzano-authproxy, Privileged != false",
          },
          <*errors.errorString | 0xc000760260>{
              s: "SecurityContext not configured correctly for pod verrazzano-authproxy-f9d8767f6-54np9, container verrazzano-authproxy, AllowPrivilegeEscalation != false",
          },
          <*errors.errorString | 0xc0007602b0>{
              s: "only NET_BIND_SERVICE capability allowed, found unexpected capabilities added to container verrazzano-authproxy in pod verrazzano-authproxy-f9d8767f6-54np9: [NET_BIND_SERVICE MKNOD]",
          },
      ]
```

### DIFF
--- a/tests/e2e/verify-install/security/pod_security_test.go
+++ b/tests/e2e/verify-install/security/pod_security_test.go
@@ -65,7 +65,7 @@ var _ = t.Describe("Ensure pod security", Label("f:security.podsecurity"), func(
 		Fail(fmt.Sprintf("Failed to get default kubeconfig path: %s", err.Error()))
 	}
 	for ns := range skipPods {
-		ns := ns
+		ns := ns // needed to avoid the spec closure from capturing and retaining the last key in the map; see Ginkgo docs
 		t.ItMinimumVersion(fmt.Sprintf("Chek security for pods in namespace %s", ns), "1.5.0", kubeconfigPath, func() {
 			var podList *corev1.PodList
 			var err error
--- a/tests/e2e/verify-install/security/pod_security_test.go
+++ b/tests/e2e/verify-install/security/pod_security_test.go
@@ -6,8 +6,6 @@ package security
 import (
 	"context"
 	"fmt"
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"strings"
 	"time"
 
@@ -26,6 +24,9 @@ var t = framework.NewTestFramework("security")
 const (
 	waitTimeout     = 3 * time.Minute
 	pollingInterval = 5 * time.Second
+
+	// Only allowed capability in restricted mode
+	capNetBindService = "NET_BIND_SERVICE"
 )
 
 var skipPods = map[string][]string{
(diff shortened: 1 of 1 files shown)


tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -57,52 +70,57 @@ var _ = BeforeSuite(beforeSuite)
...
-				}
-				Expect(expectPodSecurityForNamespace(pod)).To(Not(HaveOccurred()))
+			podList, err = clientset.CoreV1().Pods(ns).List(context.TODO(), v1.ListOptions{})
+			return podList, err
+		}, waitTimeout, pollingInterval).ShouldNot(BeNil())
+
+		var errors []error
+		pods := podList.Items
+		for _, pod := range pods {
+			t.Logs.Infof("Checking pod %s/%s", ns, pod.Name)
```
mbenmclean, line 88: May want to remove this log as it might cause the ginkgo output to be a little noisy. Can be second pass though

mbenmclean (approved):

tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -57,52 +70,57 @@ var _ = BeforeSuite(beforeSuite)
...
-
-			pods := podList.Items
-			for _, pod := range pods {
-				if shouldSkipPod(pod.Name, ns) {
-					continue
-				}
-				Expect(expectPodSecurityForNamespace(pod)).To(Not(HaveOccurred()))
+			podList, err = clientset.CoreV1().Pods(ns).List(context.TODO(), v1.ListOptions{})
+			return podList, err
+		}, waitTimeout, pollingInterval).ShouldNot(BeNil())
```
fredtibbitts, line 83: It might be better to use `BeEmpty()`. It covers nil as well as empty sets. I'm not sure how much it matters in this case though.

fredtibbitts (dismissed):

tests/e2e/verify-install/security/pod_security_test.go
michael-cico: Yeah I left this in for verification of what pods it was finding during the checks, since we had issues with that.  I'll fix this and Fred's comment as I verified it after the pipeline runs.

tests/e2e/verify-install/security/pod_security_test.go
michael-cico: the func is returning a `PodList`, I tried switching to `BeEmpty()` and it hung because that object is not a slice/array.  So I'd have to return the `PodList.Items` array instead and do a nil-check inside the test func.  I'll probably just leave it as-is for now.

tests/e2e/verify-install/security/pod_security_test.go
michael-cico: Changed to `Debugf`.  Not sure how that can be enabled, but...

PR #5047, opened 2023-01-03
VZ-3124: Added support for DNS configuration analysis in vz

neeraj-n-singh:


tools/vz/pkg/analysis/internal/util/files/parse_install_log.go
```diff
@@ -51,3 +52,26 @@ func FilterLogsByLevelComponent(level string, component string, allMessages []Lo
 	}
 	return filteredLogs, nil
 }
+
```
agv426, line 55: Why are you adding yet another file search function here (similar/duplicate of other existing file utilities in text.go?)

Why not use one of the existing searchFile* variants here (which are also setting us up for supporting time range correlated searches for when that work is completed)

neeraj-n-singh: **FilterLogsByLevelComponent** expect logs format in a particular format which external logs DNS doesn't have.
Expected format:
type LogMessage struct {
	Level     string `json:"level"`
	Timestamp string `json:"@timestamp"`
	Message   string `json:"message"`
	Component string `json:"component"`
}
e
while external-dns pods are in the following format:

time="2021-07-22T15:11:35Z" level=info msg="Instantiating new Kubernetes client"
time="2021-07-22T15:11:35Z" level=info msg="Using inCluster-config based on serviceaccount-token"
time="2021-07-22T15:11:35Z" level=info msg="Created Kubernetes client [https://10.96.0.1:443](https://10.96.0.1/)"
time="2021-07-22T15:11:37Z" level=info msg="Created Istio client"
time="2021-07-22T15:11:40Z" level=error msg="getting zones: listing zones ****

1. There is no component here.
2.  It is not in json format too.
3. field name are also not in sync (message will be msg in logs)

This method is created in a generic way to keep future logs because different components might have different logging formats.

agv426: I'm talking about SearchLogMessage

That is pretty much doing what the text.go SearchFile* stuff already does

tools/vz/pkg/analysis/internal/util/cluster/install.go
```diff
@@ -468,3 +469,31 @@ func getUniqueNamespaces(log *zap.SugaredLogger, compsNoMessages []string) []str
 	uniqueNamespaces = helpers.RemoveDuplicate(uniqueNamespaces)
 	return uniqueNamespaces
 }
+
+func analyzeExternalDNS(log *zap.SugaredLogger, clusterRoot string, issueReporter *report.IssueReporter) {
+	errorMessagesRegex := []string{}
+	errorMessagesRegex = append(errorMessagesRegex, `.*level=error.*"getting zones: listing zones in.*Service error:NotAuthorizedOrNotFound.*`)
```
agv426, line 475: nit: should these be defined as constants instead of strings inlined

neeraj-n-singh: I will be assuming that if in the future we have more patterns for the same, then we can add them into the array instead of creating multiple constants. Thoughts?

agv426: sure that is fine too, just like using constants generally so they can be found and updated more easily

neeraj-n-singh: Let me think of another approach to maintain multiple regexes.

tools/vz/pkg/analysis/internal/util/files/parse_install_log.go
neeraj-n-singh: I updated the implementation.

agv426 (approved):

PR #5048, opened 2023-01-03
VZ-7982 Implement the Istio dashboards in Grafana

davisgomes:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/grafanadashboards/dashboards.go
+++ b/platform-operator/controllers/verrazzano/component/grafanadashboards/dashboards.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package grafanadashboards
--- a/platform-operator/controllers/verrazzano/component/grafanadashboards/dashboards.go
+++ b/platform-operator/controllers/verrazzano/component/grafanadashboards/dashboards.go
@@ -1,0 +1,18 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package grafanadashboards
+
+import (
+	"github.com/verrazzano/verrazzano/pkg/bom"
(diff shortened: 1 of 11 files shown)


platform-operator/helm_config/charts/verrazzano-grafana-dashboards/templates/configmaps.yaml
```diff
@@ -10,6 +10,9 @@ items:
   {{- range $path, $_ :=  .Files.Glob  "**/*.json" }}
   {{- $fileName := base $path }}
   {{- $cmName := trimPrefix "dashboards/" $path | trimSuffix ".json" | replace "/" "-" | replace "_" "-" | replace " " "-" | lower }}
+  {{- $subdir := trimSuffix $fileName $path | trimSuffix "/" | base }}
+# If Istio is disabled, skip deploying the Istio ConfigMaps
+{{- if or $.Values.istioEnabled (not (eq $subdir "Istio")) }}
```
desagar, line 15: this logic is a bit confusing to me. Why do we need the second condition?

davisgomes: This logic is a boolean simplification of this condition:

platform-operator/controllers/verrazzano/component/grafanadashboards/dashboards.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/verrazzano/verrazzano/pkg/bom"
+	"github.com/verrazzano/verrazzano/pkg/vzcr"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+)
+
+// AppendOverrides builds the set of Grafana dashboard overrides for the Helm install
+func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	if !vzcr.IsIstioEnabled(compContext.EffectiveCR()) || !vzcr.IsIstioEnabled(compContext.EffectiveCRV1Beta1()) {
```
desagar, line 14: can't we use enabled.IsIstioEnabled? Why do we have to pass in both here?

davisgomes: The IsIstioEnabled function accepts a runtime object. I believe I have to check both of the effective CRs to make sure if it's disabled unless I'm misunderstanding.

desagar (approved):

PR #5062, opened 2023-01-05
VZ-8106 alertmanager pod security

mbenmclean:
Adds pod security for alertmanager

### DIFF
--- a/platform-operator/helm_config/overrides/prometheus-operator-values.yaml
+++ b/platform-operator/helm_config/overrides/prometheus-operator-values.yaml
@@ -56,20 +56,21 @@ prometheus:
         replacement: local
 alertmanager:
   enabled: false
-  securityContext:
-    fsGroup: 65534
-    runAsGroup: 65534
-    runAsNonRoot: true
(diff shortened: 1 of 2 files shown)


platform-operator/helm_config/overrides/prometheus-operator-values.yaml
```diff
@@ -56,6 +56,21 @@ prometheus:
...
+      runAsUser: 65534
+      seccompProfile:
+        type: RuntimeDefault
+    containers:
+      - name: alertmanager
+        securityContext:
+          privileged: false
+      - name: config-reloader
+        securityContext:
+          privileged: false
```
michael-cico, line 73: do we need a drop `ALL` here?

mbenmclean: > 

Nope most of the container security context stuff is included by default. The only thing we need to add is the privileged: false. I included alertmanager in the e2e tests and it passes.

michael-cico (approved):

PR #5071, opened 2023-01-06
VZ-8130: fix OpenSearch wait target version

adiforluls:
The ESWaitTargetVersion variable is set as an environment variable in VMO container through a helm chart option. This PR is a one liner fix to update the correct wait target version for OpenSearch

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-monitoring-operator/values.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-monitoring-operator/values.yaml
@@ -14,7 +14,7 @@ monitoringOperator:
   defaultSimpleCompReplicas: 1
   defaultPrometheusReplicas: 1
   alertManagerImage: "noimage"
-  esWaitTargetVersion: 1.2.3
+  esWaitTargetVersion: 2.3.0
   oidcAuthEnabled: true
   RequestMemory: 48Mi
 


platform-operator/helm_config/charts/verrazzano-monitoring-operator/values.yaml
```diff
@@ -14,7 +14,7 @@ monitoringOperator:
   defaultSimpleCompReplicas: 1
   defaultPrometheusReplicas: 1
   alertManagerImage: "noimage"
-  esWaitTargetVersion: 1.2.3
+  esWaitTargetVersion: 2.3.0
```
karan56625, line 17: will this change resolve the issue while upgrading. For older version, it would be 1.2.3. I guess, we should update it to 2.3 while upgrading.

adiforluls: I did a 1.4 install and upgraded to my branch and did not see that error popping in VMO that is seen currently with master. The variable is read from an env var in VMO which is set through this helm chart option. For older version of Verrazzano the VMO chart will still be pointing to 1.2.3 since we're not changing this in a release, but for master this now needs to point to 2.3.0.

karan56625: ok. If this variable is read from ENV , then changes make sense.

karan56625 (approved):

PR #5072, opened 2023-01-06
VZ-8024: update fluentd daemonset to meet pod and container security standards

adiforluls:
PR to add pod and container security context to fluentd. Tried to run the pod as non-root fluent user, but from local testing,  and from some issues reported on the internet and github, it's not feasible (Refs: Some info [here](https://github.com/fluent/fluentd-kubernetes-daemonset/issues/420) and [here](https://github.com/fluent/fluentd-kubernetes-daemonset#run-as-root)). The `/var/log` dir mounted from the host node is owned by root, we don't control enough to make this work.

After dropping all caps from the fluentd container, had to add back `DAC_OVERRIDE` cap for input tail plugin, found info [here](https://docs.fluentd.org/deployment/linux-capability)
Without this, some logs were not being sent due to permission problems.

Also updated e2e tests.


tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -51,6 +48,9 @@ var skipPods = map[string][]string{
 var skipContainers = []string{}
 var skipInitContainers = []string{"istio-init"}
 
+// Only allowed capabilities in restricted mode
+var allowedCapabilities = []string{"NET_BIND_SERVICE", "DAC_OVERRIDE"}
```
michael-cico, line 52: I don't think we want to allow DAC_OVERRIDE for everything.  This might be better as a per-service/pod exception in the `podExceptions`  map below.

michael-cico (approved):

PR #5090, opened 2023-01-06
Move release-1.2 provisioning to use the new JAO

agv426:
We are moving our Jenkins runner provisioning to be handled by the Jenkins Agent Operator now instead of the Jenkins OCI plugin.

The new labels are ones that the PRODUCTION JAO is handling.

This will address the need for manually provisioning instances for older release lines and other cases, and will make managing out cloud instance configurations much easier overall (and easier to automate)


agv426:
Closing this as we are NOT moving the older releases over to the JAO we are disabling the crons for them instead.

If we need to revive the jobs for it and move it to the JAO in the future we can simply form a new PR to change the labels

PR #5093, opened 2023-01-09
VZ-8111 - Move logic to repair MySQL pods into VPO as a monitor thread

mgianatagh:
Move logic to repair MySQL pods into VPO as a monitor thread.  Remove the corresponding work around logic from the test suite.

- Moved all MySQL repair logic to be part of the new MySQL checker thread folder
- Enabled the checking and repair of MySQL pods stuck terminating to be part of the new checker thread.  Over time all work arounds for MySQL will be part of the checker thread.  Just moving one initially.
- Changed deployment of hello-helidon app to use three replicas for the test run


platform-operator/controllers/mysqlcheck/mysqlchecker_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestStart(t *testing.T) {
+	p := newTestMySQLCheck()
+	p.Start()
```
michael-cico, line 15: nothing really to assert here?

michael-cico (approved):

platform-operator/controllers/mysqlcheck/mysqlchecker_test.go
mgianatagh: I added some asserts.  They may grow over time.

platform-operator/controllers/mysqlcheck/mysqlchecker.go
```diff
@@ -0,0 +1,86 @@
...
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+// Pause pauses the MySQLChecker if it was running.
+// It is safe to call Pause multiple times
+func (mc *MySQLChecker) Pause() {
```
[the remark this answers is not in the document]
mgianatagh: It is only used for testing for now.  I used the HealthChecker code as a reference for these changes, and it is not using Pause yet. But I may enable it in a future PR.

PR #5097, opened 2023-01-09
VZ-8112: Add repair of MySQL pods waiting for readiness gates to VPO

mgianatagh:
Add check for MySQL pods stuck waiting for readiness gates to VPO mysqlcheck thread.

Remove MySQL workarounds for readiness gates from the end-to-end tests.


platform-operator/controllers/verrazzano/mysqlcheck/mysql_workarounds.go
```diff
@@ -28,6 +28,7 @@ const (
 	helmReleaseName     = "mysql"
 	componentNamespace  = "keycloak"
 	componentName       = "mysql"
+	repairTimeoutPeriod = 5 * time.Minute
```
michael-cico, line 31: should this be tunable?

michael-cico: maybe in a follow-up

michael-cico (approved):

PR #5102, opened 2023-01-10
VZ-8104 OAM Operator pod security

mbenmclean:
This change adds pod security for the OAM operator. This required modifying the oam operator image, so this change also picks up the new image.


platform-operator/helm_config/overrides/oam-kubernetes-runtime-values.yaml
```diff
@@ -1,8 +1,20 @@
...
 
 # Removing the limits for OAM since there could be many OAM apps deployed
 resources:
   requests:
     cpu: 100m
-    memory: 20Mi
\ No newline at end of file
+    memory: 20Mi
+
+podSecurityContext:
```
michael-cico, line 10: do we need to use the user/group IDs?

mbenmclean: It sets it correctly by default. I can add it explicitly if you'd like.

mbenmclean: Actually it doesn't I'll update that.

platform-operator/helm_config/overrides/oam-kubernetes-runtime-values.yaml
```diff
@@ -1,8 +1,23 @@
...
+  fsGroup: 65532
```
pfmackin, line 11: is fsGroup really needed?

michael-cico (approved):

PR #5106, opened 2023-01-10
VZ-8101 Keycloak pod security

pfmackin:
Fix Keycloak security contexts to comply with Pod security.


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -81,6 +81,15 @@ const kcInitContainerValueTemplate = `
           mountPath: /theme
         - name: cacerts
           mountPath: /cacerts
+      securityContext:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+        privileged: false
+        runAsGroup: 0
```
mbenmclean, line 90: Is this right? Wouldn't that be root. For all the examples I've seen runAsGroup has had the same value as runAsUser

platform-operator/helm_config/overrides/keycloak-values.yaml
```diff
@@ -98,7 +98,17 @@ ingress:
       secretName: "{{ .Values.tlsSecret }}"
 
 securityContext:
-  runAsUser: 0
-  runAsNonRoot: false
+  allowPrivilegeEscalation: false
+  privileged: false
+  runAsNonRoot: true
+  runAsUser: 1000
+  runAsGroup: 0
```
mbenmclean, line 105: Same question about runAsGroup

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
michael-cico, line 90: is `runAsGroup: 0` intentional?

pfmackin: runAsGroup 0 is allowed for restricted.  In this case the user in /etc/passwd has group 0 so it has to be used.

mbenmclean: ok if you say it's fine then I trust you

mbenmclean (approved):

PR #5123, opened 2023-01-12
VZ-8099 mysql and mysql router pod security

pfmackin:
Change mysql innodbCluster template to add pod security.  There are certain capabilities that must remain.  Also fixdatadir container needs to run as root.

### DIFF
--- a/platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
+++ b/platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
@@ -10,9 +10,19 @@ spec:
   tlsUseSelfSigned: {{ $use_self_signed }}
   router:
     instances: {{ required "routerInstances is required" .Values.routerInstances }}
-  {{ if ((.Values).router.podSpec) }}
-    podSpec: {{ toYaml ((.Values).router.podSpec) | nindent 6 }}
-  {{ end }}
+    podSpec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: router
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
+            runAsNonRoot: true
+            capabilities:
+              drop:
+                - ALL
   {{- if not $use_self_signed }}
   {{- $default_secret_name := printf "%s-router-tls" $cluster_name }}
   {{- $secret_name := default $default_secret_name ((.Values.tls).routerCertAndPKsecretName) }}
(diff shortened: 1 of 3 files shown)


platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
```diff
@@ -10,9 +10,19 @@ spec:
   tlsUseSelfSigned: {{ $use_self_signed }}
   router:
     instances: {{ required "routerInstances is required" .Values.routerInstances }}
-  {{ if ((.Values).router.podSpec) }}
-    podSpec: {{ toYaml ((.Values).router.podSpec) | nindent 6 }}
-  {{ end }}
+    podSpec:
```
pshirodkar-orcl, line 13: This probably could have been overridden via https://github.com/verrazzano/verrazzano/blob/master/platform-operator/helm_config/overrides/mysql-values.yaml but I guess it's okay

tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -32,7 +32,7 @@ const (
 
 var skipPods = map[string][]string{
 	"keycloak": {
-		"mysql",
+		"mysql-0",
```
pshirodkar-orcl, line 35: do we only have zeroth STS?

platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
```diff
@@ -10,9 +10,19 @@ spec:
   tlsUseSelfSigned: {{ $use_self_signed }}
   router:
     instances: {{ required "routerInstances is required" .Values.routerInstances }}
-  {{ if ((.Values).router.podSpec) }}
-    podSpec: {{ toYaml ((.Values).router.podSpec) | nindent 6 }}
```
mbenmclean, line 14: Why delete this and not just put this in the values overrides file?

pfmackin: I have to change the template anyway so i would rather just have it all here.

tests/e2e/verify-install/security/pod_security_test.go
pfmackin: yes for that test that this runs

platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
pfmackin: I wanted to have all the changes in the template since i had to change the template anyway.  The next version of MySQL is supposed to fix these settings so we just replace the chart.

mbenmclean (dismissed):

saidu941 (approved):

pshirodkar-orcl (approved):

PR #5132, opened 2023-01-13
VZ-8115: Pod Security and container security for Kiali

pshirodkar-orcl:
Pod Security and container security for Kiali.

### DIFF
--- a/platform-operator/helm_config/overrides/kiali-server-values.yaml
+++ b/platform-operator/helm_config/overrides/kiali-server-values.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+# Copyright (c) 2021, 2023, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 # NOTE: The image you're looking for isn't here. The mysql and linux images now come from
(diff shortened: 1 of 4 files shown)


platform-operator/helm_config/overrides/kiali-server-values.yaml
```diff
@@ -18,6 +18,22 @@ deployment:
     enabled: false
   instance_name: vmi-system-kiali
   view_only_mode: true
+  pod_security_context:
```
pshirodkar-orcl, line 21: Before: Pod security warning entry for kiali can be seen.
Screenshot 2023-01-12 at 3 49 04 PM

After: Pod security warning entry for kiali is no longer seen.
Screenshot 2023-01-13 at 12 13 41 PM

mbenmclean (changes requested):
modify pod_security_test.go skipPods variable to ensure that kiali gets tested. Make the vmi-system prefix more specific so it doesn't catch Kiali and Kiali gets tested.

pfmackin (approved):

pshirodkar-orcl:
> modify pod_security_test.go skipPods variable to ensure that kiali gets tested. Make the vmi-system prefix more specific so it doesn't catch Kiali and Kiali gets tested.

I have addressed your comment now.

mbenmclean (approved):

PR #5133, opened 2023-01-13
JIRA:VZ-8071, VZ-8150 support creation of Keycloak client for registered managed clusters

jmaron99:
This PR addresses the ability to:
- configure the auth proxy to use a managed-cluster-specific Keycloak client (managed cluster)
- invoke the Keycloak API to create the managed cluster client (admin)

MC successful build:  https://build.verrazzano.io/job/verrazzano-multi-cluster-acceptance-tests/job/jmaron%252FVZ-8150/6/


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1611,15 +1706,22 @@ func populateSubdomainInTemplate(ctx spi.ComponentContext, tmpl string) (string,
...
-		return "", err
+	var dnsSubDomain string
+	if subdomain == nil {
+		dnsSubDomain, err = getDNSDomain(ctx.Client(), ctx.EffectiveCR())
+		if err != nil {
+			ctx.Log().Errorf("Component Keycloak failed retrieving DNS sub domain: %v", err)
+			return "", err
+		}
+	} else {
+		dnsSubDomain = *subdomain
```
desagar, line 1717: why the string pointer and the dereferencing? Was there any reason you could not just pass subdomain as a string instead of *string? Strings have "" as the default empty value (so you can pass that/check for that instead of nil).

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -878,7 +879,7 @@ func (r *Reconciler) isManagedClusterRegistrationSecret(o client.Object) bool {
 	if secret.Namespace != vzconst.VerrazzanoSystemNamespace || secret.Name != vzconst.MCRegistrationSecret {
 		return false
 	}
-	r.AddWatch(fluentd.ComponentJSONName, jaegeroperator.ComponentJSONName)
+	r.AddWatch(fluentd.ComponentJSONName, jaegeroperator.ComponentJSONName, authproxy.ComponentJSONName)
```
desagar, line 882: remind me why VZ component needs to watch the authProxy?

platform-operator/internal/vzconfig/ingress.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2021, 2023, Oracle and/or its affiliates.
```
desagar, line 1: this file only has a copyright change and nothing else.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
jmaron99: Yep - just general personal preference to use 'nil' instead of len(string).

platform-operator/controllers/verrazzano/reconcile/controller.go
jmaron99: It's not exactly adding authproxy as a watcher but rather a "watchee".  Invoking AddWatch() ensure that a change to the registration secret (creation, update) will trigger authproxy's reconciliation.

platform-operator/internal/vzconfig/ingress.go
jmaron99: I know - I had it in the PR, decided it wasn't required, but our build logic won't let it go...I get an error from the build if I try to take it out.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
desagar: I don't remember where I've seen this but I think the recommendation is to use "" unless you need to distinguish nil from "".

jmaron99: How strongly do you feel about me changing it?  I'll mention that a search in our code base brought up quite a few examples of the use of string pointers - it's part of the reason I felt it was acceptable.

desagar (approved):

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -501,6 +511,46 @@ func (r *VerrazzanoManagedClusterReconciler) getVerrazzanoResource() (*v1beta1.V
...
+	}
+
+	err = keycloak.LoginKeycloak(ctx, cfg, cli)
+	if err != nil {
+		return err
+	}
+
+	promHost := vmc.Status.PrometheusHost
+	if len(promHost) == 0 {
+		return fmt.Errorf("Prometheus host not yet available from VMC Status")
```
fredtibbitts, line 537: It could be a long time before the prometheus host is populated, is this going to fill the operator logs up with errors?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1049,11 +1140,16 @@ func getSecretPassword(ctx spi.ComponentContext, namespace string, secretname st
 
 // getDNSDomain returns the DNS Domain
 func getDNSDomain(c client.Client, vz *vzapi.Verrazzano) (string, error) {
+	var dnsDomain string
 	dnsSuffix, err := vzconfig.GetDNSSuffix(c, vz)
 	if err != nil {
 		return "", err
 	}
-	dnsDomain := fmt.Sprintf("%s.%s", vz.Spec.EnvironmentName, dnsSuffix)
```
fredtibbitts, line 1056: Curious about this change, was this just broken before?

platform-operator/controllers/verrazzano/component/spi/component_context.go
```diff
@@ -38,6 +38,15 @@ func NewContext(log vzlog.VerrazzanoLogger, c clipkg.Client, actualCR *v1alpha1.
 	}, nil
 }
 
+// NewMinimalContext creates a ComponentContext limited to kubernetes client interactions and logging
+func NewMinimalContext(c clipkg.Client, log vzlog.VerrazzanoLogger) (ComponentContext, error) {
+	// Generate the effective CR based ond the declared profile and any overrides in the user-supplied one
```
fredtibbitts, line 43: "ond"

fredtibbitts (dismissed):
A couple of questions, and a few of the files should have their imports formatted correctly, but otherwise seems fine to me.

cluster-operator/controllers/vmc/vmc_controller.go
jmaron99: Not certain.  I imagine we can address if it does appear to be an issue.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
jmaron99: It was modified to account for an invocation with a minimal context that does not contain the effective CR, which now can occur for a managed cluster client creation (the logic is being invoked outside the context of a component reconciliation)

cluster-operator/controllers/vmc/vmc_controller.go
desagar: good point. Maybe that should not be an errorf - maybe a debugf. In sync_prometheus.go we generally silently wait.

platform-operator/controllers/verrazzano/component/spi/component_context.go
jmaron99: Oops

cluster-operator/controllers/vmc/vmc_controller.go
jmaron99: I'm not sure I can do that - I need the reconciliation to get re-queued which generally requires an error

jmaron99: To be clear - I'm calling fmt.Errorf so I'm generating an error not logging an error

fredtibbitts: I think the base controller reconcile code will log the error though, but I'm not 100% sure about that.

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -165,7 +165,8 @@ data:
     local validators = require "resty.jwt-validators"
 
     local oidcRealm = "{{ .OidcRealm }}"
-    local oidcClient = "{{ .PKCEClientID }}"
+    local oidcClient = "{{ .OIDCClientID }}"
+    local masterOidcClient = "{{ .PKCEClientID }}"
```
desagar, line 169: minor point, but naming this variable the "master" oidcClient is a bit misleading due to the overloading of that term ("master" branch) - maybe call it the adminClusterOidcClient?

jmaron99: Sure.  I'll just re-run the 2.5 hours of tests ;)

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
```diff
@@ -48,6 +48,9 @@ spec:
         sidecar.istio.io/inject: "true"
         traffic.sidecar.istio.io/excludeOutboundPorts: 443,6443
         verrazzano.io/metricsEnabled: "true"
+        {{ if .Values.managedClusterRegistered }}
+        verrazzano.io/managedClusterRegistered: "true"
```
desagar, line 52: and this is kept up to date if the registration is changed/deleted because we have a watch for changes to the registration secret to trigger authproxy component reconciliation?

jmaron99: yes

jmaron99: https://github.com/verrazzano/verrazzano/blob/1da1729685706ea5db1eac33df61de8063dd3387/platform-operator/controllers/verrazzano/reconcile/controller.go#L882

desagar (approved):

PR #5137, opened 2023-01-16
Add verrazzano_component label to kubelet and kube-state-metrics metrics

lake-of-dreams:
The regex for each component for pod level metrics is added in pattern
- `"(<namespace>);(<componentName>);(<regex-pattern-for-pod>)(.*)"`
and for pvc level metrics
- `"(<namespace>);(<componentName>);(<regex-pattern-for-pvc>)(.*)"
so that capture group `$2` will give target value of `verrazzano_component`

Also added an e2e test to check health status of `kubelet` and `kube-state-metrics` endpoints

### DIFF
--- a/platform-operator/constants/constants.go
+++ b/platform-operator/constants/constants.go
@@ -244,3 +244,9 @@ const RancherBackupNamesSpace = "cattle-resources-system"
 const VerrazzanoManagedKey = "verrazzano.io/namespace"
 
 const ExternalDNSNamespace = "cert-manager"
+
+// ServiceMonitorNameKubelet indicates the name of serviceMonitor resource for kubelet monitoring
+const ServiceMonitorNameKubelet = "prometheus-operator-kube-p-kubelet"
+
+// ServiceMonitorNameKubeStateMetrics indicates the name of serviceMonitor resource for kube-state-metrics monitoring
+const ServiceMonitorNameKubeStateMetrics = "kube-state-metrics"
--- a/platform-operator/helm_config/overrides/kube-state-metrics-values.yaml
+++ b/platform-operator/helm_config/overrides/kube-state-metrics-values.yaml
@@ -8,6 +8,288 @@ customLabels:
 prometheus:
   monitor:
     honorLabels: true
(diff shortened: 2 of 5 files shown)


platform-operator/controllers/verrazzano/component/common/prometheus.go
```diff
@@ -19,7 +19,60 @@ import (
...
+const (
+	keycloakHTTPService = "keycloak-http"
+
+	// LabelVerrazzanoComponent is name of label containing Verrazzano component name for a metrics.
+	LabelVerrazzanoComponent = "verrazzano_component"
+)
+
+// verrazzanoComponentRelabelConfigPodTemplates is mapping of Verrazzano component names to the regular expression template for capturing pod related metrics.
+var verrazzanoComponentRelabelConfigPodTemplates = map[string]string{
+	"Alertmanager":                    "(verrazzano-monitoring);(%s);(alertmanager-[a-z0-9-]+)(.*)",
```
desagar, line 31: The label values seem named inconsistently - there’s “Verrazzano Monitoring Operator”, “Alertmanager”, “cert-manager” - I’m thinking we should align these names to the component names in the VZ CR.

fredtibbitts (commented):
The metrics relabelings can be done right in the helm chart with helm overrides. The kube-state-metrics service monitor already supports this. See platform-operator/helm_config/overrides/kube-state-metrics-values.yaml.

It seems like it would be much cleaner and simpler to put the relabeling config in the helm overrides.

platform-operator/controllers/verrazzano/component/common/prometheus.go
lake-of-dreams: fixed

lake-of-dreams:
> The metrics relabelings can be done right in the helm chart with helm overrides. The kube-state-metrics service monitor already supports this. See platform-operator/helm_config/overrides/kube-state-metrics-values.yaml.
> 
> It seems like it would be much cleaner and simpler to put the relabeling config in the helm overrides.

moved relabelling to overrides

PR #5142, opened 2023-01-17

IshaGirdhar:


tests/e2e/examples/bobsbooks/bobs_books_test.go
```diff
@@ -261,63 +277,21 @@ var _ = t.Describe("Bobs Books test", Label("f:app-lcm.oam",
 		})
 	})
 	t.Context("Metrics.", Label("f:observability.monitoring.prom"), FlakeAttempts(5), func() {
-		// Verify application Prometheus scraped metrics
+
+		kubeconfigPath, err := k8sutil.GetKubeConfigLocation()
```
kminder, line 281: Seems like this should be done int he before suite that that it doesn't generate a bunch of meaningless errors if the kubeconfig can't be found.

tests/e2e/examples/bobsbooks/bobs_books_test.go
```diff
@@ -261,63 +277,21 @@ var _ = t.Describe("Bobs Books test", Label("f:app-lcm.oam",
...
+		if err != nil {
+			Expect(err).To(BeNil(), fmt.Sprintf("Failed to get default kubeconfig path: %s", err.Error()))
+		}
+		ok, _ := pkg.IsVerrazzanoMinVersion("1.4.0", kubeconfigPath)
```
kminder, line 285: Please use a better variable name than 'ok'.
Should this also be done in the before suite?
The error should not be ignored.

kminder (approved):
Approved.

IshaGirdhar:
> 

@kminder Lost approval because of master rebasing. Could you approve it again?

kminder (approved):
Approved.

PR #5156, opened 2023-01-18
Topic argo security

preethignanesh:
Added the security standards for Argo CD to maintain in par with the Verrazzano security standards.


tests/e2e/verify-install/security/pod_security_test.go
```diff
@@ -111,6 +113,10 @@ var _ = t.AfterEach(func() {})
 
 var _ = t.Describe("Ensure pod security", Label("f:security.podsecurity"), func() {
 	testFunc := func(ns string) {
+		if ns == "argocd" && !isArgo {
```
michael-cico, line 116: This shouldn't be needed, if the pods aren't present they won't be checked.

platform-operator/thirdparty/charts/argo-cd/values.yaml
```diff
@@ -44,7 +44,7 @@ crds:
 global:
   # -- Common labels for the all resources
   additionalLabels: {}
-    # app: argo-cd
+  # app: argo-cd
```
michael-cico, line 47: now that you have the overrides file in place in `helm_config`, you should be able to revert any changes to this file... looks like a lot of formatting changes.

platform-operator/helm_config/overrides/argocd-values.yaml
```diff
@@ -1,10 +1,113 @@
...
+    allowPrivilegeEscalation: false
+    runAsGroup: 1001
+    runAsUser: 1001
+    seccompProfile:
+      type: RuntimeDefault
+    capabilities:
+      drop:
+        - ALL
+      add:
+        - NET_BIND_SERVICE
```
pfmackin, line 23: Is NET_BIND_SERVICE really needed for all these containers?

preethignanesh: We thought to add it as a standard for all the containers. Do you want this to be removed ?

platform-operator/thirdparty/charts/argo-cd/values.yaml
preethignanesh: I felt the formatting was wrong so I left the changes instead of reverting

tests/e2e/verify-install/security/pod_security_test.go
preethignanesh: removed

pfmackin (approved):

michael-cico (approved):

PR #5163, opened 2023-01-19
VZ-8116: Security context for pods and containers in istio-system namespace

pshirodkar-orcl:
VZ-8116: Security context for pods and containers in istio-system namespace.

### DIFF
--- a/platform-operator/helm_config/overrides/istio-cr.yaml
+++ b/platform-operator/helm_config/overrides/istio-cr.yaml
@@ -9,6 +9,38 @@ spec:
     egressGateways:
       - name: istio-egressgateway
         enabled: true
+        securityContext:
+          runAsUser: 1337
+          runAsGroup: 1337
+          runAsNonRoot: true
+          privileged: false
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+              - ALL
+    ingressGateways:
+      - securityContext:
+          runAsUser: 1337
+          runAsGroup: 1337
+          runAsNonRoot: true
+          privileged: false
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+              - ALL
+    pilot:
+      - securityContext:
+          runAsUser: 1337
+          runAsGroup: 1337
+          runAsNonRoot: true
+          privileged: false
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+              - ALL
 
   # Global values passed through to helm global.yaml.
   # Please keep this in sync with manifests/charts/global.yaml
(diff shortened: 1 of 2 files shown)


platform-operator/helm_config/overrides/istio-cr.yaml
```diff
@@ -9,6 +9,72 @@ spec:
...
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+          overlays:
+            - kind: Deployment
+              name: istio-egressgateway
+              patches:
+                - path: spec.template.spec.containers.[name:istio-proxy].securityContext
+                  value: |
+                    runAsUser: 1337
```
pfmackin, line 25: you don't need the runAsUser, runAsGroup, runAsNonRoot here since you have it in the pod security context.

platform-operator/helm_config/overrides/istio-cr.yaml
```diff
@@ -9,6 +9,72 @@ spec:
...
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+          overlays:
+            - kind: Deployment
+              name: istio-ingressgateway
+              patches:
+                - path: spec.template.spec.containers.[name:istio-proxy].securityContext
+                  value: |
+                    runAsUser: 1337
```
pfmackin, line 48: you don't need the runAsUser, runAsGroup, runAsNonRoot here since you have it in the pod security context.

platform-operator/helm_config/overrides/istio-cr.yaml
```diff
@@ -9,6 +9,72 @@ spec:
...
+          runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
+        overlays:
+          - kind: Deployment
+            name: istiod
+            patches:
+              - path: spec.template.spec.containers.[name:discovery].securityContext
+                value: |
+                  runAsUser: 1337
```
pfmackin, line 70: you don't need the runAsUser, runAsGroup, runAsNonRoot here since you have it in the pod security context.

platform-operator/helm_config/overrides/istio-cr.yaml
pshirodkar-orcl: In all other tickets related to container security context, we have explicitly these 3 attributes even if the pod security context mentioned the same.

pfmackin: it is not needed.  Normally you don't see it in both places.  It doesn't hurt but it is redundant, better to remove it from the pod securitycontext or the container security context.

pshirodkar-orcl: This likely will be the only place in the entire repo where things might look inconsistent from what others have implemented.

pfmackin (approved):

platform-operator/helm_config/overrides/istio-cr.yaml
pshirodkar-orcl: I have removed the three redundant RunAs settings at the container level

pfmackin (approved):

PR #5174, opened 2023-01-20
VZ-8250,VZ-8251: Update MySQL images

fredtibbitts:
Update the mysql-operator, mysql-server, and mysql-router images in the Verrazzano BOM.

### DIFF
--- a/platform-operator/thirdparty/charts/mysql/values.yaml
+++ b/platform-operator/thirdparty/charts/mysql/values.yaml
@@ -1,34 +1,74 @@
 image:
   pullPolicy: IfNotPresent
-  # The repository setting indicates where the images will be pulled
-  #repository:
   pullSecrets:
     enabled: false
     secretName:
 
-mysqlOperator:
-  image:
 
 credentials:
   root:
     user: root
 #    password: sakila
     host: "%"
-#  user:
-#    name:
-#    password:
 
 tls:
   useSelfSigned: false
 #  caSecretName:
 #  serverCertAndPKsecretName:
-#  routerCertAndPKsecretName:
+#  routerCertAndPKsecretName: # our use router.certAndPKsecretName
 
-#serverVersion: 8.0.30
+#serverVersion: 8.0.31
 serverInstances: 3
-routerInstances: 1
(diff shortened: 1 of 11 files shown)


platform-operator/thirdparty/charts/mysql/values.yaml
```diff
@@ -104,11 +145,3 @@ baseServerId: 1000
 #          prefix : /
 #          bucketName: idbcluster_backup
 #          credentials: oci-credentials
-
```
jmaron99, line 107: I guess it's ok to remove the initdbScripts value since it's set in the actual code.  Just to be safe - have you tested the 1.3 (I believe it pre-dates mysql operator) to this version upgrade path?

pfmackin (dismissed):

jmaron99:
Generally it does appear that these changes are additive, adding the newer settings to the existing charts

mgianatagh (dismissed):

stevshap (approved):

platform-operator/thirdparty/charts/mysql/values.yaml
fredtibbitts: These removed lines were commented out anyway, except for the `configurationFilesPath` setting. I have no idea what impact that may have, but it's not in the latest version of the chart. I'm running full triggered tests which does upgrade path testing, but I don't know if it specifically tests upgrading from 1.3.

jmaron99: IIRC, the configurationFilesPath plays a role in our backup processing - it adds a mount path that contains some hook scripts.  Still required @theabmitra ?

fredtibbitts: I think you're right, it does look like it's needed, but I don't understand why this is in the chart values. This should be in the overrides file and not a chart customization. I'm going to add it to the overrides. @theabmitra please let me know if there's a reason that this setting can't go in the overrides.

mgianatagh (approved):

stevshap (approved):

PR #5203, opened 2023-01-24
VZ- 8194 add ServiceTemplate to Helidon workload

mbenmclean:
This change adds a ServiceTemplate to the Helidon Workload. This allows users to customize the Helidon workload service to do things like add labels or customize port names or numbers. This change also adds an e2e test to ensure that the custom configuration of service ports and target ports are working. I had to update the app operator boilerplate.txt and that made all the generated files update their copyright as well.


application-operator/controllers/helidonworkload/helidonworkload_controller.go
```diff
@@ -273,44 +274,54 @@ func (r *Reconciler) createServiceFromDeployment(workload *vzapi.VerrazzanoHelid
...
+		if s.Spec.Type == "" {
+			s.Spec.Type = corev1.ServiceTypeClusterIP
+		}
+		if s.Spec.Ports == nil {
+			for _, container := range deploy.Spec.Template.Spec.Containers {
+				if len(container.Ports) > 0 {
+					for _, port := range container.Ports {
+						// All ports within a ServiceSpec must have unique names.
+						// When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort.
+						name := strings.ToLower(container.Name + "-" + strconv.FormatInt(int64(port.ContainerPort), 10))
```
arjav-desai, line 306: Wasn't there something about Istio using protocol from name to do routing. Just wanted to make sure, since we are removing TCP from name as it was done prior.

arjav-desai (dismissed):
LGTM.

application-operator/controllers/helidonworkload/helidonworkload_controller.go
mbenmclean: If you don't specify it Istio will detect it if it's TCP or HTTP. We discussed this change when we met.

arjav-desai: Okay.

arjav-desai (dismissed):

bryantidd (dismissed):

application-operator/controllers/helidonworkload/helidonworkload_controller.go
```diff
@@ -273,44 +274,54 @@ func (r *Reconciler) createServiceFromDeployment(workload *vzapi.VerrazzanoHelid
 				Kind:       serviceKind,
 				APIVersion: serviceAPIVersion,
 			},
-			ObjectMeta: metav1.ObjectMeta{
```
pfmackin, line 276: Line 271 should return if no contains so we can get rid of this huge indented block of code (basically the whole function)

application-operator/controllers/helidonworkload/helidonworkload_controller.go
```diff
@@ -273,44 +274,54 @@ func (r *Reconciler) createServiceFromDeployment(workload *vzapi.VerrazzanoHelid
...
+			s.Spec.Type = corev1.ServiceTypeClusterIP
+		}
+		if s.Spec.Ports == nil {
+			for _, container := range deploy.Spec.Template.Spec.Containers {
+				if len(container.Ports) > 0 {
+					for _, port := range container.Ports {
+						// All ports within a ServiceSpec must have unique names.
+						// When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort.
+						name := strings.ToLower(container.Name + "-" + strconv.FormatInt(int64(port.ContainerPort), 10))
+						protocol := corev1.ProtocolTCP
```
pfmackin, line 307: I am confused by this.  The container protocol is different from the service protocol field.  I think we are supposed to leave the service protocol empty and let istio automatically determine the protocol per https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

application-operator/controllers/helidonworkload/helidonworkload_controller.go
mbenmclean: fixed

application-operator/controllers/helidonworkload/helidonworkload_controller.go
mbenmclean: It looks liike it defaults to TCP. I can remove it.

mbenmclean: Well so the prefix is equivilent to the appProtocol field. This protocol field is the IP protocol field and is different and I believe it does relate to the container.
```
	// The IP protocol for this port. Supports "TCP", "UDP", and "SCTP".
	// Default is TCP.
	// +default="TCP"
	// +optional
	Protocol Protocol `json:"protocol,omitempty" protobuf:"bytes,2,opt,name=protocol,casttype=Protocol"`
```

pfmackin (approved):

pfmackin (approved):

PR #5205, opened 2023-01-24
Fix bug in grafana db end-to-end test

fredtibbitts:
This test had a single `Eventually` block that deleted the Grafana pod and immediately checked to see if the pod had restarted, and if it did not, the `Eventually` would rerun and the pod would be deleted again.

This fix separates the pod deletion from checking for the restarted pod so that the test does not continually delete the pod.


stevshap (approved):
LGTM

PR #5210, opened 2023-01-24
VZ-7556: Fix bug to allow users to change the VZ CR during install

marcogelle:
This addresses a bug in Verrazzano v1.3 that prevents users from making changes to the Verrazzano CR during install.

This bug was due to missing Helm labels and annotations in the Rancher component's `bootstrap-secret` Secret. This PR fixes this by ensuring that `bootstrap-secret` is deleted in the Rancher preinstall, so that its recreation during Rancher install goes smoothly.

This PR also fixes a slightly related bug where Helm CLI upgrade command was not being retried upon failure.

This PR also adds to the unit tests relevant to my changes.

### DIFF
--- a/pkg/helm/helmcli.go
+++ b/pkg/helm/helmcli.go
@@ -22,6 +22,9 @@ var Debug bool
 // cmdRunner needed for unit tests
 var runner vzos.CmdRunner = vzos.DefaultRunner{}
 
+// number of times runHelm will retry running its Helm command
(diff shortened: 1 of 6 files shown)


michael-cico (approved):

pfmackin (approved):

theabmitra (approved):
looks good. I like the max retry in helm component .

PR #5217, opened 2023-01-24
Update the pilot and proxy base images

davisgomes:

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -179,7 +179,7 @@
           "images": [
             {
               "image": "proxyv2",
-              "tag": "1.15.3-20221221225651-900a5c7b",
+              "tag": "1.15.3-20230124194836-900a5c7b",
               "helmFullImageKey": "monitoringOperator.istioProxyImage"
             },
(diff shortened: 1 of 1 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -106,12 +106,12 @@
...
             {
               "image": "pilot",
-              "tag": "1.15.3-20221221225651-900a5c7b",
+              "tag": "1.15.3-20230124194836-900a5c7b",
               "helmFullImageKey": "values.pilot.image"
             },
             {
               "image": "proxyv2",
-              "tag": "1.15.3-20221221225651-900a5c7b",
+              "tag": "1.15.3-20230124194836-900a5c7b",
```
fredtibbitts, line 114: There are two more references to proxyv2 in the BOM that also need to be updated.

pfmackin (approved):

platform-operator/verrazzano-bom.json
davisgomes: fixed

fredtibbitts (approved):

PR #5223, opened 2023-01-25
VZ-8025: PR to combine WebLogic Ingress trait and Logging Trait tests. 

Abhijithang:
We have two separate tests for Console Ingress trait and Logging Trait tests which deploys the same ToDo list application. Also, we run tests of these apps in parallel, leading to instability and issues that are harder to debug. Therefore, this PR combines these apps and removes redundant stages from the Jenkins pipelines.

### DIFF
--- a/tests/e2e/ingress/console/console_ingress_test.go
+++ b/tests/e2e/ingress/console/console_ingress_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2023, Oracle and/or its affiliates.
+// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package ingress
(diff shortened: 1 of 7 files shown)


tests/e2e/ingress/console/console_ingress_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
bobdonat, line 1: Can this source file be removed now?

tests/e2e/loggingtrait/weblogicworkload/weblogic_loggingtrait_test.go
```diff
@@ -1,10 +1,12 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
bobdonat, line 1: should the name of the weblogic_logging_* files  be renamed since the ingress trait is also part of the test now.

tests/e2e/ingress/console/console_ingress_test.go
Abhijithang: I tried removing it but the Verrazzano main build fails with a copyright error. Not sure why.

davisgomes (approved):

bobdonat (approved):

PR #5231, opened 2023-01-25
VZ-8357 Uptake of Rancher 2.7.1

mgianatagh:
Uptake Rancher 2.7.1


platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -515,3 +519,79 @@ func isRancherNamespace(ns *corev1.Namespace) bool {
 	}
 	return false
 }
+
+// deleteRancherFinalizers - delete Rancher finalizers on resources that the cleanup job
```
michael-cico, line 523: So is this largely to clean up Rancher finalizers on non-Rancher objects?  Or just both generally?

mgianatagh: Rancher puts finalizers on every Role and RoleBinding in the cluster.  The cleanup job is not finding the ones in namespaces outside of the Rancher set of namespaces. This is probably due to use trimming down the rancher-cleanup script to only process rancher namespaces.

michael-cico (approved):

PR #5232, opened 2023-01-25
VZ-8311 authorization policy empty path

mbenmclean:
This change fixes the ingresstrait controller so that when a path of "/" is passed in for the authorization policy the name does not have a trailing -

### DIFF
--- a/application-operator/controllers/ingresstrait/ingresstrait_controller.go
+++ b/application-operator/controllers/ingresstrait/ingresstrait_controller.go
@@ -780,7 +780,10 @@ func (r *Reconciler) createOrUpdateAuthorizationPolicies(ctx context.Context, ru
 	for _, path := range rule.Paths {
 		if path.Policy != nil {
 			pathSuffix := strings.Replace(path.Path, "/", "", -1)
-			policyName := fmt.Sprintf("%s-%s", namePrefix, pathSuffix)
+			policyName := namePrefix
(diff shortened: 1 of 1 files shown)


application-operator/controllers/ingresstrait/ingresstrait_controller.go
```diff
@@ -780,7 +780,10 @@ func (r *Reconciler) createOrUpdateAuthorizationPolicies(ctx context.Context, ru
 	for _, path := range rule.Paths {
 		if path.Policy != nil {
 			pathSuffix := strings.Replace(path.Path, "/", "", -1)
-			policyName := fmt.Sprintf("%s-%s", namePrefix, pathSuffix)
+			policyName := namePrefix
+			if pathSuffix != "" {
```
anders-swanson, line 784: Can this ever result in duplicate authorization policy names?

anders-swanson (approved):

application-operator/controllers/ingresstrait/ingresstrait_controller.go
mbenmclean: Well the name includes the ingresstrait name so I think only if you created two ingresstraits with the same name. That's not a new problem though if you previously did that with the same path you'd get conflicting authorization policy names. Perhaps the authorization policy should include the namespace to make sure it's unique.

PR #4401, opened 2022-10-14
VZ-7341: Uptake Kiali 1.57.1

pshirodkar-orcl:
Uptake Kiali 1.57.1.

PR #5235, opened 2023-01-26
VZ-8371: Kiali showing error connecting to Prometheus on the Graph screen

pshirodkar-orcl:
Restored the external_services section in kiali-server overrides.

### DIFF
--- a/platform-operator/helm_config/overrides/kiali-server-values.yaml
+++ b/platform-operator/helm_config/overrides/kiali-server-values.yaml
@@ -42,6 +42,14 @@ kubernetes_config:
   - "DeploymentConfig"
   - "ReplicationController"
 
+external_services:
+  prometheus:
+    url: http://prometheus-operator-kube-p-prometheus.verrazzano-monitoring.svc.cluster.local:9090
+  grafana:
+    enabled: false
+  tracing:
+    enabled: false
+
 server:
   web_root: ""
 


platform-operator/helm_config/overrides/kiali-server-values.yaml
```diff
@@ -42,6 +42,14 @@ kubernetes_config:
   - "DeploymentConfig"
   - "ReplicationController"
 
+external_services:
```
pshirodkar-orcl, line 45: I had incorrectly removed the `external_services` section previously during uptake of Kiali 1.57.1.
refer https://github.com/verrazzano/verrazzano/pull/4401/files#diff-40daaca389f1c4f8d4877b7a5a0717fed6090918e09429098f60206f2008af0d

mgianatagh (approved):

platform-operator/helm_config/overrides/kiali-server-values.yaml
```diff
@@ -42,6 +42,14 @@ kubernetes_config:
...
+  prometheus:
+    url: http://prometheus-operator-kube-p-prometheus.verrazzano-monitoring.svc.cluster.local:9090
```
anders-swanson, line 47: Prometheus may not necessarily have this name, if the user customizes their prometheus via the vz cr they can change the name.

anders-swanson (approved):

PR #5243, opened 2023-01-26
Fix PSR connection timing issue

pfmackin:
When running PSR worker in the mesh, there is a timing problem where connection can fail if the worker starts before the istio sidecar proxy. Do some connection retries to fix this issue. This fixes the OpenSearch scale and WebLogic scale problems we were seeing.


pfmackin:
Obsolete

PR #5251, opened 2023-01-27
VZ-7337: Recover from terminal failure of data load job

saketmht:
When the data load job fails during preUpgrade, the VPO is stuck reconciling forever without starting job again. 

Requirements:

- Ensure the VPO at least logs an error when the data load job terminally fails.
- Change VPO to retry the data log job after it has terminally failed.
- Ensure the VZ analysis CLI can detect when the data load job terminally fails.
- Update public documentation with instructions for manually recovering from a terminally failed data load job

Fixes:

- Requeue data load job when it fails during pre-upgrade
- Check if user created pod for data migration during pre-upgrade if VPO created pod fails to come up
- Check mysql-cluster-secret for root password when requeuing after pre-upgrade (i.e. after upgrade since mysql secret containing root password is deleted)
- Check db-migration secret for "manuallyDbMigrated" stage to notify VPO in-case user manually does the data migration
- vz analyze able to detect if the load job fails once or more times (add KeycloakDataMigrationFailure as a known issue)
- Update docs with manual steps for the user (https://github.com/verrazzano/docs/pull/831)

---------------------------------
Sample vz analyze output:

    Analyzing resources from the cluster ...

    Detected 1 issues in the cluster:

	ISSUE (KeycloakDataMigrationFailure)
		summary: Failure(s) migrating Keycloak data during mysql upgrade
		actions:
			action: Consult https://verrazzano.io/v1.5/docs/troubleshooting/diagnostictools/analysisadvice/keycloakdatamigrationfailure using supporting details identified in the report
		supportingData:
			messages:
				Failed pre-upgrade for component mysql: DB load job has failed and will be retried
			related resource(s):
				Verrazzano custom resource
				Log from pod "verrazzano-platform-operator-5874d7995d-kzhfk" in namespace "verrazzano-install"


platform-operator/controllers/verrazzano/component/mysql/legacydb.go
```diff
@@ -307,6 +308,14 @@ func isJobPodRunning(ctx spi.ComponentContext, jobName string) wait.ConditionFun
 			ctx.Log().Infof("DB load job pod is running or has completed successfully")
 			return true, nil
 		case v1.PodFailed:
+			// Check if the user manually created a pod to do the DB migration
+			if len(podList.Items) > 1 {
```
jmaron99, line 312: So the listing in this case returns more than 1 - one for the failed VPO job, one for the user manually stated job?

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
```diff
@@ -566,6 +601,21 @@ func checkDbMigrationJobCompletion(ctx spi.ComponentContext) bool {
 	}
 	for _, container := range dbMigrationPod.Status.ContainerStatuses {
 		if container.Name == dbLoadContainerName && container.State.Terminated != nil && container.State.Terminated.ExitCode == 0 {
```
jmaron99, line 603: why do we care about the istio container?  i.e. why not just query directly for the state of the migration container as it is really the only container we care about?

tools/vz/pkg/analysis/test/cluster/keycloak-data-migration-failure/cluster-snapshot/verrazzano-install/verrazzano-platform-operator-5874d7995d-kzhfk/logs.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
```
jmaron99, line 1: I don't think this need to be merged?

jmaron99: Never mind - looks like it's a test file

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
saketmht: Incase the the pod from VPO job has failed, it checks if any other pod exists with label: {job-name: load-dump}. If it does then the user has created it because normally there should be only one pod with this label created by the VPO job

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
saketmht: The loop basically goes through all the containers of the pod and if container.Name == migration-container then it checks its status. 
I haven't really touched the loop code. Just added a log statement if migration container was successfully terminated

tools/vz/pkg/analysis/test/cluster/keycloak-data-migration-failure/cluster-snapshot/verrazzano-install/verrazzano-platform-operator-5874d7995d-kzhfk/logs.txt
saketmht: Yes its a test file for the vz analyze command.

jmaron99:
I assume you went through a functional, possibly manual test to ensure this works?

platform-operator/controllers/verrazzano/component/mysql/legacydb.go
```diff
@@ -540,13 +567,21 @@ func cleanupDbMigrationJob(ctx spi.ComponentContext) error {
 func checkDbMigrationJobCompletion(ctx spi.ComponentContext) bool {
 	// check for existence of db restoration job.  If it exists, wait for its completion
 	ctx.Log().Progress("Checking status of keycloak DB restoration")
+
+	// Check if the Db Migration was done manually in case of terminal job failure
```
jmaron99, line 571: Just to be clear:  this check is at the head of this method to first check to see if a manual migration was undertaken and succeeded prior to checking the VPO migration status. Correct?

saketmht: Yes. Since this function would requeue the job if it has failed, it is better to first check if a manual migration was done or not

saketmht:
> I assume you went through a functional, possibly manual test to ensure this works?

Yes. I hung the load job during the pre upgrade and then did all the steps manually as mentioned in the docs PR and the upgrade was successful.

jmaron99 (dismissed):
LGTM

jmaron99:
One thing - any way to add a unit test for the manual migration db stage?

saketmht:
> One thing - any way to add a unit test for the manual migration db stage?

I don't think so. The VPO only checks the db-migration secret for a value added by the user to indicate that the migration was done manually. It doesn't really verify if the intermediate steps were actually done by the user or not. Not sure how that would translate into a unit test. 

- Have a fakeClient built with failed job and db-migration secret
- Initially the checkDbMigrationJobCompletion returns false
- Update the db-migration secret with the "db-migrated" value to true
- Call the checkDbMigrationJobCompletion function again and check if it returns true

This wouldn't be much of a "test" though.

agv426 (approved):

tools/vz/pkg/analysis/internal/util/report/issue.go
```diff
@@ -124,26 +126,27 @@ const (
...
+	IstioIngressPrivateSubnet:    {Type: IstioIngressPrivateSubnet, Summary: "Failed to create LoadBalancer for Istio Ingress Gateway", Informational: false, Impact: 10, Confidence: 10, Actions: []Action{KnownActions[IstioIngressPrivateSubnet]}},
+	NginxIngressPrivateSubnet:    {Type: NginxIngressPrivateSubnet, Summary: "Failed to create LoadBalancer for Nginx Ingress Controller", Informational: false, Impact: 10, Confidence: 10, Actions: []Action{KnownActions[NginxIngressPrivateSubnet]}},
+	ExternalDNSConfigureIssue:    {Type: ExternalDNSConfigureIssue, Summary: "Failed to setup DNS configuration", Informational: false, Impact: 10, Confidence: 10, Actions: []Action{KnownActions[ExternalDNSConfigureIssue]}},
+	KeycloakDataMigrationFailure: {Type: KeycloakDataMigrationFailure, Summary: "Failure(s) migrating Keycloak data during mysql upgrade", Informational: true, Impact: 10, Confidence: 10, Actions: []Action{KnownActions[KeycloakDataMigrationFailure]}},
```
bhatpmk, line 149: Use MySQL instead of mysql

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -623,6 +625,20 @@ func createMySQLInitFile(ctx spi.ComponentContext, userPwd []byte) (string, erro
 	return file.Name(), nil
 }
 
+// getRootDBPassword returns the password to access the DB as root
+func getRootDBPassword(compContext spi.ComponentContext) ([]byte, error) {
```
bhatpmk, line 629: The function grantXARecoverAdmin has the same code, can you please update it to call getRootDBPassword ?

bhatpmk (approved):
LGTM

PR #5252, opened 2023-01-27
VZ-8346: Change the default number of shards per index to 1

IshaGirdhar:
This is to change the default number of shards per index in OpenSearch from 5 to 1.

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
@@ -1003,7 +1003,7 @@ data:
         "settings":{
           "index.refresh_interval":"5s",
           "index.mapping.total_fields.limit":"2000",
-          "number_of_shards":5,
+          "number_of_shards":1,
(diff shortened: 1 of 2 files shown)


gaurav-05 (approved):

platform-operator/helm_config/charts/verrazzano-fluentd/templates/fluentd-config-configmap.yaml
```diff
@@ -1003,7 +1003,7 @@ data:
         "settings":{
           "index.refresh_interval":"5s",
           "index.mapping.total_fields.limit":"2000",
-          "number_of_shards":5,
+          "number_of_shards":1,
```
kminder, line 1006: What are the upgrade implications of this?
Will existing installs be changed from 5 to 1 during upgrade?

IshaGirdhar: So upgrade it won't work, though the CM will be updated but fluentd plugin creates the index template(with new changes) only if it doesn't exist, and in the upgrade case, it's already there with 5 shards, so it won't update the existing one.

PR #5262, opened 2023-01-30
Topic argo keycloak timeout

preethignanesh:
For Argo CD, we are increasing the access token lifespan to avoid frequent logout when logged in via keycloak


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -989,6 +989,14 @@ func configureKeycloakRealms(ctx spi.ComponentContext) error {
 		if err != nil {
 			return err
 		}
+
+		//Setting the Access Token Lifespan value to 20mins.
```
anders-swanson, line 993: What's the significance of 20 mins?

preethignanesh: Earlier the token lifespan was set to 5 mins, which implies that in 5 mins of login, UI would logout the user. So, after discussion with Kevin, we have decided to raise it to 20 mins and parallelly following up with Argo CD to use the refresh tokens after expiry.

pfmackin (approved):

PR #5269, opened 2023-01-31
VZ- 8353 LRE pipeline using Rancher Proxy

mbenmclean:
This change updates the LRE pipeline to access the cluster through the Rancher Proxy rather than through the bastion

### DIFF
--- a/ci/lre/Jenkinsfile
+++ b/ci/lre/Jenkinsfile
@@ -151,7 +151,12 @@ pipeline {
             }
         }
 
-       stage('Upgrade to stable commit') {
+        stage('Upgrade to stable commit') {
+            when {
+                expression {
+                    return params.UPGRADE_VZ
+                }
+            }
             steps {
                 script {
                     // Get the last stable commit ID that passed the triggered tests
--- a/ci/lre/Jenkinsfile
+++ b/ci/lre/Jenkinsfile
@@ -48,13 +48,6 @@ pipeline {
         OCI_CLI_KEY_FILE = credentials('oci-dev-api-key-file')
         OCI_CLI_REGION = "us-ashburn-1"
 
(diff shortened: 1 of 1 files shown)


ci/lre/Jenkinsfile
```diff
@@ -190,12 +130,28 @@ pipeline {
             }
         }
 
-        stage('Upgrade to stable commit') {
-            when {
```
bobdonat, line 194: Why was this when section removed?

ci/lre/Jenkinsfile
```diff
@@ -190,12 +130,28 @@ pipeline {
...
-                expression {
-                    return params.UPGRADE_VZ
+        stage("Create OCI config file") {
```
davisgomes, line 133: Why do we still need to set up the OCI config file?

mbenmclean: To get the latest stable commit hash

ci/lre/Jenkinsfile
mbenmclean: It wasn't it's still there further down. Think it's just a little confusing with how github is rendering the diff

bobdonat: I see this:

       stage('Upgrade to stable commit') {
            steps {
                script {
 with the "when" missing

bobdonat: If you view the full file you will see it is missing

mbenmclean: Fixed

ci/lre/Jenkinsfile
bobdonat: I think it is only needed when doing upgrade. The "when" could be added here.

mbenmclean: fixed

davisgomes (approved):

PR #5278, opened 2023-02-01
Fix Rancher verrazzano user GRB test

mbenmclean:
Fixes a bug in the Racher verrazzano user GlobalRoleBinding test to make sure that the right user is selected.

### DIFF
--- a/tests/e2e/verify-infra/restapi/rancher_url_test.go
+++ b/tests/e2e/verify-infra/restapi/rancher_url_test.go
@@ -137,15 +137,18 @@ var _ = t.Describe("rancher", Label("f:infra-lcm",
 							return false, err
 						}
 						for _, userData := range userList.Items {
-							userPrincipals, userPrincipalsOK := userData.UnstructuredContent()[rancher.UserAttributePrincipalIDs].([]interface{})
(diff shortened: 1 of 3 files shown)


davisgomes (dismissed):

davisgomes (dismissed):

tests/e2e/verify-infra/restapi/rancher_url_test.go
```diff
@@ -137,14 +137,15 @@ var _ = t.Describe("rancher", Label("f:infra-lcm",
 							return false, err
 						}
 						for _, userData := range userList.Items {
-							userPrincipals, ok := userData.UnstructuredContent()[rancher.UserAttributePrincipalIDs].([]interface{})
-							if ok {
+							userPrincipals, userPrincipalsOK := userData.UnstructuredContent()[rancher.UserAttributePrincipalIDs].([]interface{})
```
bobdonat, line 140: I would check for the username first and if t does not match then "continue" in the for loop and then the rest of the code can remain as is. You could also use userData.GetName() == "Verrazzano Admin" instead of checking for username.

mbenmclean: fixed

bobdonat (approved):

PR #5315, opened 2023-02-06
VZ-8648: Allow component disable

michael-cico:
Allow components to be disabled in the component state machine, if the validators allow it. 

The existing install/update state machine has been modified to allow disabling (uninstalling) a component, delegating the logic and uninstall-state management to the existing `uninstallSingleComponent` method.
 
The following changes have also been made
- Unify the internal component in-memory tracker states to the same type to allow re-using the uninstall-component code
- rename the `installXXXComponent` funcs to `reconcileXXXComponent`

### DIFF
--- a/platform-operator/controllers/verrazzano/reconcile/install_component.go
+++ b/platform-operator/controllers/verrazzano/reconcile/install_component.go
@@ -196,7 +196,7 @@ func (r *Reconciler) reconcileSingleComponent(spiCtx spi.ComponentContext, compS
 
 			compStateContext.state = compStateInstallEnd
 
-		case compStateUninstallStart, compStatePreUninstall, compStateUninstall, compStateWaitUninstalled, compStateUninstalleDone:
+		case compStateUninstallStart, compStatePreUninstall, compStateUninstall, compStateWaitUninstalled, compStateUninstallDone:
 			// Delegates the component uninstall work to
 			result, err := r.uninstallSingleComponent(compContext, compStateContext, comp)
 			if err != nil || result.Requeue {
--- a/platform-operator/controllers/verrazzano/reconcile/install_component.go
+++ b/platform-operator/controllers/verrazzano/reconcile/install_component.go
@@ -114,14 +114,13 @@ func (r *Reconciler) installSingleComponent(spiCtx spi.ComponentContext, compSta
 				compStateContext.state = compStateInstallEnd
 				continue
 			}
-			isInstalled, err := comp.IsInstalled(compContext)
-			if err != nil {
-				compLog.ErrorfThrottled("Error checking component installed state: %v", err)
-				return ctrl.Result{Requeue: true}
-			}
-			if isInstalled && !comp.IsEnabled(compContext.EffectiveCR()) {
-				// start uninstall of a single component
-				compStateContext.state = compStateUninstallStart
+			if !comp.IsEnabled(compContext.EffectiveCR()) {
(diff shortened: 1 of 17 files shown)


platform-operator/controllers/verrazzano/reconcile/install_component.go
```diff
@@ -177,15 +186,22 @@ func (r *Reconciler) installSingleComponent(spiCtx spi.ComponentContext, install
...
 		case compStateInstallComplete:
 			if err := r.updateComponentStatus(compContext, "Install complete", vzapi.CondInstallComplete); err != nil {
 				compLog.ErrorfThrottled("Error writing component Ready state to the status: %v", err)
 				return ctrl.Result{Requeue: true}
 			}
 
-			installContext.state = compStateInstallEnd
+			compStateContext.state = compStateInstallEnd
+
+		case compStateUninstallStart, compStatePreUninstall, compStateUninstall, compStateWaitUninstalled, compStateUninstalleDone:
```
pfmackin, line 199: Maybe these states should be removed, just keeping uninstallDone
Also typo on Uninstalle

mbenmclean: fixed typo, other thing has been discussed.

platform-operator/controllers/verrazzano/reconcile/install_component.go
```diff
@@ -77,44 +77,53 @@ func (r *Reconciler) installComponents(spiCtx spi.ComponentContext, tracker *ins
...
 
 		case compStateInstallInitReady:
 			if skipComponentFromReadyState(compContext, comp, componentStatus) {
-				installContext.state = compStateInstallEnd
+				compStateContext.state = compStateInstallEnd
 				continue
 			}
-			installContext.state = compStateWriteInstallStartedStatus
+			if !comp.IsEnabled(compContext.EffectiveCR()) {
+				if isCurrentlyInstalled(compContext, comp) {
```
pfmackin, line 118: If component not enabled then the code should immediately transition into some uninstall state where you can do the check is currentlyInstalled.

platform-operator/controllers/verrazzano/reconcile/install_component.go
```diff
@@ -177,15 +186,22 @@ func (r *Reconciler) installSingleComponent(spiCtx spi.ComponentContext, install
...
 				compLog.ErrorfThrottled("Error writing component Ready state to the status: %v", err)
 				return ctrl.Result{Requeue: true}
 			}
 
-			installContext.state = compStateInstallEnd
+			compStateContext.state = compStateInstallEnd
+
+		case compStateUninstallStart, compStatePreUninstall, compStateUninstall, compStateWaitUninstalled, compStateUninstallDone:
+			// Delegates the component uninstall work to
+			result, err := r.uninstallSingleComponent(compContext, compStateContext, comp)
```
pfmackin, line 201: So we have a install state machine context that we are passing to uninstall which uses a different state machine.  This seems wrong.

platform-operator/controllers/verrazzano/reconcile/install_component.go
mbenmclean: I'm not entirely sure what you mean.

platform-operator/controllers/verrazzano/reconcile/install_component.go
mbenmclean: No I believe part of the change is to combine all the trackers to use one componentState type

PR #5330, opened 2023-02-08
VZ-8243: E2E testing for changing VZ CR during install

marcogelle:
This PR creates a new testing pipeline `verrazzano-dynamic-config-install-tests`, specified by the Jenkinsfile `JenkinsfileUpdateDuringInstall`. This new pipeline is added to the `verrazzano-dynamic-config-suite` pipeline.

The new pipeline tests the general scenario that was fixed with https://github.com/verrazzano/verrazzano/pull/5259.

`JenkinsfileUpdateDuringInstall` runs two install tests. Each install test starts with a typo in the `WildcardDNS` domain, then changes the typo mid-installation. The tests differ only in the timing of the VZ CR change. I decided the timing of the VZ CR change to account for the error scenarios from https://github.com/verrazzano/verrazzano/pull/5259 and https://github.com/verrazzano/verrazzano/pull/5210.


ci/dynamic-updates/JenkinsfileUpdateDuringInstall
```diff
@@ -0,0 +1,578 @@
...
+                            }
+                        }
+                        failure {
+                            archiveArtifacts artifacts: "**/kind-logs/**", allowEmptyArchive: true
+                        }
+                        always {
+                            archiveArtifacts artifacts: "acceptance-test-operator.yaml,downloaded-operator.yaml", allowEmptyArchive: true
+                            // enable debug logging of Verrazzano api istio proxy
+                            script {
+                                if (params.ENABLE_API_ENVOY_LOGGING) {
```
michael-cico, line 228: do we need this block?  Had a similar comment below.

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
```diff
@@ -0,0 +1,578 @@
...
+                        echo "Downloading VZ CLI from object storage"
+                        oci --region us-phoenix-1 os object get --namespace ${OCI_OS_NAMESPACE} -bn ${OCI_OS_COMMIT_BUCKET} --name ephemeral/${env.BRANCH_NAME}/${SHORT_COMMIT_HASH}/${VZ_CLI_TARGZ} --file ${VZ_CLI_TARGZ}
+                        tar xzf ${VZ_CLI_TARGZ} -C ${GO_REPO_PATH}
+                        ${VZ_COMMAND} version
+                    """
+                }
+            }
+        }
+
+        stage('Kickoff Parallel VZ Install and VZ CR update during Keycloak Install') {
```
michael-cico, line 199: Nit, might want to shorten the stage name for readability, i.e., "Keycloak Update During Install" or something

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
```diff
@@ -0,0 +1,578 @@
...
+                timeout(time: 30, unit: "MINUTES")
+            }
+            steps {
+                sh """
+                    ${VZ_COMMAND} uninstall --timeout 30m
+                """
+            }
+        }
+
+        stage('Kickoff Parallel VZ Install and VZ CR update during Rancher Install') {
```
michael-cico, line 303: Nit, maybe shorten this stage name for readability.

michael-cico, line 303: Do we need 2 different update during install tests?  Are they testing different things?

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
```diff
@@ -0,0 +1,578 @@
...
+                                }
+                            }
+                        }
+                        failure {
+                            archiveArtifacts artifacts: "**/kind-logs/**", allowEmptyArchive: true
+                        }
+                        always {
+                            archiveArtifacts artifacts: "acceptance-test-operator.yaml,downloaded-operator.yaml", allowEmptyArchive: true
+                            // enable debug logging of Verrazzano api istio proxy
+                            script {
```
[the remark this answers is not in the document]
marcogelle: Oh yes, this is just copied from the dynamic config test pipeline. I'll go ahead and delete that part.

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
marcogelle: Agreed, we don't need this block. I'll remove it in my next set of changes

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
marcogelle: That's fair; I agree the name is too long. I'll shorten it.

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
marcogelle: Will do

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
michael-cico: just for clarification, I mean the ENABLE_ENVOY_LOGGING block and parameter

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
marcogelle: The 2 different updates only differ in the timing of the update, which affects the reproducibility of bugs in this type of scenario. For example, in the case of https://github.com/verrazzano/verrazzano/pull/5210, the bug did not occur if the update happened before the `bootstrap-secret` Secret was created.

I was thinking that including different timings would allow this pipeline to catch errors more effectively, but let me know if you still think that having 2 updates is unnecessary.

marcogelle: I think the parameterization of the condition checks is a good change. I can work on that and add it to this PR, rather than doing a second pass.

I'm not very confident in the coverage provided by only one scenario, so calling both scenarios in parallel in the wrapper pipeline (`ci/dynamic-updates/JenkinsfileTrigger`) makes the most sense to me.

Also, if we later decide that having both scenarios is overkill, then it should be easy to remove one once the test is parameterized.

michael-cico: Sounds good; I think generally, at some point in the far flung future (well hopefully not) we'd like most of our pipelines to be short and focused like we're talking about here.  For example, the other dynamic-update test pipeline could probably be physically broken up and parameterized like this if we had the time (not recommending we do that now, to be clear 😄  ).

ci/dynamic-updates/JenkinsfileUpdateDuringInstall
```diff
@@ -195,7 +193,7 @@ pipeline {
             }
         }
 
-        stage('Update During Keycloak Install') {
+        stage('Update During Install') {
```
michael-cico, line 196: Looks good!  Maybe add a comment section above this to describe what's being tested and how, for future eyes.

michael-cico (approved):

PR #5331, opened 2023-02-08
VZ-8554: updated resource usage and application dashboard

lake-of-dreams:
- Same layout for resource usage and application dashboards
- some fixes to application dashboard regarding legends
- Removed points from graph as the graph looked a bit messy on increasing the time range


kminder (dismissed):
Approved.

kminder (approved):
Approved.

PR #5334, opened 2023-02-08
VZ-8248 Added logic and tests to default resource requirements webhook to be …

AbeHern:
Adding to internal config to create condition for creating ResourceRequirementValidation webhook


platform-operator/scripts/wait4webhook.sh
```diff
@@ -6,6 +6,6 @@ while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano
...
 while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano-platform-operator-webhook:443/v1beta1-validate-mysql-install-override-values -H 'Content-Type: application/json')" != "200" ]]; do sleep 5; done
 while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano-platform-operator-webhook:443/v1alpha1-validate-mysql-install-override-values -H 'Content-Type: application/json')" != "200" ]]; do sleep 5; done
-while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano-platform-operator-webhook:443/v1beta1-validate-requirements -H 'Content-Type: application/json')" != "200" ]]; do sleep 5; done
-while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano-platform-operator-webhook:443/v1alpha1-validate-requirements -H 'Content-Type: application/json')" != "200" ]]; do sleep 5; done
+#while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' https://verrazzano-platform-operator-webhook:443/v1beta1-validate-requirements -H 'Content-Type: application/json')" != "200" ]]; do sleep 5; done
```
kminder, line 9: I'm not really sure how this script is used.  Ideally these checks would be done if the resource requirements validation webhooks were enabled.  Is there any possibility of checking the setting of from the args in the VPO deployment?

anders-swanson: This script is executed in the init container. So if the deployment resource is aware of these parameters, then yes.

AbeHern: My builds have failed regardless of this file. It's because of the conditional I wrapped around the v1beta1 webhook in run_webhook.go

PR #5341, opened 2023-02-09
VZ-8212: Update helm charts to enable k8s 1.25 support for PodDisruptionBudget

Abhijithang:
PodDisruptionBudget for the policy/v1beta1 API version is not supported in k8s v1.25. So the existing manifests should be migrated to policy/v1 API version. Therefore, this PR will use the appropriate API version based on the underlying k8s version. 
Source: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

### DIFF
--- a/platform-operator/thirdparty/charts/argo-cd/templates/argocd-application-controller/pdb.yaml
+++ b/platform-operator/thirdparty/charts/argo-cd/templates/argocd-application-controller/pdb.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.controller.pdb.enabled }}
-apiVersion: {{ include "argo-cd.apiVersion.pdb" . }}
+apiVersion: {{ ternary "policy/v1" "policy/v1beta1" (semverCompare ">=1.21.0-0" .Capabilities.KubeVersion.Version) }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ include "argo-cd.controller.fullname" . }}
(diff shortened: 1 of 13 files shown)


pshirodkar-orcl:
I am seeing references to "defaultPodDisruptionBudget" in Istio. One such reference here https://github.com/verrazzano/verrazzano/blob/e4fe87c3205cea868ce57cf1190ad4be142844c0/platform-operator/helm_config/overrides/istio-cr.yaml#L85. Any idea what happens when enabling it in k8s 1.25?

platform-operator/thirdparty/charts/argo-cd/templates/argocd-application-controller/pdb.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.controller.pdb.enabled }}
-apiVersion: {{ include "argo-cd.apiVersion.pdb" . }}
+apiVersion: {{ ternary "policy/v1" "policy/v1beta1" (semverCompare ">=1.21.0-0" .Capabilities.KubeVersion.Version) }}
```
pfmackin, line 2: What happens during upgrade to all the old PDB resources that have the older version?

Abhijithang: The above change in the PDB helm chart is an API version value, so since it is a change in the PDB helm chart definition, I think the helm upgrade will create a new resource for k8s1.25 and remove the old API version resource.

However, this need to be tested locally and if it does not remove the existing PDB resources for some reason then we will have to remove it when VPO reconciles after the underlying k8s upgrade.

In any case I think the above changes for PDB helm chart would not change. So should I take the upgrade testing in a new task and fix post-upgrade issues if the resource remain same? or it has to go in with this task?

pfmackin (approved):

Abhijithang:
> I am seeing references to "defaultPodDisruptionBudget" in Istio. One such reference here
> 
> https://github.com/verrazzano/verrazzano/blob/e4fe87c3205cea868ce57cf1190ad4be142844c0/platform-operator/helm_config/overrides/istio-cr.yaml#L85
> . Any idea what happens when enabling it in k8s 1.25?

If enabled it will create a PodDisruptionBudget resource, and based on the parameters in the config such as minAvailable/maxAavailable the pod eviction will be carried out when k8s is upgraded to 1.25. To upgrade without any issues it is recommended to increase the replicas of the istio pod before replacing the compute nodes with k8s.1.25 to make sure PDB does not prevent eviction

PR #5350, opened 2023-02-10
VZ-8625: Fix vz analyze to properly detect install failure

saketmht:
The verrazzano-resource.json file which the analyze code parses to get the state of vz components initially had the contents of VerrazzanoList type. But as part of VZ-8249, changes were made such that verrazzano-resource.json file now has contents of Verrazzano type. The analyze code still unmarshals the file into a VerrazzanoList variable which results in empty contents so it is not able to get the components status.

Added logic to unmarshal the verrazzano-resource.json to Verrazzano type as well if the contents of VerrazzanoList.Items is empty.


agv426 (approved):

tools/vz/pkg/analysis/internal/util/cluster/install.go
```diff
@@ -337,24 +338,35 @@ func getComponentsNotReady(log *zap.SugaredLogger, clusterRoot string) ([]string
 		return compsNotReady, err
 	}
 
-	// There should be only one Verrazzano resource, so the first item from the list should be good enough
-	for _, vzRes := range vzResourceList.Items {
-		if vzRes.Status.State != installv1alpha1.VzStateReady {
-			log.Debugf("Verrazzano installation is not complete, installation state %s", vzRes.Status.State)
+	var vzRes installv1alpha1.Verrazzano
```
bobdonat, line 341: is there still a need to unmarshal into a verrazzanoList?

bhatpmk (approved):
LGTM

tools/vz/pkg/analysis/internal/util/cluster/install.go
saketmht: It is for backwards compatibility. The bug reports generated using CLI from earlier versions would still generate the verrazzanl-resource.json as VerrazzanoList type

PR #5354, opened 2023-02-10
VZ-8616: ArgoCD multicluster fixes

leliyin:
VZ-8616: ArgoCD multicluster fixes

1. Use vmc name as cluster name to be shown in the argo UI
2. Check clusterroletemplatebinding exists before POST
3. Only create the token once and then on reconcile get it until it has an expiration

### DIFF
--- a/cluster-operator/controllers/vmc/argocd.go
+++ b/cluster-operator/controllers/vmc/argocd.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"time"
 
@@ -30,11 +31,13 @@ import (
 )
 
 const (
-	clusterSecretName               = "argocd-cluster-secret"    //nolint:gosec
-	argocdClusterTokenTTLEnvVarName = "ARGOCD_CLUSTER_TOKEN_TTL" //nolint:gosec
-	createTimestamp                 = "verrazzano.io/create-timestamp"
-	expiresAtTimestamp              = "verrazzano.io/expires-at-timestamp"
-	clusterroletemplatebindingsPath = "/v3/clusterroletemplatebindings"
+	clusterSecretName                      = "argocd-cluster-secret"    //nolint:gosec
+	argocdClusterTokenTTLEnvVarName        = "ARGOCD_CLUSTER_TOKEN_TTL" //nolint:gosec
+	createTimestamp                        = "verrazzano.io/create-timestamp"
+	expiresAtTimestamp                     = "verrazzano.io/expires-at-timestamp"
+	clusterroletemplatebindingsPath        = "/v3/clusterroletemplatebindings"
+	clusterroletemplatebindingByUserIDPath = "/v3/clusterroletemplatebindings?userId="
+	dataField                              = "data"
 )
 
 func (r *VerrazzanoManagedClusterReconciler) isArgoCDEnabled() (bool, error) {
@@ -170,7 +173,7 @@ func (r *VerrazzanoManagedClusterReconciler) createOrUpdateArgoCDSecret(rc *ranc
 	return err
 }
 
-func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *corev1.Secret, rc *rancherutil.RancherConfig, clusterID, cluserName string, rancherURL string, caData []byte) error {
+func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *corev1.Secret, rc *rancherutil.RancherConfig, clusterName, clusterID, rancherURL string, caData []byte) error {
 	token := rc.APIAccessToken
 	if secret.Annotations == nil {
 		secret.Annotations = map[string]string{}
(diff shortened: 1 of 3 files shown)


cluster-operator/controllers/vmc/argocd.go
```diff
@@ -30,11 +31,13 @@ import (
...
-	createTimestamp                 = "verrazzano.io/create-timestamp"
-	expiresAtTimestamp              = "verrazzano.io/expires-at-timestamp"
-	clusterroletemplatebindingsPath = "/v3/clusterroletemplatebindings"
+	clusterSecretName                      = "argocd-cluster-secret"    //nolint:gosec
+	argocdClusterTokenTTLEnvVarName        = "ARGOCD_CLUSTER_TOKEN_TTL" //nolint:gosec
+	createTimestamp                        = "verrazzano.io/create-timestamp"
+	expiresAtTimestamp                     = "verrazzano.io/expires-at-timestamp"
+	clusterroletemplatebindingsPath        = "/v3/clusterroletemplatebindings"
+	clusterroletemplatebindingByUserIDPath = "/v3/clusterroletemplatebindings?userId="
+	dataField                              = "data"
```
fredtibbitts, line 40: This constant is only used in one place, and it's very generic, I would just eliminate the const.

cluster-operator/controllers/vmc/argocd.go
```diff
@@ -170,7 +173,7 @@ func (r *VerrazzanoManagedClusterReconciler) createOrUpdateArgoCDSecret(rc *ranc
 	return err
 }
 
-func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *corev1.Secret, rc *rancherutil.RancherConfig, clusterID, cluserName string, rancherURL string, caData []byte) error {
+func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *corev1.Secret, rc *rancherutil.RancherConfig, clusterName, clusterID, rancherURL string, caData []byte) error {
```
fredtibbitts, line 176: This function is getting fairly large. It should be refactored into multiple smaller functions.

fredtibbitts (approved):
A couple of comments but they can be addressed in a separate PR. The logic seems OK to me.

cluster-operator/controllers/vmc/argocd.go
```diff
@@ -249,6 +264,28 @@ func (r *VerrazzanoManagedClusterReconciler) updateArgoCDClusterRoleBindingTempl
...
+	// Send a request to see if the clusterroletemplatebinding for the given user exists
+	reqURL := rc.BaseURL + clusterroletemplatebindingByUserIDPath + url.PathEscape(userID) + "&clusterId=" + url.PathEscape(vmc.Status.RancherRegistration.ClusterID)
+	headers := map[string]string{"Authorization": "Bearer " + rc.APIAccessToken}
+	response, body, err := rancherutil.SendRequest(http.MethodGet, reqURL, headers, "", rc, r.log)
+	if err != nil {
+		return err
+	}
+	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusNotFound {
+		return err
+	}
```
davisgomes, line 276: I'm not sure I understand this block. Won't err always be nil here?

cluster-operator/controllers/vmc/argocd.go
```diff
@@ -249,6 +264,28 @@ func (r *VerrazzanoManagedClusterReconciler) updateArgoCDClusterRoleBindingTempl
...
 	}
 
+	// Send a request to see if the clusterroletemplatebinding for the given user exists
+	reqURL := rc.BaseURL + clusterroletemplatebindingByUserIDPath + url.PathEscape(userID) + "&clusterId=" + url.PathEscape(vmc.Status.RancherRegistration.ClusterID)
+	headers := map[string]string{"Authorization": "Bearer " + rc.APIAccessToken}
+	response, body, err := rancherutil.SendRequest(http.MethodGet, reqURL, headers, "", rc, r.log)
+	if err != nil {
+		return err
+	}
+	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusNotFound {
```
davisgomes, line 274: We should not ignore status not found. This request should always return 200 even if the item does not exist

cluster-operator/controllers/vmc/argocd.go
```diff
@@ -249,6 +264,28 @@ func (r *VerrazzanoManagedClusterReconciler) updateArgoCDClusterRoleBindingTempl
...
+		return err
+	}
+
+	if response.StatusCode == http.StatusOK {
```
davisgomes, line 278: I think you can just check `if statusCode != http.StatusOK` then return a new error

davisgomes: Then you can move the rest of the code in this block down outside of the if statement.

pkg/rancherutil/rancher_config.go
```diff
@@ -233,32 +228,47 @@ func CreateTokenWithTTL(rc *RancherConfig, log vzlog.VerrazzanoLogger, ttl, clus
 		return "", "", log.ErrorfNewErr("Failed to parse response: %v", err)
 	}
 
-	return tokenResponse.Token, tokenResponse.Name, nil
+	return tokenResponse.Token, tokenResponse.Created, nil
+}
+
+type TokenAttrs struct {
```
davisgomes, line 234: Can you rename the two token structs to something more consistent i.e. `TokenPostResponse` and `TokenGetResponse`

cluster-operator/controllers/vmc/argocd.go
leliyin: will file a separate ticket to address this.

cluster-operator/controllers/vmc/argocd.go
leliyin: Done.

pkg/rancherutil/rancher_config.go
leliyin: Fixed.

cluster-operator/controllers/vmc/argocd.go
leliyin: Fixed.

cluster-operator/controllers/vmc/argocd.go
leliyin: Fixed.

davisgomes (approved):

fredtibbitts (approved):

PR #5358, opened 2023-02-10
VZ-8602 New images for the 1.5 release need to have names that are compliant with OCR

stevshap:
Updating image names under `additional-rancher` in the bom, etc.


bryantidd (approved):

tests/e2e/registry/registry_test.go
```diff
@@ -142,6 +142,9 @@ func getRegistryURL(containerImage string) (string, error) {
 	}
 	imageName := getImageName(containerImage)
 	// If the image is not defined in the bom, return an error
+	if imageRegistryMap[imageName] == "" && imageName == "shell" {
```
fredtibbitts, line 145: It would probably be good to have a comment here explaining why this is needed.

fredtibbitts (approved):
One comment but it can be addressed in a separate PR.

PR #5370, opened 2023-02-13
VZ-8575: Remove system-tools from VPO image

mgianatagh:
Remove system-tools from VPO image.  It is no longer used.

### DIFF
--- a/platform-operator/Dockerfile
+++ b/platform-operator/Dockerfile
@@ -37,11 +37,6 @@ RUN groupadd -r verrazzano \
     && mkdir /home/verrazzano \
     && chown -R 1000:verrazzano /home/verrazzano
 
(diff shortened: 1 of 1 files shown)


pshirodkar-orcl (approved):

platform-operator/Dockerfile
```diff
@@ -37,11 +37,6 @@ RUN groupadd -r verrazzano \
     && mkdir /home/verrazzano \
     && chown -R 1000:verrazzano /home/verrazzano
 
-# install Rancher system-tools (used by uninstall)
```
bobdonat, line 40: can the yum install of wget be removed too

mgianatagh: It looks like it can.  I will give that a try.

bobdonat (approved):

PR #5372, opened 2023-02-14
VZ-8629 Rancher image addition

stevshap:
Adding justification for the added logic to private registry tests

### DIFF
--- a/tests/e2e/registry/registry_test.go
+++ b/tests/e2e/registry/registry_test.go
@@ -141,10 +141,13 @@ func getRegistryURL(containerImage string) (string, error) {
 		}
 	}
 	imageName := getImageName(containerImage)
-	// If the image is not defined in the bom, return an error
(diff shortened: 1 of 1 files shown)


tests/e2e/registry/registry_test.go
```diff
@@ -141,10 +141,13 @@ func getRegistryURL(containerImage string) (string, error) {
 		}
 	}
 	imageName := getImageName(containerImage)
-	// If the image is not defined in the bom, return an error
+	// When the Rancher images changed from vz/rancher/shell to vz/rancher-shell, this logic was needed
+	// in the case of private registry upgrade. This function throws an error if the old container image
```
fredtibbitts, line 145: I don't think this is exactly right. We aren't throwing an error in this case. This logic handles looking up both the old and new shell image names in the BOM, and if neither are found, then this function returns an error.

fredtibbitts (approved):

PR #5396, opened 2023-02-16
VZ-8248 Disabling resource requirements validation

AbeHern:
Added boolean param to disabled resource requirements warnings and subsequent logic to test and toggle when desired.

### DIFF
--- a/pkg/k8sutil/apply_yaml.go
+++ b/pkg/k8sutil/apply_yaml.go
@@ -210,7 +210,7 @@ func (y *YAMLApplier) applyAction(obj *unstructured.Unstructured) error {
 			}
 		}
 		// Delete any keys in server obj not included in the client fields.
-		for key, _ := range obj.Object {
+		for key := range obj.Object {
 			if key == "kind" || key == "apiVersion" {
 				continue
 			}
--- a/pkg/k8sutil/apply_yaml.go
+++ b/pkg/k8sutil/apply_yaml.go
@@ -209,6 +209,25 @@ func (y *YAMLApplier) applyAction(obj *unstructured.Unstructured) error {
 				return err
 			}
 		}
+		// Delete any keys in server obj not included in the client fields.
+		for key, _ := range obj.Object {
+			if key == "kind" || key == "apiVersion" {
+				continue
+			}
+			keyFound := false
+			for _, clientField := range clientFields {
+				if clientField.name == key {
+					keyFound = true
+					break
+				}
+			}
+			if !keyFound {
+				err = unstructured.SetNestedField(obj.Object, nil, key)
+				if err != nil {
+					return err
+				}
+			}
+		}
 		return nil
 	})
 	if err != nil {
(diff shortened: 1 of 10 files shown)


bobdonat (approved):

anders-swanson (approved):

anders-swanson (approved):

bobdonat (approved):

platform-operator/helm_config/charts/verrazzano-platform-operator/values.yaml
```diff
@@ -39,3 +39,5 @@ webhookAffinity:
                   - verrazzano-platform-operator-webhook
           topologyKey: kubernetes.io/hostname
         weight: 100
+webhooks:
+  resourceValidation: false
```
kminder, line 43: Doesn't this need to be templated?

pkg/k8sutil/apply_yaml_test.go
```diff
@@ -141,9 +141,8 @@ func TestApplyFNonSpec(t *testing.T) {
 	assert.Equal(t, 1, len(saUpdated.ImagePullSecrets))
 	assert.Equal(t, "verrazzano-container-registry", saUpdated.ImagePullSecrets[0].Name)
 
-	assert.NotEmpty(t, saUpdated.Secrets)
-	assert.Equal(t, 1, len(saUpdated.Secrets))
-	assert.Equal(t, "verrazzano-platform-operator-token", saUpdated.Secrets[0].Name)
```
kminder, line 146: Why was this removed?

pkg/k8sutil/apply_yaml.go
```diff
@@ -209,6 +209,25 @@ func (y *YAMLApplier) applyAction(obj *unstructured.Unstructured) error {
 				return err
 			}
 		}
+		// Delete any keys in server obj not included in the client fields.
```
kminder, line 212: Why was this added?

platform-operator/helm_config/charts/verrazzano-platform-operator/values.yaml
AbeHern: Here is where the value is defaulted. Not sure if it needs to be formatted here but it is formatted in validatingwebhookconfiguration.yaml

pkg/k8sutil/apply_yaml_test.go
AbeHern: The file used in this test has no Secrets, so the test kept breaking as it was expecting to have a secret. So it was removed

pkg/k8sutil/apply_yaml.go
AbeHern: I had Bob help me with this. From what I've understood, last time I merged and broke master it was failing at Upgrade Path. There was an issue with the vz cli that was still expecting there to be a validating webhook when the new operator.yaml does not indeed have one. So This pruning essentially takes care of this.

platform-operator/helm_config/charts/verrazzano-platform-operator/values.yaml
kminder: My bad.  I was thinking this was a template.

pkg/k8sutil/apply_yaml_test.go
kminder: ok

pkg/k8sutil/apply_yaml.go
kminder: I see.  Thanks.

PR #5433, opened 2023-02-22
VZ-7954: Fix to stop recreating cluster registration tokens

saidu941:
- Changes to rancher.go to get the existing CRT token before creating one

### DIFF
--- a/cluster-operator/controllers/vmc/rancher.go
+++ b/cluster-operator/controllers/vmc/rancher.go
@@ -345,8 +345,7 @@ func getCACertFromManagedClusterSecret(rc *rancherutil.RancherConfig, clusterID,
 // getRegistrationYAMLFromRancher creates a registration token in Rancher for the managed cluster and uses the
 // returned token to fetch the registration (manifest) YAML.
 func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherClusterID string, log vzlog.VerrazzanoLogger) (string, error) {
-	action := http.MethodPost
-	payload := `{"type": "clusterRegistrationToken", "clusterId": "` + rancherClusterID + `"}`
+
 	reqURL := rc.BaseURL + clusterRegTokenPath
 	headers := map[string]string{"Content-Type": "application/json"}
 	headers["Authorization"] = "Bearer " + rc.APIAccessToken
@@ -354,10 +353,12 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
 	var token string
 	token, err := getRegistrationTokenFromRancher(rc, rancherClusterID, log)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
-
 	if token == "" {
+		action := http.MethodPost
+		payload := `{"type": "clusterRegistrationToken", "clusterId": "` + rancherClusterID + `"}`
+
 		response, manifestContent, err := rancherutil.SendRequest(action, reqURL, headers, payload, rc, log)
 		if err != nil {
 			return "", err
@@ -376,8 +377,7 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
 	}
 	// Rancher 2.5.x added the cluster ID to the manifest URL.
 	manifestURL := rc.BaseURL + manifestPath + token + "_" + rancherClusterID + ".yaml"
-
-	action = http.MethodGet
+	action := http.MethodGet
 	response, manifestContent, err := rancherutil.SendRequest(action, manifestURL, headers, "", rc, log)
 
 	if err != nil {
@@ -402,12 +402,15 @@ type ClusterRegistrationTokens struct {
 
 func getRegistrationTokenFromRancher(rc *rancherutil.RancherConfig, rancherClusterID string, log vzlog.VerrazzanoLogger) (string, error) {
 
+	log.Infof("Calling getRegistrationTokenFromRancher")
 	action := http.MethodGet
-	reqURL := rc.BaseURL + "v3/clusterregistrationtoken?state=active&&clusterId=" + rancherClusterID
+	reqURL := rc.BaseURL + clusterRegTokenPath + "?state=active&&clusterId=" + rancherClusterID
 	headers := map[string]string{"Content-Type": "application/json"}
 	headers["Authorization"] = "Bearer " + rc.APIAccessToken
 
-	response, manifestContent, err := rancherutil.SendRequest(action, reqURL, headers, "", rc, log)
+	response, manifestContent, err := rancherutil.SendRequest(action, reqURL, headers, "{}", rc, log)
+	log.Infof("Called getRegistrationTokenFromRancher")
+	log.Info(response)
 	if err != nil {
 		return "", err
 	}
@@ -417,6 +420,7 @@ func getRegistrationTokenFromRancher(rc *rancherutil.RancherConfig, rancherClust
 	}
 
 	data, err := httputil.ExtractFieldFromResponseBodyOrReturnError(manifestContent, "data", "unable to find data token in Rancher response")
+	log.Infof("Extracted data")
 	if err != nil {
 		return "", err
 	}
@@ -431,7 +435,7 @@ func getRegistrationTokenFromRancher(rc *rancherutil.RancherConfig, rancherClust
 	}
 
 	log.Infof("No active clusterRegistrationToken found")
-	return "", nil
+	return "", err
 }
 
 // createOrUpdateSecretRancherProxy simulates the controllerutil create or update function through the Rancher Proxy API for secrets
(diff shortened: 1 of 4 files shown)


cluster-operator/controllers/vmc/rancher.go
```diff
@@ -1,27 +1,26 @@
...
-	"net/http"
-
 	"github.com/Jeffail/gabs/v2"
 	cons "github.com/verrazzano/verrazzano/pkg/constants"
 	"github.com/verrazzano/verrazzano/pkg/httputil"
 	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
 	"github.com/verrazzano/verrazzano/pkg/mcconstants"
 	"github.com/verrazzano/verrazzano/pkg/rancherutil"
 	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"io"
```
fredtibbitts, line 16: This and `net/http` should not have moved from the first import section. All system module imports should be in the first section.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -352,29 +351,63 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
...
-
+	var token string
+	token, err := getRegistrationTokenFromRancher(rc, rancherClusterID, log)
 	if err != nil {
 		return "", err
 	}
 
-	err = httputil.ValidateResponseCode(response, http.StatusCreated)
+	if token == "" {
+		response, manifestContent, err := rancherutil.SendRequest(action, reqURL, headers, payload, rc, log)
```
fredtibbitts, line 361: Can you move the definition of action, reqURL, and payload inside this `if` block? They aren't needed outside.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -352,29 +351,63 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
...
+	ClusterID string `json:"clusterId"`
+	ExpiresAt string `json:"expiresAt"`
+	State     string `json:"state"`
+	Token     string `json:"token"`
+}
+
+func getRegistrationTokenFromRancher(rc *rancherutil.RancherConfig, rancherClusterID string, log vzlog.VerrazzanoLogger) (string, error) {
+
+	action := http.MethodGet
+	reqURL := rc.BaseURL + clusterRegTokenPath + "?state=active&&clusterId=" + rancherClusterID
```
fredtibbitts, line 406: Extra "&"?

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -1,27 +1,26 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
```
kminder, line 1: This should not change the first year.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -384,7 +417,22 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
...
+	json.Unmarshal([]byte(data), &items)
+	for _, item := range items {
+		if item.ClusterID == rancherClusterID && item.State == "active" {
+			log.Oncef("ClusterRegistrationToken exists for the cluster %s", rancherClusterID)
+			return item.Token, nil
+		}
+	}
+
+	log.Infof("No active clusterRegistrationToken found")
+	return "", nil
```
kminder, line 435: Should this be an error?

cluster-operator/controllers/vmc/vmc_controller_test.go
```diff
@@ -1348,6 +1348,24 @@ func TestRegisterClusterWithRancherHTTPErrorCases(t *testing.T) {
 			return resp, nil
 		})
 
+	// Expect an HTTP request to get the registration token in Rancher for the clusterId
+	mockRequestSender.EXPECT().
```
kminder, line 1352: In the future please consider using fakes instead of mocks.

pkg/httputil/httputil_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
```
kminder, line 1: Why did this need to change.  It should not have changed the first year.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -384,7 +417,22 @@ func getRegistrationYAMLFromRancher(rc *rancherutil.RancherConfig, rancherCluste
...
+	var items []ClusterRegistrationTokens
+	json.Unmarshal([]byte(data), &items)
+	for _, item := range items {
+		if item.ClusterID == rancherClusterID && item.State == "active" {
+			log.Oncef("ClusterRegistrationToken exists for the cluster %s", rancherClusterID)
+			return item.Token, nil
+		}
+	}
+
+	log.Infof("No active clusterRegistrationToken found")
```
fredtibbitts, line 434: You might want to make this `Debugf` or `Oncef`, and it should probably be "cluster registration token", and also add the cluster id to the message.

cluster-operator/controllers/vmc/vmc_controller_test.go
```diff
@@ -1348,6 +1348,24 @@ func TestRegisterClusterWithRancherHTTPErrorCases(t *testing.T) {
...
+		Do(gomock.Not(gomock.Nil()), mockmatchers.MatchesURI(clusterRegTokenPath)).
+		DoAndReturn(func(httpClient *http.Client, req *http.Request) (*http.Response, error) {
+			asserts.Contains(clusterRegTokenPath, req.URL.Path)
+
+			_, err := io.ReadAll(req.Body)
+			asserts.NoError(err)
+
+			r := io.NopCloser(bytes.NewReader([]byte(`{"data":[{"token":"unit-test-token"}]}`)))
```
fredtibbitts, line 1360: Since you aren't also returning a `state` of active, this token won't be used. Is that what you expect?

cluster-operator/controllers/vmc/vmc_controller_test.go
```diff
@@ -2171,6 +2195,27 @@ func expectRegisterClusterWithRancherHTTPCalls(t *testing.T, requestSenderMock *
...
+		Do(gomock.Not(gomock.Nil()), mockmatchers.MatchesURI(clusterRegTokenPath)).
+		DoAndReturn(func(httpClient *http.Client, req *http.Request) (*http.Response, error) {
+			asserts.Contains(clusterRegTokenPath, req.URL.Path)
+
+			// assert that the cluster ID in the request body is what we expect
+			_, err := io.ReadAll(req.Body)
+			asserts.NoError(err)
+
+			// return a response with the CRT
+			r := io.NopCloser(bytes.NewReader([]byte(`{"data":[]}`)))
```
fredtibbitts, line 2209: I'm not sure how this is working, since this mock is not returning a token.

fredtibbitts (commented):
I don't see a unit test for the case where the registration token already exists and we verify that we do not try to create a new token.

cluster-operator/controllers/vmc/vmc_controller_test.go
```diff
@@ -2171,6 +2195,26 @@ func expectRegisterClusterWithRancherHTTPCalls(t *testing.T, requestSenderMock *
 
 	manifestToken := "unit-test-manifest-token"
 
+	// Expect an HTTP request to create the registration token in Rancher
```
fredtibbitts, line 2198: Update comment.

cluster-operator/controllers/vmc/vmc_controller_test.go
```diff
@@ -2171,6 +2195,26 @@ func expectRegisterClusterWithRancherHTTPCalls(t *testing.T, requestSenderMock *
...
+	requestSenderMock.EXPECT().
+		Do(gomock.Not(gomock.Nil()), mockmatchers.MatchesURI(clusterRegTokenPath)).
+		DoAndReturn(func(httpClient *http.Client, req *http.Request) (*http.Response, error) {
+			asserts.Contains(clusterRegTokenPath, req.URL.Path)
+
+			// assert that the cluster ID in the request body is what we expect
+			_, err := io.ReadAll(req.Body)
+			asserts.NoError(err)
+
+			// return a response with the CRT
```
fredtibbitts, line 2208: Update comment.

cluster-operator/controllers/vmc/rancher.go
fredtibbitts: In this case, the request to lookup tokens succeeded but we did not find an existing, active reg token for the cluster. We then want to create a reg token. I suppose we could treat this as an error, but then the calling function would just always ignore any errors and proceed to creating a token. I guess there are arguments for both.

cluster-operator/controllers/vmc/vmc_controller_test.go
fredtibbitts: I agree, fakes are so much easier to use and understand. In this case, we're modifying existing tests that already use mocks, so converting to fakes would have been a sizable effort. I think in general as part of tech debt cleanup we need to go through all of the tests that use mocks and convert them to fakes.

fredtibbitts (approved):

PR #5435, opened 2023-02-22
Adding e2e tests for Argo CD Multicluster

preethignanesh:
Argo CD registers the managed cluster, we need to add e2e tests for Argo CD verification.
By default, Argo CD is disabled in all profiles, so we explicitly enable it in the multicluster pipeline


tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: app1
+  namespace: argocd
+spec:
+  destination:
```
desagar, line 9: where does it specify that the cluster to place this application in is the managed cluster?

desagar: also this YAML file should be in the multicluster/verify-argocd directory since that's where it's used, not in config/scripts

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,187 @@
...
+		// THEN expect that the multi-cluster resources have been created on the managed cluster
+		t.It("Has multi cluster resources", func() {
+			Eventually(func() bool {
+				return examples.VerifyMCResources(managedKubeconfig, false, true, testNamespace)
+			}, waitTimeout, pollingInterval).Should(BeTrue())
+		})
+		// GIVEN an admin cluster and at least one managed cluster
+		// WHEN the multi-cluster example application has been created on admin cluster and placed in managed cluster
+		// THEN expect that the app is deployed to the managed cluster
+		t.It("Has application placed", func() {
```
desagar, line 117: can you explain how this works? I'm not clear on how the app gets placed on the managed cluster when the Application resource doesn't have a destination name specified.

ci/multicluster/Jenkinsfile
```diff
@@ -835,7 +840,7 @@ def uninstallVerrazzanoOnCluster(count, verrazzanoConfig) {
             sh """
                 export KUBECONFIG="${KUBECONFIG_DIR}/${count}/kube_config"
                 echo "Deleting \\$verrazzanoConfig"
-                time ${VZ_COMMAND} uninstall
+                time ${VZ_COMMAND} uninstall --timeout 45m
```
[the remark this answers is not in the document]
preethignanesh: Yes, it was taking around 32 mins for the uninstall to complete. On discussion with @mgianatagh, I set it to 45mins and he would be logging a jira to get the ranhcer-cleanup job to run faster.

tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
```diff
@@ -0,0 +1,25 @@
...
+    name: ''
+    namespace: hello-helidon
```
desagar, line 11: I'd prefer if we used a namespace specific to the argo case e.g. hello-helidon-argo

tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
preethignanesh: In Argo CD we specify the destination as admin cluster and use the multi cluster examples which has the placement set to deploy on the managed cluster. We are testing if the deployment indeed happens on the managed cluster.
I can move the location of the file to that directory as you have suggested.

ci/multicluster/Jenkinsfile
```diff
@@ -460,6 +460,11 @@ pipeline {
                         runGinkgo('multicluster/verify-rancher', '${TEST_DUMP_ROOT}/verify-rancher')
                     }
                 }
+                stage('verify-argocd') {
```
desagar, line 463: since this is verifying an application, I think it would be better to move this to the Acceptance Tests part of the Jenkinsfile.

tests/e2e/multicluster/verify-argocd/argocd_test.go
preethignanesh: We are deploying the  [hello-helidon](https://github.com/verrazzano/verrazzano/tree/master/examples/multicluster/hello-helidon)/verrazzano-project.yaml

tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
desagar: oh I see, so you are not using ArgoCD to directly target the managed cluster.

ci/multicluster/Jenkinsfile
desagar: I think this becomes more important since you're not using ArgoCD to target the app to the cluster.

tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
preethignanesh: We need to use the same namespace that is specified in the application example.

tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml
preethignanesh: yes, we wanted to ensure the functionality of the placement option to be not lost even though we specify a dufferent destination in the Argo CD

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,178 @@
...
+	t.Context("Delete resources", func() {
+		t.It("Delete resources on admin cluster", func() {
+			Eventually(func() error {
+				return deleteArgoCDApplication(managedKubeconfig)
+			}, waitTimeout, pollingInterval).ShouldNot(HaveOccurred())
+		})
+
+		t.It("Verify automatic deletion on managed cluster", func() {
+			Consistently(func() bool {
+				result, err := examples.VerifyHelloHelidonInCluster(managedKubeconfig, false, true, testProjectName, testNamespace)
```
desagar, line 113: shouldn't this be checking for app deleted in the cluster? This will fail because you want the app to _not_ be running after deletion. The examples.VerifyAppDeleted function might do what you want here.

Also, why did you choose to use Consistently here instead of Eventually? It's not quite clear to me.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.install"), func() {
+	t.Context("Admin Cluster", func() {
+		t.BeforeEach(func() {
+			os.Setenv(k8sutil.EnvVarTestKubeConfig, os.Getenv("ADMIN_KUBECONFIG"))
+		})
+
+		t.It("has the expected secrets", func() {
+			secretName := fmt.Sprintf("%s-argocd-cluster-secret", managedClusterName)
+			Eventually(func() bool {
```
fredtibbitts, line 76: In general, the functions used in an `Eventually` should return an `error` as the only or 2nd return result. Otherwise we can lose useful information that indicates why the test failed. In this particular case, `findSecret` should return an `error`, the `Eventually` func should return an `error`, and the `Eventually` condition should change to `ShouldNot(HaveOccurred()`.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,174 @@
...
+				return findSecret(argoCDNamespace, secretName)
+			}, waitTimeout, pollingInterval).Should(BeTrue(), "Expected to find secret "+secretName)
+		})
+
+		t.It("secret has the data content with the same name as managed cluster", func() {
+			secretName := fmt.Sprintf("%s-argocd-cluster-secret", managedClusterName)
+			Eventually(func() bool {
```
fredtibbitts, line 83: Same comment as above, return `error` here instead.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	t.Context("Managed Cluster", func() {
+		// GIVEN an admin cluster and at least one managed cluster
+		// WHEN the multi-cluster example application has been  placed in managed cluster
+		// THEN expect that the app is deployed to the managed cluster
+		t.It("Has application placed", func() {
+			Eventually(func() bool {
+				result, err := helloHelidonPodsRunning(managedKubeconfig, testNamespace)
+				if err != nil {
+					AbortSuite(fmt.Sprintf("One or more pods are not running in the namespace: %v, error: %v", testNamespace, err))
```
fredtibbitts, line 97: We usually don't abort the suite like this. Unless there's a good reason, just return the error.

preethignanesh: Removed the abort suite

tests/e2e/multicluster/verify-argocd/argocd_test.go
preethignanesh: Modified the code as specified

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -0,0 +1,190 @@
...
+const (
+	waitTimeout          = 15 * time.Minute
+	pollingInterval      = 10 * time.Second
+	consistentlyDuration = 1 * time.Minute
+	testNamespace        = "hello-helidon-argo"
+	argoCDNamespace      = "argocd"
+)
+
+const (
+	argoCDHelidonApplicationFile = "tests/e2e/config/scripts/hello-helidon-argocd-mc.yaml"
```
fredtibbitts, line 32: I thought @desagar made a comment that the yaml file shouldn't be in the `config/scripts` directory? Maybe that was a different PR? Regardless, this does not seem like the right place for this file. For other tests, we create a `testdata` directory under the directory containing the test suite.

desagar (approved):

fredtibbitts (approved):

PR #5439, opened 2023-02-23
VZ-8755: Fix bug where argoCD URL doesn't get set in VZ CR (#5432)

mbenmclean:
(cherry picked from commit c6630682137f60abef14722b80f8b832f3c1a7a8)


kminder (approved):
Approved.

JeffTancill (approved):

PR #5461, opened 2023-02-27
VZ-7924 VZ CLI operations should automatically call bug-report on failure

knelson7:
Default behaviour of the CLI is to automatically perform a bug report when vz install, uninstall, or upgrade fail
- added a flag to disable automatically generating a bug report 
- added test cases to ensure disable flag functions as expected 
- error returned from failed cli command is still printed to console. Potential errors that occur from running the bug report command are printed to error output


tools/vz/cmd/install/install.go
```diff
@@ -196,7 +200,7 @@ func runCmdInstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelper)
 		// Sometimes we see intermittent webhook errors due to timeouts.
 		retry := 0
 		for {
-			err = client.Create(context.TODO(), vz)
+			err = client.Create(pkgcontext.TODO(), vz)
```
absBitm, line 203: I would suggest "pkgContext"

tools/vz/cmd/upgrade/upgrade.go
```diff
@@ -183,21 +187,56 @@ func runCmdUpgrade(cmd *cobra.Command, vzHelper helpers.VZHelper) error {
 		}
 
 		// Wait for the Verrazzano upgrade to complete
-		return waitForUpgradeToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vz.Namespace, Name: vz.Name}, timeout, vpoTimeout, logFormat)
+		err = waitForUpgradeToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vz.Namespace, Name: vz.Name}, timeout, vpoTimeout, logFormat)
```
absBitm, line 190: can we skip err variable here?

tools/vz/cmd/upgrade/upgrade.go
```diff
@@ -183,21 +187,56 @@ func runCmdUpgrade(cmd *cobra.Command, vzHelper helpers.VZHelper) error {
...
+		if err != nil {
+			return callVzBugReport(cmd, vzHelper, err)
```
absBitm, line 192: to avoid code redundancy, can we put this **callVzBugReport** inside helpers and call it from install/uninstall/upgrade?

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -156,11 +160,36 @@ func runCmdUninstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelpe
...
+		contextFlag, errFlag2 := cmd.Flags().GetString(constants.GlobalFlagContext)
+		if errFlag2 != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag2.Error())
+		}
+		cmd2.Flags().StringVar(&kubeconfig, constants.GlobalFlagKubeConfig, "", constants.GlobalFlagKubeConfigHelp)
+		cmd2.Flags().StringVar(&context, constants.GlobalFlagContext, "", constants.GlobalFlagContextHelp)
+		cmd2.Flags().Set(constants.GlobalFlagKubeConfig, kubeconfigFlag)
+		cmd2.Flags().Set(constants.GlobalFlagContext, contextFlag)
+		cmd2.PersistentFlags().Set(constants.ReportFormatFlagName, constants.SummaryReport)
+		analyzeErr := bugreport.RunCmdBugReport(cmd2, vzHelper)
```
absBitm, line 186: either we can skip analyzeErr or var **analyzeErr** should be **bugReportErr**

tools/vz/cmd/install/install_test.go
```diff
@@ -314,20 +314,20 @@ func TestInstallCmdOperatorFile(t *testing.T) {
...
 	svc := corev1.Service{}
-	err = c.Get(context.TODO(), types.NamespacedName{Namespace: "verrazzano-install", Name: "verrazzano-platform-operator"}, &svc)
+	err = c.Get(pkgcontext.TODO(), types.NamespacedName{Namespace: "verrazzano-install", Name: "verrazzano-platform-operator"}, &svc)
 	assert.NoError(t, err)
 
 	// Verify the vz resource is as expected
 	vz := v1beta1.Verrazzano{}
-	err = c.Get(context.TODO(), types.NamespacedName{Namespace: "default", Name: "verrazzano"}, &vz)
+	err = c.Get(pkgcontext.TODO(), types.NamespacedName{Namespace: "default", Name: "verrazzano"}, &vz)
 	assert.NoError(t, err)
```
absBitm, line 331: can we have a unit test or e2e test to verify the bug-report getting generated or not?

tools/vz/cmd/uninstall/uninstall_test.go
```diff
@@ -351,48 +351,48 @@ func createClusterRole() *rbacv1.ClusterRole {
...
 	// Expect the Cluster Role Binding not to be deleted
 	crb := rbacv1.ClusterRoleBinding{}
-	err = client.Get(context.TODO(), types.NamespacedName{Name: constants.VerrazzanoPlatformOperator}, &crb)
+	err = client.Get(pkgcontext.TODO(), types.NamespacedName{Name: constants.VerrazzanoPlatformOperator}, &crb)
 	assert.NoError(t, err)
 
 	// Expect the Cluster Role not to be deleted
 	cr := rbacv1.ClusterRole{}
-	err = client.Get(context.TODO(), types.NamespacedName{Name: constants.VerrazzanoManagedCluster}, &cr)
+	err = client.Get(pkgcontext.TODO(), types.NamespacedName{Name: constants.VerrazzanoManagedCluster}, &cr)
```
absBitm, line 396: can we have a unit test or e2e test to verify the bug-report getting generated or not?

tools/vz/cmd/upgrade/upgrade_test.go
```diff
@@ -188,19 +188,19 @@ func TestUpgradeCmdOperatorFile(t *testing.T) {
...
 
 	svc := corev1.Service{}
-	err = c.Get(context.TODO(), types.NamespacedName{Namespace: "verrazzano-install", Name: "verrazzano-platform-operator"}, &svc)
+	err = c.Get(pkgcontext.TODO(), types.NamespacedName{Namespace: "verrazzano-install", Name: "verrazzano-platform-operator"}, &svc)
 	assert.NoError(t, err)
 
 	// Verify the version got updated
-	err = c.Get(context.TODO(), types.NamespacedName{Namespace: "default", Name: "verrazzano"}, vz)
+	err = c.Get(pkgcontext.TODO(), types.NamespacedName{Namespace: "default", Name: "verrazzano"}, vz)
 	assert.NoError(t, err)
```
absBitm, line 204: can we have a unit test or e2e test to verify the bug-report getting generated or not?

tools/vz/cmd/upgrade/upgrade.go
knelson7: I don't think so. I have to save the err returned from waitForUpgradeToComplete() because that determines whether bug report gets called.

tools/vz/cmd/upgrade/upgrade.go
knelson7: I put the method (callVzBugReport) into bug report and reuse it for install, uninstall, and upgrade. It can't exist in helpers because it creates an import cycle since bug report imports helpers, and helpers would then need to import bug report to call NewCmdBugReport and RunCmdBugReport. Let me know if there's a better way to do this.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
 	}
 	return len(entries) == ignoreFilesCount
 }
+
+func CallVzBugReport(cmd *cobra.Command, vzHelper helpers.VZHelper, err error) error {
```
bhatpmk, line 256: I think it is better to export a function from bugreport.go, which can be called from install/upgrade/uninstall CLIs. There is no need set the global flags, if I understand correctly.

knelson7: You do need to set the global flags because they're accessed in the method getKubeConfigGivenCommand(), which is called from GetKubeClient(), GetClient(), and GetDynamicClient() in RunCmdBugReport()

tools/vz/cmd/install/install_test.go
```diff
@@ -67,6 +75,32 @@ func TestInstallCmdDefaultTimeout(t *testing.T) {
...
+
+// TestInstallCmdDefaultTimeoutNoBugReport
+// GIVEN a CLI install command with --timeout=2s and auto-bug-report=false
+//
+//	WHEN I call cmd.Execute for install
+//	THEN the CLI install command times out and a bug report is not generated
+func TestInstallCmdDefaultTimeoutNoBugReport(t *testing.T) {
+	c := fake.NewClientBuilder().WithScheme(helpers.NewScheme()).WithObjects(testhelpers.CreateTestVPOObjects()...).Build()
+	cmd, buf, errBuf, _ := createNewTestCommandAndBuffers(t, c)
+	cmd.PersistentFlags().Set(constants.TimeoutFlag, "2s")
```
kminder, line 90: Does this really end up waiting 2 seconds?  If so this isn't a great practice in a unit test.

tools/vz/cmd/install/install_test.go
```diff
@@ -67,6 +75,32 @@ func TestInstallCmdDefaultTimeout(t *testing.T) {
...
+	cmd.Flags().String(constants.GlobalFlagContext, testK8sContext, "")
+	cmd.PersistentFlags().Set(constants.AutoBugReportFlag, "false")
+	cmdHelpers.SetDeleteFunc(cmdHelpers.FakeDeleteFunc)
+	defer cmdHelpers.SetDefaultDeleteFunc()
+
+	// Run install command
+	err := cmd.Execute()
+	assert.Error(t, err)
+	assert.Equal(t, "Error: Timeout 2s exceeded waiting for install to complete\n", errBuf.String())
+	assert.Contains(t, buf.String(), "Installing Verrazzano version v1.3.1")
```
kminder, line 102: Where does this version string come from?  Will/should it change over time?

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
...
+	autoBugReportFlag, errFlag := cmd.Flags().GetBool(constants.AutoBugReportFlag)
+	if errFlag != nil {
+		fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag.Error())
```
kminder, line 259: Should this be continuing or returning the error?

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -49,11 +49,14 @@ vz bug-report --report-file bugreport.tgz --include-namespaces ns1,ns2 --include
 
 const minLineLength = 100
 
+var kubeconfig string
+var context string
+
 func NewCmdBugReport(vzHelper helpers.VZHelper) *cobra.Command {
```
kminder, line 55: All exported functions must have comments.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
...
+	autoBugReportFlag, errFlag := cmd.Flags().GetBool(constants.AutoBugReportFlag)
+	if errFlag != nil {
+		fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag.Error())
+	}
+	// if waitForUpgradeToComplete() returned an err and auto-bug-report is set to true
+	if autoBugReportFlag {
+		cmd2 := NewCmdBugReport(vzHelper)
+		kubeconfigFlag, errFlag := cmd.Flags().GetString(constants.GlobalFlagKubeConfig)
+		if errFlag != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag.Error())
```
kminder, line 266: Should this continue?

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
...
+		}
+		contextFlag, errFlag2 := cmd.Flags().GetString(constants.GlobalFlagContext)
+		if errFlag2 != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag2.Error())
```
kminder, line 270: Should this continue?

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -1,12 +1,13 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package uninstall
 
 import (
 	"bufio"
-	"context"
+	pkgContext "context"
```
kminder, line 8: Is this package alias required?  It is making it harder to see the real changes.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -49,11 +49,14 @@ vz bug-report --report-file bugreport.tgz --include-namespaces ns1,ns2 --include
 
 const minLineLength = 100
 
+var kubeconfig string
+var context string
```
kminder, line 53: Can a more specific name be used here?
At a minimum a comment is required because I have no idea what this string is.
A better name would probably prevent a bunch of package aliasing too.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
...
+		cmd2.Flags().StringVar(&kubeconfig, constants.GlobalFlagKubeConfig, "", constants.GlobalFlagKubeConfigHelp)
+		cmd2.Flags().StringVar(&context, constants.GlobalFlagContext, "", constants.GlobalFlagContextHelp)
+		cmd2.Flags().Set(constants.GlobalFlagKubeConfig, kubeconfigFlag)
+		cmd2.Flags().Set(constants.GlobalFlagContext, contextFlag)
+		bugReportErr := RunCmdBugReport(cmd2, vzHelper)
+		if bugReportErr != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error calling vz bug-report %s \n", bugReportErr.Error())
+		}
+	}
+	return err
```
kminder, line 281: Should comment here that this is intended to return the original passed in error.

tools/vz/cmd/upgrade/upgrade_test.go
```diff
@@ -49,12 +54,44 @@ func TestUpgradeCmdDefaultNoWait(t *testing.T) {
...
+	c := fake.NewClientBuilder().WithScheme(helpers.NewScheme()).WithObjects(append(testhelpers.CreateTestVPOObjects(), vz)...).Build()
+
+	// Send stdout stderr to a byte buffer
+	buf := new(bytes.Buffer)
+	errBuf := new(bytes.Buffer)
+	rc := testhelpers.NewFakeRootCmdContext(genericclioptions.IOStreams{In: os.Stdin, Out: buf, ErrOut: errBuf})
+	rc.SetClient(c)
+	cmd := NewCmdUpgrade(rc)
+	assert.NotNil(t, cmd)
+	cmd.PersistentFlags().Set(constants.TimeoutFlag, "2s")
```
kminder, line 73: How many seconds total do all these tests take?

tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -156,11 +158,12 @@ func runCmdUninstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelpe
 
 	// Wait for the Verrazzano uninstall to complete.
 	err = waitForUninstallToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vz.Namespace, Name: vz.Name}, timeout, vpoTimeout, logFormat, useUninstallJob)
-	if err != nil {
-		return fmt.Errorf("Failed to uninstall Verrazzano: %s", err.Error())
+	if err == nil {
```
mgianatagh, line 161: It may be cleaner for this to be "err != nil" and put lines 164-166 in the block, and return nil as the last line in the function

tools/vz/cmd/upgrade/upgrade_test.go
knelson7: 12 seconds

knelson7: I reduced the Timeout to 2ms, so now it's 8 seconds

tools/vz/cmd/install/install_test.go
knelson7: I reduced it to 2ms instead

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +252,31 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
...
+		if errFlag2 != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag2.Error())
+		}
+		cmd2.Flags().StringVar(&kubeconfig, constants.GlobalFlagKubeConfig, "", constants.GlobalFlagKubeConfigHelp)
+		cmd2.Flags().StringVar(&context, constants.GlobalFlagContext, "", constants.GlobalFlagContextHelp)
+		cmd2.Flags().Set(constants.GlobalFlagKubeConfig, kubeconfigFlag)
+		cmd2.Flags().Set(constants.GlobalFlagContext, contextFlag)
+		bugReportErr := RunCmdBugReport(cmd2, vzHelper)
+		if bugReportErr != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error calling vz bug-report %s \n", bugReportErr.Error())
```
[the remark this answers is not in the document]
knelson7: I changed it to print to error stream. If i return this error, it'll overwrite the original error returned from the install command

tools/vz/cmd/analyze/analyze_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
```
bhatpmk, line 1: There is no change in this file

tools/vz/cmd/bugreport/bugreport.go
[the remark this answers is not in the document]
knelson7: Bhat recommended printing to error stream

tools/vz/cmd/bugreport/bugreport.go
knelson7: Updated the names to be kubeconfigFlagValPointer and contextFlagValPointer since the var points to a string variable in which the value of the flag is stored

tools/vz/cmd/analyze/analyze_test.go
knelson7: I know, but for some reason, it the build fails with a copyright error if I don't change it.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -249,3 +253,34 @@ func isDirEmpty(directory string, ignoreFilesCount int) bool {
 	}
 	return len(entries) == ignoreFilesCount
 }
+
+// creates a new bug-report cobra command, initailizes and sets the required flags, and runs the new command.
+// Returns the original error that's passed in as a parameter to preserve the error received from previous cli command failure.
+func CallVzBugReport(cmd *cobra.Command, vzHelper helpers.VZHelper, err error) error {
+	autoBugReportFlag, errFlag := cmd.Flags().GetBool(constants.AutoBugReportFlag)
```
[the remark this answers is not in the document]
knelson7: What do you mean exactly? 
To call runCmdBugReport we have to create a new cobra command and set up the required flags that it needs. Also, we have to check the autobugreport flag. Because of all this logic, I added a helper method here.

tools/vz/cmd/bugreport/bugreport.go
knelson7: Updated so that it returns after printing the error

tools/vz/cmd/bugreport/bugreport.go
knelson7: Updated so that it returns after printing the error

tools/vz/cmd/bugreport/bugreport.go
knelson7: Updated so that it returns after printing the error to match other cases where string flags can't be found

tools/vz/cmd/bugreport/bugreport.go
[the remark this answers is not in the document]
bhatpmk: The new flag auto-bug-report is added to install/uninstall/upgrade, but bug-report doesn't provide that. The new function CallVzBugReport is called only from other commands, so I was just trying to understand whether there is a need to check whether auto-bug-report is set

absBitm:
@knelson7 could you please address the code smells and duplication

tools/vz/cmd/uninstall/uninstall_test.go
```diff
@@ -24,8 +23,18 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"os"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"testing"
+)
+
+const (
+	testKubeConfig    = "/tmp/kubeconfig"
```
bhatpmk, line 33: You can create a temporary directory, rather than using /tmp/kubeconfig

tools/vz/cmd/uninstall/uninstall_test.go
```diff
@@ -139,7 +155,49 @@ func TestUninstallCmdDefaultTimeout(t *testing.T) {
...
+	vpo := createVpoPod()
+	namespace := createNamespace()
+	vz := createVz()
+	webhook := createWebhook()
+	cr := createClusterRole()
+	crb := createClusterRoleBinding()
+	c := fake.NewClientBuilder().WithScheme(helpers.NewScheme()).WithObjects(deployment, vpo, vz, namespace, webhook, cr, crb).Build()
+
+	// Send stdout stderr to a byte buffer
+	buf := new(bytes.Buffer)
```
bhatpmk, line 180: We had intermittent test failures due to usage of byte buffer, please take a look at the change in https://github.com/verrazzano/verrazzano/pull/5548

tools/vz/cmd/install/install.go
```diff
@@ -88,6 +90,23 @@ func NewCmdInstall(vzHelper helpers.VZHelper) *cobra.Command {
...
+	err := installVerrazzano(cmd, vzHelper)
+	if err != nil {
+		autoBugReportFlag, errFlag := cmd.Flags().GetBool(constants.AutoBugReportFlag)
+		if errFlag != nil {
+			fmt.Fprintf(vzHelper.GetOutputStream(), "Error fetching flags: %s", errFlag.Error())
+			return err
+		}
+		if autoBugReportFlag {
+			//err returned from CallVzBugReport is the same error that's passed in, the error that was returned from installVerrazzano
+			err = bugreport.CallVzBugReport(cmd, vzHelper, err)
```
bhatpmk, line 102: As we make more progress, we should be able to find out whether we need to capture the logs from additional pods in a namespace by utilizing the following flags of bug-report.

  -l, --include-logs                 Include logs from the pods in one or more namespaces; this is specified along with the --include-namespaces flag.
  -i, --include-namespaces strings   A comma-separated list of namespaces, in addition to the ones collected by default (system namespaces), for collecting cluster information. This flag can be specified multiple times, such as --include-namespaces ns1 --include-namespaces ns...

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -26,7 +26,7 @@ import (
 
 const (
 	captureVerrazzanoErrMsg = "Capturing Verrazzano resource"
-
+
```
bhatpmk, line 29: There is no change in this file

bhatpmk (dismissed):
LGTM

bhatpmk (approved):

PR #5465, opened 2023-02-27
VZ-8671 New pipeline to build VZ CLI RPM

ayaranga:


kminder:
From my perspective this has far too much in the Jenkinsfile.  Can there not be a single target in a Makefile that does all these steps?  How would someone go about doing this in their dev env?

ci/cli/Jenkinsfile
```diff
@@ -0,0 +1,211 @@
...
+
+pipeline {
+    options {
+        timeout(time: 1, unit: 'HOURS')
+        skipDefaultCheckout true
+        timestamps()
+    }
+
+    agent {
+        label "${agentLabel}"
```
agv426, line 14: The agent label will need to be computed based on the BUILD_PLAT/BUILD_OS selections.

That will be a followup correct?

ayaranga: Yes, once we have Jenkins agent images ready for other combinations, the agent label code will also be updated accordingly.

PR #5471, opened 2023-02-28
VZ-8153: Making prometheus a-la-carte and remove dependency on NGINX and Authproxy

saidu941:
This PR includes the changes to the prometheus component files
> Conditionally check/update ingresses for prometheus based on NGINX enabled/disabled to make prometheus a-la-carte

### DIFF
--- a/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
+++ b/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
@@ -136,7 +136,7 @@ func postInstallUpgrade(ctx spi.ComponentContext) error {
 		ExtraAnnotations: common.SameSiteCookieAnnotations(prometheusName),
 	}
 
-	if vzcr.IsNGINXEnabled(ctx.EffectiveCR()) && vzcr.IsAuthProxyEnabled(ctx.EffectiveCR()) {
+	if vzcr.IsNGINXEnabled(ctx.EffectiveCR()) {
 		if err := common.CreateOrUpdateSystemComponentIngress(ctx, props); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -135,8 +135,11 @@ func postInstallUpgrade(ctx spi.ComponentContext) error {
 		// Enable sticky sessions, so there is no UI query skew in multi-replica prometheus clusters
 		ExtraAnnotations: common.SameSiteCookieAnnotations(prometheusName),
 	}
-	if err := common.CreateOrUpdateSystemComponentIngress(ctx, props); err != nil {
-		return err
+
+	if vzcr.IsNGINXEnabled(ctx.EffectiveCR()) && vzcr.IsAuthProxyEnabled(ctx.EffectiveCR()) {
```
michael-cico, line 139: I'm guessing that we'll need to replicate this pattern for other components as well; maybe we should have a short-hand `IsIngressEnabled()` call in the `vzcr` package?

or, do we need to check for both, or is just checking for NGINX enough?

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
```diff
@@ -189,6 +189,9 @@ func (c prometheusComponent) GetIngressNames(ctx spi.ComponentContext) []types.N
 
 	if vzcr.IsPrometheusEnabled(ctx.EffectiveCR()) {
 		ns := ComponentNamespace
+		if !vzcr.IsNGINXEnabled(ctx.EffectiveCR()) && !vzcr.IsAuthProxyEnabled(ctx.EffectiveCR()) {
```
michael-cico, line 192: We should add unit tests for these cases also.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
michael-cico, line 139: Add a unit test also

mbenmclean (dismissed):

michael-cico (approved):

PR #5483, opened 2023-03-02
Only register prometheus keycloak client if set by managed cluster

anders-swanson:
This would cycle with errors if managed cluster prometheus is disabled.

### DIFF
--- a/cluster-operator/controllers/vmc/vmc_controller.go
+++ b/cluster-operator/controllers/vmc/vmc_controller.go
@@ -551,6 +551,7 @@ var createClient = func(r *VerrazzanoManagedClusterReconciler, vmc *clustersv1al
 	// Skip Keycloak client generation if Prometheus isn't present in VMC status
 	// MCAgent on the managed cluster will set this if/when it is ready
 	if len(promHost) == 0 {
+		r.log.Debug("Skipping Prometheus Keycloak client creation: VMC Prometheus not found")
 		return nil
 	}
 
--- a/cluster-operator/controllers/vmc/vmc_controller.go
+++ b/cluster-operator/controllers/vmc/vmc_controller.go
@@ -547,6 +547,12 @@ func (r *VerrazzanoManagedClusterReconciler) getVerrazzanoResource() (*v1beta1.V
 // leveraged to replace method (unit testing)
 var createClient = func(r *VerrazzanoManagedClusterReconciler, vmc *clustersv1alpha1.VerrazzanoManagedCluster) error {
 	const prometheusHostPrefix = "prometheus.vmi.system"
+	promHost := vmc.Status.PrometheusHost
+	// Skip Keycloak client generation if Prometheus isn't present in VMC status
+	// MCAgent on the managed cluster will set this if/when it is ready
+	if len(promHost) == 0 {
+		return nil
+	}
 
 	// login to keycloak
 	cfg, cli, err := k8sutil.ClientConfig()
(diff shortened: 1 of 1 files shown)


cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -547,6 +547,12 @@ func (r *VerrazzanoManagedClusterReconciler) getVerrazzanoResource() (*v1beta1.V
 // leveraged to replace method (unit testing)
 var createClient = func(r *VerrazzanoManagedClusterReconciler, vmc *clustersv1alpha1.VerrazzanoManagedCluster) error {
 	const prometheusHostPrefix = "prometheus.vmi.system"
+	promHost := vmc.Status.PrometheusHost
+	// Skip Keycloak client generation if Prometheus isn't present in VMC status
+	// MCAgent on the managed cluster will set this if/when it is ready
+	if len(promHost) == 0 {
+		return nil
```
fredtibbitts, line 554: It might be useful to have a debug log statement here.

fredtibbitts (approved):

cluster-operator/controllers/vmc/vmc_controller.go
anders-swanson: Added debug

jmaron99 (approved):

cluster-operator/controllers/vmc/vmc_controller.go
anders-swanson: it should be reconciled when the status is set by mcagent, but I am currently testing that. Will let you know

desagar: I think I answered my question. https://github.com/verrazzano/verrazzano/pull/5483/files#diff-e43c1f5486c03e16914b7dcf76feb73c30b07dc75db40665e026de932d31cfc5R281

anders-swanson: given mcagent sets this in its own control loop

desagar: line 281 always requeues so hopefully the answer is it will be reconciled.

anders-swanson: > The debug log may print multiple times before then.
Sure, but its debug. So we wont see it usually. Right now this repeatedly prints errors until registration is complete

desagar: I do think that debug log will likely print several times, but since it is only a debug log it's not a major concern.

desagar (approved):

cluster-operator/controllers/vmc/vmc_controller.go
fredtibbitts: It will requeue, but the requeue delay is 3 minutes instead of a few seconds, so the behavior is going to change. I'm not sure what the impact of the longer delay will be.

fredtibbitts (approved):

cluster-operator/controllers/vmc/vmc_controller.go
anders-swanson: The VMC will change when the MCAgent updates it. That itself should trigger a reconcile (testing)

fredtibbitts: I seem to recall that changes to the status subresource do not trigger reconciles, but I can't remember for sure.

PR #5505, opened 2023-03-07
VZ-8757: Reduce uninstall time by 3 or 4 minutes

mgianatagh:
Modify VPO to removing rancher finalizers before running the rancher-cleanup job.  In my local testing and in the pipeline I saw about a 3 or 4 minute decrease in uninstall time.  An incremental step toward reducing the overall time.


platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
```diff
@@ -78,6 +78,8 @@ var postUninstallFunc postUninstallFuncSig = invokeRancherSystemToolAndCleanup
 
 var rancherCleanupJobYamlPath = defaultRancherCleanupJobYaml
 
+var rancherFinalizersDeleted = false
```
bobdonat, line 81: if uninstall is run a second time will the variable be true and finalizers not get deleted?

michael-cico (dismissed):

platform-operator/controllers/verrazzano/component/rancher/rancher_uninstall.go
michael-cico: nice catch, might be good to initialize this to false during pre-uninstall

mgianatagh: good catch, I just pushed a fix

michael-cico (approved):

PR #5509, opened 2023-03-07
VZ 8883: Upgrade webhook error

anders-swanson:
Updates the webhook error on upgrade to be more user friendly.

### DIFF
--- a/platform-operator/apis/verrazzano/validators/validate.go
+++ b/platform-operator/apis/verrazzano/validators/validate.go
@@ -102,7 +102,7 @@ func ValidateVersion(requestedVersion string) error {
 		return err
 	}
 	if !requestedSemVer.IsEqualTo(bomSemVer) {
-		return fmt.Errorf("Invalid version %s: Please use Verrazzano version %s.", requestedSemVer.ToString(), bomSemVer.ToString())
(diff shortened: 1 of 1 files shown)


platform-operator/apis/verrazzano/validators/validate.go
```diff
@@ -102,7 +102,7 @@ func ValidateVersion(requestedVersion string) error {
 		return err
 	}
 	if !requestedSemVer.IsEqualTo(bomSemVer) {
-		return fmt.Errorf("Requested version %s does not match BOM version %s", requestedSemVer.ToString(), bomSemVer.ToString())
+		return fmt.Errorf("invalid version %s: Please use Verrazzano version %s", requestedSemVer.ToString(), bomSemVer.ToString())
```
rosemarymarano, line 105: Any reason why invalid is lowercase (here and in the following error message)?

rosemarymarano (approved):
Otherwise, LGTM.

platform-operator/apis/verrazzano/validators/validate.go
anders-swanson: it's a style constraint for returning these types of error messages. Don't begin sentences with a capitalized letter, and don't end them with a full-stop.

rosemarymarano: @anders-swanson ok, thanks for the explanation.

PR #5510, opened 2023-03-07
VZ-8711 Add sidecar implementation to the Prometheus instance

davisgomes:
- Adds an optional Thanos sidecar to the Prometheus
- Adds ingress to the installation of Prometheus if Thanos sidecar is enabled
- Updates helm overrides to allow default sidecar configuration

**Local testing:**
- Installed with and without the sidecar value set to make sure nothing was installed
- Installed with the sidecar value set and removed it to see if the sidecar became disabled
- Installed with the disabled value and changed it to the sidecar to see if the sidecar got added

### DIFF
--- a/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
+++ b/platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
@@ -43,6 +43,9 @@ const (
 	prometheusHostName        = "prometheus.vmi.system"
 	prometheusCertificateName = "system-tls-prometheus"
 
+	thanosHostName        = "thanos"
+	thanosCertificateName = "monitoring-tls-thanos"
+
 	istioPrometheus = "prometheus-server"
 )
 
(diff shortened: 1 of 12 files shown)


platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -309,7 +309,7 @@ spec:
 {{- if .Values.prometheus.prometheusSpec.priorityClassName }}
   priorityClassName: {{ .Values.prometheus.prometheusSpec.priorityClassName }}
 {{- end }}
-{{- if .Values.prometheus.prometheusSpec.thanos }}
+{{- if and (.Values.prometheus.prometheusSpec.thanos) (eq .Values.prometheus.thanos.integration "sidecar") }}
```
desagar, line 312: if the user doesn't specify a "thanos:" section in the prom spec, but did specify that they want integration type of sidecar, how do we enable the sidecar?

davisgomes: I have a default override in the prometheus component that populates the thanos image value no matter what. This check only populates the thanos spec in the Prometheus CR if sidecar is selected. The user can still override the thanos section if they like.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator_component.go
```diff
@@ -186,40 +189,51 @@ func (c prometheusComponent) ValidateUpdateV1Beta1(old *installv1beta1.Verrazzan
...
+		ns = authproxy.ComponentNamespace
 	}
+	ingressNames = append(ingressNames, types.NamespacedName{
+		Namespace: ns,
+		Name:      constants.PrometheusIngress,
+	})
 
-	return ingressNames
+	if thanosEnabled, err := isThanosEnabled(ctx); !thanosEnabled || err != nil {
+		return ingressNames
```
fredtibbitts, line 206: Should we log an error here if there's an error, since errors aren't bubbled up?

platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/prometheus/prometheus.yaml
fredtibbitts: I think I'm missing something too. If the user just specifies this in the VZ CR:
```
...
prometheus:
  thanos:
    integration: sidecar
...
```
and does not specify any `prometheusSpec` `thanos` yaml, how does the sidecar get enabled?

fredtibbitts (approved):

desagar (dismissed):

fredtibbitts (approved):

PR #5513, opened 2023-03-07
VZ-8764: Use URL trigger to control the periodic and backend pipeline triggers

pshirodkar-orcl:
Use URL trigger to control the periodic and backend pipeline triggers

### DIFF
--- a/ci/JenkinsfileBackendTests
+++ b/ci/JenkinsfileBackendTests
@@ -60,7 +60,7 @@ pipeline {
 
     triggers {
         URLTrigger(
-            cronTabSpec: '*/2 * * * *',
+            cronTabSpec: branchSpecificSchedule,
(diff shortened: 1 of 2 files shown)


ci/JenkinsfileBackendTests
```diff
@@ -54,7 +59,25 @@ pipeline {
     }
 
     triggers {
-        cron(branchSpecificSchedule)
+        URLTrigger(
+            cronTabSpec: branchSpecificSchedule,
```
pshirodkar-orcl, line 63: @agv426: I am keeping the cron schedule of URL Trigger entry to their respective existing schedule for periodic and backend pipelines. Please let me know in case you want to use a different cron schedule for both.

agv426 (approved):

PR #5518, opened 2023-03-08
VZ-8349 use config map or secret for overrides in istio and nginx

AbeHern:
Added logic to allow ConfigMap and Secret overrides and validating any externalIPs present


cluster-operator/controllers/vmc/jaeger.go
```diff
@@ -1,17 +1,17 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package vmc
 
 import (
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 7: imports should be sorted

platform-operator/controllers/verrazzano/component/istio/istio_component_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/validators"
```
bobdonat, line 10: imports should be sorted

platform-operator/controllers/verrazzano/component/istio/istio_install.go
```diff
@@ -6,6 +6,7 @@ package istio
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 9: imports should be sorted

platform-operator/controllers/verrazzano/component/istio/jaeger.go
```diff
@@ -1,19 +1,19 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package istio
 
 import (
 	"bytes"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 9: imports should be sorted

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 10: imports should be sorted

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -6,6 +6,7 @@ package mysql
 import (
 	"context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 9: imports should be sorted

platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -1,11 +1,12 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2021, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package helm
 
 import (
 	ctx "context"
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/common/override"
```
bobdonat, line 9: imports should be sorted

platform-operator/controllers/verrazzano/component/istio/istio_component_test.go
bobdonat, line 10: imports should be sorted

platform-operator/controllers/verrazzano/component/istio/istio_install.go
bobdonat, line 9: imports should be sorted

platform-operator/internal/vzconfig/args.go
```diff
@@ -1,30 +1,51 @@
...
+		value, err := override.ExtractValueFromOverrideString(o, jsonPath)
 		if err != nil {
 			return err
 		}
-		if container := o.Path(jsonPath); container != nil {
+
+		/* This check that the casted value is valid is needed before setting keyPresent to true or t
+		trying to call validateExternalIP
+		*/
+		v := castValuesToString(value)
```
anders-swanson, line 45: Can you check if the typecast was successful? If it fails, return an error

anders-swanson: v might be nil if the typecast fails, in which case its an invalid external ip

platform-operator/internal/vzconfig/args.go
```diff
@@ -50,14 +71,25 @@ func CheckExternalIPsArgs(installArgs []vzapi.InstallArgs, overrides []vzapi.Ove
 // CheckExternalIPsOverridesArgs method goes through the key-value pair to detect the presence of the
 // specific key for the corresponding component that holds the external IP address
 // It also checks whether IP addresses are valid and provided in a List format
-func CheckExternalIPsOverridesArgs(overrides []v1beta1.Overrides, jsonPath, compName string) error {
-	for _, override := range overrides {
-		o, err := gabs.ParseJSON(override.Values.Raw)
+func CheckExternalIPsOverridesArgs(overrides []v1beta1.Overrides, jsonPath, compName string, namespace string) error {
+	scheme := runtime.NewScheme()
```
anders-swanson, line 75: Any way to make some common codes between these functions? Looks largely duplicated

platform-operator/internal/vzconfig/args.go
```diff
@@ -92,8 +138,31 @@ func validateExternalIP(addresses []string, key, compName string) error {
...
+
+func castValuesToString(value interface{}) []string {
+	var values []string
+	switch val := value.(type) {
+	case nil:
+		return nil
+	case string:
+		if value != nil {
+			values = append(values, value.(string))
+			fmt.Println(val)
```
anders-swanson, line 157: remmove println

platform-operator/internal/vzconfig/args_test.go
```diff
@@ -34,18 +44,23 @@ const (
 // THEN return an error, nil otherwise
 func TestCheckExternalIPsArgs(t *testing.T) {
 	asserts := assert.New(t)
+	createFakeTestClient()
```
anders-swanson, line 47: can you use parameterized tests as discussed?

platform-operator/internal/vzconfig/args_test.go
```diff
@@ -20,12 +27,15 @@ const (
 	compName                 = "istio"
 	ExternalIPArg            = "gateways.istio-ingressgateway.externalIPs"
 	specServiceJSONPath      = "spec.components.ingressGateways.0.k8s.service"
-	externalIPJsonPathSuffix = "externalIPs.0"
+	externalIPJsonPathSuffix = "externalIPs"
```
anders-swanson, line 30: why'd this change?

platform-operator/internal/vzconfig/args_test.go
```diff
@@ -73,55 +92,168 @@ func TestCheckExternalIPsOverridesArgs(t *testing.T) {
...
-		return vzapi.Verrazzano{
-			Spec: vzapi.VerrazzanoSpec{
-				Components: vzapi.ComponentSpec{
-					Istio: &vzapi.IstioComponent{},
-				},
-			},
-		}
-	}
+func TestInvalidIPWithConfigMapOverride(t *testing.T) {
+	asserts := assert.New(t)
```
anders-swanson, line 129: these look like candidates for parameterized tests

platform-operator/internal/vzconfig/args_test.go
AbeHern: A could of reasons:

1. validateExternalIPs() in args.go takes a string of IPs to validate but originally only every validated the first IP given to it in said list. I thought this was weird so changed it. 
2. To actually have more than one IP passed to validateExternalIPs() I changed the path used to access the IPs to grab the entire array and not just the first one

platform-operator/internal/vzconfig/args.go
```diff
@@ -1,30 +1,51 @@
...
+	c, err := GetControllerRuntimeClient(scheme)
+	if err != nil {
+		return err
+	}
+	overrideYAMLs, err := override.GetInstallOverridesYAMLUsingClient(c, v1beta1Overrides, namespace)
+	if err != nil {
+		return err
+	}
+	for _, o := range overrideYAMLs {
+		keyPresent = false
```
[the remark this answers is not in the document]
AbeHern: I think the only reason we set it false here is so that it's reset for each call to validateExternalIPs()?

platform-operator/internal/vzconfig/args_test.go
bobdonat: Seems like a nice to have. Probably shouldn't hold up the merge.

platform-operator/internal/vzconfig/args.go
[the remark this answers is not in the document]
bobdonat: I was thinking of the case where there are two keys. First time thru the loop the key is present but second time it is not. The value of keyPresent will be false. That does not seem correct.

platform-operator/internal/vzconfig/args.go
AbeHern: You're right if it find nil it will return nil, but the check right after stops it from trying to validate the IP and it also doesn't set keyPresent to true which outputs the expected error at line [66](https://github.com/verrazzano/verrazzano/blob/abehern/vz-8349-Use-configMap-or-secret-for-overrides-in-istio-and-nginx/platform-operator/internal/vzconfig/args.go#L65)

platform-operator/internal/vzconfig/args.go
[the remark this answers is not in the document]
AbeHern: Okay I should remove it altogether then

platform-operator/internal/vzconfig/args.go
anders-swanson: then the error is "key not found"? Should be invalid data type, expected array of strings or some such message

AbeHern: So then should I change it so an invalid data type is caught instead of what's at line 66?

bobdonat: I think @anders-swanson is suggesting you return an error if castValuesToString returns nil. Right now if castValuesToString returns nil it just goes on to the next override.  What would cause castValuesToString to return nil.  Is nil considered an error?

AbeHern: nil is returned when any of the externalIPs are nil or null. The if statement that follows catches that what was returned was nil and doesn't set keyPresent to true so that the keyPresent error can be returned later

bobdonat: If nil is not an error then I think the code is ok as is

anders-swanson (approved):
LGTM

bobdonat (approved):

PR #5521, opened 2023-03-09
VZ-8903 FIX : Unit test TestInvalidIssues always fails locally on second run

absBitm:


tools/vz/pkg/analysis/internal/util/report/report_test.go
```diff
@@ -76,6 +76,8 @@ func TestInvalidIssues(t *testing.T) {
 	errBuf := new(bytes.Buffer)
 	rc := help.NewFakeRootCmdContext(genericclioptions.IOStreams{In: os.Stdin, Out: buf, ErrOut: errBuf})
 	assert.NoError(t, GenerateHumanReport(logger, "report", constants.SummaryReport, true, true, true, 8, 8, rc))
```
bhatpmk, line 78: You can have a deferred anonymous function to delete the file.

absBitm: I guess defer isn't required here as we are asserting the file to exist otherwise test will fail

bhatpmk (approved):

PR #5542, opened 2023-03-10
NO JIRA - Extend install timeout to 45 from 30 in tests/e2e/config/sc…

bryantidd:
…ripts/wait-for-verrazzano-install.sh


tests/e2e/config/scripts/wait-for-verrazzano-install.sh
```diff
@@ -21,7 +21,7 @@ while [[ $retval_success -ne 0 ]] && [[ $retval_failed -ne 0 ]]  && [[ $i -lt 30
   i=$((i+1))
 done
 
-if [[ $retval_failed -eq 0 ]] || [[ $i -eq 30 ]] ; then
```
pfmackin, line 24: need to change line 15 also

pfmackin (approved):

PR #5549, opened 2023-03-13
VZ-8854: Update VZ CR to add OpensearchOperator component

saketmht:
This PR includes:
- Update to v1beta1 and v1alpha1 CR to include OpensearchOperator as a new component.
- Update to manifests, CRD.
- Add conversion from v1beta to v1alpha1 and vice-versa for the component.


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -431,6 +435,30 @@ type ComponentSpec struct {
 	WebLogicOperator *WebLogicOperatorComponent `json:"weblogicOperator,omitempty"`
 }
 
+// OpenSearchOperatorComponent specifies the OpenSearchOperator configuration
+type OpenSearchOperatorComponent struct {
```
adiforluls, line 439: Is this not a helm component? If it is then add `InstallOverrides` in the component spec

platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
```diff
@@ -421,6 +425,30 @@ type ComponentSpec struct {
 	WebLogicOperator *WebLogicOperatorComponent `json:"weblogicOperator,omitempty"`
 }
 
+// OpenSearchOperatorComponent specifies the OpenSearchOperator configuration
+type OpenSearchOperatorComponent struct {
```
adiforluls, line 429: Same comment about `InstallOverrides`

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: I think even if it's not a `HelmComponent`, it should support `InstallOverrides`.  The overrides are component specific, so ideally the YAML will correspond to whatever the underlying operator/component supports for configuration.

Istio is an example of this, we use `istioctl` to manage our Istio instance and support overrides for that.

gaurav-05: I dont think we have any usecase for overriding operator defaults but thinking about it seems would be better to add.

adiforluls (approved):
LGTM but wait for Kevin and others to also review.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -431,6 +435,37 @@ type ComponentSpec struct {
...
+	// If true, then OpenSearchOperator will be installed
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+	// List of Overrides for the default `values.yaml` file for the component Helm chart. Overrides are merged together,
+	// but in the event of conflicting fields, the last override in the list takes precedence over any others. You can
+	// find all possible values
+	// TODO: Add a link to the charts/values.yaml
```
bobdonat, line 446: TODO?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -431,6 +435,37 @@ type ComponentSpec struct {
 	WebLogicOperator *WebLogicOperatorComponent `json:"weblogicOperator,omitempty"`
 }
 
+// OpenSearchOperatorComponent specifies the OpenSearchOperator configuration
```
bobdonat, line 438: Since this is an API change this has to be reviewed by API review team

platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
```diff
@@ -421,6 +425,37 @@ type ComponentSpec struct {
...
+	// If true, then OpenSearchOperator will be installed
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+	// List of Overrides for the default `values.yaml` file for the component Helm chart. Overrides are merged together,
+	// but in the event of conflicting fields, the last override in the list takes precedence over any others. You can
+	// find all possible values
+	// TODO: Add a link to the charts/values.yaml
```
bobdonat, line 436: Same comment re: TODO

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: Agreed, with operators there's generally not much to configure.   But without the hooks as you suggest users will have no ability to override if they need/want to.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
saketmht: The link to the values.yaml is added here. Like platform-operator/thirdparty/charts/... which I have yet to add for the operator. Will update this in the next PR when I add the operator component code

PR #5553, opened 2023-03-13
VZ-8827: Refactor Auth.lua to move hardcoded ingress/service to environment variables.

karan56625:
- Move hardcoded service-ingress to environment variables so that we don't need to add code whenever we introduce any new ingress or modify existing ing/service.
- After this PR, if developer need to change the ingress or service name, they can just update the environment variables in `verrazzano-authproxy-env-config` configmap. And if developer needs to add new ingress, create new env in `verrazzano-authproxy-env-config` and update the mapping table `ingressServiceTable`


platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -493,53 +526,36 @@ data:
...
-        elseif backend == 'kiali' then
-            serverUrl = me.makeVmiBackendUrl(backend, '20001')
-        elseif backend == 'jaeger' then
-            serverUrl = me.makeMonitoringComponentBackendUrl('jaeger-operator-jaeger-query', '16686')
-        else
-            me.not_found("Invalid backend name '"..backend.."'")
+    function me.getBackedServerURLFromIngressHost(ingressHost)
+        local backendServerUrl = nil
+        local ingressServiceTable = {}
+        me.setTableValue(ingressServiceTable, os.getenv("VZ_VERRAZZANO_HOST"), me.getLocalKubernetesApiUrl())
```
lake-of-dreams, line 537: so everytime this function is called, we are going to repopulate this table? May be we could init the table only one?

karan56625: you mean a global table?? But Still, after every request, that table will be loaded every time

lake-of-dreams: why can't the data in the configmap be structured in a way that removes the need of extra steps of first setting these in env variables, then in a table, only to be retrieved on a single key everytime. Could the cm be read as a single file in the pod with key as `host` and value as corresponding `url`?

karan56625: initially, we we considered that approach, but there was some blocker. I am not able to recall the exact issue. Ok, let me reconsider the approach, will update you.

lake-of-dreams: another way is to load the data as a json and then use cjson which we already use

pfmackin: I have the same comment as Ravi

pfmackin: Also, the PR comment says that hard coding is removed, but that is what is happening below, env vars are hard coded.

pfmackin (dismissed):

PR #5570, opened 2023-03-14
Additional pod and container security context settings for Thanos components

fredtibbitts:
Out of the box, the Thanos components include some pod and container security context settings. This PR adds more settings to make the pods and containers more secure, and makes the security configuration consistent with other Verrazzano components.

To verify the changes, I ran the pod security e2e test which reports missing security context on pods and containers on a cluster with Thanos installed (including the Prometheus sidecar). After these changes, the test passes. I also ran an internal diagnostic tool against the verrazzano-monitoring namespace, and it reported no issues.

I ran one additional validation by labeling the verrazzano-monitoring namespace with `"pod-security.kubernetes.io/enforce"=restricted`. That also reported no issues.

### DIFF
--- a/platform-operator/helm_config/overrides/thanos-values.yaml
+++ b/platform-operator/helm_config/overrides/thanos-values.yaml
@@ -2,6 +2,65 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 query:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
+  containerSecurityContext:
+    privileged: false
+    capabilities:
+      drop:
+        - ALL
+
   # Adds the Prometheus Thanos sidecar as a Store API endpoint to Query
   stores:
     - prometheus-operator-kube-p-prometheus:10901
+
+queryFrontend:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
+  containerSecurityContext:
+    privileged: false
+    capabilities:
+      drop:
+        - ALL
+
+compactor:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
+  containerSecurityContext:
+    privileged: false
+    capabilities:
+      drop:
+        - ALL
+
+storegateway:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
+  containerSecurityContext:
+    privileged: false
+    capabilities:
+      drop:
+        - ALL
+
+ruler:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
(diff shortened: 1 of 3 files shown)


platform-operator/helm_config/overrides/thanos-values.yaml
```diff
@@ -2,6 +2,65 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 query:
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
```
desagar, line 7: Is the seccompprofile RuntimeDefault valid on all K8S versions we support? This page says it's a beta in v1.25 if I understand right: https://kubernetes.io/docs/tutorials/security/seccomp/

desagar (approved):

platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -301,6 +301,11 @@ spec:
 {{- if .Values.prometheus.prometheusSpec.containers }}
   containers:
 {{ toYaml .Values.prometheus.prometheusSpec.containers | indent 4 }}
+{{- if and (.Values.prometheus.prometheusSpec.thanos) (eq .Values.prometheus.thanos.integration "sidecar") }}
+    - name: thanos-sidecar
```
davisgomes, line 305: Are you able to use a helm value for the name of this sidecar?

platform-operator/helm_config/overrides/thanos-values.yaml
fredtibbitts: This is what we set for all pods. If that were an issue I'd think it would have shown up in other places, but I don't know the answer to your question.

platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/prometheus/prometheus.yaml
fredtibbitts: The Prometheus Operator doc says there are three legal values for the container name:

> The current container names are: prometheus, config-reloader, and thanos-sidecar

So I can make it a helm value but if the value is changed then this will break. Or are you saying there's already a helm setting for this I should use?

platform-operator/helm_config/overrides/thanos-values.yaml
fredtibbitts: @pfmackin do you know the answer to Deva's question? The doc does seem to indicate that it's new in 1.25, but we use it all over the place.

PR #5576, opened 2023-03-14
Move NetworkPolicy from Prometheus Operator install component to Helm chart

fredtibbitts:
This PR moves the NetworkPolicy that was created programmatically into the component's Helm chart. My first attempt at this failed in the upgrade tests. When the kube-prometheus-stack helm chart was installed, it failed because the NetworkPolicy already existed in the verrazzano-monitoring namespace, and since it wasn't owned by the Helm release, the Helm chart failed to install.

To fix the upgrade issue, I added a function to delete the existing NetworkPolicy, which is now run in the pre-upgrade step.

To test this, I captured the existing NetworkPolicy from an older cluster and compared it to the new one installed wih the Helm chart. The resources were identical (aside from the UID, creation timestamp, etc.)


platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/prometheus/networkpolicy.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+apiVersion: networking.k8s.io/v1
```
davisgomes, line 2: This is under the `templates/prometheus` directory. Should it be placed in `templates/verrazzano` as we've done in Thanos?

fredtibbitts: Thanks, I moved it.

davisgomes (approved):

PR #5579, opened 2023-03-14
VZ-8959 add logging to wait4webhook.sh 

pfmackin:
Add logging to wait4webhook.sh so we can get error details when it fails.
Cherry-pick of https://github.com/verrazzano/verrazzano/commit/8ccdf62e6e3e925330bef5b3b3b6f7d94ee38687


pfmackin:
Not merging, still seeing issues

PR #5586, opened 2023-03-15
VZ-8895 Update the availability check and improve comments

davisgomes:
Small change to fix the availability check and improve the comments in the Verrazzano API

### DIFF
--- a/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
@@ -150,6 +150,7 @@ type InstanceInfo struct {
 	// The Rancher URL for this Verrazzano installation.
 	RancherURL *string `json:"rancherUrl,omitempty"`
 	// The Thanos Query URL for this Verrazzano installation.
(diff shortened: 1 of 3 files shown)


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -150,6 +150,7 @@ type InstanceInfo struct {
 	// The Rancher URL for this Verrazzano installation.
 	RancherURL *string `json:"rancherUrl,omitempty"`
 	// The Thanos Query URL for this Verrazzano installation.
+	// The Thanos Query ingress gets forwarded to the Thanos Query Frontend service.
```
fredtibbitts, line 153: Should we leave this out (or make the text a little more vague) so that if in the future we use the same ingress for either Query or Query Frontend then we don't have to change it again?

desagar: I don't think this ends up in the CRD today. So, I asked Davis for this change - we should talk :)

fredtibbitts (approved):

PR #5592, opened 2023-03-16
VZ-8861 bug report changes

absBitm:
- bug report temp file generations on default cases to avoid report generation failure if report file already exists
- bug report generation on every run
- bug report file naming format changes to vz-bug-report-datetime-XXXX.tar.gz where datetime = yyyymmddhhmmss

### DIFF
--- a/tools/vz/pkg/constants/constants.go
+++ b/tools/vz/pkg/constants/constants.go
@@ -9,55 +9,43 @@ const GlobalFlagKubeConfigHelp = "Path to the kubeconfig file to use"
 
 // GlobalFlagContext - global flag for specifying which kube config context to use
 const GlobalFlagContext = "context"
-const GlobalFlagContextHelp = "The name of the kubeconfig context to use"
-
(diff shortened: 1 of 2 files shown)


tools/vz/pkg/constants/constants.go
```diff
@@ -138,7 +126,7 @@ const (
 	BugReportFileFlagValue    = ""
 	BugReportFileFlagShort    = "r"
 	BugReportFileFlagUsage    = "The report file created by the vz bug-report command, as a *.tar.gz file. Defaults to bug-report.tar.gz in the current directory."
```
bhatpmk, line 128: A change is required for the default name of the bug report

absBitm: done

agv426 (approved):

absBitm:
referencing the changes [here](https://github.com/verrazzano/verrazzano/pull/5665)

PR #5599, opened 2023-03-16
added none profile to the distribution

saidu941:


bhatpmk (approved):

release/docs/README_FULL.md
```diff
@@ -40,7 +40,8 @@ The Verrazzano distribution includes the following artifacts:
     * `profiles/`
       * `dev.yaml`: The standard `dev` profile to install Verrazzano.
       * `prod.yaml`: The standard `prod` profile to install Verrazzano.
-      * `managed-cluster.yaml`: The standard `managed-cluster` profile to install Verrazzano 
+      * `managed-cluster.yaml`: The standard `managed-cluster` profile to install Verrazzano
+      * `none.yaml`: The standard `none` profile to install Verrazzano
```
michael-cico, line 44: I would have this description be "A standard profile with all Verrazzano components disabled by default"

michael-cico: or something along those lines

saidu941: I liked your suggestion, but there was no good description for prod/dev/managed-cluster above, so was imitating the same for none

saidu941: @bhatpmk do you think we need someone's review on README related changes?

michael-cico: I think let's just go with this as-is.  If we want to change the documentation it can be a follow-up

bhatpmk: Agree

michael-cico (approved):

PR #5601, opened 2023-03-17
VZ-7460 Move to Flexible Load Balancer Usages

stevshap:
This PR changes the annotations on all Verrazzano services currently using fixed-shape LBs to use the `flexible` shape instead. This is due to OCI deprecating fixed-shape LBs in May 2023.


tests/e2e/config/scripts/create_oke_multi_cluster.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
bobdonat, line 3: shouldn't this be 2023

PR #5606, opened 2023-03-17
VZ-7916 dev component controller

mbenmclean:
This adds a controller for managing the lifecycle of dev components. This first pass will reconcile configmaps with the verrazzano.io/devComponent annotation and will use the data to install dev components.

Compatible ConfigMap:
```
apiVersion: v1
data:
  chartPath: coherence-operator
  namespace: verrazzano-system
  overrides: |-
    defaultCoherenceImage: ghcr.io/oracle/coherence-ce:22.06.1
    replicas: 1
kind: ConfigMap
metadata:
  annotations:
    verrazzano.io/devComponent: coherence
  name: dev-coherence
  namespace: default
```


platform-operator/constants/constants.go
```diff
@@ -152,12 +155,22 @@ const VerrazzanoVersion1_4_0 = "1.4.0"
...
+const VerrazzanoVersion1_6_0 = "1.6.0"
+
 // UpgradeRetryVersion is the restart version annotation field
 const UpgradeRetryVersion = "verrazzano.io/upgrade-retry-version"
 
 // ObservedUpgradeRetryVersion is the previous restart version annotation field
 const ObservedUpgradeRetryVersion = "verrazzano.io/observed-upgrade-retry-version"
 
+// VerrazzanoDevComponentAnnotationName is the name of the annotation to be used on a ConfigMap,
+// which indicates that the ConfigMap is a VZ stack configuration. The value of the annotation is
```
michael-cico, line 168: stack reference

platform-operator/constants/constants.go
```diff
@@ -152,12 +155,22 @@ const VerrazzanoVersion1_4_0 = "1.4.0"
...
+// the VZ stack name e.g. "monitoring"
+const VerrazzanoDevComponentAnnotationName = "verrazzano.io/devComponent"
```
michael-cico, line 170: I would just call it `component`, not `devcomponent`.

platform-operator/constants/constants.go
```diff
@@ -152,12 +155,22 @@ const VerrazzanoVersion1_4_0 = "1.4.0"
...
+
+const VerrazzanoDevComponentChartAnnotationName = "verrazzano.io/devComponentChart"
```
michael-cico, line 172: same comment, just `component-chart`

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	return true
+}
+
+// Reconcile fucntion for the ComponentConfigMapReconciler
+func (r *ComponentConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	verrazzanos := &vzapi.VerrazzanoList{}
```
michael-cico, line 80: nit, I'd move this to a lookup method to make the reconcile method cleaner.

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := r.List(ctx, verrazzanos); err != nil {
+		if k8serrors.IsNotFound(err) || len(verrazzanos.Items) == 0 {
+			return ctrl.Result{}, nil
+		}
+		zap.S().Errorf("Failed to get Verrazzanos %s/%s", req.Namespace, req.Name)
+		return vzctrl.NewRequeueWithDelay(2, 3, time.Second), err
+	}
+	vz := verrazzanos.Items[0]
```
michael-cico, line 88: I would add a length check guard and report an error like 

```
No Verrazzano instance found in the cluster
```

or something like that.  I'm also wondering if we should annotate the configmap with a reference to the VZ instance so we an just do a Get() on it.  Maybe 6 of 1, 1/2 dozen the other.

michael-cico: Actually, the length check is nested in the error check, I'm not sure it will be hit

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	compCtx, err := spi.NewContext(log, r.Client, &vz, nil, false)
+	if err != nil {
+		log.Errorf("Failed to create context: %v", err)
+	}
+
+	// install dev component
+	// TODO: turn this into a state machine
+	err = comp.Install(compCtx)
```
michael-cico, line 136: When the time comes for this, you can use the other POC as a guide

https://github.com/verrazzano/verrazzano/blob/mcico/v1beta2-poc/platform-operator/controllers/module/reconciler/module_reconciler.go#L40

https://github.com/verrazzano/verrazzano/blob/mcico/v1beta2-poc/platform-operator/controllers/module/reconciler/module_reconciler.go#L103

These are pretty much the same as the original module POC, maybe with a few tweaks to support uninstall

michael-cico:
General comment, maybe organize the configmap family of controllers in a hierarchical fashion, e.g., 

```
controllers/configmap/component
controllers/configmap/overrides
```

platform-operator/constants/constants.go
```diff
@@ -14,6 +14,9 @@ const VerrazzanoInstallNamespace = "verrazzano-install"
 // VerrazzanoMonitoringNamespace is the namespace for monitoring components
 const VerrazzanoMonitoringNamespace = "verrazzano-monitoring"
 
+// VerrazzanoMonitoringStackNamespace is the namespace for the monitoring stack
+const VerrazzanoMonitoringStackNamespace = "verrazzano-monitoring-stack"
```
kminder, line 18: Will the need for this be going away in a near-term second pass?

platform-operator/constants/constants.go
kminder: Don't hold getting anything merged but I'd like you to discuss with Paul and Mike C a more generic approach to labeling/annotating ConfigMaps to declare the CRD apiVersion and Kind they are "pretending" to be.

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
```
kminder, line 28: It would be ideal if there was actually a golang type that the ConfigMap content would be marshaled into.  Am I just missing that?  The idea is that any time this ConfigMap approach is taken to prototype a CRD that it should be as small a change as possible to switch to a real CRD.

michael-cico: You don't mean having a single controller for all ConfigMap-driven controller impls right?  i.e. we should be able to have different controller impls driven off of different configmap "types"?

That being said we could probably have the controller take a delegate reconciler and a Golang type and the only job of the configmap controller would be to unmarshal a payload and invoke the delegate.  Is that what you are after?

platform-operator/constants/constants.go
desagar: It doesn't appear to be use. Might have been in a cherry pick of one of my commits. It should be removed.

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
desagar: looks like a case that was missed. There should be a length check in case VZ is not installed. I believe list operations don't return an error if there are zero items found.

michael-cico: Yeah, right now it likely would never be hit, although for folks using this for WIP components may instantiate them without a VZ instance.

platform-operator/controllers/component_configmaps/dev_component.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if deployed, _ := helm.IsReleaseDeployed(h.ReleaseName, h.ChartNamespace); deployed {
+		return true
+	}
+	return false
+}
+
+// InitThirdPartyDirPath initializes the third party chart path from main.go
+func InitThirdPartyDirPath() {
```
desagar, line 83: why was this function needed? I would imagine newDevComponent could just call config.GetThirdPartyDir() like we do elsewhere.

platform-operator/constants/constants.go
mbenmclean: fixed

platform-operator/constants/constants.go
mbenmclean: I thought about this, the only thing is we install components now through the VPO registry, are going to install components through this component controller, and are working on another way to install components through our v1beta2 API. I'm not sure we want to use verrazzano.io/component for this given this is a dev feature and that's a pretty generic annotation. @kminder what do you think? This feels like something the ARB should weigh in on so don't necessarily need to decide now.

platform-operator/constants/constants.go
mbenmclean: Yes I think we need a discussion about the best way to label/annotate these configmaps

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
mbenmclean: There's a return in the embedded if statement so not sure there's a great way to do this. Leaving for now

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
mbenmclean: Reworked it. Don't think we want to rely on annotations to the configmap because that would require the user to do it

platform-operator/controllers/component_configmaps/component_configmaps_controller.go
mbenmclean: yep this is just a first pass

platform-operator/constants/constants.go
mbenmclean: Nope just forgot to clean up the constants file. Removed

PR #5608, opened 2023-03-20
MC test, extend Jaeger pod wait timeout

pfmackin:
The MC test failed because the Jaeger pod wasn't ready after 5 minutes, it was ready around 6 minutes after the OAM app was created.  Extend the timeout from 5 min to 10 min so that this doesn't fail, the failure has happened a few times.


stevshap (approved):
LGTM

PR #5609, opened 2023-03-20
Updating development version from 1.5.2 to 1.5.3

stevshap:
Updating development version from 1.5.2 to 1.5.3


bryantidd:
What about .verrazzano_developerment_version

PR #5621, opened 2023-03-21
VZ-9063 update kindest images

pfmackin:
Update the kindest images used for tests


mgianatagh (approved):

pfmackin:
These images have a issue where fluentd cannot read the systemd logs

PR #5625, opened 2023-03-22
JIRA: VZ-8333 Modify Verrazzano to leverage Helm libs rather than Helm CLI

jmaron99:
The PR has modified the helmcli.go file to leverage the Helm libs rather than the CLI.  Interactions/API from other layers remain the same for the time being (though I can envision further improvements in terms of modularity and maintainability in the future).  Most other changes involve updates to unit tests to leverage the helm API rather than override/stub out functionality.

Some specific topics for review:
- Should all components be interacting with helm layer during installation etc?  During unit test modifications discovered that some components don't actually interact with helm (though functions were stubbed out).  Primarily a discussion of helmcli.GetChartStatus and it's handling of the "no chart found" condition
- Level of logging?  During registration of Helm action config a log method is specified - should that be "info" or "debug" level logger?
- stdout/stderr returns from helm commands?


pkg/helm/helmcli.go
```diff
@@ -42,207 +44,142 @@ type HelmOverrides struct {
...
 
-	stdout, stderr, err = runHelm(log, releaseName, namespace, chartDir, "upgrade", wait, args, dryRun)
+	p := getter.All(settings)
+	vals, err := mergeValues(overrides, p)
 	if err != nil {
-		return stdout, stderr, err
+		return err
+	}
+	// load chart from the path
+	chart, err := loadChartFn(chartDir)
```
michael-cico, line 120: Should we load the chart closer to where we use it?  Looks like we don't use it until further down, and we may not need to load it if some of the checks in between fail.

jmaron99: I could.  Generally I didn't change the ordering or general structure of the methods so that is probably in the "equivalent" location to where it was previously.  But if you feel that this would be an improvement I could move it.

pkg/helm/helmcli.go
```diff
@@ -428,48 +351,119 @@ func getReleaseAppVersion(releaseName string, namespace string) (string, error)
...
+		return nil, err
+	}
+	return data.Bytes(), err
 }
 
-// SetDefaultRunner sets the command runner to default
-func SetDefaultRunner() {
-	runner = vzos.DefaultRunner{}
+// mergeValues merges values from the specified overrides
+func mergeValues(overrides []HelmOverrides, p getter.Providers) (map[string]interface{}, error) {
```
michael-cico, line 419: So the API doesn't do this for us huh?  Bummer.

jmaron99: Agreed

pkg/helm/helmcli.go
michael-cico: nah it's fine.

michael-cico (approved):

PR #5627, opened 2023-03-22
Skip OCI Node driver activation aamitra/rancher phase1

theabmitra:


platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -506,17 +506,7 @@ func (r rancherComponent) PostUpgrade(ctx spi.ComponentContext) error {
 
 // activateDrivers activates the nodeDriver oci and oraclecontainerengine kontainerDriver
 func activateDrivers(log vzlog.VerrazzanoLogger, c client.Client) error {
-	err := activateOCIDriver(log, c)
```
anders-swanson, line 509: Can you get rid of this activeOCIDriver method too?

theabmitra: done

PR #5634, opened 2023-03-23
VZ-9096: Disable startupapicheck in cert-manager

lake-of-dreams:
In cert-manager 1.9.1, `startupapicheck` job is enabled by default and pulls an image quay.io/jetstack/cert-manager-ctl:v1.9.1 which we don't distribute and hence will cause issues in air-gapped installs. This PR to disable this job till we start distributing the image.

### DIFF
--- a/platform-operator/thirdparty/charts/cert-manager/values.yaml
+++ b/platform-operator/thirdparty/charts/cert-manager/values.yaml
@@ -453,7 +453,7 @@ cainjector:
 # due to the Job never being completed because the sidecar proxy does not exit.
 # See https://github.com/cert-manager/cert-manager/pull/4414 for context.
 startupapicheck:
-  enabled: false
+  enabled: true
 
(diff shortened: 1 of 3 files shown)


platform-operator/thirdparty/charts/cert-manager/values.yaml
```diff
@@ -453,7 +453,7 @@ cainjector:
 # due to the Job never being completed because the sidecar proxy does not exit.
 # See https://github.com/cert-manager/cert-manager/pull/4414 for context.
 startupapicheck:
-  enabled: true
+  enabled: false
```
bhatpmk, line 456: Usually this kind of changes are done in the Helm overrides for cert-manager: platform-operator/helm_config/overrides/cert-manager-values.yaml

lake-of-dreams: fixed

PR #5641, opened 2023-03-24
VZ-8725: update OS/OSD Image to add Alerting plugin

karan56625:
update OS/OSD Image to add Alerting plugin

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -244,12 +244,12 @@
             },
             {
               "image": "opensearch",
-              "tag": "2.3.0-20230210125230-1ff5dd09b32",
+              "tag": "2.3.0-20230324071613-ae6dc07e5dd",
               "helmFullImageKey": "monitoringOperator.osImage"
             },
             {
               "image": "opensearch-dashboards",
-              "tag": "2.3.0-20230209072030-f9e6353395",
+              "tag": "2.3.0-20230324072109-f9e6353395",
               "helmFullImageKey": "monitoringOperator.osdImage"
             },
             {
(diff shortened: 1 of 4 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -244,12 +244,12 @@
             },
             {
               "image": "opensearch",
-              "tag": "2.3.0-20230210125230-1ff5dd09b32",
+              "tag": "2.3.0-20230324071613-ae6dc07e5dd",
```
adiforluls, line 247: There's a small test for OS plugins https://github.com/verrazzano/verrazzano/blob/16cb41eda29f416e3484fed2ff743c3b34f057ae/tests/e2e/upgrade/post-upgrade/opensearch/opensearch_test.go#L184
You can update this test to include the new plugin.

karan56625: Is there any e2e for OSD too? Also, does test cover all version of OS or OS >= 2.3.0?

adiforluls: Don't see one for OSD. Instead of checking for OS version, the check is placed for minimum verrazzano version that has the plugins, so maybe update that to 1.6.0.

adiforluls: It might be a good idea to add this for OSD, but that shouldn't hold up merge. Maybe add the test in a followup if this is urgent.

adiforluls (approved):

platform-operator/verrazzano-bom.json
karan56625: added for OS in this PR.

neeraj-n-singh (approved):
LGTM

PR #5658, opened 2023-03-27
PR to add test for WebLogic cluster app and to add in examples Jenkins

vanajamukkara:


bryantidd (commented):
Why is the istio sidecar set to debug for logging in the test?

bryantidd (commented):
Why is the port number changed? Is that required?

vanajamukkara:
> Why is the istio sidecar set to debug for logging in the test?

Used it for testing, can be removed now

vanajamukkara:
> Why is the port number changed? Is that required?

I remember seeing port conflict issues while testing.

vanajamukkara:
> > Why is the istio sidecar set to debug for logging in the test?
> 
> Used it for testing, can be removed now

removed

vanajamukkara:
> > Why is the port number changed? Is that required?
> 
> I remember seeing port conflict issues while testing.

Changed port back and tested.

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
```diff
@@ -0,0 +1,471 @@
...
+	"github.com/verrazzano/verrazzano/tests/e2e/pkg/test/framework/metrics"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+const (
+	shortWaitTimeout         = 15 * time.Minute
+	shortPollingInterval     = 10 * time.Second
+	longPollingInterval      = 20 * time.Second
+	imagePullWaitTimeout     = 50 * time.Minute
```
pfmackin, line 27: 50 min seems very long.  What is the longest image pull you saw?

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
```diff
@@ -0,0 +1,471 @@
...
+		metrics.Emit(t.Metrics.With("deployment_elapsed_time", time.Since(start).Milliseconds()))
+
+		t.Logs.Info("Container image pull check")
+		Eventually(func() bool {
+			return pkg.ContainerImagePullWait(namespace, expectedPods)
+		}, imagePullWaitTimeout, imagePullPollingInterval).Should(BeTrue())
+	}
+
+	t.Logs.Info("WebLogic Application: check expected admin server pod is running")
+	Eventually(func() bool {
```
pfmackin, line 66: This should be changed to Eventually(func() (bool, error). {

Same is true for any eventually that returns bool, but might get an error.

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
```diff
@@ -0,0 +1,471 @@
...
+
+var (
+	t                  = framework.NewTestFramework("weblogicworkload")
+	generatedNamespace = pkg.GenerateNamespace("hello-wls")
+	expectedPods       = []string{wlsAdminServer, wlsManagedServer1}
+	host               = ""
+)
+
+var beforeSuite = t.BeforeSuiteFunc(func() {
+	if !skipDeploy {
```
pfmackin, line 54: Can you move this code into a Describe block instead of the BeforeSuite.  If the BeforeSuite fails, we don't get a test failure.

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
```diff
@@ -0,0 +1,471 @@
...
+		if err != nil {
+			AbortSuite(fmt.Sprintf("WebLogic Secret %s does not exist in the namespace: %v, error: %v", helloDomainRepoCreds, namespace, err))
+		}
+		return result
+	}, shortWaitTimeout, longPollingInterval).Should(BeTrue(), "Failed to deploy the WebLogic Application: Secret does not exist")
+
+	var err error
+	// Get the host from the Istio gateway resource.
+	start := time.Now()
+	t.Logs.Info("WebLogic Application: check expected Gateway is ready")
```
pfmackin, line 130: Please get rid of all these t.Logs.Info messages, they just may the console output more verbose.

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
```diff
@@ -0,0 +1,471 @@
...
+
+		t.Logs.Info("Container image pull check")
+		Eventually(func() bool {
+			return pkg.ContainerImagePullWait(namespace, expectedPods)
+		}, imagePullWaitTimeout, imagePullPollingInterval).Should(BeTrue())
+	}
+
+	t.Logs.Info("WebLogic Application: check expected admin server pod is running")
+	Eventually(func() bool {
+		result, err := pkg.PodsRunning(namespace, []string{wlsAdminServer})
```
mbenmclean, line 67: Why do we need this? If deployWebLogicApp checks the pods are running we don't need this. If it doesn't then we don't want to AbortSuite

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
vanajamukkara: I agree. imagePullWaitTimeout was increased while testing on old Jenkins agent, didn't see longer times there too. Can be reduced to 20 min?

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
vanajamukkara: deployWebLogicApp doesn't check if the pods are running, it only creates the resources.

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
vanajamukkara: I can make those changes as enhancement

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
vanajamukkara: I see similar logs messages in other example tests too, just carried the same pattern

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
mbenmclean: Ok so why are we calling AbortSuite if this returns an error. Do we do this in our other tests?

vanajamukkara: yes

pfmackin (approved):

tests/e2e/workloads/weblogic-cluster/weblogic_workload_test.go
mbenmclean: ok

mbenmclean (approved):

PR #5659, opened 2023-03-27
Add a cron schedule to run backend tests every Friday 

Abhijithang:
This PR addes the cron schedule to the backend tests to trigger the job every Friday. The job will not be validated against last stable commit.

### DIFF
--- a/ci/JenkinsfileBackendTests
+++ b/ci/JenkinsfileBackendTests
@@ -57,6 +57,7 @@ pipeline {
             label "${agentLabel}"
         }
     }
+
     triggers {
         cron(branchSpecificSchedule)
         URLTrigger(
--- a/ci/JenkinsfileBackendTests
+++ b/ci/JenkinsfileBackendTests
@@ -580,12 +580,3 @@ def getLatestReleaseVersion() {
     echo "gitTags = ${gitTags}"
     return gitTags.pop()
 }
-
-def getCronSchedule() {
-    if (env.BRANCH_NAME.equals("master")) {
-
-        //Set cron job to run every Friday at 9 AM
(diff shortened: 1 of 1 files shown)


ci/JenkinsfileBackendTests
```diff
@@ -59,8 +59,9 @@ pipeline {
     }
 
     triggers {
+        cron(branchSpecificSchedule)
         URLTrigger(
```
pshirodkar-orcl: Please remove the entire `URLTrigger` section.

pshirodkar-orcl: The variables on lines from 38-41 can be cleaned up as well https://github.com/verrazzano/verrazzano/pull/5659/files#diff-fc401a952aba9e0ce79beeba645b31762ec2616f7a47afde22fc1e1661d3144cR38-R41. Please verify before removing.

pshirodkar-orcl (approved):

ci/JenkinsfileBackendTests
```diff
@@ -59,7 +59,8 @@ pipeline {
     }
 
     triggers {
-        URLTrigger(
+        cron(branchSpecificSchedule)
+/*        URLTrigger(
```
kminder, line 63: If this is a temporary fix please add to the comment explaining why it is being commented out.
Also using block comments should in general be avoided as it can lead to misleading results when searching code.

Abhijithang: Sure, will add a comment and refactor the comments. Thanks Kevin.

Abhijithang: Please review the comments whenever you get time. Thanks.

PR #5665, opened 2023-03-28
VZ-8861 bug report changes

absBitm:
- bug report temp file generations on default cases to avoid report generation failure if report file already exists
- bug report generation on every run
- bug report file naming format changes to vz-bug-report-datetime-XXXX.tar.gz where datetime = yyyymmddhhmmss


tools/vz/pkg/helpers/vzhelper.go
```diff
@@ -336,3 +339,13 @@ func GetVersionOut() string {
 	}
 	return verOut
 }
+
+// CheckBugReportExistsInDir checks vz bug report exists in dir or not
+func CheckBugReportExistsInDir(dir string) bool {
```
agv426, line 344: Maybe include AndRemove in the name of the function since it has the side effect of removing it.

I know that I needed to look to see where the file got cleaned up when I looked at the tests above, it would have more clear

agv426 (approved):

PR #5669, opened 2023-03-28
VZ-9142 Create Install/Upgrade tests for Thanos and Thanos Sidecar

davisgomes:
**Changes**

- Add an install and upgrade test to verify Thanos
- enable Thanos for kind installations
- Small bug fixes in Thanos and tests


tests/e2e/verify-install/promstack/promstack_test.go
```diff
@@ -120,6 +121,68 @@ func areOverridesEnabled() bool {
...
+		if ok && integration == "sidecar" {
+			return true, nil
+		}
+	}
+	t.Logs.Infof("Thanos Sidecar override not found in the Prometheus Operator overrides.")
+	return false, nil
+}
+
+// isThanosInstalled returns true if Thanos is disabled, or if Thanos is enabled and the sidecar container is found
+func isThanosInstalled() (bool, error) {
```
fredtibbitts, line 159: The comment and the function name are confusing to me. Functionally this is validating that if the sidecar is enabled in the overrides then the sidecar exists. What about something like `thanosSidecarDeployedIfEnabled`, just needs to be more descriptive I think.

tests/e2e/pkg/common.go
```diff
@@ -739,3 +741,40 @@ func IsVerrazzanoManaged(labels map[string]string) bool {
 	}
 	return false
 }
+
+func IngressesExist(vz *v1alpha1.Verrazzano, namespace string, ingressNames []string) (bool, error) {
```
fredtibbitts, line 745: This seems overly complicated. Instead of listing and dealing with maps, why not just do a `Get` of each ingress?

tests/e2e/verify-install/promstack/promstack_test.go
davisgomes: I will update this.

tests/e2e/pkg/common.go
davisgomes: I think I have a way to do the same and make it simpler

fredtibbitts (approved):

fredtibbitts (approved):

PR #5670, opened 2023-03-28
VZ-9172.  Add VMI tests to a-la-carte pipeline

michael-cico:
Adds the `verify-infra/vmi` tests to the post-app-stack install stage for the a-la-carte pipeline.

This also modifies the `wait-for-verrazzano-install.sh` script to check the VZ CR state before checking install condition.  The ability to to do a 

```
kubectl wait --for=condition
```

seems to be broken.  I will open a bug for that behavior, but this tweak should allow the test pipeline to behave as expected.

### DIFF
--- a/tests/e2e/config/scripts/wait-for-verrazzano-install.sh
+++ b/tests/e2e/config/scripts/wait-for-verrazzano-install.sh
@@ -13,13 +13,18 @@ i=0
 
 resName=$(kubectl get vz -o jsonpath='{.items[*].metadata.name}')
 echo "waiting for install of resource ${resName} to complete"
-
 while [[ $retval_success -ne 0 ]] && [[ $retval_failed -ne 0 ]]  && [[ $i -lt 45 ]]  ; do
   sleep 60
-  output=$(kubectl wait --for=condition=InstallFailed verrazzano/${resName} --timeout=0 2>&1)
-  retval_failed=$?
-  output=$(kubectl wait --for=condition=InstallComplete verrazzano/${resName} --timeout=0 2>&1)
-  retval_success=$?
+  vzstate=$(kubectl get vz my-verrazzano -o jsonpath={.status.state})
+  echo "vz/${resName} state: ${vzstate}"
+  # Only check the status if the CR is in the ready state
+  if [ "${vzstate}" == "Ready" ]; then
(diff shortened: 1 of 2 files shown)


tests/e2e/config/scripts/wait-for-verrazzano-install.sh
```diff
@@ -13,13 +13,18 @@ i=0
...
 resName=$(kubectl get vz -o jsonpath='{.items[*].metadata.name}')
 echo "waiting for install of resource ${resName} to complete"
-
 while [[ $retval_success -ne 0 ]] && [[ $retval_failed -ne 0 ]]  && [[ $i -lt 45 ]]  ; do
   sleep 60
-  output=$(kubectl wait --for=condition=InstallFailed verrazzano/${resName} --timeout=0 2>&1)
-  retval_failed=$?
-  output=$(kubectl wait --for=condition=InstallComplete verrazzano/${resName} --timeout=0 2>&1)
-  retval_success=$?
+  vzstate=$(kubectl get vz my-verrazzano -o jsonpath={.status.state})
```
saidu941, line 18: I'm assuming `my-verrazzano` is used consistently

saidu941 (approved):

pfmackin (approved):

tests/e2e/config/scripts/wait-for-verrazzano-install.sh
```diff
@@ -29,4 +34,6 @@ if [[ $retval_failed -eq 0 ]] || [[ $i -eq 45 ]] ; then
     exit 1
 fi
 
+kubectl get vz ${resName}
```
mbenmclean, line 37: only thing is this is assuming that the vz is always in the default namespace. I'm assuming that's a valid assumption for our vz CRs we use for the pipelines, but if possible could be good to get and use the namespace in the get vz call too.

mbenmclean (approved):

tests/e2e/config/scripts/wait-for-verrazzano-install.sh
michael-cico: I think the script is ancient; any pipelines using it would have to be updated to provide it.  If we need to do that we can do it as a second pass maybe, or when needed.

PR #5672, opened 2023-03-28
VZ-9133-experimental-module

mbenmclean:
This change adds an optional module CRD and validating webhook based off a new experimental.module values.yaml value


kminder (approved):
Approved.

PR #5680, opened 2023-03-30
VZ-7993 istio explicitly set sidecar.istio.io/inject=false for system components

pfmackin:
There are certain system components existing in namespaces that are labeled `istio-injection=enabled`, where the components should not be in the mesh.  Previously, we were using the istio-system/istio-sidecar-injector configmap to omit such components from the mesh. Change this to explicitly set the pod label `sidecar.istio.io/inject=false`.

### DIFF
--- a/platform-operator/helm_config/overrides/jaeger-operator-values.yaml
+++ b/platform-operator/helm_config/overrides/jaeger-operator-values.yaml
@@ -72,3 +72,6 @@ containerSecurityContext:
   capabilities:
     drop:
       - ALL
+
+extraLabels:
+  sidecar.istio.io/inject: "false"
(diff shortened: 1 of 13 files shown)


platform-operator/helm_config/overrides/jaeger-operator-values.yaml
```diff
@@ -71,4 +71,7 @@ containerSecurityContext:
   allowPrivilegeEscalation: false
   capabilities:
     drop:
-      - ALL
\ No newline at end of file
+      - ALL
```
michael-cico, line 74: Is there a place to document the change in charts without READMEs?

michael-cico (approved):

PR #5693, opened 2023-03-31
Add HelmComponent function for fetching computed values

fredtibbitts:
This PR updates the HelmComponent Upgrade function to return the Helm Release object. I also added a new function that calls Upgrade with the dry run flag set to true, and coalesces the chart values to produce a map of computed Helm values.

This new GetComputedValues function can be used by install components to inspect the computed Helm values at install/upgrade/reconcile time.

I did some local cluster testing with a modified install component and it seems to work as expected.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/helm/helm_component.go
+++ b/platform-operator/controllers/verrazzano/component/helm/helm_component.go
@@ -25,6 +25,7 @@ import (
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/secret"
 	"github.com/verrazzano/verrazzano/platform-operator/internal/config"
 
+	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/release"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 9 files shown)


platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -668,6 +669,60 @@ func (h HelmComponent) preInstallUpgrade(ctx spi.ComponentContext) error {
...
+	if err != nil {
+		return nil, err
+	}
+	if found {
+		stdout, err := helm.GetValues(context.Log(), h.ReleaseName, resolvedNamespace)
+		if err != nil {
+			return nil, err
+		}
+
+		tmpFile, err := vzos.CreateTempFile("helm-overrides-values-*.yaml", stdout)
```
pfmackin, line 702: where is this deleted

fredtibbitts: I copied this code and when I first looked at it I had the same question. It's deleted on line 686 by this defer:
```
defer vzos.RemoveTempFiles(context.Log().GetZapLogger(), `helm-overrides.*\.yaml`)
```
That regex matches the temp file pattern.

PR #5694, opened 2023-03-31
VZ-9168: Update the Dev Component Controller to allow for components implementing the Component interface

bobdonat:
This pull request updates the Dev Component Controller to allow for components implementing the spi Component interface.  Components that implement the spi Component interface can be installed specifying a shim component.  

A sample config map using a shim component is shown below:

```
apiVersion: v1
data:
  name: capi
kind: ConfigMap
metadata:
  labels:
    experimental.verrazzano.io/configmap-kind: ShimComponent
    experimental.verrazzano.io/configmap-apiversion: v1beta2
  name: dev-capi
  namespace: default

```


platform-operator/controllers/configmaps/components/dev_helm_component.go
```diff
@@ -70,33 +69,3 @@ func newDevHelmComponent(cm *v1.ConfigMap) (devComponent, error) {
 		},
 	}, nil
 }
-
```
michael-cico, line 73: So anyone using a `devComponent` will get the `HelmComponent` impls for these now, correct?  I can't recall why we added these here.

michael-cico: nevermind, I see these just got moved.

michael-cico (approved):

platform-operator/controllers/configmaps/components/dev_helm_component.go
bobdonat: the difference is in the contents of the configMap

PR #5699, opened 2023-04-03
VZ-9212: Fix E2E tests to work for Kind suite for phase-1 and phase-2 Rancher changes

mgianatagh:
Tests are passing the Kind test suite.  Not sure if they will also pass the MultiCluster suite, that suite is currently broken due to an unrelated issue.


tests/e2e/security/netpol/network_policy_test.go
```diff
@@ -158,11 +159,13 @@ var _ = t.Describe("Test Network Policies", Label("f:security.netpol"), func() {
 				err = testAccess(metav1.LabelSelector{MatchLabels: map[string]string{"app.kubernetes.io/instance": "ingress-controller"}}, "ingress-nginx", metav1.LabelSelector{MatchLabels: map[string]string{"app": "rancher"}}, "cattle-system", 443, true, true)
 				Expect(err).To(BeNil(), fmt.Sprintf("FAIL: Test rancher ingress failed: reason = %s", err))
 			},
-			func() {
```
bobdonat, line 161: Could the service workaround for starting the webhook be added to the rancher helm chart instead of disabling these tests.  This would allow the tests to work and local installs to work without the workaround. Also, I noticed that @anders-swanson had commented out the dependency checks for Rancher which could then be un-commented out.

bobdonat (approved):

PR #5706, opened 2023-04-03
JIRA: VZ-9169 create Rancher credential from auth CAPI credential

jmaron99:
This PR addresses the need to create a Rancher OCI credential from an existing CAPI auth credential during Rancher installation (creation of cattle-global-data NS).

There needs to be another PR to complete the JIRA to address to create the CAPI secret from the Rancher credential in the postInstall phase of the CAPI component installation.


platform-operator/controllers/verrazzano/component/rancher/rancher_component_test.go
```diff
@@ -519,6 +522,111 @@ func TestInstall(t *testing.T) {
 	}
 }
 
+func TestCreateAdminCloudCredentialSecret(t *testing.T) {
+	//userList.SetGroupVersionKind(schema.GroupVersionKind{
```
anders-swanson, line 526: remove comments

platform-operator/controllers/verrazzano/component/rancher/rancher_component_test.go
```diff
@@ -519,6 +522,111 @@ func TestInstall(t *testing.T) {
...
+metadata:
+  resourceVersion: ""
+`
+	userListMap := make(map[string]interface{})
+	err := yaml.Unmarshal([]byte(userListYaml), &userListMap)
+	assert.NoError(t, err)
+	userList := &unstructured.UnstructuredList{}
+	userList.SetUnstructuredContent(userListMap)
+
+	// NOTE: Had to resort to leveraging mocks since deep in the reflection layers of the fake client creation the
```
anders-swanson, line 554: can you use a fake client, add a single user, and then get by label selector? Not sure why you need the unstructured list.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -901,6 +906,31 @@ func (r *Reconciler) watchThanosInternalUserSecret(namespace string, name string
 	)
 }
 
+func (r *Reconciler) watchRancherGlobalDataNamespace(namespace string, name string) error {
```
anders-swanson, line 909: I think we might need to reconcile rancher if the CAPI secret is created

platform-operator/controllers/verrazzano/component/rancher/rancher_component_test.go
jmaron99: I'd rather not introduce the go.mod dependency on the rancher lib to create a user - it isn't used anywhere else as far as I can tell.

anders-swanson (approved):

platform-operator/controllers/verrazzano/reconcile/controller.go
jmaron99: We discussed and put in the JIRA handling that case with a CAPI component post install

platform-operator/controllers/verrazzano/component/rancher/rancher_component_test.go
anders-swanson: something like: 
```
fakeClient := fake.NewClientBuilder().WithScheme(k8scheme.Scheme).WithObjects(
		&unstructured.Unstructured{ //user data },
)
```

anders-swanson (approved):

platform-operator/controllers/verrazzano/component/rancher/rancher_component_test.go
jmaron99: May work - haven't tried.  Not sure whether the fake client's list implementation is sophisticated enough to return a list with a single registered object.  I assumed a LIst() would need a formed list to return.

PR #5707, opened 2023-04-03
VZ-8840: Grafana datasource support for Thanos

fredtibbitts:
This PR adds Thanos to the list of configured datasources in Grafana. If Thanos Query Frontend is enabled, Thanos is added to the Grafana datasources list and Thanos is set as the default datasource. Dashboards will then fetch metrics from Thanos. If Thanos Query Frontend is not enabled but Prometheus is enabled, Prometheus will be the only configured datasource.

To test, I installed in a local cluster with the following configurations and validated the datasources in Grafana:
- Thanos component enabled, Thanos Query Frontend enabled, Prometheus enabled
- Thanos component enabled, Thanos Query Frontend disabled, Prometheus enabled
- Thanos disabled, Prometheus enabled
- Thanos disabled, Prometheus disabled

I ran the same scenarios but upgrading from Verrazzano 1.5.2 and confirmed the datasources were updated correctly.

### DIFF
--- a/platform-operator/thirdparty/manifests/grafana/datasources-configmap.yaml
+++ b/platform-operator/thirdparty/manifests/grafana/datasources-configmap.yaml
@@ -1,0 +1,31 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+data:
+  datasource.yaml: |2-
+    apiVersion: 1
+    datasources:
+{{ if .isThanosQueryEnabled }}
+    - name: Thanos
+      type: prometheus
+      orgId: 1
+      access: proxy
+      url: {{ .thanosQueryURL }}:{{ .thanosQueryPort }}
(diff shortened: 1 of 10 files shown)


platform-operator/thirdparty/manifests/grafana/datasources-configmap.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+data:
+  datasource.yaml: |2-
+    apiVersion: 1
+    datasources:
+{{ if .isThanosQueryEnabled }}
```
desagar, line 12: since this is statically done in the Helm chart, what happens if Thanos query is disabled after a VZ install? Does that always go through the Grafana install?

desagar (approved):

platform-operator/thirdparty/manifests/grafana/datasources-configmap.yaml
fredtibbitts: I haven't tested this scenario (althoughI can), but I believe what will happen is that the Grafana component would get reconciled again and it would update the configmap without a Thanos datasource. I'll try it.

davisgomes (approved):

PR #5709, opened 2023-04-04
VZ-9109: Fix the issue with PATH variable in daily scanning job

bhatpmk:
This PR sets the latest Grype and Trivy to PATH, ahead of all other entries, so that the daily scan always uses the latest version of these scanners.

### DIFF
--- a/ci/scripts/get_branch_scan_results.sh
+++ b/ci/scripts/get_branch_scan_results.sh
@@ -85,6 +85,18 @@ SCAN_RESULTS_BASE_DIR=${WORKSPACE}/scan-results
 export SCAN_RESULTS_DIR=${SCAN_RESULTS_BASE_DIR}/latest
 mkdir -p ${SCAN_RESULTS_DIR}
 
+if [[ -z "${SCANNER_PATH}" ]]; then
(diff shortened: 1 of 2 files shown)


ci/scripts/get_branch_scan_results.sh
```diff
@@ -85,6 +85,18 @@ SCAN_RESULTS_BASE_DIR=${WORKSPACE}/scan-results
 export SCAN_RESULTS_DIR=${SCAN_RESULTS_BASE_DIR}/latest
 mkdir -p ${SCAN_RESULTS_DIR}
 
+if [[ -z "${SCANNER_PATH}" ]]; then
+  echo "Environment variable SCANNER_PATH is not set"
+else
+  export PATH="${SCANNER_PATH}:${PATH}"
+fi
+echo "PATH in get_branch_scan_results.sh ${PATH}"
```
lake-of-dreams, line 93: I think this will cause unnecessary noise..

bhatpmk: Removed

lake-of-dreams (approved):

lake-of-dreams (approved):

PR #5732, opened 2023-04-06
VZ-8551: VPO status Init:CrashLoopBackOff after applying operator.yaml

pshirodkar-orcl:
Set the kubectl.kubernetes.io/last-applied-configuration annotation when doing install or upgrade using vz cli for correct merge calculation of Kubernetes resources.

This will allow us to use any combination of install & upgrade using either kubectl or vz cli.

### DIFF
--- a/pkg/kubectlutil/kubectl_util.go
+++ b/pkg/kubectlutil/kubectl_util.go
@@ -1,0 +1,26 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package kubectlutil
+
+import (
+	"fmt"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/util"
+)
+
+// SetLastAppliedConfigurationAnnotation applies the kubectl.kubernetes.io/last-applied-configuration annotation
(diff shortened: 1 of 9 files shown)


pkg/kubectlutil/kubectl_util.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package kubectlutil
```
pshirodkar-orcl, line 4: I creted a new package since I was getting cyclic import errors when I tried to put the new function in `pkg/k8sutil` etc.

pkg/kubectlutil/kubectl_util.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/util"
+)
+
+// SetLastAppliedConfigurationAnnotation applies the kubectl.kubernetes.io/last-applied-configuration annotation
```
mgianatagh, line 15: Can there be a unit test for this new function?

pshirodkar-orcl: Added unit test.

pshirodkar-orcl: I added more unit tests. Please review when available.

mgianatagh (approved):

PR #5738, opened 2023-04-06
VZ-9247 - Enable ServiceMonitors for Thanos components, add needed network and auth policies, related Helm chart changes

desagar:


platform-operator/controllers/verrazzano/component/thanos/thanos.go
```diff
@@ -59,9 +59,19 @@ func AppendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
 	}
 	kvs = append(kvs, image...)
 
+	kvs = appendVerrazzanoOverrides(ctx, kvs)
```
davisgomes, line 62: `appendVerrazzanoOverrides` is a bit confusing, but I can't come up with a better name. What about `appendVerrazzanoGeneratedOverrides`. I still don't really like it...

platform-operator/controllers/verrazzano/component/thanos/thanos_test.go
```diff
@@ -250,6 +287,19 @@ func TestAppendOverrides(t *testing.T) {
 	}
 }
 
+func mergeMaps(startWithMap map[string]string, overlayMaps ...map[string]string) map[string]string {
```
davisgomes, line 290: I don't think you really need a start with map in this function

platform-operator/helm_config/overrides/thanos-values.yaml
```diff
@@ -101,3 +101,11 @@ receive:
...
       drop:
         - ALL
+
+metrics:
+  enabled: true
+  serviceMonitor:
+    enabled: true
+    labels:
+      release: prometheus-operator
+    namespace: verrazzano-monitoring
```
davisgomes, line 111: Won't this default to the release namespace which is verrazzano-monitoring? Leaving it out makes the helm deployment more generic

pkg/vzcr/enabled.go
```diff
@@ -66,6 +66,21 @@ func IsIstioEnabled(cr runtime.Object) bool {
 	return true
 }
 
+// IsIstioInjectionEnabled - Returns false only if Istio injection is explicitly disabled in the CR
```
fredtibbitts, line 69: Should be something like "if both Istio and Istio injection are explicitly disabled in the CR"?

platform-operator/thirdparty/charts/thanos/templates/verrazzano/authorizationpolicy.yaml
```diff
@@ -62,6 +73,17 @@ spec:
...
+    - from:
+        - source:
+            namespaces:
+              - verrazzano-monitoring
+            principals:
+              - cluster.local/ns/verrazzano-monitoring/sa/prometheus-operator-kube-p-prometheus
+      to:
+        - operation:
+            ports:
+              - "10902"
```
davisgomes, line 86: Could you add this to the namespaces and principles above?

fredtibbitts (commented):
For all of our other monitors, we add a default verrazzano cluster label like so:
```
      - action: replace
        targetLabel: verrazzano_cluster
        replacement: local
```
Do we need to do the same with these monitors?

platform-operator/controllers/verrazzano/component/thanos/thanos_test.go
desagar: good point :-) will remove it

pkg/vzcr/enabled.go
desagar: no, if istio is disabled it implies injection is as well. 
If Istio is enabled, injection is enabled by default.

So the only case we would return false is if someone explicitly disabled Istio injection OR Istio itself. I'll modify to reflect that.

platform-operator/helm_config/overrides/thanos-values.yaml
desagar: hm...I'm not sure. I guess I can do that for now, and if we decide to move the whole thing to another namespace we'll have to deal with it then.

platform-operator/thirdparty/charts/thanos/templates/verrazzano/authorizationpolicy.yaml
desagar: I could, but I think this auth policy thing is quite confusing as it is. I like to keep these separate so we can comment them individually. This authorization policy needs a revisit because it's a bit muddled between query/queryfrontend and storegateway

platform-operator/helm_config/overrides/thanos-values.yaml
fredtibbitts: The way the ServiceMonitors templates are written, the namespace defaults to the release namespace, so I agree, we should remove the namespace setting.

pkg/vzcr/enabled.go
desagar: done

fredtibbitts (dismissed):

platform-operator/controllers/verrazzano/component/thanos/thanos.go
desagar: added a comment as we discussed

davisgomes (approved):

PR #5740, opened 2023-04-07
VZ-9275: Update Keycloak image with SnakeYAML upgraded to v2.0

bhatpmk:
Refreshes the Keycloak 20.0.1 image built with SnakeYAML v2.0 in https://github.com/verrazzano/keycloak/pull/10


lake-of-dreams (approved):

bhatpmk:
VZ-9275

PR #5765, opened 2023-04-10
VZ-8892 Update the Thanos VMC sync to remove Prometheus federation

davisgomes:
**Changes**

- Add a cluster label to the managed cluster store targets for selection
- Change the update and create process to factor in labels
- Add an ingress host flag to override the Rancher API calls for running in IDE
- Update and add unit tests to test new update and delete process

**Testing**

- Run a multicluster setup in kind
- Turn on and off the Thanos component on the managed cluster
- Turn on and off the Thanos component on the admin cluster
- Make sure that the Thanos query store configmap gets modified correctly
- Make sure that the Prometheus additional configs get modified correctly
- Deregister and make sure both Thanos and Prometheus targets get cleaned up

### DIFF
--- a/application-operator/mcagent/mcagent.go
+++ b/application-operator/mcagent/mcagent.go
@@ -176,9 +176,9 @@ func (s *Syncer) updateVMCStatus() error {
 	if err != nil {
 		return fmt.Errorf("Failed to get Thanos query URL to update VMC %s: %v", vmcName, err)
 	}
-	if thanosAPIHost != "" {
-		vmc.Status.ThanosQueryStore = thanosAPIHost
-	}
+
+	// If Thanos is disabled, we want to empty the host so Prometheus federation returns
+	vmc.Status.ThanosQueryStore = thanosAPIHost
 
 	// update status of VMC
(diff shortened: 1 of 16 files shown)


application-operator/mcagent/mcagent.go
```diff
@@ -176,9 +176,9 @@ func (s *Syncer) updateVMCStatus() error {
...
 	}
-	if thanosAPIHost != "" {
-		vmc.Status.ThanosQueryStore = thanosAPIHost
-	}
+
+	// If Thanos is disabled, we want to empty the host so Prometheus federation returns
+	vmc.Status.ThanosQueryStore = thanosAPIHost
 
 	// update status of VMC
 	return s.AdminClient.Status().Update(s.Context, &vmc)
```
fredtibbitts, line 184: Is this going to do an update on every mcagent loop iteration? I guess there's nothing like controllerutil CreateOrUpdate for status subresources? It would be nice if we only did the update when necessary.

fredtibbitts (approved):

desagar (approved):

PR #5772, opened 2023-04-10
VZ-9280: E2E test for Thanos Store Gateway

fredtibbitts:
This PR adds an end-to-end test for the Thanos Store Gateway to the Kind acceptance tests. If an environment variable to enable the test is set, the script creates a secret with the Thanos storage provider configuration and updates the VZ CR to enable the Store Gateway and point to the config secret. I enabled the flag in the main Kind AT pipeline.

I tested by running the test against a cluster with and without the Thanos Store Gateway enabled and confirmed that it checked for the running Store Gateway pod as needed.

Note that if the Store Gateway cannot mount the secret or if the secret contents don't pass minimal validation, the Store Gateway pod will go into CrashLoopBackoff and the test will fail, so the check for running (and ready) Store Gateway pod seems sufficient.

### DIFF
--- a/tests/e2e/verify-install/thanos/thanos_test.go
+++ b/tests/e2e/verify-install/thanos/thanos_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Jeffail/gabs/v2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/verrazzano/verrazzano/pkg/k8sutil"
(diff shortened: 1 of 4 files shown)


tests/e2e/verify-install/thanos/thanos_test.go
```diff
@@ -51,8 +53,33 @@ var beforeSuite = t.BeforeSuiteFunc(func() {
...
+	if isThanosInstalled {
+		isStoreGatewayEnabled, err = isStoreGatewayEnabledInOverrides(inClusterVZ.Spec.Components.Thanos.InstallOverrides.ValueOverrides)
+		if err != nil {
+			AbortSuite(fmt.Sprintf("Failed to process VZ CR Thanos overrides: %v", err))
+		}
+	}
 })
 
+// isStoreGatewayEnabledInOverrides returns true if the Thanos Store Gateway is enabled in the VZ CR overrides
+func isStoreGatewayEnabledInOverrides(overrides []v1alpha1.Overrides) (bool, error) {
```
desagar, line 66: do we do this kind of thing anywhere else in the tests that you know of? i.e. examine the overrides and extract meaning out of them?

PR #5779, opened 2023-04-11
VZ-7993 Add Istio test to a-la-carte pipeline

pfmackin:
Install Istio and test hello-helidon as part of a-la-cart
This branch includes a fix the MySQL check where MySQL is not ready so MySQL operator needs to be restarted.


ci/a-la-carte/Jenkinsfile
```diff
@@ -310,6 +310,9 @@ pipeline {
 
                     kubectl apply -f ${WORKSPACE}/appstack.yaml
 
+                    # wait for Verrazzano install to start reconciling
+                    ${WORKSPACE}/tests/e2e/config/scripts/wait-for-verrazzano-reconciling.sh
```
mbenmclean, line 314: why is this necessary?

ci/a-la-carte/Jenkinsfile
```diff
@@ -456,6 +459,93 @@ pipeline {
         }
 
 
+
+        stage('Install Istio App stack') {
```
mbenmclean, line 463: This is in the wrong place. This should be after install app stack and before install none

ci/a-la-carte/Jenkinsfile
pfmackin: there is a race condition where we apply a CR but before the VPO changes state to reconciling, we check for complete and think it is done.  This makes sure we wait until it is reconciling.

ci/a-la-carte/Jenkinsfile
pfmackin: fixed and testing

mbenmclean (approved):

mbenmclean (approved):

PR #5781, opened 2023-04-11
VZ-9282 Check for Grafana datasource in VMI test

davisgomes:
Adds a test to ensure that the proper datasource is enabled in Grafana using the API.

### DIFF
--- a/tests/e2e/verify-infra/vmi/vmi_test.go
+++ b/tests/e2e/verify-infra/vmi/vmi_test.go
@@ -378,6 +378,19 @@ var _ = t.Describe("VMI", Label("f:infra-lcm"), func() {
 						Eventually(assertAdminRole, waitTimeout, pollingInterval).Should(BeTrue())
 					}
 				})
+
+				t.It("Grafana should have a datasource present", func() {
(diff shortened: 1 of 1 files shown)


tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -681,3 +696,47 @@ func verrazzanoSecretRequired(vz *vzalpha1.Verrazzano) bool {
...
+		return false, nil
+	}
+	resp, err := pkg.GetWebPageWithBasicAuth(*vz.Status.VerrazzanoInstance.GrafanaURL+"/api/datasources", "", "verrazzano", password, kubeconfigPath)
+	if err != nil {
+		t.Logs.Errorf("Failed to get Grafana datasources: %v", err)
+		return false, err
+	}
+
+	var datasources []map[string]interface{}
+	err = json.Unmarshal(resp.Body, &datasources)
```
fredtibbitts, line 717: Does the response also include the "is default" flag? If so I think it would be good to also check that the datasource is marked as default.

davisgomes: It does. I can check that too.

fredtibbitts (dismissed):

PR #5790, opened 2023-04-12
VZ-7914: a la carte pipeline updates with ginkgo

mbenmclean:
This change updates the a la carte pipeline to use ginkgo to perform updates instead of the Jenkinsfile. It also adds the cluster management stage.


kminder (approved):
Approved.

PR #5795, opened 2023-04-12
Prometheus a la carte doc changes and postuninstall bug fix

mbenmclean:
This change adds the Prometheus a la carte CHANGES.md entry, and also fixes a bug in the postuninstall hook that deletes the legacy ingress

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,8 @@
 ### v1.6.0
 Features:
 
-- Prometheus stack is now able to be dynamically enabled or disabled at any point during the Verrazzano installation lifecycle
+- The Prometheus components can now be enabled or disabled at any point in the Verrazzano lifecycle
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,4 +1,8 @@
 ### v1.6.0
+Features:
+
+- Prometheus stack is now able to be dynamically enabled or disabled at any point during the Verrazzano installation lifecycle
```
kminder, line 4: Prometheus stack can now be enabled or disabled at any point in the Verrazzano lifecycle

michael-cico: Do we need to add a line about the `none` profile here?

mbenmclean: Yes we probably should I can add that

michael-cico (approved):

PR #5812, opened 2023-04-17
VZ-9234 add VPO watch for Capi secret to fire rancher reconcile and create admin cloud credential

knelson7:


anders-swanson (approved):
LGTM!

jmaron99:
Admittedly I may be confused...

This PR seems to support CAPI coming up after rancher, having its own auth credential, and creating a cloud credential based on the CAPI secret (potentially overwriting an existing admin cloud credential?).  

I thought we were trying to address a situation in which CAPI comes up, does no yet have a auth secret, and is leveraging the existing rancher cloud credential to create the auth secret?  @anders-swanson

jmaron99:
> 
> There should be no existing admin cloud credential if CAPI comes up second
> 

I'm assuming that "cloud credential" references the credential in cattle-global-data namespace

> Same case as above applies here, there will be no CAPI admin cloud credential if CAPI comes up second?

Given my definition above - rancher can come up with no cloud credential in cattle-global-data?  And CAPI will always come up with an auth credential assuming the installation of OCI provider?

PR #5826, opened 2023-04-19
VZ-9366 VZ CLI should not automatically call bug report and analyze upon certain failures 

knelson7:
Example cases that should not automatically run analyze or bug report 
- Invoking vz uninstall when vz is not installed 
- Invoking vz install, uninstall, upgrade with incorrect CLI arguments 
- Invoking vz install with multiple versions 
- Invoking vz install when an install has already started 

Example cases that SHOULD automatically run analyze or bug report 
- timeout after waiting for install, uninstall, upgrade to complete 
- timeout after waiting for platform operator 
- failing to create the vz install resource

### DIFF
--- a/tools/vz/cmd/uninstall/uninstall.go
+++ b/tools/vz/cmd/uninstall/uninstall.go
@@ -163,7 +163,7 @@ func runCmdUninstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelpe
 				return failedToUninstallErr(err)
 			}
 		} else {
-			return autoBugReport(cmd, vzHelper, err)
+			return bugreport.AutoBugReport(cmd, vzHelper, err)
 		}
 	}
 	_, _ = fmt.Fprintf(vzHelper.GetOutputStream(), "Uninstalling Verrazzano\n")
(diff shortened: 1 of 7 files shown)


tools/vz/cmd/uninstall/uninstall.go
```diff
@@ -193,13 +163,31 @@ func uninstallVerrazzano(cmd *cobra.Command, vzHelper helpers.VZHelper) error {
...
-	return waitForUninstallToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vz.Namespace, Name: vz.Name}, timeout, vpoTimeout, logFormat, useUninstallJob)
+	err = waitForUninstallToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vz.Namespace, Name: vz.Name}, timeout, vpoTimeout, logFormat, useUninstallJob)
+	if err != nil {
+		return autoBugReport(cmd, vzHelper, err)
+	}
+	return nil
+}
+
+// autoBugReport checks that AutoBugReportFlag is set and then kicks off vz bugreport CLI command. It returns the same error that is passed in
+func autoBugReport(cmd *cobra.Command, vzHelper helpers.VZHelper, err error) error {
```
mgianatagh, line 180: I notice there is a autoBugReport function for each cmd.  Can there be a common function for all commands?

knelson7: Fixed! I put it in bugreport. Originally I had wanted to put it in vzhelpers or cmdhelpers, but ran into import cycles.

mgianatagh (approved):

PR #5831, opened 2023-04-20
VZ-9334: make CAPI install component non-experimental

bobdonat:
This pull request moves the CAPI install component from experimental to non-experimental.

In addition, the ocne provider manifests are updated, the oci provider was updated to 0.8.1 which does not require oci credentials, and all CAPI providers are now installed into the verrazzano-capi namespace as agreed upon.

The CAPI install component is not installed by default.  To install the CAPI component include the following in the vz CR:
```
  components:
    capi:
      enabled: true
```


pkg/vzcr/enabled.go
```diff
@@ -84,6 +84,20 @@ func IsIstioInjectionEnabled(cr runtime.Object) bool {
 	return true
 }
 
+// IsCAPIEnabled - Returns false only if CAPI is explicitly disabled by the user
```
mgianatagh, line 87: Would this function be more appropriate to be in the API code?  My concern is needing to account for the different API revisions.

bobdonat: Other components had the same function here so I followed that pattern.

PR #5846, opened 2023-04-24
VZ-8699: Keycloak integration for operator based OpenSearch via AuthProxy

saketmht:
The goal of this PR is to integrate operator based OS with Verrazzano without any CRD changes i.e. the operator based OS will be installed via the dev component controller for now. These code changes should not affect the current Verrazzano workflow in any way until the operator based OS is installed and the VZ CR is updated to use the new OS services.

Changes:

- Update keycloak to create and add new roles - vz_log_pusher for verrazzano-es-internal user and opensearch-admin for verrazzano user
- Update keycloak to add new ingress hosts
```
https://osd.logging.{{.DNSSubDomain}}
https://opensearch.logging.{{.DNSSubDomain}}
```

- Update authproxy to allow redirection for new operator based OS ingresses and service
- Update authproxy to forward proxy headers and roles mapped to a user
- Update cluster-operator to use the operator based OS ingress for managed clusters if fluentd and jaeger are using operator based OS authproxy service URL


cluster-operator/controllers/vmc/sync_registration_secret.go
```diff
@@ -171,28 +178,28 @@ func (r *VerrazzanoManagedClusterReconciler) getVzESURLSecret(vzList *vzapi.Verr
...
 
-// Get the VMI opensearch URL.
-func (r *VerrazzanoManagedClusterReconciler) getVmiESURL(vzList vzapi.VerrazzanoList) (URL string, err error) {
+// Get the opensearch URL.
+func (r *VerrazzanoManagedClusterReconciler) getESURL(vzList vzapi.VerrazzanoList, ingressName string) (URL string, err error) {
 	if len(vzList.Items) == 0 {
 		return "", nil
 	}
-	if !vzcr.IsOpenSearchEnabled(&vzList.Items[0]) {
+	if ingressName == vmiIngest && !vzcr.IsOpenSearchEnabled(&vzList.Items[0]) {
```
adiforluls, line 186: Why is this needed? How are you getting the vmi-os-ingest url now?

adiforluls: Got it

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -933,6 +941,18 @@ func configureKeycloakRealms(ctx spi.ComponentContext) error {
...
 	}
 
+	// Create Verrazzano Log Pusher Role
+	err = createVerrazzanoRole(ctx, cfg, cli, vzLogPusherRole)
+	if err != nil {
+		return err
+	}
+
+	// Create opensearch-admin role
+	err = createVerrazzanoRole(ctx, cfg, cli, opensearchAdminRole)
```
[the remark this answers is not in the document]
saketmht: The operator based OS comes with security plugin which provides access to the users based on the roles they have. So for verrazzano user we are creating a role opensearch-admin which provides all access. Similarly, a vz_log_pusher role for the verrazzano-es-internal user which the fluentd uses to push logs.

adiforluls (approved):

cluster-operator/controllers/vmc/jaeger.go
```diff
@@ -71,9 +71,20 @@ func (r *VerrazzanoManagedClusterReconciler) getJaegerOpenSearchConfig(vzList *v
 
 	// Decide which OpenSearch URL to use.
 	// If the Jaeger OpenSearch URL is the default URL, use VMI OpenSearch ingress URL.
+	// If the Jaeger OpenSearch URL is the default operator based URL, use operator based OpenSearch ingress URL.
 	// If the Jaeger OpenSearch URL  is not the default, meaning it is a custom OpenSearch, use the external OpenSearch URL.
 	if jsc.OSURL == vzconstants.DefaultJaegerOSURL {
-		jc.URL, err = r.getVmiESURL(*vzList)
+		jc.URL, err = r.getESURL(*vzList, vmiIngest)
```
kminder, line 77: Are these two code paths now identical?
If so they should be collapsed.

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
```diff
@@ -1866,3 +1866,19 @@ func TestAddClientRoleToUser(t *testing.T) {
...
+func TestAddRealmRoleToUser(t *testing.T) {
+	k8sutil.ClientConfig = fakeRESTConfig
+	k8sutil.NewPodExecutor = k8sutilfake.NewPodExecutor
+	cfg, cli, _ := fakeRESTConfig()
+
+	c := fake.NewClientBuilder().WithScheme(k8scheme.Scheme).Build()
+	ctx := spi.NewFakeContext(c, testVZ, nil, false)
+
+	err := addRealmRoleToUser(ctx, cfg, cli, "test-user", "test-realm", "test-role")
+	assert.NoError(t, err)
```
kminder, line 1883: Just asserting that no error occurred isn't a sufficient unit test.

IshaGirdhar (dismissed):

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
```diff
@@ -454,6 +468,10 @@ data:
             if hostname and #hostname > 0 then
                 first, last, backend_name = hostname:find("^([^.]+)")
                 if backend_name and #backend_name > 0 then
+                    -- Append '-logging' if 'logging' is in the ingressHost and not already present in backend_name
+                    if string.find(ingressHost, 'logging') and not string.find(backend_name, 'logging') then
```
lake-of-dreams, line 472: seems error prone, should probably check for exact string in `ingressHost` rather than contains `logging` , there can be any ingress that contains `logging`

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
lake-of-dreams: This test does not seem correct, you should add the expectation from podExec like in [fakeGetArgoCDClientSecretFromKeycloak](https://github.com/verrazzano/verrazzano/blob/b3bbf5e420da7486799a47c6d3cc1350eca5c6e6/platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go#L276) and then test this

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1966,6 +1998,20 @@ func updateRancherClientSecretForKeycloakAuthConfig(ctx spi.ComponentContext) er
 	return common.UpdateKeycloakOIDCAuthConfig(ctx, authConfig)
 }
 
+// addRealmRoleToUser adds a realm role to the given user in the target realm
+func addRealmRoleToUser(ctx spi.ComponentContext, cfg *restclient.Config, cli kubernetes.Interface, userName, targetRealm, roleName string) error {
+	kcPod := keycloakPod()
+	addRoleCmd := kcAdminScript + " add-roles -r " + targetRealm + " --uusername " + userName + " --rolename " + roleName
```
lake-of-dreams, line 2004: could this use `fmt.Sprintf`?

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -48,11 +48,13 @@ const (
 	vzMonitorGroup          = "verrazzano-monitors"
 	vzSystemGroup           = "verrazzano-system-users"
 	vzAPIAccessRole         = "vz_api_access"
+	vzLogPusherRole         = "vz_log_pusher"
 	vzUserName              = "verrazzano"
 	vzInternalPromUser      = "verrazzano-prom-internal"
 	vzInternalEsUser        = "verrazzano-es-internal"
 	keycloakPodName         = "keycloak-0"
 	realmManagement         = "realm-management"
+	opensearchAdminRole     = "opensearch-admin"
```
lake-of-dreams, line 57: is `opensearch-admin` a standard role in Opensearch i.e. does it exist for non-verrazzano installations? If not and is verrazzano specific, probably it should have `vz` or `verrazzano` prefixed

pkg/constants/constants.go
```diff
@@ -131,6 +131,12 @@ const DefaultOpensearchURL = "http://verrazzano-authproxy-opensearch:8775"
 // Default Jaeger OpenSearch URL
 const DefaultJaegerOSURL = "http://verrazzano-authproxy-opensearch.verrazzano-system:8775"
 
+// Default OpenSearch URL for opensearch-operator based OpenSearch
+const DefaultOperatorOSURL = "http://verrazzano-authproxy-opensearch-logging:8775"
```
lake-of-dreams, line 135: does this service exist in any other namespace apart from `verrazzano-system` ? If not, can we just have one url like `http://verrazzano-authproxy-opensearch-logging.verrazzano-system:8775` ?

saketmht: The current VMO two different urls so I was just following the pattern. But yeah it makes sense to just have one. Will change that

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy-configmap.yaml
saketmht: I am not sure if we should be considering any 'other' ingress going via authproxy other than what we create with verrazzano install. This is more of a WA since we will two OS. VMO OS and the operator based OS both resolving to backend = opensearch. So adding '-logging' to the operator based ingress host as it will be in the 'verrazzano-logging' ns

lake-of-dreams: by any ingress I meant ingress created via verrazzano install only, this logic can easily be overlooked if a component is introduced that creates another ingress with host like `xyz-logging.xx.xx.xx.xx.xx` , if we know the first part then probably should split the hostname and exactly match the first part,

saketmht: Okay I get your point. Will update the code

pkg/constants/constants.go
saketmht: Thinking again, we need the user to update the svc url in upgrade cases so I will leave both the urls for now as both are valid and should work.

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
saketmht: Updated

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
saketmht: Sure

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
saketmht: Updated the test to check for some keywords in the command

lake-of-dreams (dismissed):
Approving but still would request someone with idea of Jaeger and OS integration to check if indeed two service URLs are required https://github.com/verrazzano/verrazzano/pull/5846/files#r1176035541

platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -1249,13 +1281,13 @@ func createVerrazzanoRole(ctx spi.ComponentContext, cfg *restclient.Config, cli
 	}
 	role := "name=" + roleName
 	createRoleCmd := kcAdminScript + " create roles -r " + vzSysRealm + " -s " + role
-	ctx.Log().Debugf("createVerrazzanoRole: Create Verrazzano API Access Role Cmd = %s", createRoleCmd)
+	ctx.Log().Debugf("createVerrazzanoRole: Create %s Role Cmd = %s", roleName, createRoleCmd)
 	stdout, stderr, err := k8sutil.ExecPod(cli, cfg, kcPod, ComponentName, bashCMD(createRoleCmd))
 	if err != nil {
-		ctx.Log().Errorf("Component Keycloak failed creating Verrazzano API Access Role: stdout = %s, stderr = %s", stdout, stderr)
+		ctx.Log().Errorf("Component Keycloak failed creating %s Role: stdout = %s, stderr = %s", roleName, stdout, stderr)
```
kminder, line 1287: Use consistent case for things in log messages.  Here I think 'Role' should be 'role' since the usage isn't a proper noun.

kminder (dismissed):
Approved.  Please address my comment about log message changes in a follow up change.

IshaGirdhar (approved):

PR #5851, opened 2023-04-24
VZ-9461 Verrazzano must co-exist with existing ingress-nginx

pfmackin:
Verrazzano must co-exist with existing ingress-nginx for new installations. 

1. Install Ingress NGINX into verrazzano-ingress-nginx
2. Upgrade should not move Ingress NGINX.  If Ingress NGINX was installed into ingress-nginx, don't move it since that would cause a new LB (and IP) to be created.

Here is the change:

1. At operator startup, determine what namespace should be used for Ingress NGINX.  This is done by looking at the Helm releases.   If a release exists in `ingress-nginx` where the ingress class is `verrazzano-ingress` then continue to use the `ingress-nginx` namespace, otherwise use `verrazzano-ingress-nginx`
2. Write that namespace to a var in nginxutil package.
3. At runtime, all of the VZ code will use the global namepace var (via func)


michael-cico (approved):

agv426 (dismissed):

pkg/nginxutil/nginxutil.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	type Controller struct {
+		IngressClassResource `json:"ingressClassResource"`
+	}
+	type helmValues struct {
+		Controller `json:"controller"`
+	}
+
+	// See if NGINX is installed in the ingress-nginx namespace
+	found, err := helm2.IsReleaseInstalled(releaseName, namespace)
```
mbenmclean, line 58: It's possible that the user has an existing nginx release in the ingress-nginx namespace with the same name we use. We probably should ensure that we installed this release. Not exactly sure the best way to do that. Perhaps we could look for the correct image in the values.

pfmackin: I would say it is extremely unlikely

mbenmclean: I'm changing this to look at the namespace label not the release so this is moot

pkg/constants/constants.go
```diff
@@ -181,8 +179,11 @@ const ArgoCDClusterRancherSecretName = "verrazzano-argocd-secret"
 // ArgoCDClusterRancherUsername is the username in Rancher used to identify the Verrazzano Argo CD cluster user
 const ArgoCDClusterRancherUsername = "vz-argoCD-reg"
 
-// DefaultRancherIngressHost is the default host used for Rancher API requests
-const DefaultRancherIngressHost = "ingress-controller-ingress-nginx-controller.ingress-nginx"
+// DefaultRancherIngressHostPrefix is the default internal Ingress host prefix used for Rancher API requests
+const DefaultRancherIngressHostPrefix = "ingress-controller-ingress-nginx-controller."
+
+// DefaultRancherIngressHost is the default internal Ingress host used for Rancher API requests
+const DefaultRancherIngressHost = DefaultRancherIngressHostPrefix + constants.IngressNginxNamespace
```
pfmackin, line 186: Aren't these 2 consts only used in rancherutil?  If so, it is better to move them there to prevent anyone other code from using them.

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,30 @@
...
+// CheckIfVerrazzanoManagedNamespaceExists returns true if the namespace exists and has the verrazzano.io/namespace label
+func CheckIfVerrazzanoManagedNamespaceExists(nsName string) (bool, error) {
+	client, err := k8sutil.GetCoreV1Client()
+	if err != nil {
+		return false, fmt.Errorf("Error creating corev1 client: %v", err)
+	}
+
+	namespace, err := client.Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return false, fmt.Errorf("Unexpected error checking for namespace %s: %v", nsName, err)
```
pfmackin, line 24: change to "Failed with ...

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckIfVerrazzanoManagedNamespaceExists returns true if the namespace exists and has the verrazzano.io/namespace label
+func CheckIfVerrazzanoManagedNamespaceExists(nsName string) (bool, error) {
+	client, err := k8sutil.GetCoreV1Client()
+	if err != nil {
+		return false, fmt.Errorf("Error creating corev1 client: %v", err)
```
pfmackin, line 19: change to "Failed creating...

platform-operator/thirdparty/charts/ingress-nginx/templates/clusterrolebinding.yaml
```diff
@@ -7,11 +7,11 @@ metadata:
     {{- with .Values.controller.labels }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
-  name: {{ include "ingress-nginx.fullname" . }}
+  name: {{ include "ingress-nginx.namespacedfullname" . }}
```
michael-cico, line 10: this is to distinguish it from earlier installs during upgrade right?

michael-cico (approved):

pkg/constants/constants.go
mbenmclean: moved

pkg/namespace/namespace.go
mbenmclean: fixed

platform-operator/thirdparty/charts/ingress-nginx/templates/clusterrolebinding.yaml
mbenmclean: fixed

pkg/namespace/namespace.go
mbenmclean: fixed

michael-cico (approved):

michael-cico (approved):

PR #5859, opened 2023-04-25
VZ-9459.  OCI DNS Cert-Manager webhook integration 

michael-cico:


platform-operator/controllers/verrazzano/component/common/certmanager.go
```diff
@@ -55,3 +57,13 @@ func CertManagerCrdsExist() (bool, error) {
 	// Found required CRDs
 	return crdsExist, nil
 }
+
+func NewFakeAPIExtTestClient(objs ...runtime.Object) func(log ...vzlog.VerrazzanoLogger) (apiextensionsv1client.ApiextensionsV1Interface, error) {
```
michael-cico, line 61: Maybe move these under the k8sutil where the client getter lives.

PR #5875, opened 2023-04-28
VZ-9434: add seLinuxOptions in Fluentd Daemonset

karan56625:
To add seLinuxOptions to get it overridden in case selinux is enforcing.
There will be no values passed to it by default. But customer can override these if they want to provide seLinuxOptions in case seLinux is enforcing.

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-fluentd/templates/daemonset.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-fluentd/templates/daemonset.yaml
@@ -82,11 +82,13 @@ spec:
           image: {{ .Values.logging.fluentdImage }}
           imagePullPolicy: IfNotPresent
           securityContext:
+{{- if .Values.seLinuxOptions }}
             seLinuxOptions:
-              type: {{ .Values.securityContext.seLinuxOptions.type }}
-              level: {{ .Values.securityContext.seLinuxOptions.level }}
-              role: {{ .Values.securityContext.seLinuxOptions.role }}
-              user: {{ .Values.securityContext.seLinuxOptions.user }}
+              type: {{ .Values.seLinuxOptions.type }}
+              level: {{ .Values.seLinuxOptions.level }}
+              role: {{ .Values.seLinuxOptions.role }}
+              user: {{ .Values.seLinuxOptions.user }}
+{{- end }}
             privileged: false
             allowPrivilegeEscalation: false
             capabilities:
--- a/platform-operator/helm_config/charts/verrazzano-fluentd/templates/daemonset.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-fluentd/templates/daemonset.yaml
@@ -82,6 +82,11 @@ spec:
           image: {{ .Values.logging.fluentdImage }}
           imagePullPolicy: IfNotPresent
           securityContext:
+            seLinuxOptions:
+              type: {{ .Values.securityContext.seLinuxOptions.type }}
+              level: {{ .Values.securityContext.seLinuxOptions.level }}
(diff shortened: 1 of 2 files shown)


platform-operator/helm_config/charts/verrazzano-fluentd/templates/daemonset.yaml
```diff
@@ -82,6 +82,13 @@ spec:
           image: {{ .Values.logging.fluentdImage }}
           imagePullPolicy: IfNotPresent
           securityContext:
+{{- if .Values.seLinuxOptions }}
```
adiforluls, line 85: We point customers to look at `values.yaml` file for components to figure out how to set value overrides. Can you add comments in fluentd `values.yaml` for these values?

See argoCD for example https://github.com/verrazzano/verrazzano/blob/v1.5.2/platform-operator/thirdparty/charts/argo-cd/values.yaml

gaurav-05: > We point customers to look at `values.yaml` file for components to figure out how to set value overrides. Can you add comments in fluentd `values.yaml` for these values?
> 
> See argoCD for example https://github.com/verrazzano/verrazzano/blob/v1.5.2/platform-operator/thirdparty/charts/argo-cd/values.yaml

Are you referring to the Ref links

adiforluls: > > We point customers to look at `values.yaml` file for components to figure out how to set value overrides. Can you add comments in fluentd `values.yaml` for these values?
> > See argoCD for example https://github.com/verrazzano/verrazzano/blob/v1.5.2/platform-operator/thirdparty/charts/argo-cd/values.yaml
> 
> Are you referring to the Ref links

See InstallOverrides field under https://verrazzano.io/latest/docs/reference/api/vpo-verrazzano-v1beta1/#install.verrazzano.io/v1beta1.FluentdComponent

karan56625: comments are added. Also we are gonna have Doc PR for the customer to know how these seLinuxOptions can be overriden.

platform-operator/helm_config/charts/verrazzano-fluentd/values.yaml
```diff
@@ -24,3 +24,15 @@ fluentd:
 monitoring:
   enabled: true
   useIstioCerts: true
+
+# In the environment where SELinux is enforcing, you can consider to override the default SELinux security context for the Fluentd so that it can have the required read/write privileges.
+# By customizing these options, you can ensure that Fluentd operates within the appropriate SELinux security context.
+seLinuxOptions:
+  # -- String to override the SELinux type for the Fluentd process. This determines the access permissions and restrictions for the Fluentd.
+  type: "" # spc_t
```
kminder, line 32: Is it ok that these are defaulted to the empty string?
In other words should these just be comment or defaults?
The way the template is written I think this will result in the presence of a seLinuxOptions with empty type, etc. in the daemonset resource.

karan56625: yes, empty string are also ok. this will be treated as null for seLinuxOptions object .It would be like
```
securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            add:
            - DAC_OVERRIDE
            drop:
            - ALL
          privileged: false
          seLinuxOptions: {}
```
 so default `container_t` will be used for the Fluentd.

we can comment these if that seems fine.

kminder (approved):
Approved.

PR #5877, opened 2023-04-28
VZ-9459.  External CM e2e tests

michael-cico:


platform-operator/controllers/verrazzano/component/helm/helm_component.go
```diff
@@ -213,12 +213,20 @@ func (h HelmComponent) GetMinVerrazzanoVersion() string {
...
-	if context.IsDryRun() {
-		context.Log().Debugf("IsInstalled() dry run for %s", h.ReleaseName)
+func (h HelmComponent) IsInstalled(ctx spi.ComponentContext) (bool, error) {
+	if ctx.IsDryRun() {
+		ctx.Log().Debugf("IsInstalled() dry run for %s", h.ReleaseName)
 		return true, nil
 	}
-	installed, _ := helm.IsReleaseInstalled(h.ReleaseName, h.resolveNamespace(context))
+	// check to make sure we own the namespace first
+	vzManaged, err := namespace.CheckIfVerrazzanoManagedNamespaceExists(h.resolveNamespace(ctx))
```
pfmackin, line 222: are you sure that every namespace for the components that we install have the verrazzano managed label?

pkg/istio/istioctl.go
```diff
@@ -90,6 +92,15 @@ func Uninstall(log vzlog.VerrazzanoLogger) (stdout []byte, stderr []byte, err er
 
 // IsInstalled returns true if Istio is installed
 func IsInstalled(log vzlog.VerrazzanoLogger) (bool, error) {
+	// check to make sure we own the namespace first
+	vzManaged, err := namespace.CheckIfVerrazzanoManagedNamespaceExists(constants.IstioSystemNamespace)
```
pfmackin, line 96: Why are we changing Isito code?

ci/scripts/install_third_party_components.sh
```diff
@@ -0,0 +1,49 @@
...
+
+echo "Installing cert-manager via helm chart"
+
+kubectl create ns cert-manager
+kubectl apply -f platform-operator/thirdparty/manifests/cert-manager
+
+controllerTag=$(cat platform-operator/verrazzano-bom.json | jq '.components[] | select(.name=="cert-manager")' | jq '.subcomponents[0].images[] | select(.image=="cert-manager-controller")' | jq .tag -r)
+cainjectorTag=$(cat platform-operator/verrazzano-bom.json | jq '.components[] | select(.name=="cert-manager")' | jq '.subcomponents[0].images[] | select(.image=="cert-manager-cainjector")' | jq .tag -r)
+webhookTag=$(cat platform-operator/verrazzano-bom.json | jq '.components[] | select(.name=="cert-manager")' | jq '.subcomponents[0].images[] | select(.image=="cert-manager-webhook")' | jq .tag -r)
+helm upgrade cert-manager -n cert-manager platform-operator/thirdparty/charts/cert-manager \
```
michael-cico, line 15: why are we using our own CM chart instead of just doing a `kubect apply` of a an OS release of CM?

michael-cico: I know we talked about it, I just can't remember why.

michael-cico: e.g.,

```
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
```

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -73,7 +73,8 @@ func CreateOrUpdateVMI(ctx spi.ComponentContext, updateFunc VMIMutateFunc) error
 	if vzcr.IsNGINXEnabled(effectiveCR) {
 		dnsSuffix, err = vzconfig.GetDNSSuffix(ctx.Client(), effectiveCR)
 		if err != nil {
-			return ctx.Log().ErrorfNewErr("Failed getting DNS suffix: %v", err)
+			err = fmt.Errorf("Failed getting DNS suffix: %v", err)
+			return ctx.Log().ErrorfThrottledNewErr("Failed getting DNS suffix: %v", err)
```
michael-cico, line 77: why not just return `ErrorfThrottledNewErr`?  Seems like the `fmt.Error` is duplicating the throttled error.

platform-operator/controllers/verrazzano/component/helm/helm_component.go
michael-cico: If we aren't labeling all the namespaces we install, we probably should right?

pfmackin: yes, install and upgrade

pkg/istio/istioctl.go
mbenmclean: Try to keep it consistent with the other IsInstalled code. I think we should check to make sure we own the namespace before installing something in it.

ci/scripts/install_third_party_components.sh
mbenmclean: I just did it because it's easier. I can change that if we think it's better

platform-operator/controllers/verrazzano/component/common/vmi.go
mbenmclean: Oh I should delete the fmt.Error line

ci/scripts/install_third_party_components.sh
mbenmclean: I mean the base chart we have for cert-manger is just a third party chart we copied over

michael-cico: For now it's OK, but we may need to update this; I think to validate the final changes we'll need to install CM into a different namespace from `cert-manager`.

pfmackin (approved):

platform-operator/controllers/verrazzano/component/networkpolicies/netpol_component.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
michael-cico, line 1: copyright should be `2022, 2023,` right?

mbenmclean (approved):

PR #5891, opened 2023-05-02
VZ-8962 Change Rancher config to reuse existing user token when applicable

stevshap:
This PR modifies the RancherConfig struct to look for an existing user token before automatically generating a new one. This reduces the number of Rancher API requests by ~50% and fixes the reported ArgoCD bug


platform-operator/controllers/verrazzano/component/clusteroperator/cluster_operator_test.go
```diff
@@ -350,6 +356,9 @@ func adminTokenMock(httpMock *mocks.MockRequestSender) *mocks.MockRequestSender
 
 // TestCreateVZClusterUser tests the creation of the VZ cluster user through the Rancher API
 func TestCreateVZArgoCDClusterUser(t *testing.T) {
+	// clear any cached user auth tokens when the test completes
+	defer rancherutil.DeleteStoredTokens()
```
fredtibbitts, line 360: I think you can remove this since you have the same `defer` below in the `t.Run`?

pkg/rancherutil/rancher_config.go
```diff
@@ -56,6 +58,9 @@ var DefaultRetry = wait.Backoff{
 	Jitter:   0.1,
 }
 
+var userToken = make(map[string]string)
```
fredtibbitts, line 61: Might be nice to have a comment as to what this is used for, something like "// userToken is an in-memory cache of Rancher user names to API authentication tokens". 

Also I'm not crazy about the name. How about `userTokens` or `userTokenCache`?

fredtibbitts (approved):

PR #5896, opened 2023-05-03
Draft: Karak/add fluent operator install

karan56625:


platform-operator/thirdparty/manifests/fluent-operator/filter-parser/rancher-filter-parser.yaml
```diff
@@ -0,0 +1,163 @@
...
+    fluentbit.fluent.io/enabled: "true"
+  name: rancherFilter
+  namespace: {{ .cattleFleetNamespace }}
+spec:
+  filters:
+    - parser:
+        keyName: log
+        reserveData: true
+        parser: rancherFilter1,rancherFilter2,rancherFilter3
+  match: "kubernetes.**cattle-system** kubernetes.**fleet-system** kubernetes.**local-path-provisioner**"
```
adiforluls, line 70: The match rule doesn't look correct, this is fluentd way. FluentBit tags are like this
`kube.var.log.containers.<pod_name>_<namespace_name>_<container_name>.log`

adiforluls: Also I don't think you can define multiple match patterns in a single match rule, you need a `matchRegex` for that

platform-operator/thirdparty/manifests/fluent-operator/application-output.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: fluentbit.fluent.io/v1alpha2
+kind: ClusterOutput
+metadata:
+  labels:
+    fluentbit.fluent.io/enabled: "true"
+  name: verrazzano-application
+spec:
+  matchRegex: '^(.*)\.(.*)_(?!(keycloak|istio-system|verrazzano\-system|...))(.*)\-.*$'
```
adiforluls, line 10: The trailing `...` will need to be replaced by actual system namespaces

platform-operator/controllers/controller_utils.go
```diff
@@ -87,3 +92,31 @@ func ProcDeletedOverride(statusUpdater vzstatus.Updater, c client.Client, vz *in
...
+		args["systemNamespace"] = vzConstans.VerrazzanoSystemNamespace
+		args["monitoringNamespace"] = vzConstans.VerrazzanoMonitoringNamespace
+		args["certManagerNamespace"] = vzConstans.CertManagerNamespace
+		args["istioNamespace"] = vzConstans.IstioSystemNamespace
+		args["keycloakNamespace"] = vzConstans.KeycloakNamespace
+		args["ingressNamespace"] = constants.IngressNginxNamespace
+		args["rancherNamespace"] = vzConstans.RancherSystemNamespace
+		args["cattleFleetNamespace"] = "cattle-fleet-system"
+		args["cattleFleetLocalNamespace"] = "cattle-fleet-local-system"
+		filterAndParser := path.Join(config.GetThirdPartyManifestsDir(), "fluent-operator/"+fileName)
```
adiforluls, line 110: This might not go in the final code that's going to be checked-in but probably wouldn’t our templates/manifests in thirdparty dir.

platform-operator/controllers/verrazzano/component/registry/registry.go
```diff
@@ -68,6 +69,7 @@ func ResetGetComponentsFn() {
 func InitRegistry() {
 	componentsRegistry = []spi.Component{
 		networkpolicies.NewComponent(), // This must be first, don't move it.  see netpol_components.go
+		fluentoperator.NewComponent(),
```
adiforluls, line 72: Note: Making fluent-operator the 2nd component in this list guarantees that it will be the 2nd component to be upgraded after netpol, but doesn't really guarantee anything about install (install is not sequential, upgrade is). You'd also need to make every other component dependent on Fluent Operator such that it gets installed first.

platform-operator/thirdparty/manifests/fluent-operator/filter-parser/oam-filter-parser.yaml
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+apiVersion: fluentbit.fluent.io/v1alpha2
```
adiforluls, line 3: You'd also need to create a namespace `FluentBitConfig` CR per system namespace to select these configs. We might need additional helm value to add `namespaceFluentBitCfgSelector` in FluentBit CR via helm override in the upstream I'm thinking.

platform-operator/thirdparty/manifests/fluent-operator/filter-parser/oam-filter-parser.yaml
```diff
@@ -0,0 +1,42 @@
...
+    fluentbit.fluent.io/enabled: "true"
+  name: oamFilter
+  namespace: {{ .systemNamespace }}
+spec:
+  filters:
+    - parser:
+        keyName: log
+        reserveData: true
+        parser: oamParser1,oamParser2
+  match: "kubernetes.**oam-kubernetes-runtime**verrazzano-system_oam-kubernetes-runtime**"
```
adiforluls, line 16: Match rule needs to be changed accordingly for FluentBit.

platform-operator/controllers/verrazzano/component/thanos/thanos_component.go
```diff
@@ -153,6 +155,22 @@ func (t ThanosComponent) PreUpgrade(ctx spi.ComponentContext) error {
 	return t.HelmComponent.PreUpgrade(ctx)
 }
 
+// PostInstall deletes the remaining resources after uninstallation
```
adiforluls, line 158: Probably the comment is about post uninstall?

platform-operator/controllers/verrazzano/component/thanos/thanos_component.go
```diff
@@ -153,6 +155,22 @@ func (t ThanosComponent) PreUpgrade(ctx spi.ComponentContext) error {
...
+func (t ThanosComponent) PostInstall(context spi.ComponentContext) error {
+	if err := controllers.ExecuteFluentFilterAndParser(context, fluentOperatorFilterFile, false); err != nil {
+		return err
+	}
+	return nil
+}
+
+// PostUninstall deletes the remaining resources after uninstallation
+func (t ThanosComponent) PostUninstall(context spi.ComponentContext) error {
+	if err := controllers.ExecuteFluentFilterAndParser(context, fluentOperatorFilterFile, true); err != nil {
```
adiforluls, line 168: This seems wrong?

karan56625: it deletes the resource after component uninstall.

platform-operator/controllers/verrazzano/component/fluentoperator/fluentoperator.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+const (
+	fluentbitDaemonset         = "fluent-bit"
+	fluentOperatorFullImageKey = "operator.container.repository"
+	fluentOperatorInitImageKey = "fluentbit.image.repository"
+	fluentBitFullImageKey      = "operator.initcontainer.repository"
+	fluentOperatorImageTag     = "operator.container.tag"
+	fluentOperatorInitTag      = "fluentbit.image.tag"
+	fluentBitImageTag          = "operator.initcontainer.tag"
```
adiforluls, line 29: Would wanna swap the above two const? `fluentOperatorInitTag = "operator.initcontainer.tag"`

PR #5933, opened 2023-05-05
VZ-9459.  Reorganize validators in CertManager components

michael-cico:
Reorganize the validators according to component function and create a common package for shared helpers and test utils.

In preparation for the ExternalCertManager changes.

The are 3 components:

- certmanager/controller - Manages the lifecycle of the Verrazzano CertManager instance
- certmanager/config - Manages the Verrazzano ClusterIssuer configuration
- certmanager/ocidns - Manages the lifecycle of the VZ OCI DNS webhook for CertManager

The `certmanager/common` package is created to share some common configuration and validation helpers, and some common test helpers.


platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -0,0 +1,190 @@
...
+// is not greater than 64 characters
+func ValidateLongestHostName(effectiveCR runtime.Object) error {
+	envName := getEnvironmentName(effectiveCR)
+	dnsSuffix, wildcard := getDNSSuffix(effectiveCR)
+	spaceOccupied := preOccupiedspace
+	longestHostName := fmt.Sprintf("%s.%s.%s", longestSystemURLPrefix, envName, dnsSuffix)
+	if len(longestHostName) > 64 {
+		if wildcard {
+			spaceOccupied = spaceOccupied + len(dnsSuffix)
+			return fmt.Errorf("spec.environmentName %s is too long. For the given configuration it must have at most %v characters", envName, 64-spaceOccupied)
```
pfmackin, line 59: Errors should start with "Failed.. "  This is true for the other errors that don't follow that rule.

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -0,0 +1,190 @@
...
+		dnsSuffix = cr.Spec.Components.DNS.OCI.DNSZoneName
+	} else if cr.Spec.Components.DNS.External != nil {
+		wildcard = false
+		dnsSuffix = cr.Spec.Components.DNS.External.Suffix
+	}
+	return dnsSuffix, wildcard
+}
+
+// ValidateConfiguration Checks if the configuration is valid and is a CA configuration
+// - returns true if it is a CA configuration, false if not
```
pfmackin, line 104: comment is wrong

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -0,0 +1,190 @@
...
+// - returns an error if both CA and ACME settings are configured
+func ValidateConfiguration(certConfig v1beta1.Certificate) (err error) {
+	// Check if Ca or Acme is empty
+	caNotEmpty, err := checkExactlyOneIssuerConfiguration(certConfig)
+	if err != nil {
+		return err
+	}
+
+	if caNotEmpty {
+		if err := validateCAConfiguration(certConfig.CA, constants.CertManagerNamespace); err != nil {
```
pfmackin, line 114: can be simplified 
return validate ...

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -0,0 +1,190 @@
...
+			return err
+		}
+		return nil
+	}
+
+	if err := validateAcmeConfiguration(certConfig.Acme); err != nil {
```
pfmackin, line 120: can be simplified 
return validate ...

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -0,0 +1,190 @@
...
+func checkExactlyOneIssuerConfiguration(certConfig v1beta1.Certificate) (isCA bool, err error) {
+	// Check if Ca or Acme is empty
+	caNotEmpty := certConfig.CA != v1beta1.CA{}
+	acmeNotEmpty := certConfig.Acme != v1beta1.Acme{}
+	if caNotEmpty && acmeNotEmpty {
+		return false, errors.New("certificate object Acme and CA cannot be simultaneously populated")
+	} else if !caNotEmpty && !acmeNotEmpty {
+		return false, errors.New("Either Acme or CA certificate authorities must be configured")
+	}
+	return caNotEmpty, nil
```
pfmackin, line 135: What if CA is empty but ACME is not?  Do you want to return false?

pfmackin (approved):

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
michael-cico: It's the original behavior; I think the assumption was we only want to validate the CA if it's the one configured or the ACME one otherwise.

michael-cico: I'll try to make that clearer in the code.

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
michael-cico: Fixed and pushed

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
michael-cico: Fixed and pushed

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
michael-cico: The way it is/was written is that we only expect one or the other to be configured; it's how we know which way to set up the ClusterIssuer.

The calling code uses the boolean to know if it's a CA config, if not ACME is assumed and we use that.  then we validate the ACME config.  The one thing I did change is this also used to do the validation at the same time, I separated the 2 because the check needs to be made at times where the validation doesn't.

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
michael-cico: Fixed and pushed

pfmackin (approved):

PR #5969, opened 2023-05-10
Thanos Istio proxy container Prometheus scrape fix

fredtibbitts:
While I was doing manual testing of the Prometheus Operator upgrade, I noticed that Prometheus was not able to scrape metrics from the Thanos Istio proxy containers in the Query and Query Frontend pods. My test cluster had Calico enabled.

The problem was the NetworkPolicy resources were blocking access to the Istio proxy metrics ports in those pods. I have made the following changes to the Thanos NetworkPolicies:
- Updated the Query and Query Frontend policies to add a label match selector on the component. Without this, the policies were selecting all Thanos pods.
- Updated the Query and Query Frontend policies to include the Istio proxy metrics port.

I tested by installing in cluster with Calico enabled and validated that the scrape targets for the Thanos and Istio proxy containers in the Thanos pods were healthy. I also navigated to the Query Frontend UI and validated that the Store Gateway showed healthy in the list of stores (confirming that the Query NetworkPolicy was not needed).

### DIFF
--- a/platform-operator/thirdparty/charts/thanos/templates/verrazzano/networkpolicy.yaml
+++ b/platform-operator/thirdparty/charts/thanos/templates/verrazzano/networkpolicy.yaml
@@ -51,9 +51,13 @@ spec:
         # api port
         - port: 10902
           protocol: TCP
+        # Istio proxy metrics port
+        - port: 15090
+          protocol: TCP
   podSelector:
     matchLabels:
       app.kubernetes.io/name: {{ include "common.names.fullname" . }}
+      app.kubernetes.io/component: query-frontend
   policyTypes:
     - Ingress
 ---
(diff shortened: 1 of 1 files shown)


davisgomes (approved):

pfmackin (dismissed):

desagar (approved):

platform-operator/thirdparty/charts/thanos/templates/verrazzano/networkpolicy.yaml
```diff
@@ -95,8 +99,12 @@ spec:
         # api port
```
desagar, line 99: a few lines above this, there's a rule for ingress to the storegateway API endpoint 10901. This rule will only match query now, so is that going to be an issue?

fredtibbitts: Good question, I think what we really need is a separate NetworkPolicy for Thanos Store Gateway. These two NetworkPolicies were applying to all Thanos pods before my change, and I think we need separate policies. Instead of trying to address that here, I'm just going to fix the original bug I found (not being able to scrape Istio proxy) and remove the component label selectors. I'll create a ticket to clean up the NetworkPolicies and have one per component.

desagar (approved):

PR #5974, opened 2023-05-10
VZ-9645.  Use verrazzano-system namespace for OCI DNS webhook

michael-cico:
Run the Verrazzano OCI DNS solver webhook in the `verrazzano-system` namespace
- give it permission to read secrets in the Cert-Manager `clusterResourceNamespace` (e.g., the OCI DNS auth secret)
- modify the RBAC permissions to allow Cert-Manager to discover and invoke the webhook APIService
- turn off Istio injection for the deployment
- Add E2E tests cases for external-dns and the webhook deployments if they are enabled

Other changes:
- Modify `IsEnabled` to check that the webhook is deployed IFF CM is enabled, OCI DNS is configured, and ACME is configured
- align validation functions on v1alpha1 API for now.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
@@ -42,8 +42,25 @@ func ResetCoreV1ClientFunc() {
 
 // IsCA - Check if cert-type is CA, if not it is assumed to be Acme
 func IsCA(compContext spi.ComponentContext) (bool, error) {
-	comp := vzapi.ConvertCertManagerToV1Beta1(compContext.EffectiveCR().Spec.Components.CertManager)
-	return checkExactlyOneIssuerConfiguration(comp.Certificate)
+	componentSpec := compContext.EffectiveCR().Spec.Components
+	if componentSpec.CertManager != nil {
+		return checkExactlyOneIssuerConfiguration(componentSpec.CertManager.Certificate)
+	}
+	// If the stanza isn't present assume the Self-signed CA issuer config
+	return true, nil
+}
+
+// IsACMEConfig - Check if cert-type is ACME
+func IsACMEConfig(vz interface{}) (bool, error) {
+	vzv1alpha1, err := convertIfNecessary(vz)
+	if err != nil {
+		return false, err
+	}
+	if vzv1alpha1.Spec.Components.CertManager != nil {
+		isCA, err := checkExactlyOneIssuerConfiguration(vzv1alpha1.Spec.Components.CertManager.Certificate)
+		return !isCA, err
+	}
+	return false, nil
 }
 
 // ValidateLongestHostName - validates that the longest possible host name for a system endpoint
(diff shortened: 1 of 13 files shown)


platform-operator/controllers/verrazzano/component/certmanager/ocidns/certmanagerocidns.go
```diff
@@ -26,3 +29,24 @@ func isCertManagerOciDNSReady(context spi.ComponentContext) bool {
 	prefix := fmt.Sprintf("Component %s", context.GetComponent())
 	return ready.DeploymentsAreReady(context.Log(), context.Client(), deployments, 1, prefix)
 }
+
+func (c certManagerOciDNSComponent) postUninstall(ctx spi.ComponentContext) error {
```
pfmackin, line 33: What about post upgrade?

platform-operator/helm_config/charts/verrazzano-cert-manager-ocidns-webhook/values.yaml
```diff
@@ -11,6 +11,7 @@ groupName: verrazzano.io
 
 certManager:
   namespace: cert-manager
+  clusterResourceNamespace: cert-manager
```
mbenmclean, line 14: How do we ensure that the cert-manager namespace exists? Can this chart get installed if cert-manager is disabled but ExternalDNS is enabled? Isn't that why you moved externalDNS to verrazzano system. I thought the whole point of that was so we wouldn't put stuff in the cert-manager namespace

michael-cico: There's more work on this in the other branch that is implementing the new API changes; we probably need some validation that whatever CM namespace(s) they point us at exist. 

I did notice I need to update the `IsEnabled` check in this component to only deploy IFF both
- OCI DNS is configured
- the ACME issuer is configured

Right now it deploys if OCI DNS is configured.

As to putting stuff in the clusterResourceNamespace, we have to put cluster-scoped secrets into that namespace.  It's a Cert-Manager convention.  When the challenge comes in for a ClusterIssuer-generated certificate, the namespace is clusterResourceNamespace.

platform-operator/controllers/verrazzano/component/certmanager/ocidns/certmanagerocidns.go
michael-cico: Yes, we'll need that hook.  Right now it likely works because we flag it to install before upgrade.

michael-cico: actually, not sure what we'd do post-upgrade... this is post uninstall, not post-install.  Were you thinking this was post-install?

michael-cico: I'll add a pre-upgrade in a follow up commit, looks like we need that.

platform-operator/controllers/verrazzano/component/certmanager/common/validate.go
```diff
@@ -133,7 +161,7 @@ func checkExactlyOneIssuerConfiguration(certConfig v1beta1.Certificate) (isCACon
 	return caNotEmpty, nil
```
pfmackin, line 161: shouldn't this return true, nil?
what if caNotEmpty is false, byt acmeNotEmpty is true

michael-cico: The above conditionals should validate that they are not the same value (both true, or both false), but one or the other can still be true.

Probably the conditional can be simplified to

```
if caNotEmpty == acmeNotEmpty {
  // error, only one or the other can be set
}
return caNotEmpty, nil
```

pfmackin: I still don’t understand, i think your conditional suggestion is correct, but it should return true unless i am missing something.

pfmackin (approved):

PR #5985, opened 2023-05-12
VZ-9598.  API updates for External CM configuration

michael-cico:
Add API for integrating with an user-managed Cert-Manager instance.

Adds
- External CM configuration to inform Verrazzano where CM is installed, and where the `clusterResourceNamespace` is
- Allow specifying the CM ServiceAccount (used to set proper RBAC permissions on the OCI DNS webhook when it is in use)
- Allow overrides for the OCI DNS webhook for both the (Verrazzano-managed) CertManagerComponent and ExternalCertManagerComponent configurations


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -349,6 +350,11 @@ type ComponentSpec struct {
 	// +optional
 	Elasticsearch *ElasticsearchComponent `json:"elasticsearch,omitempty"`
 
+	// Defines the settings for an externally-managed Cert-Manager instance to be used by this Verrazzano installation;
```
pfmackin, line 353: normal format for comment is "ExternalCertManager ..."  but i see that we not following that convention in general.

pfmackin (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
michael-cico: Will fix this in a follow-up PR.

PR #5994, opened 2023-05-15
VZ-9702.  Limit OCI DNS webhook secrets access to configured auth secret

michael-cico:
Updates the RBAC for the OCI DNS Auth webhook to scope access to secrets in the Cert-Manager `clusterResourceNamespace` to only what is required/configured by Verrazzano.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/ocidns/certmanagerocidns.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/ocidns/certmanagerocidns.go
@@ -5,6 +5,7 @@ package ocidns
 
 import (
 	"fmt"
+	"github.com/verrazzano/verrazzano/pkg/bom"
 	"github.com/verrazzano/verrazzano/pkg/constants"
 	"github.com/verrazzano/verrazzano/pkg/k8s/ready"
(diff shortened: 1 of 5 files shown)


platform-operator/controllers/verrazzano/component/certmanager/ocidns/certmanagerocidns.go
```diff
@@ -30,6 +31,32 @@ func isCertManagerOciDNSReady(context spi.ComponentContext) bool {
...
 
+func appendOCIDNSOverrides(ctx spi.ComponentContext, _ string, namespace string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+
+	secretName, err := GetOCIDNSSecretName(ctx)
+	if err != nil {
+		return kvs, err
+	}
+
+	overrides := []bom.KeyValue{
+		{Key: "ociAuthSecrets[0]", Value: secretName},
```
pfmackin, line 42: is that really the key?

michael-cico: yes.  Should it be different?  This is a key into our own chart for the webhook.

pfmackin (approved):

PR #6043, opened 2023-05-19
Rancher Image update

anders-swanson:
- OKE Kontainer driver is built in to rancher, always active
- New Rancher-UI version with compartment tree fix
- custom drivers are auto updated during rancher upgrade.

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -148,13 +148,13 @@
             {
               "image": "rancher",
               "dashboard": "v2.7.2-20230509143745-870a44e17",
-              "tag": "20230519180423-7b84dcaeb",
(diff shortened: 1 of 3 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -148,13 +148,13 @@
             {
               "image": "rancher",
               "dashboard": "v2.7.2-20230509143745-870a44e17",
-              "tag": "v2.7.3-20230517224130-9edb4a892",
+              "tag": "20230519180423-7b84dcaeb",
```
mgianatagh, line 151: This tag doesn't look right, it should be prefixed with 2.7.3.

anders-swanson: updated the Rancher image.

mgianatagh (approved):

PR #6072, opened 2023-05-24
Update the ArgoCD chart and image

davisgomes:
**Changes**
- Updated the ArgoCD image to the latest BFS
- Updated the ArgoCD chart to a version to match the image BFS


platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
fredtibbitts, line 1: Looks like this mysql chart change snuck in here.

platform-operator/thirdparty/charts/argo-cd/Chart.yaml
```diff
@@ -1,28 +1,31 @@
...
-appVersion: v2.5.3
-kubeVersion: ">=1.21.0-0"
-description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
-name: argo-cd
-version: 5.16.1
+appVersion: v2.7.2
+dependencies:
+- condition: redis-ha.enabled
+  name: redis-ha
+  repository: https://dandydeveloper.github.io/charts/
```
fredtibbitts, line 13: Can we change this to a file:// URL or something, so it references the `charts/argo-cd/charts/redis-ha` directory? Otherwise I don't think will work in an air-gapped env if `redis-ha.enabled` is set.

davisgomes: Yes. Good point

platform-operator/thirdparty/charts/mysql/templates/deployment_cluster.yaml
davisgomes: Thanks. I removed it.

fredtibbitts: Is `file://redis-ha` right? I can't tell from looking through the docs and other links if it should be that, or one of these:
`file://charts/redis-ha`
`file://../charts/redis-ha`
`file://../redis-ha`

davisgomes: I think this is correct based on the Thanos local chart: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/thirdparty/charts/thanos/Chart.yaml#L8

fredtibbitts: Nice, thanks.

fredtibbitts (approved):

PR #6074, opened 2023-05-24
VZ-9724.  Add new CertManager components to profiles, implement Certificate conversion 

michael-cico:
Adds the new CertManager components to the Verrazzano profiles:

* ClusterIssuerComponent
* CertManagerWebhookOCI

In addition, logic has been added to the `effectiveCR` merge in the VPO to  preserve legacy `certManager.certificate` configurations and convert it forward to the `clusterIssuer` when the the effective configuration is computed.  This will allow VPO components internally to align on the `ClusterIssuerComponent` and phase out the `CertManagerComponent.Certificate` field.

Additional changes:
- Some additional `ClusterIssuerComponent` utils have been added in the API packages
- Added support for merging the `InstallOverrides` for the `CertManagerWebhookOCI` component
- Added `vzcr.IsClusterIssuerEnabled` and `vzcr.IsCertManagerWebhookOCIEnabled` functions
- fixed a bug in the `none.yaml` profiles where the CertManager override was wiping out the `base.yaml` values for the `certificate` field (merge strategy for that field is replace)

### DIFF
--- a/pkg/profiles/merge.go
+++ b/pkg/profiles/merge.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package profiles
@@ -192,6 +192,12 @@ func AppendComponentOverrides(actual, profile *v1alpha1.Verrazzano) {
 	profileCertManager := profile.Spec.Components.CertManager
 	if actualCertManager != nil && profileCertManager != nil {
 		actualCertManager.ValueOverrides = mergeOverrides(actualCertManager.ValueOverrides, profileCertManager.ValueOverrides)
+	}
+
+	actualOCIWebhook := actual.Spec.Components.CertManagerWebhookOCI
+	profileOCIWebhook := profile.Spec.Components.CertManagerWebhookOCI
+	if actualOCIWebhook != nil && profileOCIWebhook != nil {
+		actualOCIWebhook.ValueOverrides = mergeOverrides(actualOCIWebhook.ValueOverrides, profileOCIWebhook.ValueOverrides)
 	}
 
 	actualCoherenceOperator := actual.Spec.Components.CoherenceOperator
--- a/platform-operator/apis/verrazzano/v1alpha1/issuer_utils.go
+++ b/platform-operator/apis/verrazzano/v1alpha1/issuer_utils.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package v1alpha1
+
+import (
+	"fmt"
+	"github.com/verrazzano/verrazzano/pkg/constants"
+)
+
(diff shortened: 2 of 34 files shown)


pkg/profiles/merge.go
```diff
@@ -194,6 +194,12 @@ func AppendComponentOverrides(actual, profile *v1alpha1.Verrazzano) {
 		actualCertManager.ValueOverrides = mergeOverrides(actualCertManager.ValueOverrides, profileCertManager.ValueOverrides)
 	}
 
+	actualOCIWebhook := actual.Spec.Components.CertManagerWebhookOCI
+	profileOCIWebhook := profile.Spec.Components.CertManagerWebhookOCI
+	if actualOCIWebhook != nil && profileOCIWebhook != nil {
+		actualOCIWebhook.ValueOverrides = mergeOverrides(actualOCIWebhook.ValueOverrides, profileOCIWebhook.ValueOverrides)
```
pfmackin, line 200: Does this have any precedence ramifications, meaning does the order matter? I would think so, and would expect the actual to take precedence.

platform-operator/apis/verrazzano/v1beta1/issuer_utils.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
pfmackin, line 1: I am surprised that we have to duplicate this from v1alpha1.  Don't we convert to one or the other internally and work off that?

platform-operator/apis/verrazzano/v1beta1/issuer_utils.go
```diff
@@ -0,0 +1,48 @@
...
+func (c *ClusterIssuerComponent) IsCAIssuer() (bool, error) {
+	if c.CA != nil && c.LetsEncrypt != nil {
+		return false, fmt.Errorf("Illegal state, can not configure CAIssuer and LetsEncrypt issuer simultaneously")
+	}
+	return c.CA != nil, nil
+}
+
+// IsLetsEncryptIssuer returns true of the issuer configuration is for a LetsEncrypt issuer, or an error if it is misconfigured
+func (c *ClusterIssuerComponent) IsLetsEncryptIssuer() (bool, error) {
+	isCAIssuer, err := c.IsCAIssuer()
```
pfmackin, line 33: why not just return c.LetsEncrypt != nil ?

pfmackin (approved):

platform-operator/apis/verrazzano/v1beta1/issuer_utils.go
michael-cico: We do, I supposed I could remove it from v1alpha1.  I think I mainly wanted to have it for both APIs, and I think it has to live with the type.  I can revisit after.

platform-operator/apis/verrazzano/v1alpha1/issuer_utils.go
```diff
@@ -0,0 +1,47 @@
...
+		IssuerConfig: IssuerConfig{
+			CA: &CAIssuer{
+				SecretName: constants.DefaultVerrazzanoCASecretName,
+			},
+		},
+	}
+}
+
+// IsCAIssuer returns true of the issuer configuration is for a CA issuer, or an error if it is misconfigured
+func (c *ClusterIssuerComponent) IsCAIssuer() (bool, error) {
```
mbenmclean, line 23: Is this called from a validator. If it's called from reconcile it will be too late obviously

PR #6080, opened 2023-05-25
VZ-9564 - New OCNE CAPI images with module operator addon install 

theabmitra:

### DIFF
--- a/platform-operator/capi/bootstrap-ocne/v0.1.0/bootstrap-components.yaml
+++ b/platform-operator/capi/bootstrap-ocne/v0.1.0/bootstrap-components.yaml
@@ -30,122 +30,122 @@ spec:
           namespace: capi-ocne-bootstrap-system
           path: /convert
       conversionReviewVersions:
-      - v1
-      - v1beta1
+        - v1
+        - v1beta1
   group: bootstrap.cluster.x-k8s.io
   names:
     categories:
-    - cluster-api
+      - cluster-api
     kind: OCNEConfig
     listKind: OCNEConfigList
     plural: ocneconfigs
     singular: ocneconfig
   scope: Namespaced
   versions:
-  - additionalPrinterColumns:
-    - description: Cluster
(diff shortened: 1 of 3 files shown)


platform-operator/capi/control-plane-ocne/v0.1.0/control-plane-components.yaml
```diff
@@ -2872,6 +2917,17 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
jmaron99, line 2920: I don't think this modification was necessary since this charts are not created via the same mechanism as in the cluster provider repo but I suppose it does mimic the same result.

platform-operator/capi/bootstrap-ocne/v0.1.0/bootstrap-components.yaml
```diff
@@ -2534,8 +2534,13 @@ spec:
...
         command:
         - /manager
-        image: ghcr.io/verrazzano/cluster-api-ocne-bootstrap-controller:v0.1.0-20230421153459-16a5418
-        imagePullPolicy: Always
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        image: ghcr.io/verrazzano/cluster-api-ocne-bootstrap-controller:v0.1.0-20230525020552-c2d7135
```
jmaron99, line 2542: Is this image value overwritten by the value in the clusterctl yaml being generated by the VPO processing?

bobdonat: It should be.  The clusterctl.yaml values are used for overrides,

platform-operator/capi/control-plane-ocne/v0.1.0/control-plane-components.yaml
theabmitra: this modification was not international, this is a result of the `make generate` so that means  in capi main this was already present.

bobdonat (approved):

PR #6094, opened 2023-05-25
VZ-9796 (VZ-9747) Allow the application operator to be run in cluster agent mode, which only starts up the managed cluster agent. Also made the mcagent a separate controller instead of a goroutine.

desagar:


application-operator/main.go
```diff
@@ -74,6 +77,11 @@ func main() {
 		"Runs in webhook mode; if false, runs the main operator reconcile loop")
 	flag.BoolVar(&runWebhookInit, "run-webhook-init", false,
 		"Runs the webhook initialization code")
+	flag.BoolVar(&runAppOperator, "run-app-operator", true,
```
michael-cico, line 80: Looks like it's possible to run both with `true`... should it be either-or?  I'm assuming that enabling both gives us what we have now.

desagar: You are correct - as we discussed in today's review, I plan to make it either-or, once the new component is ready. If we do it now while we're only running one copy of the app operator, we'll break stuff.

application-operator/mcagent/mcagent.go
```diff
@@ -30,93 +33,136 @@ import (
...
+func (r *Reconciler) createAgentPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return r.isAgentSecret(e.Object)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return r.isAgentSecret(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return r.isAgentSecret(e.ObjectNew)
```
davisgomes, line 75: Why do we use `e.ObjectNew` here and `e.Object` for the rest?

application-operator/mcagent/mcagent.go
```diff
@@ -30,93 +33,136 @@ import (
...
+		Log:                 r.Log,
+		Context:             ctx,
+		ProjectNamespaces:   []string{},
+		StatusUpdateChannel: r.AgentChannel,
+		ManagedClusterName:  managedClusterName,
+	}
+
+	// Read current agent state from config map
+	mcAgentStateConfigMap := corev1.ConfigMap{Data: map[string]string{}}
+	if err := r.Get(ctx, mcAgentStateConfigMapName, &mcAgentStateConfigMap); client.IgnoreNotFound(err) != nil {
```
davisgomes, line 132: I'm not sure I've found where this configmap gets created. Should we try to create it if we have not found it, or would this lead to a bad state?

application-operator/mcagent/mcagent.go
pfmackin: because it is an update

application-operator/mcagent/mcagent.go
```diff
@@ -30,93 +33,136 @@ import (
...
-		}
-		return nil
+		r.Log.Errorf("failed processing multi-cluster resources: %v", err)
 	}
-	err = validateAgentSecret(&secret)
+	return clusters.NewRequeueWithRandomDelay(requeueDelayMinSeconds, requeueDelayMaxSeconds), nil
+}
+
+// ProcessAgentThread - process one iteration of the agent thread
+func (r *Reconciler) ProcessAgentThread(ctx context.Context, agentSecret corev1.Secret) error {
```
pfmackin, line 117: This is not really a thread, it is inline with the Reconcile code (which controller-runtime runs in a go routine)

application-operator/mcagent/mcagent.go
```diff
@@ -186,6 +256,17 @@ func (s *Syncer) updateVMCStatus() error {
 
 // SyncMultiClusterResources - sync multi-cluster objects
 func (s *Syncer) SyncMultiClusterResources() {
```
pfmackin, line 258: it would be good to have Syncer in syncer.go (or someting).  It is not a good practice to have 2 different method receivers in the same file.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -465,6 +465,31 @@ func createOrUpdateSecretRancherProxy(secret *corev1.Secret, rc *rancherutil.Ran
...
 
+// deleteSecretRancherProxy deletes a secret on a remote cluster through the Rancher Proxy API and
+// returns true if deletion occurred, false otherwise
+func deleteSecretRancherProxy(secret *corev1.Secret, rc *rancherutil.RancherConfig, clusterID string, log vzlog.VerrazzanoLogger) (bool, error) {
+	if secret == nil {
+		return false, log.ErrorNewErr("Failed to get secret, nil value passed to get request")
+	}
+	reqURL := constructSecretURL(secret, rc.Host, clusterID, false)
+	headers := map[string]string{"Authorization": "Bearer " + rc.APIAccessToken}
+	resp, _, err := rancherutil.SendRequest(http.MethodDelete, reqURL, headers, "", rc, log)
```
pfmackin, line 476: Is this deleting a secret on the Admin cluster?  If so, I thought we were using the Kubernetes API?

desagar: This code has been removed after your review, because it caused some other issue - so this is now moot. But to answer the earlier question, we were deleting something on the managed cluster here, using a request exactly like the K8S API request, but via the Rancher proxy.

application-operator/mcagent/mcagent.go
desagar: Moved the receiver methods for syncer to the mcagent_syncer.go.

bryantidd (approved):

PR #6095, opened 2023-05-26
JIRA: VZ-9788 optimize helm release storage and update upgrade completion check

jmaron99:
- Modified helm runtime release storage to 1 previous release to improve helm execution performance
- Update status check at end of upgrade to include upgrade completion condition


pfmackin (approved):

pkg/helm/helmcli.go
```diff
@@ -390,6 +391,7 @@ func getActionConfig(log vzlog.VerrazzanoLogger, settings *cli.EnvSettings, name
 	if err := actionConfig.Init(settings.RESTClientGetter(), namespace, os.Getenv("HELM_DRIVER"), log.Debugf); err != nil {
 		return nil, err
 	}
+	actionConfig.Releases.MaxHistory = 1
```
michael-cico, line 394: Wonder if we should open a task to maybe make this configurable in the future?

michael-cico (approved):

pkg/helm/helmcli.go
```diff
@@ -135,6 +135,7 @@ func Upgrade(log vzlog.VerrazzanoLogger, releaseName string, namespace string, c
 		client.Namespace = namespace
 		client.DryRun = dryRun
 		client.Wait = wait
+		client.MaxHistory = 1
```
mbenmclean, line 138: I thought we decided on 2 for history

jmaron99: That was mentioned but towards the end of the meeting I thought the conclusion had been to stick with 1. @pfmackin - do you remember?

PR #6105, opened 2023-05-27
VZ-9847.  Align remaining VPO components on new CertManager APIs

michael-cico:
Update components in the VPO to use the new clusterIssuer component for certificates configuration.  This largely consists of the Rancher code which sometimes needs to set up additional CA bundles to communicate with other components internally and remote managed clusters.

The `ComponentContext.EffectiveCR` is used to organize the other VPO components around the certificate issuer configuration using the new `ClusterIssuerComponent`.  That change was introduced in an earlier PR.  When the `EffectiveCR` is computed, the existing `CertManager` configuration is merged into the `ClusterIssuerComponent` configuration.  This prevents component code from having to switch between the `CertManagerComponent.Certificate` field and the `ClusterIssuerComponent`, and preserves  backwards compatibility with earlier release configurations.

Changes:
- Convert the Rancher component code over to the `ClusterIssuerComponent` for certificate issuer configuration
- Where appropriate update the unit tests to use a fake `ComponentContext` to compute the EffectiveCR for the scenario
- Update the `spi.NewFakeContext()` constructor to compute the effectiveCRs automatically if appropriate

Other changes:
- Reverted a change to `HelmComponent` that was caching the resolved namespaces of a component; it's not clear if that's necessary at this point and might introduce subtle bugs
- Fixes Rancher Keycloak SSO using a Custom CA using the new APIs (see VZ-9882)
- Adds the Cert-Manager ACME API scheme to the VPO controllerruntime client scheme

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher_install.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher_install.go
@@ -18,8 +18,8 @@ import (
 
 // patchRancherIngress annotates the Rancher ingress with environment specific values
 func patchRancherIngress(c client.Client, vz *vzapi.Verrazzano) error {
-	cm := vz.Spec.Components.CertManager
-	if cm == nil {
+	clusterIssuer := vz.Spec.Components.ClusterIssuer
+	if clusterIssuer == nil {
 		return errors.New("CertificateManager was not found in the effective CR")
 	}
 	dnsSuffix, err := vzconfig.GetDNSSuffix(c, vz)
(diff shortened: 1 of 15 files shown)


platform-operator/controllers/verrazzano/component/rancher/rancher_preinstall.go
```diff
@@ -43,15 +42,35 @@ func createCattleSystemNamespace(log vzlog.VerrazzanoLogger, c client.Client) er
...
+		return err
+	}
+	if isCAIssuer {
+		caSecretNamespace := clusterIssuer.ClusterResourceNamespace
+		caSecretName := clusterIssuer.CA.SecretName
+		namespacedName := types.NamespacedName{
+			Namespace: caSecretNamespace,
+			Name:      caSecretName,
+		}
+		certKey := caCert
```
michael-cico, line 62: The CA bundle key in the secret is different that the default self-signed secret when using a customer-provided CA.  we may need to investigate if this needs to be configurable.

platform-operator/controllers/verrazzano/component/rancher/rancher_install.go
```diff
@@ -43,15 +43,24 @@ func patchRancherIngress(c client.Client, vz *vzapi.Verrazzano) error {
...
 	}
 	return c.Patch(context.TODO(), ingress, ingressMerge)
 }
 
-// addAcmeIngressAnnotations annotate ingress with ACME specific values
+func isLetsEncryptIssuer(clusterIssuer *vzapi.ClusterIssuerComponent) bool {
+	if clusterIssuer == nil {
+		return false
+	}
+	isLEIssuer, _ := clusterIssuer.IsLetsEncryptIssuer()
```
michael-cico, line 59: maybe pass in the logger here and record the error?

pfmackin (approved):

PR #6123, opened 2023-05-31
VZ-9546: Rename component capi to clusterAPI

prasad-shirodkar:
Rename component capi to clusterAPI.


anders-swanson (dismissed):

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -24,13 +24,13 @@ import (
...
 // ComponentName is the name of the component
-const ComponentName = "capi"
+const ComponentName = "clusterAPI"
 
 // Namespace for CAPI providers
 const ComponentNamespace = constants.VerrazzanoCAPINamespace
 
 // ComponentJSONName is the JSON name of the component in CRD
-const ComponentJSONName = "capi"
+const ComponentJSONName = "clusterAPI"
```
bobdonat, line 33: I think this is the name that will show up on the vz resource component status.  Is that the name we want?

prasad-shirodkar: yes that's exactly we want now.

prasad-shirodkar: or should it be something like `cluster-api` keeping in line with other names?

bobdonat: not sure, a similar question came up last week for another component. @kminder what do you think?

kminder: Unfortunately in spec it needs to be clusterAPI and in status needs to be cluster-api.  I'm not sure what needs to change in the code to accomplish this.

bobdonat (dismissed):

bobdonat (approved):

PR #6128, opened 2023-05-31
VZ-9633.  Cert ready-checks, Prom Operator webhook certs with customer-managed CertManager

michael-cico:
The VPO currently does checks on Certificate status when the CertManager component is enabled; however this breaks when using a customer-managed Cert-Manager instance.

Additionally, the Prometheus Operator uses Cert-Manager certs if it is available, however it can no longer rely on checking if the Verrazzano `certManager` component is enabled.

In both cases we shift to using the `clusterIssuer` component to check for these behaviors.  The `clusterIssuer` is what unifies both cases, the Verrazzano-managed and customer-managed Cert-Manager instances.

Also fixes up related unit tests.

Manually verified the behavior in the debugger for the customer-managed CM case, and examined the VPO logs for the default case.

### DIFF
--- a/pkg/k8s/ready/certificates_test.go
+++ b/pkg/k8s/ready/certificates_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 package ready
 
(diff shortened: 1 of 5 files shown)


mbenmclean (approved):

pkg/k8s/ready/certificates_test.go
```diff
@@ -147,16 +151,27 @@ func TestCheckCertificatesNotReadyCertManagerDisabled(t *testing.T) {
...
 		Spec: v1alpha1.VerrazzanoSpec{
 			Components: v1alpha1.ComponentSpec{
-				CertManager: &v1alpha1.CertManagerComponent{Enabled: &cmEnabled},
+				CertManager:   &v1alpha1.CertManagerComponent{Enabled: &disabled},
+				ClusterIssuer: &v1alpha1.ClusterIssuerComponent{Enabled: &disabled},
 			},
 		},
 	}
 
+	//cmEnabled := false
```
pfmackin, line 165: should remove at some point

pfmackin (approved):

fredtibbitts (approved):

PR #6130, opened 2023-06-01
vz 9316 integration test for VZ CLI Custom Resource validation

AbeHern:
Adding an integration test to the existing dynamic config pipeline


ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,6 +228,51 @@ pipeline {
...
+                                    cd "\$WORKSPACE"
+                                    ./vz install --filename "\$WORKSPACE"/ci/dynamic-updates/invalidCR.yaml --manifests ${WORKSPACE}/downloaded-vpo.yaml
+                                    else
+                                    cd "\$GO_REPO_PATH"/verrazzano/tools/vz
+                                    GO111MODULE=on GOPRIVATE=github.com/verrazzano go run main.go install --filename "\$WORKSPACE"/invalid.yaml --manifests ${WORKSPACE}/downloaded-vpo.yaml
+                                    fi
+                                    """
+                                } catch (error) {
+                                    sh """
+                                    kind delete clusters ${CLUSTER_NAME}
```
prasad-shirodkar, line 268: Please remove both statements. The kind cluster is already getting in the `always` section under `post` refer [here](https://github.com/verrazzano/verrazzano/blob/eab88fc5bf5f37fa4235d6497248807c5afaf76d/ci/dynamic-updates/Jenkinsfile#L670).

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,13 +229,63 @@ pipeline {
         stage('Acceptance Tests') {
 
             stages {
+                stage('VZ CLI Custom Resource Validation Tests') {
+                    environment {
+                        KIND_KUBERNETES_CLUSTER_VERSION="${params.KUBERNETES_CLUSTER_VERSION}"
+                        OCI_OS_LOCATION="ephemeral/${env.BRANCH_NAME}/${SHORT_COMMIT_HASH}"
+                    }
+                    steps {
+                        script{
```
michael-cico, line 238: If possible I'd break this out into a function, or better a separate script instead of embedding it inline in the stage.

michael-cico: I'm also wondering if this is overcomplicating the pipeline; it's already too complicated as it is.  If we truly need an e2e test, maybe we should investigate a smaller more focused pipeline for just the CLi where we can do small use cases like this.

ci/scripts/prepare_jenkins_at_environment.sh
```diff
@@ -164,17 +164,19 @@ if [ $? -ne 0 ]; then
...
-  ./vz install --filename ${WORKSPACE}/acceptance-test-config.yaml --manifests ${TARGET_OPERATOR_FILE} --timeout ${INSTALL_TIMEOUT_VALUE}
-else
-  cd ${GO_REPO_PATH}/verrazzano/tools/vz
-  GO111MODULE=on GOPRIVATE=github.com/verrazzano go run main.go install --filename ${VZ_INSTALL_FILE} --manifests ${TARGET_OPERATOR_FILE} --timeout ${INSTALL_TIMEOUT_VALUE}
-fi
-result=$?
-if [[ $result -ne 0 ]]; then
-  exit 1
+# This flag is defaulted to false so that the VZ install proceeds as usual
+if ! [[ ${CR_VALIDATION_TEST} ]]; then
```
michael-cico, line 168: Might be better to generalize this switch, e.g., SKIP_VERRAZZANO_INSTALL, and set it from the pipeline's param

ci/scripts/prepare_jenkins_at_environment.sh
```diff
@@ -29,110 +29,114 @@ TEST_OVERRIDE_SECRET_FILE="./tests/e2e/config/scripts/pre-install-overrides/test
...
-echo "Create Kind cluster"
-cd ${TEST_SCRIPTS_DIR}
-./create_kind_cluster.sh "${CLUSTER_NAME}" "${GO_REPO_PATH}/verrazzano/platform-operator" "${KUBECONFIG}" "${KIND_KUBERNETES_CLUSTER_VERSION}" true true true $INSTALL_CALICO "NONE" ${KIND_NODE_COUNT}
-if [ $? -ne 0 ]; then
-    mkdir $WORKSPACE/kind-logs
-    kind export logs $WORKSPACE/kind-logs
-    echo "Kind cluster creation failed"
-    exit 1
-fi
+clusterNames=$(kind get clusters)
```
michael-cico, line 32: Probably move this inside the `create_kind_cluster.sh`  script, but maybe in a second pass.

ci/scripts/prepare_jenkins_at_environment.sh
```diff
@@ -29,110 +29,114 @@ TEST_OVERRIDE_SECRET_FILE="./tests/e2e/config/scripts/pre-install-overrides/test
...
+  fi
+  # With the Calico configuration to set disableDefaultCNI to true in the KIND configuration, the control plane node will
+  # be ready only after applying calico.yaml. So wait for the KIND control plane node to be ready, before proceeding further,
+  # with maximum wait period of 5 minutes.
+  kubectl wait --for=condition=ready nodes/${CLUSTER_NAME}-control-plane --timeout=5m --all
+  kubectl wait --for=condition=ready pods/kube-controller-manager-${CLUSTER_NAME}-control-plane -n kube-system --timeout=5m
+  echo "Listing pods in kube-system namespace ..."
+  kubectl get pods -n kube-system
+
+  echo "Install metallb"
```
michael-cico, line 62: Why is metallb inside the CALICO conditional?  it should probably be outside it. You don't need Calico to install MetalLB.

ci/scripts/prepare_jenkins_at_environment.sh
michael-cico: ok, I see why you have it here, you can disregard this.

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,10 +228,59 @@ pipeline {
...
+                            if (params.SKIP_VERRAZZANO_INSTALL_PARAM == true) {
```
michael-cico, line 238: I think this should probably broken out into a separate Groovy func to make the pipeline more readable; ideally this could be wrapped in a Ginkgo test also.

I think another CLI enhancement would be to allow a `--dry-run=server` mode to just validate the operation and return any errors from the webhooks without applying them.  `kubectl` has that kind of feature.

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,10 +228,59 @@ pipeline {
...
+                                    yq -i eval '.spec.components.fluentd.enabled = true' ${INSTALL_CONFIG_FILE_KIND}
+                                    yq -i eval '.spec.components.prometheus.enabled = true' ${INSTALL_CONFIG_FILE_KIND}
+                                    yq -i eval '.spec.components.grafana.enabled = true' ${INSTALL_CONFIG_FILE_KIND}
+                                    yq -i eval '.spec.components.console.enabled = true' ${INSTALL_CONFIG_FILE_KIND}
+                                    yq -i eval '.spec.components.verrazzano.enabled = true' ${INSTALL_CONFIG_FILE_KIND}
+                                    oci --region us-phoenix-1 os object get --namespace ${OCI_OS_NAMESPACE} -bn ${OCI_OS_COMMIT_BUCKET} --name ephemeral/${env.BRANCH_NAME}/${SHORT_COMMIT_HASH}/operator.yaml --file ${WORKSPACE}/downloaded-vpo.yaml
+                                    if [ -f "$WORKSPACE/vz" ]; then
+                                    cd "\$WORKSPACE"
+                                    ./vz install --filename "\$WORKSPACE"/ci/dynamic-updates/invalidCR.yaml --manifests ${WORKSPACE}/downloaded-vpo.yaml
```
bobdonat, line 264: I think you want to check the status of each vz install and verify that the validation error message returned is as expected.

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,10 +228,59 @@ pipeline {
...
+                                    ./vz install --set profile=trash --manifests ${WORKSPACE}/downloaded-vpo.yaml
+                                    else
+                                    cd "\$GO_REPO_PATH"/verrazzano/tools/vz
```
bobdonat, line 267: Why is there a need to have two ways to run the install command here?

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,10 +228,59 @@ pipeline {
...
+                stage('VZ CLI Custom Resource Validation Tests') {
+                    environment {
+                        KIND_KUBERNETES_CLUSTER_VERSION="${params.KUBERNETES_CLUSTER_VERSION}"
+                        OCI_OS_LOCATION="ephemeral/${env.BRANCH_NAME}/${SHORT_COMMIT_HASH}"
+                    }
+                    steps {
+                        script{
+                            if (params.SKIP_VERRAZZANO_INSTALL_PARAM == true) {
+                                try {
+                                     if(params.CRD_API_VERSION == "v1alpha1"){
```
bobdonat, line 240: Can't you just calll 
```
ci/scripts/prepare_jenkins_at_environment.sh ${params.CREATE_CLUSTER_USE_CALICO} ${params.WILDCARD_DNS_DOMAIN}
```

ci/dynamic-updates/Jenkinsfile
```diff
@@ -225,10 +228,24 @@ pipeline {
         stage('Acceptance Tests') {
 
             stages {
+                stage('VZ CLI Custom Resource Validation Tests') {
```
bobdonat, line 231: Should use when expression to actually skip stage instead of have stage run.
Again, can be made in followup PR.

ci/scripts/validate_custom_resource.sh
```diff
@@ -0,0 +1,28 @@
...
+VALIDATION_ERROR="ValidationError"
+
+cd "$WORKSPACE"
+ERROR1=$(./vz install --filename "$INVALID_OPERATOR_YAML" --manifests "$VPO_YAML" 2>&1 >/dev/null)  
+if [[ "$ERROR1" =~ .*"$VALIDATION_ERROR".* ]]; then
+    echo "Expected Error: $VALIDATION_ERROR, Actual Error: $VALIDATION_ERROR"
+    echo "Error: $VALIDATION_ERROR was caught"
+else 
+    echo "Expected Error: $VALIDATION_ERROR from $INVALID_OPERATOR_YAML, Actual Error: $ERROR1" 
+    exit 1
```
bobdonat, line 18: I think both validations should happen and then exit 1 if either failed.

bobdonat (approved):

PR #6132, opened 2023-06-01
Make cluster issuer namespace configurable in a-la-carte pipeline

mbenmclean:
This change makes the cluster issuer namespace configurable in the a-la-carte pipeline, but defaults it to my-cert-manager. It also changes the namespace for the third party cert-manager to my-cert-manager.

### DIFF
--- a/ci/a-la-carte/Jenkinsfile
+++ b/ci/a-la-carte/Jenkinsfile
@@ -263,6 +263,7 @@ pipeline {
             environment {
                 KIND_KUBERNETES_CLUSTER_VERSION = "${params.KUBERNETES_CLUSTER_VERSION}"
                 OCI_OS_LOCATION = "ephemeral/${env.BRANCH_NAME}/${SHORT_COMMIT_HASH}"
+                CLUSTER_RESOURCE_NAMESPACE = "${params.CLUSTER_RESOURCE_NAMESPACE}"
             }
             steps {
                 sh """
--- a/ci/a-la-carte/Jenkinsfile
+++ b/ci/a-la-carte/Jenkinsfile
@@ -28,14 +28,18 @@ pipeline {
                 // 1st choice is the default value
                 choices: [ "1.25", "1.24", "1.23", "1.22" ])
         string (name: 'GIT_COMMIT_TO_USE',
-                        defaultValue: 'NONE',
-                        description: 'This is the full git commit hash from the source build to be used for all jobs',
-                        trim: true)
+                defaultValue: 'NONE',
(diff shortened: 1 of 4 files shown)


ci/a-la-carte/Jenkinsfile
```diff
@@ -28,14 +28,18 @@ pipeline {
...
+                description: 'This is the full git commit hash from the source build to be used for all jobs',
+                trim: true)
         string (name: 'VERRAZZANO_OPERATOR_IMAGE',
-                        defaultValue: 'NONE',
-                        description: 'Verrazzano platform operator image name (in ghcr.io repo).  If not specified, the operator.yaml from Verrazzano repo will be used to create Verrazzano platform operator',
+                defaultValue: 'NONE',
+                description: 'Verrazzano platform operator image name (in ghcr.io repo).  If not specified, the operator.yaml from Verrazzano repo will be used to create Verrazzano platform operator',
                 trim: true)
         booleanParam (description: 'Whether to capture full cluster snapshot on test failure', name: 'CAPTURE_FULL_CLUSTER', defaultValue: false)
+        string (name: 'CLUSTER_RESOURCE_NAMESPACE',
```
michael-cico, line 39: Should there be separate parameters for where Cert-Manager gets installed, and for the `clusterResourceNamespace`?  They can be different.  VZ only needs the `clusterResourceNamespace`, but you may need a separate param for where it CM gets installed to.

michael-cico: We can do that in a follow up if that's preferred, to get this change in.

michael-cico (dismissed):

ci/a-la-carte/Jenkinsfile
michael-cico: OK, this is just if we want to make the CRS different from where the CM is installed.

ci/a-la-carte/Jenkinsfile
```diff
@@ -28,14 +28,18 @@ pipeline {
...
+                defaultValue: 'my-cert-manager',
+                description: 'The name of the namespace to install the cluster issuer into',
```
michael-cico, line 41: Update this description; this is for where CM will look to create cluster-scoped resources (i.e., secrets for `ClusterIssuer` objects). By default it will be the same as where Cert-Manager gets installed, but it can be different.

E.g.,

```
The namespace for cluster-scoped Cert-Manager resources; defaults to where Cert-Manager is installed.
```

mbenmclean: fixed

michael-cico (approved):

PR #6133, opened 2023-06-01
VZ-9840 Create the Cluster Agent component

davisgomes:
**Changes**
- Add the Cluster agent component to the install and install by default
- Add the Cluster agent field to the Verrazzano CR
- Updated switch to run the Cluster agent mode for the agent deployment
- Clusterrole updated for the Cluster agent to manage Verrazzano Projects

**Testing**
- Installed on the admin and managed cluster and verify that the registration was successful
- Deployed a multicluster Verrazzano application to make sure everything gets synced


platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent.go
```diff
@@ -0,0 +1,60 @@
...
+			Value: envImageOverride,
+		})
+	} else {
+		// Get Application Operator image from the BOM
+		bomFile, err := bom.NewBom(config.GetDefaultBOMFilePath())
+		if err != nil {
+			return nil, err
+		}
+
+		images, err := bomFile.BuildImageOverrides("verrazzano-application-operator")
```
desagar, line 36: I'm not sure I understand what this is doing. The app operator itself calls this function with "verrazzano" as the parameter.

platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent_component_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+package clusteragent
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+)
+
```
desagar, line 12: Would be good to add a TestIsReady. Let's not hold up this merge but we can add it in another PR.

platform-operator/controllers/verrazzano/testdata/test_bom.json
```diff
@@ -328,11 +328,6 @@
               "image": "VERRAZZANO_APPLICATION_OPERATOR_IMAGE",
               "tag": "VERRAZZANO_APPLICATION_OPERATOR_TAG",
               "helmFullImageKey": "image"
-            },
```
desagar, line 331: curious why this change was needed.

desagar: I think I get this now - we need this change since there's no BOM entry for cluster agent.

desagar (approved):

platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/verrazzano/verrazzano/pkg/bom"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1beta1"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// AppendClusterAgentOverrides Honor the APP_OPERATOR_IMAGE env var if set; this allows an explicit override
```
michael-cico, line 18: comment seems off.  Do we need to honor an env var though like we do for App operator?  Not even sure we do that anymore.

platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent.go
```diff
@@ -0,0 +1,60 @@
...
+// of the verrazzano-application-operator image when set.
+func AppendClusterAgentOverrides(compContext spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	// Respect the env variable override for the application operator image
+	// If it is not set, default to the image given in the BOM
+	envImageOverride := os.Getenv(constants.VerrazzanoAppOperatorImageEnvVar)
```
michael-cico, line 23: should this be a different env var?

michael-cico: oh wait, I forgot it's the same image running in a different context, my bad.

platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent.go
michael-cico: I forgot it's the same image running in a different context

application-operator/main.go
```diff
@@ -99,7 +96,7 @@ func main() {
 	} else if runWebhooks {
 		exitErr = operatorinit.StartWebhookServer(metricsAddr, log, enableLeaderElection, certDir, scheme)
 	} else {
-		exitErr = operatorinit.StartApplicationOperator(metricsAddr, enableLeaderElection, defaultMetricsScraper, runAppOperator, runClusterAgent, log, scheme)
+		exitErr = operatorinit.StartApplicationOperator(metricsAddr, enableLeaderElection, defaultMetricsScraper, runClusterAgent, log, scheme)
```
michael-cico, line 99: This seems like it's not following the `main` pattern; instead of passing in the flag here, should we have a `StartClusterAgent` call and segregate what's running there from the `StartApplicationOperator` code?  Or is the App operator a superset of ClusterAgent functionality?

michael-cico: If we do want to separate them, it can be a follow-up PR.  I just think if we can separate the modes more cleanly it might make more sense.

platform-operator/controllers/verrazzano/component/clusteragent/cluster_agent.go
desagar: No, this uses the app operator env var for at least the first release.

application-operator/main.go
desagar: good suggestion. we'll take this as a follow on.

PR #6142, opened 2023-06-02
VZ-9453: Add multi field with correct type in OS index template

saketmht:
Added an extra multi field with correct type for few values in the OS index template.


kminder (approved):
Approved.

PR #6146, opened 2023-06-02
Fix v1beta1 none profile

michael-cico:
I found a couple of new components that were not added to the `none` profile for `v1beta1`.  

I added those, and updated the unit test to also check the `v1beta1` effective CR to catch these in the future.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/registry/registry_test.go
+++ b/platform-operator/controllers/verrazzano/component/registry/registry_test.go
@@ -669,39 +669,6 @@ func TestComponentDependenciesMetStateCheckCompDisabled(t *testing.T) {
 	runDepenencyStateCheckTest(t, v1alpha1.CompStateDisabled, false)
 }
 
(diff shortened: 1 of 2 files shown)


fredtibbitts (dismissed):

platform-operator/controllers/verrazzano/component/registry/registry_test.go
```diff
@@ -656,6 +669,39 @@ func TestComponentDependenciesMetStateCheckCompDisabled(t *testing.T) {
 	runDepenencyStateCheckTest(t, v1alpha1.CompStateDisabled, false)
 }
 
+// TestV1Beta1NoneProfileInstalledAllComponentsDisabled Tests the EffectiveV1Beta1CR
```
mbenmclean, line 672: remove this commented code

michael-cico: fixed

mbenmclean (approved):

PR #6154, opened 2023-06-04
VZ-9942 VZ-9721 Exclude VZ 1.3 release from upgrade testing

ayaranga:
Also, the following changes are done:
- Fix the logic in parsing `excludeReleaseTags` in the `derive_upgrade_version.go` code.
- Update the `derive_upgrade_version.go` code to handle the case where there are no more than two minor versions supported for the upgrade. 
- Do not run the VMI prometheus pre-upgrade tests for versions greater than or equal to vz 1.4 as VMI prometheus support has been moved to Prometheus operator in 1.4.

### DIFF
--- a/ci/chaos/JenkinsfileResiliencyTrigger
+++ b/ci/chaos/JenkinsfileResiliencyTrigger
@@ -33,7 +33,8 @@ pipeline {
                         description: 'This is the wildcard DNS domain',
                         trim: true)
         string (name: 'EXCLUDE_RELEASES',
-                defaultValue: "v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.1.0, v1.1.1, v1.1.2, v1.2.0, v1.2.1, v1.2.2",
+                defaultValue: "v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.1.0, v1.1.1, v1.1.2, v1.2.0, v1.2.1, v1.2.2," +
+                        " v1.3.0, v1.3.1, v1.3.2, v1.3.3, v1.3.4, v1.3.5, v1.3.6, v1.3.7, v1.3.8",
                 description: 'This is to exclude the specified releases from upgrade tests.', trim: true)
         string (name: 'TAGGED_TESTS',
                 defaultValue: '',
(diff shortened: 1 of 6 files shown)


ci/chaos/JenkinsfileResiliencyTrigger
```diff
@@ -33,7 +33,8 @@ pipeline {
                         description: 'This is the wildcard DNS domain',
                         trim: true)
         string (name: 'EXCLUDE_RELEASES',
-                defaultValue: "v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.1.0, v1.1.1, v1.1.2, v1.2.0, v1.2.1, v1.2.2",
+                defaultValue: "v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.1.0, v1.1.1, v1.1.2, v1.2.0, v1.2.1, v1.2.2," +
```
lake-of-dreams, line 36: should this instead be changed to something like VALID_RELEASES and can contain regex like 1.4.*, or even if we want to do EXCLUDE_RELEASES it could be an regex like v1.0.*, v1.3* etc

ci/tools/derive_upgrade_version.go
```diff
@@ -164,6 +166,19 @@ func getInterimRelease(releaseTags []string) string {
 	majorVersionValue := parseInt(latestReleaseTagSplit[0])
 	minorInterimVersionValue := parseInt(latestReleaseTagSplit[1]) - 1
 
+	// Handles the case where only two or less minor releases are available for upgrade
```
lake-of-dreams, line 169: should this code instead be going to github and actually fetch what releases are available there rather than calculating here?

ci/chaos/JenkinsfileResiliencyTrigger
ayaranga: I think this can be taken up as part of the refactoring work that we plan to do for upgrade resiliency test pipeline code.

ci/tools/derive_upgrade_version.go
ayaranga: The `releaseTags` passed to this function is actually calculated from the github tags and the excludeReleaseTags configured. The logic used here is not that user-friendly and this entire code needs refactoring. This is planned to be taken up as a separate ticket.

lake-of-dreams (approved):

PR #6176, opened 2023-06-07
Add certificates to the Fluent Operator in the managed cluster

karan56625:
The Fluent Operator needs to communicate with Opensearch via an Ingress in the managed cluster. Certificates are added to Fluentbit by injecting the certificate from the verrazzano-cluster-registration secret into the Fluentbit pod. These certificates will be used by the Fluentbit system and the application cluster's output


adiforluls (approved):

platform-operator/controllers/verrazzano/component/fluentbitosoutput/fluentopensearchoutput_component.go
```diff
@@ -147,12 +159,38 @@ func AppendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
...
+	kvs = append(kvs, bom.KeyValue{Key: OverrideApplicationHostKey, Value: host})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemHostKey, Value: host})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideApplicationPortKey, Value: port})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemPortKey, Value: port})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideApplicationPasswordKey, Value: constants.MCRegistrationSecret})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemPasswordKey, Value: constants.MCRegistrationSecret})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideApplicationUserKey, Value: constants.MCRegistrationSecret})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemUserKey, Value: constants.MCRegistrationSecret})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemTLSKey, Value: "true"})
+	kvs = append(kvs, bom.KeyValue{Key: OverrideSystemCAFileKey, Value: CACertPath + "/" + CACertName})
```
michael-cico, line 190: So do these overrides just mount the secret/bundle in the pod?  Think I'm missing a little bit of context

michael-cico: OK, I see that below.  So, `es-ca-bundle` exists in the `verrazzano-system` namespace, and is kept in sync by the secrets controller in the cluster-agent right?

michael-cico (approved):

platform-operator/controllers/verrazzano/component/fluentbitosoutput/fluentopensearchoutput_component.go
karan56625: yes, clusterRegistrationSecret contains the es-ca-bundle. We are mounting the secret only in case of managed cluster as clusterRegistrationSecret gets created in managed cluster only once cluster registration is done.

PR #6187, opened 2023-06-08
VZ-9970: Change OS Nodepool replicas type to *int32

saketmht:
Problem:
Currently, when the user configures the nodes for the opensearch component in the VZ CR, the number of replicas needs to explicitly defined even for our default nodePools. If `replicas` is not defined, it is assumed to be zero and all the running pods are terminated. For our default nodePools, if `replicas` is not defined, it should fall back to what is defined in our prod/dev profiles instead of 0.

Default nodes for prod for profile:
```
    opensearch:
      nodes:
        - name: es-data
          replicas: 3
          resources:
            requests:
              memory: 4.8Gi
          roles:
            - data
          storage:
            size: 50Gi
        - name: es-ingest
          replicas: 1
          resources:
            requests:
              memory: 2.5Gi
          roles:
            - ingest
        - name: es-master
          replicas: 3
          resources:
            requests:
              memory: 1.4Gi
          roles:
            - master
          storage:
            size: 50Gi
      enabled: true
```

Now if the user want to change, lets say, the java opts for the data nodes, the most likely change to VZ CR would be like this:
```
opensearch:
  nodes:
    - name: es-data
       javaOpts: -Xms732422k -Xmx732422k -Dhttps.proxyHost=172.31.8.214 -Dhttps.proxyPort=3128
```

Currently, this change would terminate all the data nodes.

This PR is to change the type of `replicas` field to *int32 so that 0 and nil value can be differentiated and handled accordingly i.e if the `replicas` field is nil, it should fallback to our defined default replicas count.


pkg/profiles/merge.go
```diff
@@ -298,31 +298,29 @@ func AppendComponentOverrides(actual, profile *v1alpha1.Verrazzano) {
...
+func mergeOSNodesV1alpha1(merged *v1alpha1.Verrazzano) {
 	profileOpensearch := merged.Spec.Components.Elasticsearch
-	if actualOpensearch != nil && profileOpensearch != nil {
-		for i := range actualOpensearch.Nodes {
-			for j := range profileOpensearch.Nodes {
-				if actualOpensearch.Nodes[i].Name == profileOpensearch.Nodes[j].Name {
-					profileOpensearch.Nodes[j].Replicas = actualOpensearch.Nodes[i].Replicas
-				}
+	var v int32
+	if profileOpensearch != nil {
```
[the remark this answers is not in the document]
saketmht: 1. Yes, the files are merged before this func. The merged effectiveCR is passed to this func.
2. So this function previously was working around the omitempty for the `replicas` field. Since the type was int32(not a pointer), both nil and 0 value was omitted i.e. was set to nil in the acutalCR. So when the actualCR was merged with the profile, it defaulted to whatever is in the base yamls. So setting replicas to 0 didn't have any effect. This function was basically setting it back to 0 in the effectiveCR after the merge.

adiforluls (dismissed):
LGTM

pkg/profiles/merge.go
```diff
@@ -50,7 +50,7 @@ func MergeProfiles(actualCR *v1alpha1.Verrazzano, profileFiles ...string) (*v1al
 		return nil, err
 	}
 
-	mergeOSNodesV1alpha1(actualCR, &newCR)
+	mergeOSNodesV1alpha1(&newCR)
```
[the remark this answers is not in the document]
saketmht: Updated the function name.

platform-operator/apis/verrazzano/v1alpha1/convert_to.go
```diff
@@ -402,7 +402,7 @@ func mergeOpenSearchNodes(src, dst *v1beta1.OpenSearchNode) {
 	if src.Storage != nil {
 		dst.Storage = src.Storage
 	}
-	if src.Replicas > 0 {
+	if *src.Replicas > 0 {
```
anders-swanson, line 405: possible nil pointer?

pkg/profiles/merge.go
```diff
@@ -298,31 +298,29 @@ func AppendComponentOverrides(actual, profile *v1alpha1.Verrazzano) {
...
-		for i := range actualOpensearch.Nodes {
-			for j := range profileOpensearch.Nodes {
-				if actualOpensearch.Nodes[i].Name == profileOpensearch.Nodes[j].Name {
-					profileOpensearch.Nodes[j].Replicas = actualOpensearch.Nodes[i].Replicas
-				}
+	var v int32
+	if profileOpensearch != nil {
+		for i := range profileOpensearch.Nodes {
+			if profileOpensearch.Nodes[i].Replicas == nil {
+				profileOpensearch.Nodes[i].Replicas = &v
```
anders-swanson, line 309: any weirdness associated with setting these values to the exact same pointer?

platform-operator/apis/verrazzano/v1alpha1/convert_to.go
```diff
@@ -486,7 +489,7 @@ func convertInstallArgsToOSNodes(args []InstallArgs) (map[string]v1beta1.OpenSea
 
 	nodes := map[string]v1beta1.OpenSearchNode{}
 	addNode := func(node *v1beta1.OpenSearchNode) {
-		if node.Replicas > 0 {
+		if *node.Replicas > 0 {
```
anders-swanson, line 492: nil pointer?

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -347,7 +347,7 @@ func IsMultiNodeOpenSearch(vz *vzapi.Verrazzano) (bool, error) {
 // addNodeGroupReplicas iterates through each OpenSearch node and sums the replicas
 func addNodeGroupReplicas(os *vzapi.ElasticsearchComponent, replicas *int32) {
 	for _, node := range os.Nodes {
-		*replicas += node.Replicas
+		*replicas += *node.Replicas
```
anders-swanson, line 350: possible nil pointer?

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -56,7 +56,7 @@ func nodesToObjectKeys(vz *vzapi.Verrazzano) *ready.AvailabilityObjects {
 	objects := &ready.AvailabilityObjects{}
 	if vzcr.IsOpenSearchEnabled(vz) && vz.Spec.Components.Elasticsearch != nil {
 		for _, node := range vz.Spec.Components.Elasticsearch.Nodes {
-			if node.Replicas < 1 {
+			if *node.Replicas < 1 {
```
anders-swanson, line 59: possible nil pointer?

platform-operator/apis/verrazzano/v1alpha1/convert_to.go
saketmht: I have explicitly set the replica count to 0 for the src variables here, in a function that is called before this. So it will not be a nil pointer

platform-operator/apis/verrazzano/v1alpha1/convert_to.go
saketmht: The replica is set to 0 before hand, so it will not be a nil pointer

platform-operator/controllers/verrazzano/component/common/vmi.go
saketmht: The values being used here is from the effectiveCR, and I am setting the replicas to 0 in the effectiveCR if it is nil. So there should be no nil pointers.

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
saketmht: Setting nil pointers to 0 value in effectiveCR so not a nil pointer

pkg/profiles/merge.go
saketmht: updated the code

platform-operator/apis/verrazzano/v1alpha1/convert_to.go
michael-cico: it's usually good to have a nil-check regardless, just to avoid a potential panic.

saketmht: Added nil checks here and at several other palces

anders-swanson (approved):

kminder (approved):
Approved.

PR #6191, opened 2023-06-08
VZ-9799 Move managed-cluster-configurations example location, minor readme fix

desagar:


kminder (approved):
Approved.

PR #6205, opened 2023-06-09
VZ-9996 Extend clusterAPI to support customization of provider images

mgianatagh:


platform-operator/apis/verrazzano/testdata/clusterapi/v1alpha1.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - values:
+            global:
+              registry: ghcr.io
+              imagePullSecrets:
+                - name: secret1
+                - name: secret2
+              pullPolicy: IfNotPresent
+            defaultProviders:
+              ocne:
+                version: 1.7
```
anders-swanson, line 22: can you remind me the significance of the ocne.version field?

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -645,6 +645,14 @@ type ClusterAPIComponent struct {
 	// If true, then Cluster API Providers will be installed.
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// List of Overrides for the default `values.yaml` file for the component Helm chart. Overrides are merged together,
```
bobdonat, line 649: There is no values.yaml or helm chart for cluster api.

platform-operator/apis/verrazzano/v1beta1/verrazzano_types.go
```diff
@@ -630,6 +630,14 @@ type ClusterAPIComponent struct {
 	// If true, then Cluster API Providers will be installed.
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// List of Overrides for the default `values.yaml` file for the component Helm chart. Overrides are merged together,
```
bobdonat, line 634: Same comment as above.  There is no values.yaml or helm chart for cluster api.

platform-operator/apis/verrazzano/testdata/clusterapi/v1alpha1.yaml
mgianatagh: It will be used to ready an OCNE release asset (how that happens needs to be worked out) default the values of bootstrap and ControlPlane.  I will add a comment line for that.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
mgianatagh: good catch, fixed in both places.

anders-swanson (approved):

theabmitra (approved):
Approved. do we need to make this work for alpha1 ?

PR #6209, opened 2023-06-12
VZ-10041: Add Pod Security for Fluent Operator

karan56625:
Add Pod Security for Fluent Operator and Fluentbit pods.

- No container level security context for Fluentbit. As it is missing in fluentbit CRD in flunet Operator Helmchart.
- kept running Fluentbit as root user because it requires write access to the directory /fluent-bit/etc/tail that has write access to root only.

### DIFF
--- a/platform-operator/helm_config/overrides/fluent-operator-values.yaml
+++ b/platform-operator/helm_config/overrides/fluent-operator-values.yaml
@@ -61,10 +61,5 @@ operator:
     runAsGroup: 1025
     runAsNonRoot: true
     runAsUser: 1025
-    privileged: false
-    allowPrivilegeEscalation: false
-    capabilities:
-      drop:
-        - ALL
     seccompProfile:
       type: RuntimeDefault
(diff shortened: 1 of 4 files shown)


platform-operator/helm_config/overrides/fluent-operator-values.yaml
```diff
@@ -50,3 +50,16 @@ fluentbit:
   envVars:
     - name: CLUSTER_NAME
       value: {{ .clusterName }}
+  # Pod security context for Fluentbit Pod. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  podSecurityContext:
+    seccompProfile:
+      type: RuntimeDefault
+
+operator:
```
pfmackin, line 58: can you disable all capabilities?

karan56625: FluentOperator uses CRD:https://github.com/verrazzano/fluent-operator/blob/200b7db7ca48c60865a95940166ff00e6b06cde2/charts/fluent-operator/charts/fluent-bit-crds/crds/fluentbit.fluent.io_fluentbits.yaml#L4310 to create fluentbit instance. It doesn't contain any field for container level security context.

pfmackin (approved):

PR #6242, opened 2023-06-14
VZ-9048: Add FluentBit operator e2e tests.

IshaGirdhar:


tests/e2e/fluentbit-operator/infra/infra_test.go
```diff
@@ -0,0 +1,102 @@
...
+		Fail(err.Error())
+	}
+
+	m := FluentBitOperatorEnabledModifier{}
+	update.UpdateCRWithRetries(m, longPollingInterval, longWaitTimeout)
+
+	// GIVEN a VZ custom resource in dev profile,
+	// WHEN FluentBit operator is enabled,
+	// THEN pods for fluentbit-operator components gets created.
+	update.ValidatePods(fluentBitOperatorLabelValue, fluentBitComponentLabel, constants.VerrazzanoSystemNamespace, 1, false)
```
adiforluls, line 69: Should the `ValidatePods` check be done in BeforeSuite instead of `Describe/It`? If that's the pattern we follow in other tests then it's fine.

tests/e2e/fluentbit-operator/infra/infra_test.go
```diff
@@ -0,0 +1,102 @@
...
+	t.It("Verify recent Opensearch log record exists", func() {
+		Eventually(func() bool {
+			return pkg.LogRecordFound("verrazzano-system", time.Now().Add(-5*time.Minute), map[string]string{
+				"kubernetes.namespace_name": "verrazzano-system"})
+		}, longWaitTimeout, longPollingInterval).Should(BeTrue(), "Expected to find a recent log record for k8s-sidecar container")
+	})
+
+})
+
+func marshalCRToString(cr interface{}) string {
```
adiforluls, line 95: So this func is being used to debug right? Could you remove this if that's the case?

tests/e2e/fluentbit-operator/infra/infra_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	m := FluentBitOperatorEnabledModifier{}
+	update.UpdateCRWithRetries(m, longPollingInterval, longWaitTimeout)
+
+	// GIVEN a VZ custom resource in dev profile,
+	// WHEN FluentBit operator is enabled,
+	// THEN pods for fluentbit-operator components gets created.
+	update.ValidatePods(fluentBitOperatorLabelValue, fluentBitComponentLabel, constants.VerrazzanoSystemNamespace, 1, false)
+	update.ValidatePods(fluentBitLabelValue, fluentBitComponentLabel, constants.VerrazzanoSystemNamespace, nodeCount, false)
```
adiforluls, line 70: I think it would be a good idea to also validate that Fluentd was uninstalled. This can be done as a followup if there's no existing method to test that (checking that the helm release and pods aren't there anymore).

tests/e2e/fluentbit-operator/infra/infra_test.go
IshaGirdhar: yes it's the pattern that is followed, but I think that in case of an update, maybe I can move it to Describe/It as for us pods are coming up for first time.

IshaGirdhar: Or on second thoughts, I think we can keep it there itself as we are updating the CR in any case.

tests/e2e/fluentbit-operator/infra/infra_test.go
IshaGirdhar: done

adiforluls (approved):

pkg/constants/constants.go
```diff
@@ -122,6 +122,9 @@ const MaxTimesVMCAgentPollingTime = 3
 // FluentdDaemonSetName - The name of the Fluentd DaemonSet
 const FluentdDaemonSetName = "fluentd"
 
+// FluentdDaemonSetName - The name of the Fluentd DaemonSet
```
karan56625, line 125: minor one. fluentd -> fluentbit.

karan56625 (approved):

adiforluls (approved):

PR #6250, opened 2023-06-15
VZ-10072: Fix upgrade from tip of release branches to master

Abhijithang:
This PR fixes the broken upgrade job from tip of release branches to master.


AbeHern (approved):
LGTM

PR #6254, opened 2023-06-15
VZ:10047 update security context of fluentbit container

karan56625:
Add container level security context to fluentbit container.

### DIFF
--- a/platform-operator/helm_config/overrides/fluent-operator-values.yaml
+++ b/platform-operator/helm_config/overrides/fluent-operator-values.yaml
@@ -54,6 +54,12 @@ fluentbit:
   podSecurityContext:
     seccompProfile:
       type: RuntimeDefault
+  securityContext:
+    allowPrivilegeEscalation: false
+    privileged: false
+    capabilities:
+      drop:
+      - ALL
 
 operator:
   # Pod security context for Fluent Operator pod. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
--- a/platform-operator/thirdparty/charts/fluent-operator/values.yaml
+++ b/platform-operator/thirdparty/charts/fluent-operator/values.yaml
@@ -85,6 +85,8 @@ fluentbit:
   secrets: []
   # Pod security context for Fluent Bit pods. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   podSecurityContext: {}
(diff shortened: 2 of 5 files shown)


platform-operator/helm_config/overrides/fluent-operator-values.yaml
```diff
@@ -54,6 +54,12 @@ fluentbit:
   podSecurityContext:
     seccompProfile:
       type: RuntimeDefault
+  securityContext:
+    allowPrivilegeEscalation: false
+    privileged: false
+    capabilities:
+      drop:
+      - ALL
```
adiforluls, line 62: Simply dropping all capabilities works right (i.e. you don't have to add any capability back)?

adiforluls (approved):

platform-operator/helm_config/overrides/fluent-operator-values.yaml
karan56625: yes

pfmackin: can you add 
    runAsGroup: 1025
    runAsNonRoot: true
    runAsUser: 1025

platform-operator/thirdparty/charts/fluent-operator/values.yaml
```diff
@@ -85,6 +85,8 @@ fluentbit:
   secrets: []
   # Pod security context for Fluent Bit pods. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   podSecurityContext: {}
+  # Security context for Fluent Bit container. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  securityContext: {}
```
pfmackin, line 89: So for fluentbit itself there is no securityContext?

karan56625: We have added support to add securityContext for fluentbit in upstream. We have cherry-picked that here. So we are adding fluentbit Securitycontext

platform-operator/helm_config/overrides/fluent-operator-values.yaml
karan56625: fluentbit is having issue while running as non-root user. We have tried running it as non root and providing some capabilities, but that's not enough. so running it as root user and dropping all capabilities.

karan56625: will raise follow up PR if we able to run the fluentbit as non-root user.

PR #6271, opened 2023-06-17
VZ-9965: Add ability uptake to new OCNE versions without upgrading Verrazzano

mgianatagh:


platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -154,8 +172,8 @@ func (c clusterAPIComponent) IsOperatorInstallSupported() bool {
 
 // IsInstalled checks to see if ClusterAPI is installed
 func (c clusterAPIComponent) IsInstalled(ctx spi.ComponentContext) (bool, error) {
-	daemonSet := &appsv1.Deployment{}
-	err := ctx.Client().Get(context.TODO(), types.NamespacedName{Namespace: ComponentNamespace, Name: capiCMDeployment}, daemonSet)
+	deployment := &appsv1.Deployment{}
```
jmaron99, line 175: curious what supured the change from daemon set to deployment?

michael-cico: looks like it was always a deployment, but the variable was named `daemonSet`.  Looks like it just fixed the var name

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_overrides.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	// Return default value
+	return formatProviderURL(false, provider.Image.Repository, provider.Name, provider.Image.BomVersion, provider.MetaddataFile)
+}
+
+// formatProviderURL - return the provider URL using the following format
+// https://github.com/{owner}/{Repository}/releases/{version-tag}/{componentsClient.yaml}
+func formatProviderURL(remote bool, owner string, repo string, version string, metadataFile string) string {
+	if remote {
+		return fmt.Sprintf("https://github.com/%s/%s/releases/%s/%s", owner, repo, version, metadataFile)
```
michael-cico, line 188: Will assuming `github.com` work for the air-gapped case, or is this something that's not expected to work in an air-gapped env?

michael-cico: OK, I see it's guarded by a `remote` flag?

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
mgianatagh: Yes, just fixing the variable name.

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_overrides.go
mgianatagh: Yes, you only get to this function if expecting to pull from the internet, or using the default path (which is baked into the VPO image).  Remote is true when expecting to pull from GitHub.

michael-cico (approved):

PR #6287, opened 2023-06-20
VZ-10099: format the output timestamp to add timezone offset

karan56625:
format the output timestamp to add timezone offset

### DIFF
--- a/platform-operator/thirdparty/charts/fluent-operator/templates/verrazzano/fluentbit-output-filter.yaml
+++ b/platform-operator/thirdparty/charts/fluent-operator/templates/verrazzano/fluentbit-output-filter.yaml
@@ -2,7 +2,7 @@
 apiVersion: fluentbit.fluent.io/v1alpha2
 kind: ClusterFilter
 metadata:
-  name: vz-output-filter
+  name: vz-timestamp-format-filter
   labels:
     fluentbit.fluent.io/enabled: "true"
 spec:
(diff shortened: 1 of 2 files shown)


adiforluls (commented):
Could you paste the timestamp format without this change, the timestamp format with this change and the timestamp format we get in fluentd

platform-operator/thirdparty/charts/fluent-operator/templates/verrazzano/fluentbit-output-filter.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.fluentbit.enable -}}
+apiVersion: fluentbit.fluent.io/v1alpha2
+kind: ClusterFilter
+metadata:
+  name: vz-output-filter
```
adiforluls, line 5: Maybe the name should be something like `vz-timestamp-modifier` or `vz-timestamp-format-filter` to convey what this filter is for just by name

karan56625:
> Could you paste the timestamp format without this change, the timestamp format with this change and the timestamp format we get in fluentd

fluentd:
"@timestamp": "2023-06-15T06:47:03.965000000+00:00"
fluentbit:
"@timestamp": "2023-06-15T09:37:37.000Z"
fluentbit(after this PR)
"@timestamp": "2023-06-20T16:00:20.973099067+00:00"

platform-operator/thirdparty/charts/fluent-operator/templates/verrazzano/fluentbit-output-filter.yaml
karan56625: renamed it to vz-timestamp-format-filter. that should be fine?

adiforluls (approved):

PR #6298, opened 2023-06-21
VZ-10037: update internal CAPI templates with image tags from bom.

theabmitra:
this is not a functionality change, but a change to maintain parity


kminder (approved):
Approved.

PR #6300, opened 2023-06-21
VZ-9994 Add optional external cert-manager to dynamic update pipeline

mbenmclean:
This adds external cert-manager as an option to the dynamic update pipeline. It also includes it as a configuration of the dynamic updates suite that runs from the periodics.

### DIFF
--- a/ci/dynamic-updates/JenkinsfileTrigger
+++ b/ci/dynamic-updates/JenkinsfileTrigger
@@ -116,10 +116,9 @@ pipeline {
                     steps {
                         retry (count: JOB_PROMOTION_RETRIES) {
                             script {
-                                startDCTestJob([booleanParam(name: 'RUN_INSTALL_TESTS', value: true),
+                                startDCTestJob([booleanParam(name: 'RUN_POST_UPDATE1', value: true),
+                                                booleanParam(name: 'RUN_POST_UPDATE2', value: true),
                                                 booleanParam(name: 'RUN_INFRA_TESTS', value: true),
-                                                booleanParam(name: 'RUN_SECURITY_TESTS', value: true),
-                                                booleanParam(name: 'RUN_APP_TESTS', value: true),
                                                 booleanParam(name: 'RUN_NGINX_ISTIO', value: true)])
                             }
                         }
@@ -129,10 +128,7 @@ pipeline {
                     steps {
                         retry (count: JOB_PROMOTION_RETRIES) {
                             script {
-                                startDCTestJob([booleanParam(name: 'RUN_INSTALL_TESTS', value: true),
-                                                booleanParam(name: 'RUN_INFRA_TESTS', value: true),
-                                                booleanParam(name: 'RUN_SECURITY_TESTS', value: true),
-                                                booleanParam(name: 'RUN_APP_TESTS', value: true),
+                                startDCTestJob([booleanParam(name: 'RUN_INFRA_TESTS', value: true),
                                                 booleanParam(name: 'RUN_AVAILABILITY_STATUS', value: true),
                                                 booleanParam(name: 'RUN_AUTHPROXY', value: true),
                                                 booleanParam(name: 'RUN_CERT_MANAGER', value: true),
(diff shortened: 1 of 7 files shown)


ci/dynamic-updates/JenkinsfileTrigger
```diff
@@ -127,19 +128,36 @@ pipeline {
...
+                                startDCTestJob([booleanParam(name: 'RUN_INFRA_TESTS', value: true),
+                                                booleanParam(name: 'RUN_AVAILABILITY_STATUS', value: true),
                                                 booleanParam(name: 'RUN_AUTHPROXY', value: true),
                                                 booleanParam(name: 'RUN_CERT_MANAGER', value: true),
                                                 booleanParam(name: 'RUN_FLUENTD', value: true)])
                             }
                         }
                     }
                 }
+                stage ('Availability Status, AuthProxy, CertManager, and Fluentd Update Tests With External cert-manager') {
```
michael-cico, line 140: Should we run this here, which runs on every periodic, or as part of the weeklies?

ci/scripts/prepare_jenkins_at_environment.sh
```diff
@@ -137,6 +138,10 @@ EOF
     exit 1
   fi
 
+  if [ $INSTALL_EXTERNAL_CERT_MANAGER == true ]; then
```
michael-cico, line 141: not for this pass, but if this is installing CM and NGINX, should it be something like `INSTALL_THIRDPARTY_COMPONENTS` ?

Don't want to be pedantic, but if it's installing more than one thing it should probably be named appropriately.

ci/dynamic-updates/JenkinsfileTrigger
michael-cico: I guess we could run it here for now, but move it to the weeklies once we're confident it's stable.

tests/e2e/security/netpol/network_policy_test.go
```diff
@@ -167,9 +169,17 @@ var _ = t.Describe("Test Network Policies", Label("f:security.netpol"), func() {
...
-				t.Logs.Info("Test cert-manager ingress rules")
-				err := testAccess(metav1.LabelSelector{MatchLabels: map[string]string{kubernetesAppLabel: "prometheus"}}, vzconst.PrometheusOperatorNamespace, metav1.LabelSelector{MatchLabels: map[string]string{"app": vzconst.CertManagerNamespace}}, vzconst.CertManagerNamespace, 9402, false, true)
-				Expect(err).To(BeNil(), fmt.Sprintf("FAIL: Test cert-manager ingress failed: reason = %s", err))
+				kubeconfig, err := k8sutil.GetKubeConfigLocation()
+				Expect(err).To(BeNil())
+				vz, err := pkg.GetVerrazzanoInstallResourceInCluster(kubeconfig)
+				if err != nil {
+					Expect(err).To(BeNil())
+				}
+				if vzcr.IsComponentStatusEnabled(vz, certmanager.ComponentName) {
```
michael-cico, line 178: Is this testing the network access from prometheus to CM for scraping?

ci/dynamic-updates/JenkinsfileTrigger
mbenmclean: Probably easier to run here. We don't run this from the weeklies right now. Also doesn't really get checked

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -406,110 +404,8 @@ var _ = t.Describe("VMI", Label("f:infra-lcm"), func() {
 			})
 		}
 	})
-
```
michael-cico, line 409: why are these tests removed?

ci/scripts/prepare_jenkins_at_environment.sh
mbenmclean: Ok will do as follow up

tests/e2e/security/netpol/network_policy_test.go
mbenmclean: It looks like it. I didn't add this test so I'm not entirely sure of the purpose

michael-cico (approved):

PR #6301, opened 2023-06-21
VZ-10055 address issues with rancher cleanup job and repeated install/uninstall

jmaron99:
- Modified Rancher uninstall logic to only invoke post uninstall cleanup if Rancher is enabled (supports uninstallation but continued contact from Rancher UI to OCNE cluster)
- Made modifications to uninstall logic to allow for multiple install/uninstall sequences


platform-operator/verrazzano-bom.json
```diff
@@ -187,7 +187,7 @@
             },
             {
               "image": "rancher-cleanup",
-              "tag": "v1.0.0-20230403191516-fa35c50"
+              "tag": "20230615200719-abed260"
```
mgianatagh, line 190: this should be a v1.0.0 style tag

jmaron99: Thanks for reminder. Need to merge cleanup PR first.  Will do that now

mgianatagh (approved):

PR #6302, opened 2023-06-21
Enable capi for prod and dev profiles - release 1.6

pfmackin:
Enable capi for prod and dev profiles by default.


kminder (approved):
Approved.

stevshap (approved):

PR #6316, opened 2023-06-23
Aamitra/skipdrivercheck npe

theabmitra:


theabmitra:
fixed elsewhere

PR #6327, opened 2023-06-26
VZ-10091: Create Effective CR ConfigMap

aruneshapk:
Added createOrUpdateEffectiveConfigCM function which generates the Effective CR and stores it in a ConfigMap named <vz-name>-effective-config and is stored in the same namespace as <vz-namespace> as a YAML file.

Added Unit Tests for the function createOrUpdateEffectiveConfigCM()


platform-operator/Makefile
```diff
@@ -66,7 +66,7 @@ SHELL = bash
 # Run against the configured Kubernetes cluster in ~/.kube/config
 .PHONY: run
 run:
-	$(GO) run main.go --kubeconfig=${KUBECONFIG} --zap-log-level=debug
+	$(GO) run main.go --kubeconfig=${KUBECONFIG} --zap-log-level=debug --vz-root-dir=/Users/arunepra/Desktop/verrazzano/
```
karan56625, line 69: remove this change.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -157,6 +159,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	log.Oncef("Finished reconciling Verrazzano resource %v", req.NamespacedName)
 	metricsexporter.AnalyzeVerrazzanoResourceMetrics(log, *vz)
 
+	err = r.reconcileEffCRConfig(ctx, vz, log)
+	if err != nil {
+		zap.S().Error(err)
```
karan56625, line 164: consider to increase the error counter in case of failure.  consider the same error handling that we are doing for other function call in the Reconcile.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1077,52 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
 	defer r.WatchMutex.RUnlock()
 	return r.WatchedComponents[compName]
 }
+
+// The reconcileEffCRConfig takes in the Actual CR and reconciles a configmap
+// If no configmap exists, it will create one, else it updates the configmap with
+// the effective CR
+func (r *Reconciler) reconcileEffCRConfig(ctx context.Context, vz *installv1alpha1.Verrazzano, log vzlog.VerrazzanoLogger) error {
```
karan56625, line 1084: rename the function to be more meaningful. May be `createOrUpdateEffectiveConfigCM`  or may be something else.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1077,52 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
...
+
+	// Get the Effective CR from the Verrazzano CR provided
+	effCR, err := transform.GetEffectiveCR(vz)
+	if err != nil {
+		log.Errorf(err.Error())
+		return err // test must contain AssertError
```
karan56625, line 1090: remove redundant comment

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1077,52 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
...
+	}
+
+	// Create a Configmap Object that stores the Effective CR Specs
+	effCRConfigmap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      (vz.ObjectMeta.Name + "-effective-config"),
+			Namespace: (vz.ObjectMeta.Namespace),
+		},
+		Data: map[string]string{
+			"effective-config.yaml": string(effCRSpecs),
```
karan56625, line 1107: try to create a constant for "effective-config.yaml" as it is getting used at multiple places

platform-operator/controllers/verrazzano/reconcile/uninstall.go
```diff
@@ -96,6 +97,13 @@ func (r *Reconciler) reconcileUninstall(log vzlog.VerrazzanoLogger, cr *installv
 	}
 	tracker := getUninstallTracker(cr)
 	done := false
+
+	// Delete the ConfigMap
+	err = pkg.DeleteConfigMap(cr.ObjectMeta.Namespace, cr.ObjectMeta.Name)
```
karan56625, line 102: does it throw error incase, CM doesn't exists? Take one round of uninstall pipeline.

platform-operator/controllers/verrazzano/reconcile/uninstall.go
```diff
@@ -96,6 +97,13 @@ func (r *Reconciler) reconcileUninstall(log vzlog.VerrazzanoLogger, cr *installv
...
+	if err != nil {
+		log.Errorf(err.Error())
```
karan56625, line 104: fix this error handling.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1079,51 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
...
+			Namespace: (vz.ObjectMeta.Namespace),
+		},
+		Data: map[string]string{
+			effConfigYaml: string(effCRSpecs),
+		},
+	}
+
+	// Update the configMap if a ConfigMap already exists
+	// In case, there's no ConfigMap, the IsNotFound() func will return true and then it will create one.
+	err = r.Client.Update(ctx, effCRConfigmap)
```
adiforluls, line 1114: You can use controllerutil's `CreateOrUpdate` func instead of calling `Update` func or `Create` func based on conditions.

adiforluls: See the definition and the example provided https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

platform-operator/controllers/verrazzano/reconcile/controller_test.go
```diff
@@ -2377,3 +2384,151 @@ func TestReconcilerProcReadyState(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	log, err := vzlog.EnsureResourceLogger(&vzlog.ResourceConfig{
```
adiforluls, line 2439: Should probably use `vzlog.DefaultLogger` in tests

platform-operator/controllers/verrazzano/reconcile/controller_test.go
```diff
@@ -2377,3 +2384,151 @@ func TestReconcilerProcReadyState(t *testing.T) {
...
+		Scheme:            k8scheme.Scheme,
+		Controller:        nil,
+		DryRun:            false,
+		WatchedComponents: nil,
+		WatchMutex:        nil,
+		Bom:               nil,
+		StatusUpdater:     nil,
+	}
+
+	err = r.createOrUpdateEffectiveConfigCM(context.TODO(), vz, log)
```
adiforluls, line 2473: Missing assertion for this error?

aruneshapk: no, this wasn't meant to assert any error, will get rid of these lines

aruneshapk: We're already checking this error, so no assertion is required

platform-operator/controllers/verrazzano/reconcile/controller_test.go
```diff
@@ -2377,3 +2394,139 @@ func TestReconcilerProcReadyState(t *testing.T) {
 		})
 	}
 }
+
+var trueValue = true
+
+// Tests ReconcileEffCRConfig Function
```
saketmht, line 2400: Use GIVEN/WHEN/THEN format for test function comment. See other tests for examples.
Nitpick: function comment usually starts with function name followed by what it does.

platform-operator/controllers/verrazzano/reconcile/controller_test.go
```diff
@@ -2377,3 +2394,139 @@ func TestReconcilerProcReadyState(t *testing.T) {
...
+		Controller:        nil,
+		DryRun:            false,
+		WatchedComponents: nil,
+		WatchMutex:        nil,
+		Bom:               nil,
+		StatusUpdater:     nil,
+	}
+	err := r.createOrUpdateEffectiveConfigCM(context.TODO(), vz, log)
+	if err != nil {
+		t.Errorf(err.Error())
```
saketmht, line 2475: In tests, its a better practice to assert the err to be Nil instead of logging it.

platform-operator/controllers/verrazzano/reconcile/controller_test.go
```diff
@@ -2377,3 +2394,139 @@ func TestReconcilerProcReadyState(t *testing.T) {
...
+	assert.NoError(t, err)
+	// check if configmap contains the effective CR in yaml format
+	assert.Contains(t, cm.Data, effConfigYaml)
+
+	// check if effective CR doesn't contain uname in its byte type or even in decoded format
+	effectiveCR := cm.Data[effConfigYaml]
+	assert.NotContains(t, effectiveCR, uname)
+	decodedUname, err := base64.StdEncoding.DecodeString(uname)
+	if err != nil {
+		t.Errorf(err.Error())
```
saketmht, line 2488: assert instead of logging

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -75,6 +78,9 @@ type Reconciler struct {
 // Name of finalizer
 const finalizerName = "install.verrazzano.io"
 
+// Name of Effective Configmap Data Key
+const effConfigYaml = "effective-config.yaml"
```
michael-cico, line 82: nit, maybe call this `effectiveConfigKey`

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1083,44 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
 	defer r.WatchMutex.RUnlock()
 	return r.WatchedComponents[compName]
 }
+
+// The createOrUpdateEffectiveConfigCM takes in the Actual CR, gets the Effective CR and stores it in a configmap
+// If no configmap exists, it will create one, otherwise it updates the configmap with the effective CR
+func (r *Reconciler) createOrUpdateEffectiveConfigCM(ctx context.Context, vz *installv1alpha1.Verrazzano, log vzlog.VerrazzanoLogger) error {
+
+	// Get the Effective CR from the Verrazzano CR supplied
+	effCR, err := transform.GetEffectiveCR(vz)
```
michael-cico, line 1092: I think the effective CR already exists in the `ctx` object right?

Although we may want the `v1beta1` version stored in the configmap, which isn't generated by default.

platform-operator/controllers/verrazzano/reconcile/uninstall.go
```diff
@@ -94,8 +95,22 @@ func (r *Reconciler) reconcileUninstall(log vzlog.VerrazzanoLogger, cr *installv
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// Delete the ConfigMap
+	effCRConfigmap := &corev1.ConfigMap{
```
michael-cico, line 100: Wonder if we should delete this last, after the uninstall is successful?

Also, there are delete utilities under `pkg/k8s/resource` that simplify resource cleanup

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -157,6 +163,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	log.Oncef("Finished reconciling Verrazzano resource %v", req.NamespacedName)
 	metricsexporter.AnalyzeVerrazzanoResourceMetrics(log, *vz)
 
+	// Create or Update Configmap that'll store our Effective CR
+	err = r.createOrUpdateEffectiveConfigCM(ctx, vz, log)
+	if err != nil {
+		log.Error(err.Error())
```
bobdonat, line 169: Same error is going to be logged twice - once here and once in createOrUpdateEffectiveConfigC

aruneshapk: @bobdonat Hey Bob, Thanks for taking the time to review my PR, I'll fix this, but I had one more question for you - As you can see that I have not returned any error, because I felt that this was just a log level error that we would want to see. But would like to have your suggestion in insisting whether to return err or not ?

platform-operator/controllers/verrazzano/reconcile/uninstall.go
aruneshapk: Thanks for Reviewing ! I had one more query, do you insist returning an error in case Delete doesn't get executed successfully ?

platform-operator/controllers/verrazzano/reconcile/controller.go
aruneshapk: @michael-cico Hi, I think the ctx you're talking about is spi.ComponentContext. We're only processing context.Context here.

michael-cico: Yup, my bad.

platform-operator/controllers/verrazzano/reconcile/controller.go
michael-cico: I think that's a good question; IIUC the question is whether or not we want to have the code requeue and reprocess the event if the CM create/update fails, right?

If we move this CM creation to the start of the reconcile as was talked about at the review, then I think it might be OK to requeue and try again; we haven't done anything yet with the resource.  A `newRequeueWithDelay` return like some of the others here will have us retry.

Bob what do you think?  Worth a retry or just log it and keep going?

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1083,44 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
...
+
+	// Get the Effective CR from the Verrazzano CR supplied
+	effCR, err := transform.GetEffectiveCR(vz)
+	if err != nil {
+		log.Errorf(err.Error())
+		return err
+	}
+
+	// Marshal Indent it to format the Effective CR Specs
+	effCRSpecs, err := json.MarshalIndent(effCR.Spec, "", " ")
```
michael-cico, line 1099: To marshal to YAML instead, see

https://github.com/verrazzano/verrazzano/blob/master/pkg/profiles/merge.go#L35

platform-operator/controllers/verrazzano/reconcile/controller.go
aruneshapk: Yes @michael-cico. I think both ways have their own pros and cons. Maybe @bobdonat could insinuate us at something. Thanks

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -1070,3 +1086,51 @@ func (r *Reconciler) IsWatchedComponent(compName string) bool {
...
+	v1beta1Actual := &v1beta1.Verrazzano{}
+	err := vz.ConvertTo(v1beta1Actual)
+	if err != nil {
+		log.Errorf("Failed Converting v1alpha1 Verrazzano to v1beta1: %v", err)
+	}
+
+	effCR, err := transform.GetEffectiveV1beta1CR(v1beta1Actual)
+	if err != nil {
+		log.Errorf("Failed retrieving the Effective CR: %v", err)
+		return err
```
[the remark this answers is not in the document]
aruneshapk: Yeah, we can do this too. Or maybe, My two cents of advise would be to log a general message in the parent function call only. And then let these error handling cases in the child function return an `fmt.Errorf("<message>", err). `

michael-cico (approved):

bobdonat (approved):

PR #6340, opened 2023-06-26
vz 9481 default jaeger install

AbeHern:


platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -88,12 +89,16 @@ func NewComponent() spi.Component {
 			SupportsOperatorUninstall: true,
 			MinVerrazzanoVersion:      constants.VerrazzanoVersion1_3_0,
 			ImagePullSecretKeyname:    "image.imagePullSecrets[0]",
-			ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "jaeger-operator-values.yaml"),
-			Dependencies:              []string{networkpolicies.ComponentName, cmconstants.CertManagerComponentName, opensearch.ComponentName, fluentoperator.ComponentName},
-			AppendOverridesFunc:       AppendOverrides,
-			GetInstallOverridesFunc:   GetOverrides,
+			//ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "jaeger-operator-values.yaml"),
```
anders-swanson, line 92: remove commented line

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -77,6 +77,7 @@ var (
 )
 
 func NewComponent() spi.Component {
+	//j := jaegerOperatorComponent{
```
anders-swanson, line 80: remove commented line

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -88,12 +89,16 @@ func NewComponent() spi.Component {
...
-			Dependencies:              []string{networkpolicies.ComponentName, cmconstants.CertManagerComponentName, opensearch.ComponentName, fluentoperator.ComponentName},
-			AppendOverridesFunc:       AppendOverrides,
-			GetInstallOverridesFunc:   GetOverrides,
+			//ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "jaeger-operator-values.yaml"),
+			Dependencies:            []string{networkpolicies.ComponentName, cmconstants.CertManagerComponentName, opensearch.ComponentName, fluentoperator.ComponentName},
+			AppendOverridesFunc:     AppendOverrides,
+			GetInstallOverridesFunc: GetOverrides,
 		},
 	}
+	//j.AppendOverridesFunc = func(context spi.ComponentContext, releaseName string, namespace string, chartDir string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
```
anders-swanson, line 98: remove commented lines

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -268,6 +282,21 @@ func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ str
 			if err != nil {
 				return nil, err
 			}
+		} else {
+			allInOneImage := filepath.Join(config.GetHelmOverridesDir(), "jaeger-all-in-one-operator-values.yaml")
```
anders-swanson, line 286: values file, not an image?

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -268,6 +287,23 @@ func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ str
...
+			createInstance, err = isCreateAllInOneJaegerInstance(compContext)
+			if err != nil {
+				return nil, err
+			}
+			if createInstance {
+				jaegerCRTemplate, err := template.New("jaeger").Parse(jaegerAllInOneTemplate)
+				if err != nil {
+					return nil, err
+				}
+				data := jaegerData{OpenSearchURL: openSearchURL, SecretName: globalconst.DefaultJaegerSecretName, OpenSearchReplicaCount: 0}
```
prasad-shirodkar, line 300: `jaegerAllInOneTemplate ` does not have these 3 fields. Why are you trying to fill them in the template?

platform-operator/helm_config/overrides/jaeger-all-in-one-operator-values.yaml
```diff
@@ -0,0 +1,86 @@
...
+      containerSecurityContext:
+        runAsUser: 1001
+        runAsGroup: 65534
+        runAsNonRoot: true
+        privileged: false
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+          - ALL
+    collector:
```
prasad-shirodkar, line 31: Jaeger's all-in-one image contains collector and query functionality. Why are you specifying these in the spec?

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -93,6 +94,14 @@ const extraEnvValueTemplate = `extraEnv:
     value: "{{.AllInOneImage}}"
 `
 
+// A template to define Jaeger override for using the JAEGER-ALL-IN-ONE-IMAGE to create a Jaeger instance.
+const jaegerAllInOneTemplate = `jaeger:
+  create: true
+  spec:
+    strategy: allInOne
```
ayaranga, line 101: See if you can use storage as "in-memory" so that the overrides file remains same. See https://www.jaegertracing.io/docs/1.8/deployment/#memory for details.

platform-operator/helm_config/overrides/jaeger-all-in-one-operator-values.yaml
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
```
ayaranga, line 1: Why do we need a new override file? We can use `jaeger-operator-values.yaml` itself right?

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
AbeHern: I tried both `in-memory` and `memory` in the storage option and noticed no difference. They both work in starting up the All-in-one instance

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -93,6 +94,14 @@ const extraEnvValueTemplate = `extraEnv:
...
+    storage: in-memory
```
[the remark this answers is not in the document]
AbeHern: I tried both `in-memory` and `memory` in the storage option and noticed no difference. They both work in starting up the All-in-one instance

platform-operator/helm_config/overrides/jaeger-all-in-one-operator-values.yaml
ayaranga: Are you saying that you still need a new override file `jaeger-all-in-one-operator-values.yaml`? If so, what's the exact issue you are facing?

AbeHern: Using the `in-memory` and/or `memory` as the storage type didn't require the `jaeger-all-in-one-operator-values.yaml` file. I was able to use the existing `jaeger-operator-values.yaml` file and still deploy the all-in-one without issues.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
AbeHern: Removed it

platform-operator/helm_config/overrides/jaeger-all-in-one-operator-values.yaml
AbeHern: removed the file entirely, it was not needed

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator_component.go
```diff
@@ -88,7 +88,6 @@ func NewComponent() spi.Component {
 			SupportsOperatorUninstall: true,
 			MinVerrazzanoVersion:      constants.VerrazzanoVersion1_3_0,
 			ImagePullSecretKeyname:    "image.imagePullSecrets[0]",
-			ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "jaeger-operator-values.yaml"),
```
ayaranga, line 91: You can revert this change as there are no multiple value override files now.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -268,6 +281,26 @@ func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ str
 			if err != nil {
 				return nil, err
 			}
+		} else {
+			createInstance, err = isCreateAllInOneJaegerInstance(compContext)
```
ayaranga, line 285: Let us make this logic simple. We do not need to add a new function ` isCreateAllInOneJaegerInstance`. In isCreateDefaultJaegerInstance(), you can remove the logic of checking for OpenSearch storage and add `if canUseVZOpenSearchStorage(ctx) {` logic here to choose the appropriate template.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -268,6 +281,26 @@ func AppendOverrides(compContext spi.ComponentContext, _ string, _ string, _ str
...
+			if err != nil {
+				return nil, err
+			}
+			if createInstance {
+				jaegerValuesFile := filepath.Join(config.GetHelmOverridesDir(), "jaeger-operator-values.yaml")
```
prasad-shirodkar, line 290: When using all-in-one strategy, the collector and query components should not be specified in the Jaeger CR spec.  Refer https://www.jaegertracing.io/docs/1.42/operator/#quick-start---deploying-the-allinone-image.

`esIndexCleaner` should also be disabled since the ES index cleaner job does not need to be run. Refer https://github.com/verrazzano/verrazzano/blob/f6e6f2c82427a49cebdc2ae82b4edd55a87cc2fe/platform-operator/helm_config/overrides/jaeger-operator-values.yaml#L70

AbeHern: That would mean that there would be no values file used then for the all-in-one strategy?

prasad-shirodkar: The resulting Jaeger CR spec should still have `ingress` component specified. Refer https://github.com/verrazzano/verrazzano/blob/f6e6f2c82427a49cebdc2ae82b4edd55a87cc2fe/platform-operator/helm_config/overrides/jaeger-operator-values.yaml#L14

The helm overrides should still contain these as well:
- https://github.com/verrazzano/verrazzano/blob/f6e6f2c82427a49cebdc2ae82b4edd55a87cc2fe/platform-operator/helm_config/overrides/jaeger-operator-values.yaml#L3-L7
- https://github.com/verrazzano/verrazzano/blob/f6e6f2c82427a49cebdc2ae82b4edd55a87cc2fe/platform-operator/helm_config/overrides/jaeger-operator-values.yaml#L95-L113

AbeHern: I don't think I'm getting my fundamental question answered. I'm being told on one hand to remove the `jaeger-all-in-one-operator-values.yaml` and use the `jaeger-operator-values.yaml` file for both the default and all-in-one instance. But without `jaeger-all-in-one-operator-values.yaml` where I can specify the values for query and collector, I'm not sure which plan is desired here.

prasad-shirodkar: FYI I had already mentioned the below around 2 weeks ago. It is essentially irrespective of the implementation https://github.com/verrazzano/verrazzano/pull/6340#discussion_r1257630181

> Jaeger's all-in-one image contains collector and query functionality. Why are you specifying these in the spec?

ayaranga: > When using all-in-one strategy, the collector and query components should not be specified in the Jaeger CR spec. Refer https://www.jaegertracing.io/docs/1.42/operator/#quick-start---deploying-the-allinone-image.
> 
> `esIndexCleaner` should also be disabled since the ES index cleaner job does not need to be run. Refer
> 
> https://github.com/verrazzano/verrazzano/blob/f6e6f2c82427a49cebdc2ae82b4edd55a87cc2fe/platform-operator/helm_config/overrides/jaeger-operator-values.yaml#L70

I recommend to check what happens if we specify the collector and query components when using all-in-one strategy. Does it ignore or any error seen or it has to specified under different section? If it ignores, then I think we can leave it in the values file as it is.
For  `esIndexCleaner`,  doesn't it ignore if the storage is memory? When `esIndexCleaner` is enabled, if the Jaeger is creating a index cleaner job irrespective of the storage type, then enabling of esIndexCleaner can be moved out of values file and can be added to templates in Go code.

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
AbeHern: I made an attempt to simplify some pieces please review again

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -466,8 +485,7 @@ func isJaegerCREnabled(ctx spi.ComponentContext) (bool, error) {
 	if jaegerCreateOverride != nil {
 		return jaegerCreateOverride.(bool), nil
 	}
-	// Jaeger instance would be created if Verrazzano's OpenSearch can be used as storage
-	return canUseVZOpenSearchStorage(ctx), nil
+	return false, nil
```
ayaranga, line 488: @prasad-shirodkar If we return false here, then Jaeger ingresses would not be created as expected right?

ayaranga (approved):

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
```diff
@@ -821,3 +855,34 @@ func newScheme() *runtime.Scheme {
...
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			k8sInstanceNameLabel: instanceName,
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &clipkg.ListOptions{
+		Namespace:     ComponentNamespace,
```
anders-swanson, line 869: Is this something the user can change?

anders-swanson: Same with the labels, could the jaeger cr override these?

anders-swanson (approved):

platform-operator/controllers/verrazzano/component/jaeger/operator/jaeger_operator.go
ayaranga: We do not provide an option to override the namespace and we have a webhook validation check to disallow the override of Jaeger instance name. See `validateOverrideYamls()` in jaeger_operator.go for details.

PR #6342, opened 2023-06-27
Fix network policies link in API docs

bobdonat:
This pull request fixes a link to network policies in the API docs.


rosemarymarano (approved):
Looks ok to me (but I'm not sure about the relative dirs).

rosemarymarano (approved):

bobdonat:
> Looks ok to me (but I'm not sure about the relative dirs).

Ran locally with hugo and updated url is correct.

PR #6353, opened 2023-06-28
Fix dynamic config pipeline error

mbenmclean:
This changes fixes a method too large exception in the dynamic config pipeline by moving  code from the main pipeline to separate functions.

### DIFF
--- a/ci/dynamic-updates/Jenkinsfile
+++ b/ci/dynamic-updates/Jenkinsfile
@@ -330,6 +330,20 @@ pipeline {
                 stage('Verrazzano Update1') {
                     steps {
                         sh """
+                            cd ${GO_REPO_PATH}/verrazzano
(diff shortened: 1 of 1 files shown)


michael-cico (approved):

agv426 (approved):

ci/dynamic-updates/Jenkinsfile
```diff
@@ -1066,3 +1017,66 @@ def runCustomResourceValidation(){
        ci/scripts/validate_custom_resource.sh ${WORKSPACE}/ci/dynamic-updates/invalidCR.yaml ${WORKSPACE}/downloaded-vpo.yaml
     """
 }
+
+def modifyConfigForInstallation() {
```
theabmitra, line 1021: instead of so many yq statements you can also point to an installation yaml where components need to be enabled/disabled without a condition

mbenmclean: Not trying optimize the pipeline here just want to fix the broken test suite

PR #6369, opened 2023-06-29
VZ-10107 - Change the Rancher cattle-system namespace's verrazzano.io/namespace label to equal the namespace name, to match network policy

desagar:

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
@@ -166,7 +166,9 @@ func labelNamespace(c client.Client) error {
 		}
 		if _, found := nsList.Items[i].Labels[constants.VerrazzanoManagedKey]; isRancherNamespace(&(nsList.Items[i])) && !found {
 			nsList.Items[i].Labels[constants.VerrazzanoManagedKey] = nsList.Items[i].Name
-			return c.Update(context.TODO(), &(nsList.Items[i]), &client.UpdateOptions{})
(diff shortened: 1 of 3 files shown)


platform-operator/controllers/verrazzano/component/rancher/rancher_postinstall.go
```diff
@@ -166,7 +166,7 @@ func labelNamespace(c client.Client) error {
 		}
 		if _, found := nsList.Items[i].Labels[constants.VerrazzanoManagedKey]; isRancherNamespace(&(nsList.Items[i])) && !found {
 			nsList.Items[i].Labels[constants.VerrazzanoManagedKey] = nsList.Items[i].Name
-			c.Update(context.TODO(), &(nsList.Items[i]), &client.UpdateOptions{})
+			return c.Update(context.TODO(), &(nsList.Items[i]), &client.UpdateOptions{})
```
fredtibbitts, line 169: This will exit the for loop. I think you need to check for non-nil err and only return on error.

davisgomes, line 169: I think this should only return if an error occurs since we are looping over the namespace list. This will exit on the first item that is "found"

desagar: oh, good catch, yikes

desagar: fixed

fredtibbitts (approved):

PR #6385, opened 2023-07-04
VZ-7644: Rotation of VPO webhook certs when expired

neeraj-n-singh:
* Added annotations in the webhook secrets so that it can filter the events only for labeled only,
* Created a cert controller which will watch the events and in every 24 hours will check the certificate's validation if it has an expiry of fewer than 15 days, then delete those secrets and restart the webhook deployment.

### DIFF
--- a/platform-operator/controllers/certrotation/certrotation_actions.go
+++ b/platform-operator/controllers/certrotation/certrotation_actions.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package certrotation
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	vzconst "github.com/verrazzano/verrazzano/pkg/constants"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	clipkg "sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+// ValidateCertDate validates the certs.
+func (r *CertificateRotationManagerReconciler) ValidateCertDate(certContent []byte) (bool, error) {
+	block, _ := pem.Decode(certContent)
+	if block == nil {
+		return false, r.log.ErrorfNewErr("unable to parse the certificate")
+	}
+	certs, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return false, r.log.ErrorfNewErr(err.Error())
+	}
+	deadline := time.Now().Add(r.CompareWindow)
+	if deadline.After(certs.NotAfter) {
+		r.log.Progressf("certificate for %s expires in less than %s hours",
+			certs.Subject.CommonName,
+			certs.NotAfter.Format(time.RFC3339))
+		return true, nil
+
+	}
+	r.log.Progressf("certificate for %s has validity %s",
+		certs.Subject.CommonName,
+		certs.NotAfter.Sub(deadline).Round(time.Hour))
+	return false, nil
+}
+
+// GetSecretData checks if secret exists in namespace or not
+// if exists then return secret data else return nil
+func (r *CertificateRotationManagerReconciler) GetSecretData(ctx context.Context, secretName string) (corev1.Secret, []byte) {
+	sec := corev1.Secret{}
+	err := r.Get(ctx, clipkg.ObjectKey{
+		Namespace: r.WatchNamespace,
+		Name:      secretName,
+	}, &sec)
+	if err != nil && apierrors.IsNotFound(err) {
+		r.log.Errorf("an error while listing the certificate secret %v", secretName)
+		return sec, nil
+	}
+	r.log.Debugf("successfully found the certificate secret %v", secretName)
+	if val, ok := sec.Data["tls.crt"]; ok {
+		return sec, val
+	}
+	return sec, nil
+}
+
+// DeleteSecret deletes the tls secret.
+func (r *CertificateRotationManagerReconciler) DeleteSecret(ctx context.Context, secretName corev1.Secret) error {
+	if err := r.Delete(ctx, &secretName, &clipkg.DeleteOptions{}); err != nil {
+		return r.log.ErrorfNewErr("an error while deleting the certificate secret %v in namespace %v with error %v", secretName.Name, r.WatchNamespace, err)
+	}
+	r.log.Infof("successfully deleted the secret %v in namespace %v ", secretName.Name, r.WatchNamespace)
+	return nil
+}
+
+func (r *CertificateRotationManagerReconciler) RolloutRestartDeployment(ctx context.Context) error {
+	deployment := appsv1.Deployment{}
+	var err error
+	err = r.Get(ctx, clipkg.ObjectKey{Namespace: r.TargetNamespace, Name: r.TargetDeployment}, &deployment)
+	r.log.Debugf("deployment listed with name %v", deployment.Name)
+	if err != nil && apierrors.IsNotFound(err) {
+		r.log.Errorf("an error while listing the deployment in namespace %v", r.TargetNamespace)
+		return err
+	}
+	time := time.Now()
+	_, err = controllerruntime.CreateOrUpdate(context.TODO(), r.Client, &deployment, func() error {
+		if deployment.Spec.Template.ObjectMeta.Annotations == nil {
+			deployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+		}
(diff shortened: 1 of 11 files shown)


platform-operator/main.go
```diff
@@ -98,7 +98,11 @@ func main() {
 	flag.Int64Var(&config.MySQLRepairTimeoutSeconds, "mysql-repair-timeout", config.MySQLRepairTimeoutSeconds,
 		"MySQL repair timeout seconds")
 	flag.BoolVar(&config.ExperimentalModules, "experimental-modules", config.ExperimentalModules, "enable experimental modules")
-
+	// Certificate expiry check
+	config.CertificateExpiryCheckPeriodDuration = flag.Duration("cert-expiry-check-period", internalconfig.DefaultCheckPeriod,
+		"Webhook certificate expiration check period in seconds, default 24 hours")
+	config.CertificateExpiryCheckWindowDuration = flag.Duration("cert-rotation-window", internalconfig.DefaultExpiryWindow,
+		"The window of time to rotate the webhook certificates before expiration in seconds, default 336h (2 weeks)")
```
michael-cico, line 105: Is the duration in seconds or hours?   The description says seconds in the first part, but then the default as 336h.

platform-operator/internal/k8s/certificate/certificate.go
```diff
@@ -79,6 +80,9 @@ func createTLSCert(log *zap.SugaredLogger, kubeClient kubernetes.Interface, comm
 	existingSecret, err := secretsClient.Get(context.TODO(), OperatorTLS, metav1.GetOptions{})
 	if err == nil {
 		log.Infof("Secret %s exists, using...", OperatorTLS)
+		log.Info("Adding labels in existing secret")
+		_ = updateSecretLabel(secretsClient, existingSecret)
```
michael-cico, line 84: Why do we ignore the error here?

platform-operator/internal/k8s/certificate/certificate.go
```diff
@@ -224,12 +235,22 @@ func createCACertSecretIfNecessary(log *zap.SugaredLogger, secretsClient corev1.
 	webhookCA.Data = make(map[string][]byte)
 	webhookCA.Data[CertKey] = caPEMBytes
 	webhookCA.Data[PrivKey] = caKeyPEMBytes
+	if webhookCA.Labels == nil {
+		webhookCA.Labels = make(map[string]string)
+	}
+	webhookCA.Labels[OperatorCertLabelKey] = OperatorCertLabel
 
 	_, createError := secretsClient.Create(context.TODO(), &webhookCA, metav1.CreateOptions{})
```
michael-cico, line 243: MIght be good to replace this whole block with a `controllerruntime.CreateOrUpdate()` call, I probably should have done it that way from the beginning.  Not sure why I didn't do that from the beginning.  Would simplify this a bit.

platform-operator/controllers/certrotation/certrotation_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	if len(certList) < len(r.CertificatesList) {
+		// If there are no matching certificates, we don't need to re-queue
+		r.log.Debugf("No matching certificate secrets found")
+		return newRequeueWithDelay(30, 60, time.Second), nil
+	}
+
+	err = r.CheckCertificateExpiration(ctx, certList)
+	if err != nil {
+		result := newRequeueWithDelay(5, 10, time.Second)
+		r.log.Debugf("Error while certification validation: %v", result.RequeueAfter)
```
michael-cico, line 82: "while certification" ==> "while checking certificate status", or something along those lines

platform-operator/controllers/certrotation/certrotation_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		return result, err
+	}
+
+	result := ctrl.Result{
+		Requeue:      true,
+		RequeueAfter: r.CheckPeriod, // Check period is in duration already
+	}
+	r.log.Infof("Successfully validated the certificate: %v", result.RequeueAfter)
```
michael-cico, line 90: Maybe change to "Successfully validated webhook certificates", or something like that.

platform-operator/controllers/certrotation/certrotation_actions.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	deadline := time.Now().Add(r.CompareWindow)
+	if deadline.After(certs.NotAfter) {
+		r.log.Progressf("certificate for %s expires in less than %s hours",
+			certs.Subject.CommonName,
+			certs.NotAfter.Format(time.RFC3339))
+		return true, nil
+
+	}
+	r.log.Progressf("certificate for %s has validity %s",
```
michael-cico, line 37: Perhaps say "Certificate %s is valid for %s hours", or something like that

platform-operator/internal/k8s/certificate/certificate.go
```diff
@@ -26,11 +26,12 @@ import (
...
-	OperatorName    = "verrazzano-platform-operator-webhook"
-	OldOperatorName = "verrazzano-platform-operator"
-	OperatorCA      = "verrazzano-platform-operator-ca"
-	OperatorTLS     = "verrazzano-platform-operator-tls"
-
+	OperatorName         = "verrazzano-platform-operator-webhook"
+	OldOperatorName      = "verrazzano-platform-operator"
+	OperatorCA           = "verrazzano-platform-operator-ca"
+	OperatorTLS          = "verrazzano-platform-operator-tls"
+	OperatorCertLabelKey = "install.verrazzano.io/vpo-webhook"
```
michael-cico, line 33: Because we're adding an annotation, this will require an API review.

platform-operator/controllers/certrotation/certrotation_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		r.log.Debugf("Reconciling %v, time between reconciles: %v", req.NamespacedName, lastReconcile.Sub(*r.lastReconcile))
+	}
+	r.lastReconcile = &lastReconcile
+	if ctx == nil {
+		ctx = context.TODO()
+	}
+	if result, err := r.initLogger(componentNamespace); err != nil {
+		return result, err
+	}
+	certList, err := r.getCertSecretList(ctx)
```
michael-cico, line 68: So it seems like generally, on each cycle we look at all certificates based on the annotation, and that way avoid the situation where we were going through multiple updates/requeues for each cert?

platform-operator/main.go
neeraj-n-singh: That allows any duration values. I will update with hours in document

platform-operator/internal/k8s/certificate/certificate.go
neeraj-n-singh: I have handled the error now.

platform-operator/controllers/certrotation/certrotation_controller.go
neeraj-n-singh: Sure,

platform-operator/controllers/certrotation/certrotation_controller.go
neeraj-n-singh: Sure

platform-operator/controllers/certrotation/certrotation_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		status[secret] = mustRotate
+		if mustRotate {
+			err = r.DeleteSecret(ctx, sec)
+			if err != nil {
+				return r.log.ErrorfNewErr("an error deleting the certificate")
+			}
+		}
+	}
+	for s := range status {
+		if status[s] {
```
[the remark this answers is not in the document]
neeraj-n-singh: In the case of multiple secrets, it was observed that if one of the certs expired and the other is yet to be expired in a few seconds, the last `mustrotate` will have false as the last status. Hence I make a map of all secret statuses and if anyone is expired then I am restarting deployment and use break statements to get it out from the same.

platform-operator/controllers/certrotation/certrotation_controller.go
neeraj-n-singh: Yes.

platform-operator/internal/k8s/certificate/certificate.go
neeraj-n-singh: Sure.

platform-operator/controllers/certrotation/certrotation_actions.go
neeraj-n-singh: sure, done

platform-operator/internal/k8s/certificate/certificate.go
michael-cico: I think perhaps instead of `install.verrazzano.io/vpo-webhook` maybe something like

```
install.verrazzano.io/webhook-certificate = "verrazzano-platform-operator-webhook"
```

or, alternately

```
install.verrazzano.io/certificate = "verrazzano-platform-operator-webhook"
```

neeraj-n-singh: I added this "install.verrazzano.io/certificate" = "verrazzano-platform-operator-webhook" combination.

PR #6386, opened 2023-07-05
VZ-9014 Fix Istio config errors seen in Kiali

ayaranga:
The Kiali console was reporting Istio config errors with the default installation profiles as authorization policy rules had references to the service accounts of the uninstalled components. Changes are done to fix the Istio config errors seen for authorization policies `verrazzano-authproxy-authzpol` and `vmi-system-prometheus-authzpol` by adding the rules only if the corresponding components are enabled.


mgianatagh (approved):
These changes appear ok to me

michael-cico (approved):

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -627,6 +627,36 @@ func createOrUpdatePrometheusAuthPolicy(ctx spi.ComponentContext) error {
 	authPol := istioclisec.AuthorizationPolicy{
 		ObjectMeta: metav1.ObjectMeta{Namespace: ComponentNamespace, Name: prometheusAuthPolicyName},
 	}
+	var rules []*securityv1beta1.Rule
+
+	// allow Auth Proxy, Grafana, and Kiali to access Prometheus
```
pfmackin, line 632: Can you make sure this is only done if Istio is enabled

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -779,3 +744,19 @@ func deleteNetworkPolicy(ctx spi.ComponentContext) error {
 	}
 	return nil
 }
+
+func constructAuthPolicyRule(namespaces []string, fromPrincipals []string, toPorts []string) *securityv1beta1.Rule {
```
pfmackin, line 748: Is there already common code to do this (since other components need to do the same)

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
kminder: Would it be better to check for the existence of the CRD instead of checking if Istio is enabled?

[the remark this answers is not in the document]
kminder: So why no move these SA creations to their respective components?

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_values.go
```diff
@@ -162,3 +165,15 @@ type jaegerOperatorValues struct {
 type ingressNGINXValues struct {
 	Namespace string `json:"namespace"`
 }
+
+type fluentdValues struct {
+	Enabled bool `json:"enabled"` // Always write
```
kminder, line 170: What does Always write mean?  Please clarify the comment.

platform-operator/helm_config/charts/verrazzano/templates/authorizationpolicy.yaml
```diff
@@ -24,14 +24,22 @@ spec:
       to:
         - operation:
             ports: ["{{ .Values.api.port }}", "{{ .Values.api.grpcPort }}"]
+{{- if or .Values.fluentd.enabled (and .Values.fluentOperator.enabled .Values.fluentbitOpensearchOutput.enabled) }}
     # verrazzano-authproxy:8775 <- fluentd, fluentBit
     - from:
         - source:
             namespaces: ["{{ .Release.Namespace }}"]
+            {{- if and .Values.fluentd.enabled (and .Values.fluentOperator.enabled .Values.fluentbitOpensearchOutput.enabled) }}
```
kminder, line 32: Can these somehow be distributed to the related components/charts?

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
ayaranga: Yes, the following code check already exists in lines 621-625 
```
if istio != nil && istio.Enabled != nil && !*istio.Enabled {
		return nil
	}
```

platform-operator/controllers/verrazzano/component/verrazzano/verrazzano_values.go
ayaranga: I have copied it from the already existing code in the same file. I assumed it would be used while marshalling/unmarshaling. @pfmackin can you please confirm.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
ayaranga: No, I moved this function to common package and made Kiali also use it.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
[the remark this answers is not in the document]
ayaranga: Added functions to generate SA names in common package as suggested.

platform-operator/helm_config/charts/verrazzano/templates/authorizationpolicy.yaml
ayaranga: We will try to take care of this as part of moving to more modular approach.

pfmackin (approved):

PR #6393, opened 2023-07-05
VZ-10183 Ignore VMC list if the CRD does not exist

davisgomes:
**Changes**
- Ignore the VMC list if no kind error is returned from the list request

**Testing**
- Install 1.5.3 with the cluster operator disabled
- Upgrade to this branch and verify that upgrade completes

### DIFF
--- a/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
+++ b/platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2021, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package appoper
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
```diff
@@ -110,6 +111,11 @@ func (c applicationOperatorComponent) PostUpgrade(ctx spi.ComponentContext) erro
 	// the system for multicluster.
 	vmcList := vmcv1alpha1.VerrazzanoManagedClusterList{}
 	err := ctx.Client().List(clientCtx, &vmcList)
+	// Ignore if CRD doesn't exist
+	if _, ok := err.(*meta.NoKindMatchError); ok {
```
fredtibbitts, line 115: Is there a way to unit test this?

fredtibbitts (approved):

platform-operator/controllers/verrazzano/component/appoper/app_operator_component.go
davisgomes: I think I might be able to do it with mocks

fredtibbitts (approved):

mgianatagh (approved):

PR #6395, opened 2023-07-05
vz-10001 cli should validate commands

AbeHern:
Invalidating the install process when missing the filename flag on `vz install VZ_CR.yaml`

### DIFF
--- a/tools/vz/cmd/analyze/analyze.go
+++ b/tools/vz/cmd/analyze/analyze.go
@@ -43,6 +43,14 @@ func NewCmdAnalyze(vzHelper helpers.VZHelper) *cobra.Command {
 	cmd.PersistentFlags().String(constants.ReportFileFlagName, constants.ReportFileFlagValue, constants.ReportFileFlagUsage)
 	cmd.PersistentFlags().String(constants.ReportFormatFlagName, constants.SummaryReport, constants.ReportFormatFlagUsage)
 	cmd.PersistentFlags().BoolP(constants.VerboseFlag, constants.VerboseFlagShorthand, constants.VerboseFlagDefault, constants.VerboseFlagUsage)
+
+	// Check that command args are not set at the creation of the command
+	cmd.Args = func(cmd *cobra.Command, args []string) error {
+		if len(args) != 0 {
+			return fmt.Errorf("invalid arguments specified: %s", args)
+		}
+		return nil
+	}
 	return cmd
 }
 
--- a/tools/vz/cmd/analyze/analyze.go
+++ b/tools/vz/cmd/analyze/analyze.go
@@ -44,13 +44,9 @@ func NewCmdAnalyze(vzHelper helpers.VZHelper) *cobra.Command {
 	cmd.PersistentFlags().String(constants.ReportFormatFlagName, constants.SummaryReport, constants.ReportFormatFlagUsage)
 	cmd.PersistentFlags().BoolP(constants.VerboseFlag, constants.VerboseFlagShorthand, constants.VerboseFlagDefault, constants.VerboseFlagUsage)
 
-	// Check that command args are not set at the creation of the command
-	cmd.Args = func(cmd *cobra.Command, args []string) error {
-		if len(args) != 0 {
-			return fmt.Errorf("invalid arguments specified: %s", args)
-		}
-		return nil
-	}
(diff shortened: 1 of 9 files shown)


tools/vz/cmd/install/install.go
```diff
@@ -279,13 +279,18 @@ func installVerrazzano(cmd *cobra.Command, vzHelper helpers.VZHelper, vz clipkg.
...
-func getVerrazzanoYAML(cmd *cobra.Command, vzHelper helpers.VZHelper, version string) (vz clipkg.Object, obj *unstructured.Unstructured, err error) {
+func getVerrazzanoYAML(cmd *cobra.Command, vzHelper helpers.VZHelper, version string, args []string) (vz clipkg.Object, obj *unstructured.Unstructured, err error) {
 	// Get the list yaml filenames specified
 	filenames, err := cmd.PersistentFlags().GetStringSlice(constants.FilenameFlag)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// Check to invalidate the install process when a Custom Resource is passed to without the filename flag (-f)
+	if len(filenames) == 0 && len(args) > 0 {
```
bobdonat, line 290: This solution assumes that the -f flag is being used when args are greater than zero.  It is possible that the user intended to use --set or --manifest or some other flag that required a value and then the error message would not make any sense.

I think a more general solution is needed and that same needed for the other commands like upgrade, uninstall, etc.

Have you looked at doing something like this when creating each command.
	cmd.Args = func(cmd *cobra.Command, args []string) error {
		if len(args) != 0 {
			return fmt.Errorf("invalid arguments specified: %s", args)
		}

		return nil
	}
This could be a common function called by each command.

tools/vz/cmd/install/install.go
```diff
@@ -110,6 +110,13 @@ func NewCmdInstall(vzHelper helpers.VZHelper) *cobra.Command {
 	// Hide the flag for overriding the default wait timeout for the platform-operator
 	cmd.PersistentFlags().MarkHidden(constants.VPOTimeoutFlag)
 
+	// Check that command args are not set at the creation of the command
+	cmd.Args = func(cmd *cobra.Command, args []string) error {
```
prasad-shirodkar, line 114: I checked out your branch and the built the cli locally. One thing I noticed is that when the flags are incorrect, the error displayed is misleading.

The set flag should have two dashes.
```shell
linux-amd64 $ ./vz install -f ~/v8o/dev.yaml -set profile=dev
Error: invalid arguments specified: [profile=dev]
```

The file flag should have two dashes. 
```shell
linux-amd64 $ ./vz install -file ~/v8o/dev.yaml --set profile=dev
Error: invalid arguments specified: [/home/pshirodk/v8o/dev.yaml]
```

bobdonat: For both of the above examples the shorthand flag is being used: -s and -f.
When using -set, the -s shorthand is being has a value of "et"
When using -file, the -f shorthand is being has a value of "ile"

```
vz install -file
Installing Verrazzano version v1.6.0
Error: open ile: no such file or directory
```

AbeHern: So does this fall in the scope of my fixes then?

tools/vz/cmd/analyze/analyze.go
```diff
@@ -43,6 +43,14 @@ func NewCmdAnalyze(vzHelper helpers.VZHelper) *cobra.Command {
 	cmd.PersistentFlags().String(constants.ReportFileFlagName, constants.ReportFileFlagValue, constants.ReportFileFlagUsage)
 	cmd.PersistentFlags().String(constants.ReportFormatFlagName, constants.SummaryReport, constants.ReportFormatFlagUsage)
 	cmd.PersistentFlags().BoolP(constants.VerboseFlag, constants.VerboseFlagShorthand, constants.VerboseFlagDefault, constants.VerboseFlagUsage)
+
+	// Check that command args are not set at the creation of the command
+	cmd.Args = func(cmd *cobra.Command, args []string) error {
```
prasad-shirodkar, line 48: Instead of copy-pasting the same function everywhere, please create a utility function that can be called from wherever necessary. This will adhere to DRY principle.

bobdonat (approved):

PR #6398, opened 2023-07-06
VZ-9203: Fixes for Alertmanager smtp alerts and UI

lake-of-dreams:
- Creates a new ingress for Alertmanager when it is enabled
- Alertmanager URL added to VZ cr status
- Prometheus only able to send alerts to Alertmanager when Alertmanager is not in istio-mesh (as described in [issue](https://discuss.istio.io/t/issue-with-http-communication-from-prometheus-to-alertmanager/10796/10) , therefore added the annotation in default overrides to not inject istio side car
- The alertmanager [template](https://github.com/verrazzano/verrazzano/blob/master/platform-operator/thirdparty/charts/prometheus-community/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml#L26)  defaults the `version` field to tag of image. The [code](https://github.com/prometheus-operator/prometheus-operator/blob/906669420e97573a4ea7534fb3fdf0ce8192ae03/pkg/alertmanager/amcfg.go#L1495) in prometheus operator checks if the version is greater than or equal to `0.25.0` and only then the `smtp_auth_password_file` configuration is included. In our case, we name tags as `v0.25.0-<timestamp>-<commit>`. As per semantic versioning this turns out to be a lesser version that 0.25.0 and hence the operator drops the `smtp_auth_password_file` field. To fix this, added the `version` to default overrides and changed the template to first look at `version` followed by `tag`.


michael-cico (approved):

pkg/vzcr/enabled.go
```diff
@@ -652,3 +655,42 @@ func IsComponentStatusEnabled(cr runtime.Object, componentName string) bool {
 	}
 	return false
 }
+
+// IsAlertmanagerEnabled returns true only if the Alertmanager is explicitly enabled in the CR
+func IsAlertmanagerEnabled(cr runtime.Object, ctx spi.ComponentContext) (bool, error) {
```
kminder, line 660: This needs a unit test.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -751,20 +751,45 @@ func createOrUpdateServiceMonitors(ctx spi.ComponentContext) error {
...
 		// Enable sticky sessions, so there is no UI query skew in multi-replica prometheus clusters
 		ExtraAnnotations: common.SameSiteCookieAnnotations(prometheusName),
 	}
-	return common.CreateOrUpdateSystemComponentIngress(ctx, promProps)
+	err := common.CreateOrUpdateSystemComponentIngress(ctx, promProps)
+	if err != nil {
+		return err
+	}
+
+	alertmanagerEnabled, err := vzcr.IsAlertmanagerEnabled(ctx.EffectiveCR(), ctx)
```
kminder, line 773: Seems like this should be extracted to a separate function and that function should be unit tested.

mbenmclean:
We can discuss at the API review for this, and this would certainly be a follow up, but we should consider including alertmanager as a top level component so that users don't need to go add it under the prometheus operator overrides to enable it. We can then go modify the promtheus operator overrides under the hood. We also may want to provide some user friendly ways to set up receivers because configuring them from the prometheus operator overrides is pretty un-user-friendly from what I remember.

pkg/vzcr/enabled.go
```diff
@@ -652,3 +655,42 @@ func IsComponentStatusEnabled(cr runtime.Object, componentName string) bool {
 	}
 	return false
 }
+
+// IsAlertmanagerEnabled returns true only if the Alertmanager is explicitly enabled in the CR
```
mbenmclean, line 659: perhpas mention that it looks at the prometheus operator overrides in the comment

platform-operator/apis/verrazzano/v1alpha1/convert_from.go
```diff
@@ -93,17 +93,18 @@ func convertVerrazzanoInstanceFromV1Beta1(instance *v1beta1.InstanceInfo) *Insta
...
+		KeyCloakURL:     instance.KeyCloakURL,
+		RancherURL:      instance.RancherURL,
+		ElasticURL:      instance.OpenSearchURL,
+		KibanaURL:       instance.OpenSearchDashboardsURL,
+		GrafanaURL:      instance.GrafanaURL,
+		PrometheusURL:   instance.PrometheusURL,
+		KialiURL:        instance.KialiURL,
+		JaegerURL:       instance.JaegerURL,
+		ThanosQueryURL:  instance.ThanosQueryURL,
+		AlertmanagerURL: instance.AlertmanagerURL,
```
mbenmclean, line 107: Perhaps we should alphabetize these?

platform-operator/constants/constants.go
```diff
@@ -269,3 +269,6 @@ const ThanosInternalUserName = "verrazzano-thanos-internal"
 
 // VerrazzanoPlatformOperatorHelmName is the Helm release name of the Verrazzano Platform Operator
 const VerrazzanoPlatformOperatorHelmName = "verrazzano-platform-operator"
+
+// AlertmanagerIngress is the name of the ingress for Alertmanager
+const AlertmanagerIngress = "vmi-system-alertmanager"
```
mbenmclean, line 274: Do we want to have vmi in here? I know it matches prometheus but the name vmi is defunct and we're just creating more technical debt. We'll eventually want to remove references to vmi

pkg/vzcr/enabled.go
lake-of-dreams: added

pkg/vzcr/enabled.go
lake-of-dreams: mentioned

platform-operator/apis/verrazzano/v1alpha1/convert_from.go
lake-of-dreams: I think in code these should be in the order in which they are introduced for ease of diff/patch etc. In CR, they anyways appear in alphabetic order..

platform-operator/constants/constants.go
lake-of-dreams: removed vmi

lake-of-dreams:
>We can discuss at the API review for this, and this would certainly be a follow up, but we should consider including alertmanager as a top level component so that users don't need to go add it under the prometheus operator overrides to enable it. We can then go modify the promtheus operator overrides under the hood. We also may want to provide some user friendly ways to set up receivers because configuring them from the prometheus operator overrides is pretty un-user-friendly from what I remember.

I think having the top level component also modify prometheus operator overrides would leave some room for inconsistency as now there will be two ways to configure alertmanager - one via top level component and other via directly changing the overrides. Also I believe currently changing receiver and other configuration is in-line with syntax defined at https://prometheus.io/docs/alerting/latest/configuration/ and probably should be kept that way for easy migration. Will be discussed and taken up as follow ups.

kminder (approved):
Approved.

PR #6428, opened 2023-07-10
VZ-10153 MetricsTrait Converter

AnshulDalua:
Added implementation of MetricsTrait to service monitor conversion


examples/hello-helidon/hello-helidon-app.yaml
```diff
@@ -4,6 +4,7 @@ apiVersion: core.oam.dev/v1alpha2
 kind: ApplicationConfiguration
 metadata:
   name: hello-helidon
+  namespace: hello-helidon
```
anders-swanson, line 7: Please remove this ns

go.mod
```diff
@@ -228,4 +228,4 @@ replace (
 	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.13.1
 	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.8.0
 	sigs.k8s.io/kind => github.com/verrazzano/kind v0.0.0-20221129215948-885481909133
-)
+)
```
anders-swanson, line 231: missing newline at end of file

go.sum
```diff
@@ -1278,4 +1278,4 @@ sigs.k8s.io/structured-merge-diff/v4 v4.2.3 h1:PRbqxJClWWYMNV1dhaG4NsibJbArud9kF
 sigs.k8s.io/structured-merge-diff/v4 v4.2.3/go.mod h1:qjx8mGObPmV2aSZepjQjbmb2ihdVs8cGKBraizNC69E=
 sigs.k8s.io/yaml v1.2.0/go.mod h1:yfXDCHCao9+ENCvLSE62v9VSji2MKu5jeNfTrofGhJc=
 sigs.k8s.io/yaml v1.3.0 h1:a2VclLzOGrwOHDiV8EfBGhvjHvP46CtW5j6POvhYGGo=
-sigs.k8s.io/yaml v1.3.0/go.mod h1:GeOyir5tyXNByN85N/dRIT9es5UQNerPYEKK56eTBm8=
+sigs.k8s.io/yaml v1.3.0/go.mod h1:GeOyir5tyXNByN85N/dRIT9es5UQNerPYEKK56eTBm8=
```
anders-swanson, line 1281: missing newline at end of file

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
 )
 
+var (
+	weblogicPortNames = []string{"tcp-cbt", "tcp-ldap", "tcp-iiop", "tcp-snmp", "tcp-default", "tls-ldaps",
+		"tls-default", "tls-cbts", "tls-iiops", "tcp-internal-t3", "internal-t3"}
+)
+
+// var cli client.Reader
 func main() {
-	err := app.ConfData()
```
anders-swanson, line 11: Can you sync with Vrushali about loading the OAM components, application configurations?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+					}
+					metricsTraits = append(metricsTraits, metricsTrait)
+				}
+
+			}
+		}
+	}
+
+	//Create child resources of each metric trait
+	for _, trait := range metricsTraits {
```
anders-swanson, line 121: Similar to Ingress Trait, let's move common logic out of main function to packages

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+		createMetricsChildResources(trait)
+		//Put metricsTrait method
+	}
+
+	//Create child resources of each ingress trait
+	for _, trait := range ingressTraits {
+
+		fmt.Printf("Trait API Version: %s\n", trait.APIVersion)
+		fmt.Printf("Trait name: %s\n", trait.Name)
+		//createIngressChildResources(trait)
```
anders-swanson, line 134: Why is this commented out?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+		}
+	}
+
+	//Create child resources of each metric trait
+	for _, trait := range metricsTraits {
+
+		fmt.Printf("Trait API Version: %s\n", trait.APIVersion)
+		fmt.Printf("Trait name: %s\n", trait.Name)
+		createMetricsChildResources(trait)
+		//Put metricsTrait method
```
anders-swanson, line 126: Is a method missing here?

pkg/metrics/service_monitor.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/verrazzano/verrazzano/application-operator/constants"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	prometheusClusterNameLabel = "verrazzano_cluster"
+)
+
+// ScrapeInfo captures the information needed to construct the service monitor for a generic workload
+type ScrapeInfo struct {
```
anders-swanson, line 20: If these types and methods are copied from application operator, please update application operator to use this new package.

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+	}
+
+}
+func createMetricsChildResources(metricstrait *vzapi.MetricsTrait) {
+
+	createService()
+	createServiceMonitor(metricstrait)
+}
+// createChildResources creates the Gateway, VirtualService, DestinationRule and AuthorizationPolicy resources
+func createIngressChildResources(ingresstrait *vzapi.IngressTrait) {
```
anders-swanson, line 144: I thought Vrushali had moved these methods out already?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+func createServiceMonitor(trait *vzapi.MetricsTrait) (monitor promoperapi.ServiceMonitor){
+	// Creating a service monitor with name and namespace
+	pmName, err := createServiceMonitorName(trait, 0)
+	if err != nil {
+		print(err)
+	}
+
+	// Fetch workload resource using information from the trait
+	var workload *unstructured.Unstructured
+	workload, err = FetchWorkloadFromTrait(trait)
```
anders-swanson: We need to parse the associated component workload from the input - this is going out to the cluster itself.

anders-swanson: I misunderstood this, since the method name is the same as the application operator method name. disregard this comment

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+	}
+
+	vzPromLabels := !wlsWorkload
+
+	scrapeInfo := metrics.ScrapeInfo{
+		Ports:              len(getPortSpecs(trait, traitDefaults)),
+		BasicAuthSecret:    secret,
+		IstioEnabled:       &useHTTPS,
+		VZPrometheusLabels: &vzPromLabels,
+		//sClusterName:        clusters.GetClusterName(ctx, r.Client),
```
anders-swanson: Not sure what to use for cluster name, you can use "default" as a placeholder.

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+	serviceMonitor := promoperapi.ServiceMonitor{}
+	serviceMonitor.SetName(pmName)
+	serviceMonitor.SetNamespace(workload.GetNamespace())
+	metrics.PopulateServiceMonitor(scrapeInfo, &serviceMonitor)
+	return serviceMonitor
+
+}
+
+func createService(){
+
```
anders-swanson: empty?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+				authzPolicyName := fmt.Sprintf("%s-rule-%d-authz", ingresstrait.Name, index)
+				createVirtualService(ingresstrait, rule, vsHosts, vsName, services, gateway)
+				createDestinationRule(ingresstrait, rule, drName, services)
+				createAuthorizationPolicies(ingresstrait, rule, authzPolicyName, allHostsForTrait)
+			}
+		}
+	}
+}
+//creates Server Monitor Instance
+func createServiceMonitor(trait *vzapi.MetricsTrait) (monitor promoperapi.ServiceMonitor){
```
anders-swanson: This is more appoper code... can it be shared between the two apps somehow?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+
+	return resource, nil
+}
+func FetchContainedWorkload(workload *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	apiVersion, kind, _, err := GetContainedWorkloadVersionKindName(workload)
+	if err != nil {
+		return nil,err
+	}
+	_ = ""
+	u := &unstructured.Unstructured{}
```
anders-swanson: There's no workload here?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+	}
+
+	return resource, nil
+}
+func FetchContainedWorkload(workload *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	apiVersion, kind, _, err := GetContainedWorkloadVersionKindName(workload)
+	if err != nil {
+		return nil,err
+	}
+	_ = ""
```
anders-swanson: this is leftover from something?

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+	// If no secret name explicitly provided use the default secret name.
+	if secretName == nil && traitDefaults != nil {
+		secretName = traitDefaults.Secret
+	}
+	// If neither an explicit or default secret name provided do not fetch a secret.
+	if secretName == nil {
+		return nil, nil
+	}
+	// Use the workload namespace for the secret to fetch.
+	//secretNamespace, found, err := unstructured.NestedString(workload.Object, "metadata", "namespace")
```
anders-swanson: code is commented

tools/oam-converter/main.go
```diff
@@ -4,14 +4,1346 @@
...
+		return nil,err
+	}
+	_ = ""
+	u := &unstructured.Unstructured{}
+	u.SetAPIVersion(apiVersion)
+	u.SetKind(kind)
+
+	return u, nil
+}
+func GetContainedWorkloadVersionKindName(workload *unstructured.Unstructured) (string, string, string, error) {
```
anders-swanson: It's very difficult to tell which code came from app operator, which code is new, and which came from app operator but was modified. Some organization will help here, as well as all possible code reuse between app operator and oam converter if code does need to be shared.

Suggest at least to start with putting app operator code in a different package so it is easier to identify, it is difficult to review currently.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -93,7 +132,25 @@ func ConfData() error {
...
 }
-
+func iterateDirectory(path string) ([]string, []string) {
+	var appFiles []string
+	var compFiles []string
+	filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+		if strings.Contains(info.Name(), "yaml"){
```
anders-swanson, line 142: Can you check for YAML files by verifying that the file ends with ".yaml" or ".yml". Both are valid file extensions for YAML.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -93,7 +132,25 @@ func ConfData() error {
...
+			if strings.Contains(info.Name(), "app"){
```
anders-swanson, line 143: To find application configuration and components, we need to iterate through the documents in each YAML file. Note that a given directory may have more than one application configuration, and it may not be in a file named "app". Same goes for components, we can discuss.

examples/hello-helidon/hello-helidon-app.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, 2022, Oracle and/or its affiliates.
```
anders-swanson, line 1: Copyright should be 2020, 2023. Our convention is to keep the earliest copyright year.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
-		fmt.Println("Failed to read YAML file:", err)
-		return err
+
+	//check if CLI has 2 vars
+	if len(os.Args) != 3 {
+		fmt.Println("Not enough args to run tool")
+		return nil
+	} else {
+		inputDirectory = os.Args[1]
+		//outputDirectory = os.Args[2]
```
anders-swanson, line 47: removed commented line

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
 	helidonWorkloads := []*vzapi.VerrazzanoHelidonWorkload{}
 	coherenceWorkloads := []*vzapi.VerrazzanoCoherenceWorkload{}
-	//Read app File
-	appData, err := ioutil.ReadFile("../../examples/bobs-books/bobs-books-app.yaml")
-	if err != nil {
-		fmt.Println("Failed to read YAML file:", err)
-		return err
+
+	//check if CLI has 2 vars
+	if len(os.Args) != 3 {
```
anders-swanson, line 42: typically we want to check inputs first thing, would suggest to make a separate function just for validating inputs

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
 
 func ConfData() error {
 
+	var inputDirectory string
+	//var outputDirectory string
```
anders-swanson, line 37: remove commented line

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+		inputDirectory = os.Args[1]
+		//outputDirectory = os.Args[2]
 	}
 
-	//Read Comp file
-	compData, err := ioutil.ReadFile("../../examples/bobs-books/bobs-books-comp.yaml")
-	if err != nil {
-		fmt.Println("Failed to read YAML file:", err)
-		return err
+	var appDataArr[](map[string]interface{})
```
anders-swanson, line 50: `appDataArr[] map[string]interface{}` is the appropriate style

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+	var compDataArr [](map[string]interface{})
+
+	//iterate through user inputted directory
+	appFiles, compFiles := iterateDirectory(inputDirectory)
+
+	//Loop through all app files and store into appDataArr
+	for _, input := range appFiles {
+		appData, err := ioutil.ReadFile(input)
```
anders-swanson, line 58: note that an applciation configuration may not be the only thing in a YAML file. consider the following YAML file:
example.yaml with multiple resources:
```yaml
# example.yaml
kind: ApplicationConfiguration
# ... application configuration definition
---
kind: Component
# ... component definition
---
kind: Service
# ... service definition
```

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+		if err != nil {
+			fmt.Println("Failed to read YAML file:", err)
+			return err
+		}
+		appMap := make(map[string]interface{})
+		err = yaml.Unmarshal(appData, &appMap)
+		if err != nil {
+			log.Fatalf("Failed to unmarshal YAML: %v", err)
```
anders-swanson, line 66: why do we return err sometimes, but log.Fatalf other times? I would suggest return err always, and let the caller handle the error. If its main function, then we can log.Fatalf to terminate the program on errors that the program can't handle

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+		}
+		if value, ok := appMap["kind"]; ok {
```
anders-swanson, line 68: Kind is not enough to assume ApplicationConfiguration (though this is a major corner case). API Group and Version should be checked as well for a comprehensive verification. I would suggest to create function IsApplicationConfiguration to segment this logic out (you can do similar for Component or any other type).

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+	for _, input := range compFiles {
+		compData, err := ioutil.ReadFile(input)
+		if err != nil {
+			fmt.Println("Failed to read YAML file:", err)
+			return err
+		}
+		compMap := make(map[string]interface{})
+		err = yaml.Unmarshal(compData, &compMap)
+		if err != nil {
 		log.Fatalf("Failed to unmarshal YAML: %v", err)
```
anders-swanson, line 85: indentation is wrong here

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+				compDataArr = append(compDataArr, compMap)
+			}
+		}
 	}
 
-	workloadTraitMap, componentNames, ingressTraits, metricsTraits := extract.HandleYAMLStructurePanic(appMap)
+
+	workloadTraitMap, componentNames, ingressTraits, metricsTraits := extract.HandleYAMLStructurePanic(appDataArr[1])
 	fmt.Print(ingressTraits)
+	fmt.Print(metricsTraits)
```
anders-swanson, line 97: Is this for debugging?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -91,7 +132,25 @@ func ConfData() error {
 	}
 	return nil
 }
-
+func iterateDirectory(path string) ([]string, []string) {
```
anders-swanson, line 135: I'd like to see a different tack for this method, something along these lines:
1. Walk the directory
2. For each YAML file, split the file into a buffer, and split the bufer into documents based on the `---` delimiter:
  a. For each document, determine whether its an ApplicationConfiguration, Component, or something else
  b. Store the resulting objects in a struct, marshalling application configurations and components into their relevant structs:
```
type InputDocuments struct {
  // the list of input application configurations
  applicationConfigurations []ApplicationConfiguration
  // components is a map of component names to component objects
  components map[string]Component
  // documents is the list of YAML documents that are neither application configurations nor components
  documents []map[string]interface{}
}
```

Now, the next pipeline of the program can begin to process the application configurations and components to generate the metrics traits, ingress traits, and any other converted resources.

tools/oam-converter/pkg/bobs-books-app.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: core.oam.dev/v1alpha2
```
anders-swanson, line 1: Did you mean to commit this file? You can use the path directly to the examples for testing.

tools/oam-converter/pkg/bobs-books-comp.yaml
```diff
@@ -0,0 +1,518 @@
+apiVersion: core.oam.dev/v1alpha2
```
anders-swanson, line 1: please remove from commit with above file and use examples directory instead

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -29,35 +33,71 @@ func contains(arr []string, target string) bool {
...
+	appFiles, compFiles := iterateDirectory(inputDirectory)
+
+	//Loop through all app files and store into appDataArr
+	for _, input := range appFiles {
+		appData, err := ioutil.ReadFile(input)
+		if err != nil {
+			fmt.Println("Failed to read YAML file:", err)
+			return err
+		}
+		appMap := make(map[string]interface{})
```
anders-swanson, line 63: would recommend to unmarshall to unstructured.Unstructured objects. This has a lot of helpful methods for working with k8s objects. Check this sample codes to unmarshall YAML bytes to an unstructured, using apimachinery:

```go
import (
  "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
  "k8s.io/apimachinery/pkg/runtime"
  apiyaml "k8s.io/apimachinery/pkg/util/yaml"
)

...
func toUnstructured(o []byte) (*unstructured.Unstructured, error) {
  j, err := apiyaml.ToJSON(o)
  if err != nil {
    return nil, err
  }
  obj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, j)
  return obj, err
}
```

anders-swanson: This way you can work with the unstructured object, rather than a raw map

tools/oam-converter/pkg/helidon-config-app.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2021, 2022, Oracle and/or its affiliates.
```
anders-swanson, line 1: please remove these from the oam-converter, should use examples/hello-helidon instead

tools/oam-converter/pkg/resources/coherenceresources/coherenceresources.go
```diff
@@ -81,3 +82,69 @@ func createDestinationFromRuleOrCoherenceWorkload(rule vzapi.IngressRule, cohere
...
+		}
+	}
+	return &vzapi.MetricsTraitSpec{
+		Ports: []vzapi.PortSpec{{
+			Port: &port,
+			Path: &path,
+		}},
+		Path:   &path,
+		Secret: secret,
+		//Scraper: &r.Scraper
```
anders-swanson, line 109: commented, why?

tools/oam-converter/pkg/resources/coherenceresources/coherenceresources.go
```diff
@@ -81,3 +82,69 @@ func createDestinationFromRuleOrCoherenceWorkload(rule vzapi.IngressRule, cohere
...
+	}, nil
+}
+func fetchCoherenceMetricsSpec(workload *unstructured.Unstructured) (*bool, *int, *string, error) {
```
anders-swanson, line 112: when you have more than 2 or 3 return values, suggest to return as a struct, e.g.,
```go
type CoherenceMetricsParams {
  Port *int
  Enabled *bool
  Secret *string
}
```

tools/oam-converter/pkg/resources/metrics/metrics.go
```diff
@@ -0,0 +1,14 @@
...
+
+package metrics
+
+import (
+	vzapi "github.com/verrazzano/verrazzano/application-operator/apis/oam/v1alpha1"
+	monitor "github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/servicemonitor"
+)
+
+func CreateMetricsChildResources(metricstrait *vzapi.MetricsTrait) {
+	//createService()
```
anders-swanson, line 12: commented, why?

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,400 @@
...
+	// Populate the keep labels to match the oam pod labels
+	scrapeInfo.KeepLabels = map[string]string{
+		"__meta_kubernetes_pod_label_app_oam_dev_name":      trait.Labels[oam.LabelAppName],
+		"__meta_kubernetes_pod_label_app_oam_dev_component": trait.Labels[oam.LabelAppComponent],
+	}
+
+	serviceMonitor.SetName(pmName)
+	serviceMonitor.SetNamespace(workload.GetNamespace())
+	PopulateServiceMonitor(scrapeInfo, &serviceMonitor)
+	PrintServiceMonitor(&serviceMonitor)
```
anders-swanson, line 94: Is this for debugging?

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,400 @@
...
+}
+func createServiceMonitorName(trait *vzapi.MetricsTrait, portNum int) (string, error) {
+	sname, err := createJobOrServiceMonitorName(trait, portNum)
+	if err != nil {
+		return "", err
+	}
+	return strings.Replace(sname, "_", "-", -1), nil
+}
+func createJobOrServiceMonitorName(trait *vzapi.MetricsTrait, portNum int) (string, error) {
+	namespace := getNamespaceFromObjectMetaOrDefault(trait.ObjectMeta)
```
anders-swanson, line 106: Trait doesn't have the namespace. It should come from application configuration

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,400 @@
...
+func getNamespaceFromObjectMetaOrDefault(meta metav1.ObjectMeta) string {
+	name := meta.Namespace
+	if name == "" {
+		return "default"
+	}
+	return name
+}
+
+// ScrapeInfo captures the information needed to construct the service monitor for a generic workload
+type ScrapeInfo struct {
```
anders-swanson, line 139: We have this type defined twice, I am confused

pkg/metrics/service_monitor.go
```diff
@@ -0,0 +1,226 @@
...
+	// The map to generate keep labels
+	// This matches the expected pod labels to the scrape config
+	KeepLabels map[string]string
+	// The name of the cluster for the selected workload
+	ClusterName string
+}
+
+// PopulateServiceMonitor populates the Service Monitor to prepare for a create or update
+// the Service Monitor reflects the specifications defined in the ScrapeInfo object
+func PopulateServiceMonitor(info ScrapeInfo, serviceMonitor *promoperapi.ServiceMonitor) error {
```
anders-swanson, line 40: some of this code seems copied twice, can you please clean it up

tools/oam-converter/pkg/resources/weblogicresources/weblogicresources.go
```diff
@@ -6,61 +6,76 @@ package weblogicresources
...
+		return nil, err
+	}
+	return &vzapi.MetricsTraitSpec{
+		Ports: []vzapi.PortSpec{{
+			Port: &port,
+			Path: &path,
+		}},
+		Path:   &path,
+		Secret: secret,
+		//Scraper: &r.Scraper
```
anders-swanson, line 69: copied from above?  also, commented line here

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,407 @@ package traits
...
+	secretName := trait.Spec.Secret
+	// If no secret name explicitly provided use the default secret name.
+	if secretName == nil && traitDefaults != nil {
+		secretName = traitDefaults.Secret
+	}
+	// If neither an explicit or default secret name provided do not fetch a secret.
+	if secretName == nil {
+		return nil, nil
+	}
+	// Use the workload namespace for the secret to fetch.
```
anders-swanson, line 374: commented code?

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,407 @@ package traits
...
+		"oam.verrazzano.io/v1alpha1.VerrazzanoCoherenceWorkload": {Group: "coherence.oracle.com", Version: "v1", Kind: "Coherence"},
+	}
+	key := fmt.Sprintf("%s.%s", apiVersion, kind)
+	gvk, ok := workloadToContainedGVKMap[key]
+	if ok {
+		return &gvk
+	}
+	return nil
+}
+func FetchTraitDefaults(workload *unstructured.Unstructured) (*vzapi.MetricsTraitSpec, bool, error) {
```
anders-swanson, line 300: app operator code...?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
+			if err != nil {
+				log.Fatalf("Failed to unmarshal YAML: %v", err)
+			}
+			compKind, found, err := unstructured.NestedString(component, "kind")
+			if !found || err != nil {
+				return err
+			}
+			compApiVersion, found, err := unstructured.NestedString(component, "apiVersion")
+			if !found || err != nil {
+				return err
```
anders-swanson, line 53: if !found err will be nil. Suggest return custom error here, like `errors.New("apiVersion was not found as a string field")` or similar.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
+			}
+			if compKind == "Component" && compApiVersion == "core.oam.dev/v1alpha2" {
+				components = append(components, component)
+			}
+			if compKind == "ApplicationConfiguration" && compApiVersion == "core.oam.dev/v1alpha2" {
+				appData = append(appData, component)
+			}
```
anders-swanson, line 60: Can you add a TODO here that if the kind is something else, let's put it in another array to write out as-is. This will satisfy the case when the application has K8s resources that are not OAM-specific, which is a valid use case. E.g., the user has an application configuration, a component, and a service. Just put the TODO here so we know to do it later.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
 	}
 
-	for _, trait := range metricsTraits {
+	outputResources, err := resources.CreateResources(conversionComponents)
 
-		fmt.Printf("Trait API Version: %s\n", trait.APIVersion)
-		fmt.Printf("Trait name: %s\n", trait.Name)
-		//Put metricsTrait method
-	}
+	printDirectory(outputDirectory, outputResources)
```
anders-swanson, line 77: Is this for logging? Any other reason this would be here?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
-				weblogicWorkload := &vzapi.VerrazzanoWebLogicWorkload{}
-				workloadJSON, err := json.Marshal(workload)
-
-				if err != nil {
-					log.Fatalf("Failed to marshal trait: %v", err)
-
-				}
+	filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.Fatalf(err.Error())
```
anders-swanson, line 90: return error here, let callers handle

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
-func segregateWorkloads(weblogicMap map[string]*vzapi.VerrazzanoWebLogicWorkload, components []map[string]interface{}, componentNames []string, helidonWorkloads []*vzapi.VerrazzanoHelidonWorkload, coherenceWorkloads []*vzapi.VerrazzanoCoherenceWorkload) ([]*vzapi.VerrazzanoCoherenceWorkload, []*vzapi.VerrazzanoHelidonWorkload, map[string]*vzapi.VerrazzanoWebLogicWorkload, error) {
-	//A weblogic map with the key as component name and value as a VerrazzanoWeblogicWorkload struct
-
-	for _, comp := range components {
-
-		var name string
-
-		kind := comp["spec"].(map[string]interface{})["workload"].(map[string]interface{})["kind"].(string)
-		name = comp["metadata"].(map[string]interface{})["name"].(string)
+func iterateDirectory(path string) []string {
```
anders-swanson, line 85: return `([]string, error)`

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
+}
+func printDirectory(directoryPath string, outputResources *types.KubeRecources)(error){
+	fileName := "output.yaml"
+	filePath := filepath.Join(directoryPath, fileName)
+	f, err := os.Create(filePath)
+	var output string
+	if err != nil {
+		return err
+	}
+	if outputResources.VirtualServices != nil {
```
anders-swanson, line 107: why is this stuff commented?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
+			log.Fatalf(err.Error())
+		}
+		if strings.Contains(info.Name(), "yaml") || strings.Contains(info.Name(), "yml") {
+			files = append(files, path)
+		}
+		return nil
+	})
+	return files
+}
+func printDirectory(directoryPath string, outputResources *types.KubeRecources)(error){
```
anders-swanson, line 99: can we call this writeKubeResources or something, and rename `directoryPath` to `outputDirectory` or similar.

tools/oam-converter/pkg/app/confdata.go
anders-swanson: not implemented yet?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -5,201 +5,143 @@ package app
...
+			out, err := yaml.JSONToYAML(r)
+			if err != nil {
+				return err
 			}
+			output = output + "---\n" + string(out)
 		}
 	}
-	return coherenceWorkloads, helidonWorkloads, weblogicMap, nil
-}
+	defer f.Close()
```
anders-swanson, line 140: Place the defer call earlier, right after checking the Open call error

tools/oam-converter/pkg/controllers/metricstrait_utils.go
```diff
@@ -0,0 +1,118 @@
...
+	secretName := trait.Spec.Secret
+	// If no secret name explicitly provided use the default secret name.
+	if secretName == nil && traitDefaults != nil {
+		secretName = traitDefaults.Secret
+	}
+	// If neither an explicit or default secret name provided do not fetch a secret.
+	if secretName == nil {
+		return nil, nil
+	}
+	// Use the workload namespace for the secret to fetch.
```
anders-swanson, line 58: Remove or explain commented code

tools/oam-converter/pkg/controllers/workloads.go
```diff
@@ -0,0 +1,117 @@
...
+	port := 8080
+	path := "/metrics"
+	return &vzapi.MetricsTraitSpec{
+		Ports: []vzapi.PortSpec{{
+			Port: &port,
+			Path: &path,
+		}},
+		Path:    &path,
+		Secret:  nil,
+		//Scraper: &r.Scraper
```
anders-swanson, line 115: Why is this commented?

tools/oam-converter/pkg/controllers/workloads.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	return &vzapi.MetricsTraitSpec{
+		Ports: []vzapi.PortSpec{{
+			Port: &port,
+			Path: &path,
+		}},
+		Path:   &path,
+		Secret: secret,
+		//Scraper: &r.Scraper
```
anders-swanson, line 63: why is this commented?

tools/oam-converter/pkg/controllers/workloads.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, err
+	}
+	return &vzapi.MetricsTraitSpec{
+		Ports: []vzapi.PortSpec{{
+			Port: &port,
+			Path: &path,
+		}},
+		Path:   &path,
+		Secret: secret,
+		//Scraper: &r.Scraper
```
anders-swanson, line 26: ditto here

tools/oam-converter/pkg/resources/coherenceresources/coherenceresources.go
```diff
@@ -5,79 +5,44 @@ package coherenceresources
...
 	azp "github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/authorizationpolicy"
 	gw "github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/gateway"
-	istio "istio.io/api/networking/v1beta1"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/types"
 )
 
-func CreateIngressChildResourcesFromCoherence(traitName string, ingressTrait *vzapi.IngressTrait, cohereneWorkload *vzapi.VerrazzanoCoherenceWorkload) error {
-	rules := ingressTrait.Spec.Rules
+func CreateIngressChildResourcesFromCoherence(conversionComponent *types.ConversionComponents) error {
+	rules := conversionComponent.IngressTrait.Spec.Rules
```
anders-swanson, line 15: IngressTrait can be nil here?

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, conversionComponent := range conversionComponents {
+		if conversionComponent.MetricsTrait != nil {
+			serviceMonitor, err := metrics.CreateMetricsChildResources(conversionComponent)
+			if err != nil {
+				return &outputResources,errors.New("failed to create Child resources from Metrics Trait")
+			}
+			outputResources.ServiceMonitors = append(outputResources.ServiceMonitors, serviceMonitor)
+		}
+		if conversionComponent.WeblogicworkloadMap != nil {
+			//conversionComponent.AppNamespace, conversionComponent.AppName, conversionComponent.ComponentName, conversionComponent.IngressTrait, conversionComponent.WeblogicworkloadMap[conversionComponent.ComponentName]
```
anders-swanson, line 26: remove or explain commented code

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"errors"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/coherenceresources"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/helidonresources"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/metrics"
+	weblogic "github.com/verrazzano/verrazzano/tools/oam-converter/pkg/resources/weblogicresources"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/types"
+)
+
+func CreateResources(conversionComponents []*types.ConversionComponents) (*types.KubeRecources, error) {
```
anders-swanson, line 15: CreateKubeResources

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+		if conversionComponent.Coherenceworkload != nil {
+			err := coherenceresources.CreateIngressChildResourcesFromCoherence(conversionComponent)
+			if err != nil {
+				return &outputResources,errors.New("failed to create Child resources from Coherence workload")
+
+			}
+		}
+		if conversionComponent.Helidonworkload != nil {
```
anders-swanson, line 40: What about ingress trait for generic workload?

anders-swanson: @shahvrushali22

tools/oam-converter/pkg/app/confdata.go
AnshulDalua: This is to print the new resources into the directory the user requests

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -0,0 +1,49 @@
...
+
+func CreateResources(conversionComponents []*types.ConversionComponents) (*types.KubeRecources, error) {
+	outputResources := types.KubeRecources{}
+	for _, conversionComponent := range conversionComponents {
+		if conversionComponent.MetricsTrait != nil {
+			serviceMonitor, err := metrics.CreateMetricsChildResources(conversionComponent)
+			if err != nil {
+				return &outputResources,errors.New("failed to create Child resources from Metrics Trait")
+			}
+			outputResources.ServiceMonitors = append(outputResources.ServiceMonitors, serviceMonitor)
```
anders-swanson, line 23: Is there a point to segregating the output resources by type? If we are only writing them out, you can put them all in a `[]runtime.Object`.

tools/oam-converter/pkg/resources/gateway/secret.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	coallateHosts "github.com/verrazzano/verrazzano/pkg/ingresstrait"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/types"
+	vsapi "istio.io/client-go/pkg/apis/networking/v1beta1"
+)
+
+func CreateCertificateAndSecret(conversionComponent *types.ConversionComponents) (*vsapi.Gateway, []string, error) {
+
+	// If there are no rules, create a single default rule
```
anders-swanson, line 14: Is this supposed to reference a block of code?

tools/oam-converter/pkg/resources/helidonresources/destinationrule.go
```diff
@@ -105,27 +109,28 @@ func mutateDestinationRuleFromHelidonWorkload(destinationRule *istioclient.Desti
...
-
-	if helidonWorkload.Spec.DeploymentTemplate.PodSpec.Containers[0].Ports[0].ContainerPort != 0 {
-		dest := istio.HTTPRouteDestination{
-			Destination: &istio.Destination{Host: helidonWorkload.Name}}
-		// Set the port to rule destination port
-		dest.Destination.Port = &istio.PortSelector{Number: uint32(helidonWorkload.Spec.DeploymentTemplate.PodSpec.Containers[0].Ports[0].ContainerPort)}
-		return &dest, nil
-	}
+func createDestinationFromHelidonWorkload(rulePort uint32, helidonWorkload *unstructured.Unstructured) (*istio.HTTPRouteDestination, error) {
+	//	Spec.DeploymentTemplate.PodSpec.Containers[0].Ports[0].ContainerPort != 0 {
```
anders-swanson, line 128: Unsure what is going on here

tools/oam-converter/pkg/types/types.go
```diff
@@ -0,0 +1,31 @@
...
+	AppNamespace        string
+	IngressTrait        *vzapi.IngressTrait
+	MetricsTrait        *vzapi.MetricsTrait
+	Helidonworkload     *unstructured.Unstructured
+	Coherenceworkload   *unstructured.Unstructured
+	WeblogicworkloadMap map[string]*unstructured.Unstructured
+}
+
+type KubeRecources struct {
+	VirtualServices  []*vsapi.VirtualService
```
anders-swanson, line 26: consider using `[]runtime.Object` if we are only writing these to a file.

tools/oam-converter/pkg/types/types.go
```diff
@@ -0,0 +1,31 @@
...
+
+type ConversionComponents struct {
+	AppName             string
+	ComponentName       string
+	AppNamespace        string
+	IngressTrait        *vzapi.IngressTrait
+	MetricsTrait        *vzapi.MetricsTrait
+	Helidonworkload     *unstructured.Unstructured
+	Coherenceworkload   *unstructured.Unstructured
+	WeblogicworkloadMap map[string]*unstructured.Unstructured
```
anders-swanson, line 22: How are generic workloads handled here, e.g., Deployment or other that might have traits?

tools/oam-converter/pkg/resources/helidonresources/helidonresources.go
```diff
@@ -5,60 +5,43 @@ package helidonresources
...
 		if err != nil {
 			print(err)
 			return err
 		}
-		gateway, err := gw.CreateGateway(traitName, ingressTrait, allHostsForTrait, gwName, secretName)
+
+		vsName := fmt.Sprintf("%s-rule-%d-vs", conversionComponent.ComponentName, index)
+		drName := fmt.Sprintf("%s-rule-%d-dr", conversionComponent.ComponentName, index)
+		authzPolicyName := fmt.Sprintf("%s-rule-%d-authz", conversionComponent.ComponentName, index)
+		err = createVirtualServiceFromHelidonWorkload(conversionComponent.IngressTrait, rule, vsHosts, vsName, gateway, conversionComponent.Helidonworkload)
```
anders-swanson, line 34: either ingresstrait or helidonworkload can be nil here?

anders-swanson: consider patterns like making `createVirtualServiceFromHelidonWorkload` a member function a ConversionComponent. You don't have to do this, but it's an organizational thing.

tools/oam-converter/pkg/resources/helidonresources/helidonresources.go
```diff
@@ -5,60 +5,43 @@ package helidonresources
...
-			}
-			err = azp.CreateAuthorizationPolicies(ingressTrait, rule, authzPolicyName, allHostsForTrait)
-			if err != nil {
-				return err
-			}
+		err = createDestinationRuleFromHelidonWorkload(conversionComponent.IngressTrait, rule, drName, conversionComponent.Helidonworkload)
+		if err != nil {
+			return err
+		}
+		err = azp.CreateAuthorizationPolicies(conversionComponent.IngressTrait, rule, authzPolicyName, allHostsForTrait)
```
anders-swanson, line 42: What about generic workloads?

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,100 @@
...
+	// Creating a service monitor with name and namespace
+	pmName, err := utils.CreateServiceMonitorName(trait, conversionComponent.AppName, conversionComponent.ComponentName, 0)
+	if err != nil {
+		return &serviceMonitor, err
+	}
+
+	// Fetch workload resource using information from the trait
+	var workload *unstructured.Unstructured
+
+	if conversionComponent.Helidonworkload != nil {
```
anders-swanson, line 35: I thought we were moving to using unstrcutured for handling generic workloads, if we are still splitting by type I am not understanding the purpose.

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
-	defer func() {
-		if err := recover(); err != nil {
-			log.Println("panic occurred in YAML structure:", err)
+// ExtractTrait - Extract traits from the app map
+func ExtractTrait(appMaps []map[string]interface{}) ([]*types.ConversionComponents, error) {
+	conversionComponents := []*types.ConversionComponents{}
+	for _, appMap := range appMaps {
+		appMetadata, found, err := unstructured.NestedMap(appMap, "metadata")
+		if !found || err != nil {
+			return nil, errors.New("app metadata doesn't exist")
```
anders-swanson, line 23: doesn't exist, or may exist but be wrong type

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
+		}
+		workload, found, err := unstructured.NestedMap(spec, "workload")
+		if !found || err != nil {
+			return nil, errors.New("workload in a component doesn't exist")
+		}
+		kind, found, err := unstructured.NestedString(workload, "kind")
+		if !found || err != nil {
+			return nil, errors.New("workload kind in a component doesn't exist")
+		}
+		//name = comp["metadata"].(map[string]interface{})["name"].(string)
```
anders-swanson, line 129: remove commented line

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
+						log.Fatalf("Failed to marshal trait: %v", err)
+					}
+
+					err = json.Unmarshal(workloadJSON, &helidonWorkload)
+					if err != nil {
+						fmt.Printf("Failed to unmarshal: %v\n", err)
+
+					}
+					conversionComponents[i].Helidonworkload = helidonWorkload
+				case "VerrazzanoCoherenceWorkload":
```
anders-swanson, line 184: 1. JSON unmarshalling is not case specific, and should be moved up.
2. There is no handling for generic components here

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
 
-	// Create an ingress map
-	ingressMap := make(map[string]*vzapi.IngressTrait)
+		//Checking if the specific component name is present in the component names array
+		//where component names array is the array of component names
+		//which has ingress traits applied on it
 
-	var componentNames []string
+		for i := range conversionComponents {
+			if conversionComponents[i].ComponentName == name {
```
anders-swanson, line 144: nested loop may not be necessary here, if you store component information from apps differently.

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
+// ExtractWorkload - Extract workload from comp map
+func ExtractWorkload(components []map[string]interface{}, conversionComponents []*types.ConversionComponents) ([]*types.ConversionComponents, error) {
+	weblogicMap := make(map[string]*unstructured.Unstructured)
+	for _, comp := range components {
 
-	ingressTraits := []*vzapi.IngressTrait{} //Array of ingresstraits
-	metricsTraits := []*vzapi.MetricsTrait{}
+		spec, found, err := unstructured.NestedMap(comp, "spec")
+		if !found || err != nil {
+			return nil, errors.New("spec key in a component doesn't exist")
```
anders-swanson, line 119: similar notice about error handling as mentioned earlier

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
-		componentTraits, ok := componentMap[consts.YamlTraits].([]interface{})
-		if ok && len(componentTraits) > 0 {
-			for _, trait := range componentTraits {
-				traitMap := trait.(map[string]interface{})
-				traitSpec := traitMap[consts.TraitComponent].(map[string]interface{})
-				traitKind := traitSpec[consts.TraitKind].(string)
-				if traitKind == consts.IngressTrait {
-					ingressTrait := &vzapi.IngressTrait{}
-					traitJSON, err := json.Marshal(traitSpec)
+					//weblogicWorkload := &vzapi.VerrazzanoWebLogicWorkload{}
```
anders-swanson, line 149: commented code

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -5,76 +5,203 @@ package traits
...
+						fmt.Printf("Failed to unmarshal: %v\n", err)
+
 					}
-					fmt.Println(componentMap["componentName"].(string))
 
-					//Assigning ingresstrait to a map with the key as component name
-					ingressMap[componentMap["componentName"].(string)] = ingressTrait
+					//putting into map of workloads whose key is the component name and
+					//value is the weblogic workload
+					weblogicMap[name] = weblogicWorkload
```
anders-swanson, line 166: why are we doing this for weblogic, but no other workload type?

anders-swanson (changes requested):
Please see comments, and look for ways to make the code more generic.

examples/hello-helidon/hello-helidon-app-custom-port.yaml
```diff
@@ -1,29 +0,0 @@
-# Copyright (c) 2023, Oracle and/or its affiliates.
-# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-apiVersion: core.oam.dev/v1alpha2
-kind: ApplicationConfiguration
-metadata:
-  name: hello-helidon
```
davisgomes, line 6: Is this example meant to be deleted?

examples/hello-helidon/hello-helidon-app-scaler-trait.yaml
```diff
@@ -1,35 +0,0 @@
-# Copyright (c) 2020, 2023, Oracle and/or its affiliates.
-# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-
-apiVersion: core.oam.dev/v1alpha2
```
davisgomes, line 4: Same here. Should this be deleted?

examples/hello-helidon/hello-helidon-app.yaml
```diff
@@ -1,6 +1,5 @@
-# Copyright (c) 2020, 2023, Oracle and/or its affiliates.
+# Copyright (c) 2023, Oracle and/or its affiliates.
```
davisgomes, line 1: This should remain in this format (first year is the creation year and second year is the last updated year):
`# Copyright (c) 2020, 2023, Oracle and/or its affiliates.`

examples/hello-helidon/hello-helidon-app-custom-port.yaml
anders-swanson: please restore changes to examples directory

AnshulDalua: yes this was meant to be deleted, not sure why its here again

AnshulDalua: no not sure why this got deleted, will fix

tools/oam-converter/pkg/app/confdata.go
davisgomes: Even better if you are able to use flags. I think it would help make it more obvious how the tool is used: https://pkg.go.dev/flag

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -26,9 +28,9 @@ func ConfData() error {
 
 	//Check the length of args
 	if len(os.Args) != 3 {
-		fmt.Println("Not enough args to run tool")
-		return nil
+		return errors.New("Not enough args to run tool")
```
davisgomes, line 31: This error message is too vauge. What args? What is the tool expecting from the user?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -39,18 +41,16 @@ func ConfData() error {
...
 	files, err := iterateDirectory(inputDirectory)
 	if err != nil {
 		fmt.Println("Error in iterating over directory", err)
 	}
-	var data []byte
-
 	//Read each file from the input directory
 	for _, input := range files {
-		data, _ = ioutil.ReadFile(input)
+		data, _ := ioutil.ReadFile(input)
```
davisgomes, line 53: Handle this error if that is the second argument.

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -103,163 +99,90 @@ func ConfData() error {
...
-
+	r, err := json.Marshal(object)
+	if err != nil {
+		return err
+	}
+	output, err := yaml.JSONToYAML(r)
+	if err != nil {
+		return err
 	}
+	_, err = f.WriteString(string(output))
```
davisgomes, line 180: Does this support multiple objects per file? I.e. two Gateways or two Service Monitors.

davisgomes: This seems like it will just append the objects together without splitting them. Can you test this scenario?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -103,163 +99,90 @@ func ConfData() error {
...
+	case *promoperapi.ServiceMonitor:
+		object = index
+		fileName = "servicemonitor.yaml"
+		filePath = filepath.Join(outputDirectory, fileName)
+	case []any:
+		object = index
+		fileName = "output.yaml"
+		filePath = filepath.Join(outputDirectory, fileName)
+	default:
+		return fmt.Errorf("Wrong data type%v", val)
```
davisgomes, line 160: Check output formatting on this error message, and maybe "Unsupported data type" would be more clear?

tools/oam-converter/pkg/controllers/metricstrait_utils.go
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+// source file: application-operator/controllers/metricstrait/metricstrait_utils.go
+package controllers
+
+import (
+	"context"
+	"fmt"
+	vzapi "github.com/verrazzano/verrazzano/application-operator/apis/oam/v1alpha1"
+	vznav "github.com/verrazzano/verrazzano/application-operator/controllers/navigation"
```
davisgomes, line 10: A lot of this code seems copied from the application operator. Can you extract the code in a way that both can use the shared code?

anders-swanson: I'd previously asked Anshul to try to do this, and it was it was not possible to share the code, to identify the app operator copied code somehow so we know where it was sourced from. Maybe Anshul can clarify why this code needed to be copied rather than shared

tools/oam-converter/pkg/resources/gateway/gateway.go
```diff
@@ -57,12 +58,25 @@ func buildCertificateName(trait *vzapi.IngressTrait, appNamespace string) string
 
 // buildLegacyCertificateName will generate a cert name
 func buildLegacyCertificateName(trait *vzapi.IngressTrait, appNamespace string, appName string) string {
+
+	_, ok := trait.Labels[oam.LabelAppName]
```
anders-swanson, line 62: I don't think the trait will have these labels? This is OAM specific?

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -103,163 +99,90 @@ func ConfData() error {
...
-		}
+func writeToDirectory(outputDirectory string, index any) error {
+	//fileName := "" + index.TypeMeta.Kind + ".yaml"
+	var fileName string
+	var filePath string
+	var object any
+
+	//check to find out what resource is being manipulated and printed
+	switch val := index.(type) {
+	case map[string]interface{}:
```
anders-swanson, line 135: would like to see this handled a little differently, something like this:
1. for each object, get its name and gvk. Since its a list of any, you might need to turn it to an unstructured.Unstructured first.
2. construct a filename based on the gvk and name. 
3. convert the object to bytes, and write it out to the file.

this will handle any type of resource

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -103,163 +99,90 @@ func ConfData() error {
...
+		object = index.(map[string]interface{})
+		fileName = "gateway.yaml"
+		filePath = filepath.Join(outputDirectory, fileName)
+	case []*vsapi.VirtualService:
```
anders-swanson, line 139: why do we have a nested list of virtual service? Since its a list of any in the outer loop, something is going wrong in the append

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -35,46 +30,52 @@ func CreateResources(conversionComponents []*types.ConversionComponents) (*types
...
+			continue
+		}
 		if conversionComponent.Weblogicworkload != nil {
 
-			virtualService, destinationRule, authzPolicy, err = createChildResources(conversionComponent, gateway, allHostsForTrait)
-			virtualServices = append(virtualServices, virtualService...)
-			destinationRules = append(destinationRules, destinationRule...)
-			authzPolicies = append(authzPolicies, authzPolicy...)
+			virtualService, destinationRule, authzPolicy, err := createChildResources(conversionComponent, gateway, allHostsForTrait)
+			kubeResources = append(kubeResources, virtualService)
```
anders-swanson, line 48: Please keep the append calls the way they were. This is appending a slice to a slice, such that the appended slice becomes an element of the outer slice. We want the _elements_ of the slice to be appended.

tools/oam-converter/pkg/controllers/metricstrait_utils.go
AnshulDalua: I had to copy this code because some functions I needed to modify to work with the converter. For example in createJobOrServiceMonitorName I changed the inputs so I can get the app name and component name which we already have while in application operator this function accesses this by doing app, found := trait.Labels[oam.LabelAppName]. For the functions that are copied verbatim from app operator, should I call those from app operator instead?

davisgomes: It sounds like you could write a wrapper function for the createJobOrServiceMonitorName that collects the app and component name to then call the underlying function in the app operator. This isn't always feasible but if you can find a way to tweak the function to support both, it could help.

To share the code, I'd suggest extracting the functions that are shared to an outer level package (i.e. one that does not live in the `application-operator/` or `tools/` directory; something like `/pkg/metricstrait/metricstrait.go`). Then you can import these functions in both the app operator and your oam tool.

tools/oam-converter/pkg/resources/gateway/gateway.go
AnshulDalua: I believe it is like this in app operator

tools/oam-converter/pkg/app/confdata.go
AnshulDalua: I think Vrushali needed It to be in a list of virtual services as she said multiple objects may be created

tools/oam-converter/pkg/app/confdata.go
AnshulDalua: sounds good, thank you!

tools/oam-converter/pkg/resources/gateway/gateway.go
anders-swanson: Yes, but is it applicable to OAM Converter? I don't think so, since app operator adds these labels for OAM

tools/oam-converter/pkg/app/confdata.go
anders-swanson: This is depending on how you append to the slice. If you append a slice to slice, then yes. If you append the elements of a slice to a slice, then no.

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,109 @@
...
+			return &serviceMonitor, err
+		}
+	}
+	if conversionComponent.Weblogicworkload != nil {
+		workload = conversionComponent.Weblogicworkload
+		traitDefaults, err = utils.NewTraitDefaultsForWLSDomainWorkload(workload)
+		if err != nil {
+			return &serviceMonitor, err
+		}
+	}
```
anders-swanson, line 59: Are there any trait defaults for generic workloads?

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,109 @@
...
+	// Fetch the secret by name if it is provided in either the trait or the trait defaults.
+	secret, err := operator.FetchSourceCredentialsSecretIfRequired(context.TODO(), trait, traitDefaults, workload, cli)
+	if err != nil {
+		return &serviceMonitor, err
+	}
+
+	//fetch if trait uses Istio
+	useHTTPS := types.InputArgs.IstioEnabled
+
+	//useHTTPS, err := operator.UseHTTPSForScrapeTarget(context.TODO(), cli, trait)
```
anders-swanson, line 70: remove commented code

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -80,6 +83,27 @@ func ExtractTrait(appMaps []map[string]interface{}) ([]*types.ConversionComponen
 							IngressTrait:  ingressTrait,
 						})
 					}
+					if traitKind == consts.MetricsTrait {
+						metricsTrait := &vzapi.MetricsTrait{}
+						traitJSON, err := json.Marshal(traitSpec)
+
+						if err != nil {
+							fmt.Printf("Failed to marshal trait: %v", err)
```
anders-swanson, line 91: return err

tools/oam-converter/pkg/traits/extracttrait.go
```diff
@@ -80,6 +83,27 @@ func ExtractTrait(appMaps []map[string]interface{}) ([]*types.ConversionComponen
...
+						}
+
+						err = json.Unmarshal(traitJSON, metricsTrait)
+
+						if err != nil {
+							fmt.Printf("Failed to unmarshal trait: %v", err)
```
anders-swanson, line 97: return err

application-operator/controllers/metricsbinding/metricsbinding_update.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
anders-swanson, line 1: Copyright (c) 2022, 2023,

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -74,202 +61,149 @@ func ConfData() error {
...
+}
 
-		_, err = f.WriteString(string(gatewayYaml))
+func convertAnytoUnstructured(anyResources []any) []unstructured.Unstructured {
+	var unstructuredResources []unstructured.Unstructured
+	for _, obj := range anyResources {
+		resource, err := ToUnstructured(obj)
 		if err != nil {
-			return fmt.Errorf("failed to write YAML to file: %w", err)
+			print("Null obj")
```
anders-swanson, line 135: should return an error here

tools/oam-converter/pkg/app/confdata.go
```diff
@@ -74,202 +61,149 @@ func ConfData() error {
...
+func ToUnstructured(o any) ([]unstructured.Unstructured, error) {
+	j, err := json.Marshal(o)
+	if err != nil {
+		return nil, err
+	}
+	obj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, j)
+	if err != nil {
+		return nil, err
+	}
+	if u, ok := obj.(*unstructured.Unstructured); ok {
```
anders-swanson, line 179: Is the slice wrapping behavior necessary here - I don't think the any should ever be a slice?

tools/oam-converter/pkg/app/confdata.go
anders-swanson: This comment still needs to be addressed

tools/oam-converter/pkg/resources/createresources.go
```diff
@@ -37,39 +32,48 @@ func CreateResources(cli client.Client, conversionComponents []*types.Conversion
...
 	}
-	//Appending it to Kube Resources to print the output
-	outputResources.DestinationRules = destinationRules
-	outputResources.AuthPolicies = authzPolicies
-	outputResources.VirtualServices = virtualServices
-	outputResources.Gateway = listGateway
-	return &outputResources, nil
+	return kubeResources, nil
+}
+func addResourcesToKubeResources(virtualService []*vsapi.VirtualService, destinationRule []*istioclient.DestinationRule, authzPolicy []*clisecurity.AuthorizationPolicy) {
```
anders-swanson, line 67: why are we not using the `...` operator on the slice like why were before?

tools/oam-converter/pkg/resources/createresources_test.go
```diff
@@ -109,8 +109,17 @@ func TestCreateResources(t *testing.T) {
 	assert.NotNil(t, kubeResources, "kubeResources is nil")
 
 	// Assert that the kubeResources fields are not empty or nil
-	assert.NotEmpty(t, kubeResources.VirtualServices, "VirtualServices is empty")
-	assert.NotNil(t, kubeResources.Gateway, "Gateway is nil")
+	for _, resource := range kubeResources {
```
anders-swanson, line 112: I'm not sure this is working the way you think: we want to assert that a virtual service and a gateway are present in the array. This code is asserting that if a virtual service and a gateway are in the array, they're NotEmpty.

Also, since the gateway is a map[string]interface{}, we should be asserting that the Kind is gateway.

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor_test.go
```diff
@@ -0,0 +1,159 @@
...
+	tests := []struct {
+		name         string
+		input        types.ConversionComponents
+		istioEnabled bool
+		workload     unstructured.Unstructured
+	}{
+		{
+			name: "empty",
+			input: types.ConversionComponents{
+				AppName:       "test-appconf",
```
anders-swanson, line 30: suggest to create a helper method for generating your test input. Most of the ConvertionComponents structs are the same, excepting the workload

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor_test.go
```diff
@@ -0,0 +1,159 @@
...
+				},
+			},
+			istioEnabled: true,
+		},
+	}
+	// Call the function with the sample inputs
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			serviceMonitor, err := CreateServiceMonitor(&tt.input)
+			assert.Nil(t, err, "Unexpected error returned from CreateServiceMonitor")
```
anders-swanson, line 119: use assert.NoError

tools/oam-converter/pkg/resources/servicemonitor/servicemonitor_test.go
```diff
@@ -0,0 +1,159 @@
...
+	operator "github.com/verrazzano/verrazzano/application-operator/controllers/metricstrait"
+	"github.com/verrazzano/verrazzano/tools/oam-converter/pkg/types"
+	k8smeta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"testing"
+)
+
+func TestCreateServiceMonitor(t *testing.T) {
+	port := 7001
+	types.InputArgs.IstioEnabled = false
```
anders-swanson, line 19: This use of mutable, global state can cause confusion, especially if other tests/parts of the program are using it. Suggest to make this parameterized, or at least use `defer` to reset `types.InputArgs.IstioEnabled` once the test method returns.
In the case of tests, they may be run in random orders or be parallelized, so we cannot guarantee mutable global state is consistent between runs. In general this kind of construct should also be avoided.

tools/oam-converter/pkg/traits/extracttrait.go
anders-swanson: the print can be removed now

anders-swanson (changes requested):
See comments

PR #6432, opened 2023-07-11
VZ-10295 - UpgradeCAPI OCI drivers

theabmitra:
- Add support for OKE Self managed nodes and remove OKE experimental flag
- Add Kubernetes 1.26 support in Addons template and self managed cluster 

https://github.com/oracle/cluster-api-provider-oci/releases


platform-operator/capi/infrastructure-oci/v0.12.0/metadata.yaml
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
```
bobdonat, line 1: Do the 0.9.0 files need to be removed?

theabmitra: no, they will be required for upgrade

mgianatagh (dismissed):

bobdonat (approved):

platform-operator/capi/infrastructure-oci/v0.12.0/metadata.yaml
theabmitra: ok, removed them. it works on an upgrade

PR #6441, opened 2023-07-11
vz 8838 

agv426:
This moves the distribution scanning out of the release stage 2 and into the periodic runs. 

- Scanning is done on release-* branches only by default. master and user branches can do it with an explicit parameter set
- We are still running scanning on the internal agent, but it also has a job and runs fine on the other internal ephemeral automation as well. We may move to use those in the future for this job, that will require additional work to trigger them
- There is no coordination in release stage 2 currently, it will not check if it is complete or copy the artifacts. For now that will need to be covered as a manual release process step. additional work will need to be done to automate waiting and copying (possible to do, just didn't add in this pass)
- To avoid complexity and race conditions, we store the scan results in the per-commit ephemeral store only, we do not copy it into the latest-clean-periodic location afterwards.


George-Aeillo (dismissed):
It looks good to me.

stevshap (approved):

bhatpmk (approved):

stevshap (approved):

PR #6463, opened 2023-07-12
VZ-10309 Add VPO module integration first-pass, controlled by feature gate.

pfmackin:
First pass of VPO module integration.  This code is in the experimental folder and controlled by the moduleIntegration arg.


platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -229,6 +229,9 @@ const (
 
 	// CondUpgradeComplete means the upgrade has completed successfully
 	CondUpgradeComplete ConditionType = "UpgradeComplete"
+
+	// CompStateReconciling is the state when a module is reconciling
+	CompStateReconciling CompStateType = "Reconciling"
```
michael-cico, line 234: not blocking, but looks like this should be with the other `CompStateType` declarations instead of the `ConditionType`s?

michael-cico: Or is this supposed to be a Condition?

platform-operator/internal/operatorinit/run_operator.go
```diff
@@ -208,3 +230,25 @@ func createVPOHelmChartConfigMap(kubeClient kubernetes.Interface, configMap *cor
...
+	for _, comp := range registry.GetComponents() {
+		// TODO
+		// Add function to component SPI to get WatchDescriptors
+		// Pass the descriptors to the InitController so that each component can watch whatever it wants
+		// and react
+		// END TODO
+
+		// init controller
+		if err := module.InitController(mgr, modulehandler.NewModuleHandlerInfo(), moduleapi.ModuleClassType(comp.Name())); err != nil {
+			log.Errorf("Failed to start the Calico controller", err)
```
michael-cico, line 249: Non-blocking, messages says `Calico`

michael-cico (approved):

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
pfmackin: good catch, it is state.  I will fix next pass.

platform-operator/internal/operatorinit/run_operator.go
pfmackin: will fix next pass

PR #6464, opened 2023-07-12
VZ-10035.  Roll rancher pods on update if cacerts is out of sync

michael-cico:
When Verrazzano is updated to change between private CA configurations, the secret ` cattle-system/tls-ca` will be updated with the new private CA, but the Rancher pods will not be rotated to pick it up as the deployment remains unchanged.  This will cause SSO errors logging into Rancher via Keycloak until the pods are rotated to pick up the bundle change.

With this change, immediately after running Helm in the `Install()` and `Upgrade()` methods we
- compare the contents of the `cacerts` `Settings` object used by Rancher with the data in `cattle-system/tls-ca`
- if they are different, and a rolling update of the `cattle-system/rancher` deployment is not already in progress, start a rolling restart of the Rancher deployment to pick up the new value

Other changes:
- Adds an override of the `Upgrade()` func so we can have a chance to restart the Rancher deployment if necessary right after calling Helm
- return the error from invoking `HelmComponent.PostInstall()` directly without logging it to avoid logging retryable errors and clean up the logs
- Cleans up and fixes a few other unit tests

**NOTE** that this does not yet handle the Let's Encrypt staging case, that is being investigated as part of VZ-10166.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher.go
@@ -521,7 +521,7 @@ func createOrUpdateResource(ctx spi.ComponentContext, nsn types.NamespacedName, 
 	return nil
 }
 
-// getSettingValue Returns a Rancher Settings objecgt value
+// getSettingValue Returns a Rancher Settings object value with the leading and trailing whitespace trimmed
 func getSettingValue(c client.Client, settingName string) string {
 	resource := unstructured.Unstructured{}
 	resource.SetGroupVersionKind(common.GVKSetting)
--- a/platform-operator/controllers/verrazzano/component/rancher/rancher.go
+++ b/platform-operator/controllers/verrazzano/component/rancher/rancher.go
@@ -522,19 +522,22 @@ func createOrUpdateResource(ctx spi.ComponentContext, nsn types.NamespacedName, 
 }
 
 // getSettingValue Returns a Rancher Settings object value with the leading and trailing whitespace trimmed
-func getSettingValue(c client.Client, settingName string) string {
+func getSettingValue(c client.Client, settingName string) (string, error) {
 	resource := unstructured.Unstructured{}
 	resource.SetGroupVersionKind(common.GVKSetting)
 	resource.SetName(settingName)
-	err := c.Get(context.Background(), types.NamespacedName{Name: resource.GetName()}, &resource)
-	if err != nil {
-		return ""
+
+	if err := c.Get(context.Background(), types.NamespacedName{Name: resource.GetName()}, &resource); err != nil {
+		if client.IgnoreNotFound(err) != nil {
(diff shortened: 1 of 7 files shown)


mgianatagh (approved):

platform-operator/controllers/verrazzano/component/rancher/rancher.go
```diff
@@ -520,6 +521,25 @@ func createOrUpdateResource(ctx spi.ComponentContext, nsn types.NamespacedName,
 	return nil
 }
 
+// getSettingValue Returns a Rancher Settings object value with the leading and trailing whitespace trimmed
+func getSettingValue(c client.Client, settingName string) (string, error) {
+	resource := unstructured.Unstructured{}
+	resource.SetGroupVersionKind(common.GVKSetting)
```
pfmackin, line 527: would be good if this GVK name has Rancher in it, i had to go look it up.

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
```diff
@@ -828,3 +841,108 @@ func createOrUpdateRancherUser(ctx spi.ComponentContext) error {
...
+	}
+	currentCACerts, found := tlsCASecret.Data[key]
+	if !found {
+		return "", false, ctx.Log().ErrorfThrottledNewErr("Did not find %s key in % secret", key,
+			clipkg.ObjectKeyFromObject(tlsCASecret))
+	}
+	return strings.TrimSpace(string(currentCACerts)), true, nil
+}
+
+func restartRancherDeployment(log vzlog.VerrazzanoLogger, c clipkg.Client) error {
```
pfmackin, line 913: platformRestart.go has restart code that can be used for this.  See RestartComponents
Not sure if you want to change this or not

pfmackin (approved):

platform-operator/controllers/verrazzano/component/rancher/rancher.go
michael-cico: Will do that as a follow-up

platform-operator/controllers/verrazzano/component/rancher/rancher_component.go
michael-cico: I'll look into that as a follow-up.  I have other stuff to do in this area for VZ-10166

fredtibbitts: RestartComponents restarts everything in a namespace so it would need some refactoring so it could be shared. There is one thing though that RestartComponents is doing, it checks if the deployment is paused and if so, it does not attempt to restart the deployment. I'm not sure how important that check is, but I'm wondering if it should be added to this function too (if this function ends up remaining here)?

fredtibbitts (approved):

PR #6472, opened 2023-07-13
VZ-10280: Have VZ analyze process Rancher cluster management resources

mgianatagh:


tools/vz/pkg/analysis/internal/util/cluster/rancher.go
```diff
@@ -0,0 +1,152 @@
...
+	var issueReporter = report.IssueReporter{
+		PendingIssues: make(map[string]report.Issue),
+	}
+
+	return analyzeRancherClusters(log, clusterRoot, &issueReporter)
+}
+
+// analyzeRancherClusters - analyze the status of Rancher clusters
+func analyzeRancherClusters(log *zap.SugaredLogger, clusterRoot string, issueReporter *report.IssueReporter) error {
+	clusterPath := files.FindFileInClusterRoot(clusterRoot, "default/cluster.json")
```
bobdonat, line 56: Do we actually capture this file?

mgianatagh: We do as of this PR

tools/vz/pkg/analysis/internal/util/cluster/rancher.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	issueReporter.Contribute(log, clusterRoot)
+	return nil
+}
+
+// analyzeRancherCluster - analyze a single Rancher cluster and report any issues
+func analyzeRancherCluster(clusterRoot string, cluster rancherCluster, issueReporter *report.IssueReporter) error {
+
+	var messages []string
```
bobdonat, line 92: Which cluster resource is this?  I have 5 cluster.xxx crds on my cluster.  Should the api group be checked to make it is the correct type of cluster resource

tools/vz/pkg/analysis/internal/util/cluster/rancher.go
bobdonat: Missed that.

bobdonat: I noticed that we are capturing this resource in the default namespace when it is a cluster resource.  Same was true for the other CAPI cluster resources.  I was expecting to not see a namespace for the cluster resources.

tools/vz/pkg/analysis/internal/util/cluster/rancher.go
mgianatagh: The content of clusters.json is guaranteed to have "clusters.management.cattle.io", that is what vz capture will do.  Should I change the name of the file to something else?  Like cattleclusters.json?

bobdonat (approved):

PR #6490, opened 2023-07-13
VZ-10325: Update rancher image

theabmitra:
update managed profile from rancher UI


mgianatagh (approved):

theabmitra:
closing since build was not correct

PR #6503, opened 2023-07-14
Gfaeillo/vz 10005

George-Aeillo:
Fix For ArgoCD/Rancher Certificate Refresh Bug
The change implemented in this code is now that when the GetTokenWithFilter function is called, the token selected is the token that is most recent based on its timestamp and the token has to match the clusterID and the userID of the caller.


pkg/rancherutil/rancher_config.go
```diff
@@ -308,12 +309,30 @@ func GetTokenWithFilter(rc *RancherConfig, log vzlog.VerrazzanoLogger, userID, c
...
+
+		if tokenToReturn == nil && len(item.Created) > 0 {
+			tokenToReturn = &(items[i])
+			continue
+		}
+		// Find the latest token based on creation timestamp
+		timeOfCurrentTokenToReturn, _ := time.Parse(time.RFC3339, tokenToReturn.Created)
+		timeOfTokenBeingExamined, _ := time.Parse(time.RFC3339, item.Created)
+
+		if timeOfCurrentTokenToReturn.Unix() < timeOfTokenBeingExamined.Unix() {
```
fredtibbitts, line 327: The time package has an After function that compares two Time instances, it might read better if you used that. Something like `if timeOfTokenBeingExamined.After(timeOfCurrentTokenToReturn) { ...`

pkg/rancherutil/bodyfortokentest.json
fredtibbitts, line 1: I think in general we put test data files in a `testdata` subdirectory, so `pkg/rancherutil/testdata/bodyfortokentest.json`

fredtibbitts (commented):
A couple of minor comments otherwise looks good to me.

pkg/rancherutil/rancher_config.go
George-Aeillo: Thank you! I have addressed this issue.

pkg/rancherutil/bodyfortokentest.json
George-Aeillo: Thank you! I have addressed this issue.

pkg/rancherutil/testdata/bodyfortokentest.json
```diff
@@ -0,0 +1,155 @@
...
+            "links": {
+                "remove": "…/v3/tokens/tokenid1",
+                "self": "…/v3/tokens/tokenid1",
+                "update": "…/v3/tokens/tokenid1"
+            },
+            "name": "tokenid1",
+            "ttl": 14400000,
+            "type": "token",
+            "userId": "usertestone",
+            "userPrincipal": "map[displayName:ArgoCD Cluster Registrar loginName:vz-argoCD-reg me:true metadata:map[creationTimestamp:<nil> name:local://usertest] principalType:user provider:local]",
```
mbenmclean, line 87: vz-argoCD-reg should probably be changed if that is the actual name we use

pkg/rancherutil/rancher_config.go
```diff
@@ -308,12 +309,30 @@ func GetTokenWithFilter(rc *RancherConfig, log vzlog.VerrazzanoLogger, userID, c
 	if err != nil {
 		return "", "", log.ErrorfNewErr("Failed to parse response: %v", err)
 	}
-	for _, item := range items {
-		if item.ClusterID == clusterID {
-			return item.Created, item.ExpiresAt, nil
+	var tokenToReturn *TokenGetResponse
+	tokenToReturn = nil
```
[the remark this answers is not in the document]
George-Aeillo: I have resolved this issue.

pkg/rancherutil/testdata/bodyfortokentest.json
George-Aeillo: Thanks! I have resolved this issue.

pkg/rancherutil/rancher_config.go
```diff
@@ -308,12 +309,30 @@ func GetTokenWithFilter(rc *RancherConfig, log vzlog.VerrazzanoLogger, userID, c
...
+		// Find the latest token based on creation timestamp
+		timeOfCurrentTokenToReturn, _ := time.Parse(time.RFC3339, tokenToReturn.Created)
+		timeOfTokenBeingExamined, _ := time.Parse(time.RFC3339, item.Created)
+
+		if timeOfTokenBeingExamined.After(timeOfCurrentTokenToReturn) {
+			tokenToReturn = &(items[i])
 		}
+
+	}
+	if tokenToReturn == nil {
```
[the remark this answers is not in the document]
fredtibbitts: This function is called by `mutateArgoCDClusterSecret` and if we return an error here, it will never try to create a new token, so that's bad. Maybe if `tokenToReturn` is nil, return nil token and no error and then `mutateArgoCDClusterSecret` can check for a nil token and create a new one? Looks like you can just set `createNewToken` to `true` and it will fall through and create a new token.

mbenmclean (approved):

pkg/rancherutil/rancher_config.go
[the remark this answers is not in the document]
George-Aeillo: I have tried to address this comment, is this what you were looking for?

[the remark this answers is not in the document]
mbenmclean: @fredtibbitts ^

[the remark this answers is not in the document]
fredtibbitts: We connected offline. I asked for a unit test for the "no matching tokens" case and George is adding that now.

cluster-operator/controllers/vmc/argocd_test.go
```diff
@@ -276,6 +397,7 @@ func generateClientObject(objs ...runtime.Object) client.WithWatch {
 				},
 			},
 		},
+		user.DeepCopyObject(),
```
fredtibbitts, line 400: Why does this need to be deep copied?

George-Aeillo: This needs to be deep copied to convert it into a runtime.Object.

PR #6509, opened 2023-07-14
VZ-10226: Upgrade Cluster API components only if the images are outdated

Abhijithang:
This PR fixes the issue where Cluster API components are upgraded only if the images do not match.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
+++ b/platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
@@ -262,7 +262,6 @@ func (c *PodMatcherClusterAPI) matchAndPrepareUpgradeOptions(ctx spi.ComponentCo
 		return applyUpgradeOptions, err
 	}
 
-	ctx.Log().Info("LIST OF PODS", podList.String())
 	const formatString = "%s/%s:%s"
 	for _, pod := range podList.Items {
 		for _, co := range pod.Spec.Containers {
@@ -280,10 +279,17 @@ func (c *PodMatcherClusterAPI) matchAndPrepareUpgradeOptions(ctx spi.ComponentCo
 			}
 			ctx.Log().Info(clusterAPIOCIControllerImage, co.Image, c.infrastructureProvider)
 			if isImageOutOfDate(ctx.Log(), clusterAPIOCIControllerImage, co.Image, c.infrastructureProvider) == OutOfDate {
-				applyUpgradeOptions.InfrastructureProviders = append(applyUpgradeOptions.ControlPlaneProviders, fmt.Sprintf(formatString, ComponentNamespace, ociProviderName, overrides.GetOCIVersion()))
+				applyUpgradeOptions.InfrastructureProviders = append(applyUpgradeOptions.InfrastructureProviders, fmt.Sprintf(formatString, ComponentNamespace, ociProviderName, overrides.GetOCIVersion()))
 			}
 		}
 	}
 	ctx.Log().Info("upgrade options", applyUpgradeOptions.CoreProvider, "=>", applyUpgradeOptions.BootstrapProviders, "=>", applyUpgradeOptions.InfrastructureProviders, "==>", applyUpgradeOptions.ControlPlaneProviders)
 	return applyUpgradeOptions, nil
+}
+
+func isUpgradeOptionsEmpty(upgradeOptions clusterapi.ApplyUpgradeOptions) bool {
+	return len(upgradeOptions.CoreProvider) == 0 ||
+		len(upgradeOptions.BootstrapProviders) == 0 ||
+		len(upgradeOptions.ControlPlaneProviders) == 0 ||
+		len(upgradeOptions.InfrastructureProviders) == 0
 }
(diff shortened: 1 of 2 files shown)


mgianatagh:
I am wondering if there may be a simpler approach.  Could the code instead populate the input yaml to clusterctl and compare it to the existing yaml on the VPO image.  If there are differences then assume there are changes.

Abhijithang:
> I am wondering if there may be a simpler approach. Could the code instead populate the input yaml to clusterctl and compare it to the existing yaml on the VPO image. If there are differences then assume there are changes.

Not sure what other changes yaml contains. We are looking at just a image change so I am only verifying those changes.

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
```diff
@@ -168,3 +199,93 @@ func applyTemplate(templateContent string, params interface{}) (bytes.Buffer, er
 	// Return the result containing the processed template
 	return buf, nil
 }
+
+func getImage(subComponent, imageName string) (string, error) {
+	bomFile, err := bom.NewBom(config.GetDefaultBOMFilePath())
```
bobdonat, line 204: I think this needs to take in account the cluster API overrides that can be specified in the vz CR

Abhijithang: I discussed with Michael G about the source of truth for VZ CR image Overrides, and it is understood to be a BOM file.

Abhijithang: When we specific overrides in the VZ CR, it retrieves the image details from the BOM file right? 
`apiVersion: install.verrazzano.io/v1beta1
kind: Verrazzano
metadata:
  name: example-verrazzano
spec:
  profile: dev
  components:
   clusterAPI:
       enabled: true
       overrides:
       - values:
           defaultProviders:
               ocneBootstrap:
                   version: 1.6.1
               ocneControlPlane:
                   version: 1.6.1
               oci:
                   version: v0.9.0`

bobdonat: My understanding is that you can override the BOM values with clusterAPI image overrides in the VZ CR.  @mgianatagh is more knowledgeable in this area.

Abhijithang: Sure, will wait for @mgianatagh's advise on the BOM image override concern. Thanks.

mgianatagh: The BOM functions will not take into account overrides made in the CR.  There is logic in the CAPI component which knows how to handle overrides made to the CAPI component in the CR.

See platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_overrides.go

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
```diff
@@ -168,3 +199,93 @@ func applyTemplate(templateContent string, params interface{}) (bytes.Buffer, er
...
+	podList := &v1.PodList{}
+	if err := ctx.Client().List(context.TODO(), podList, &client.ListOptions{Namespace: ComponentNamespace}); err != nil {
+		return applyUpgradeOptions, err
+	}
+
+	const formatString = "%s/%s:%s"
+	for _, pod := range podList.Items {
+		for _, co := range pod.Spec.Containers {
+			if isImageOutOfDate(ctx.Log(), clusterAPIControllerImage, co.Image, c.coreProvider) == OutOfDate {
+				applyUpgradeOptions.CoreProvider = fmt.Sprintf(formatString, ComponentNamespace, clusterAPIProviderName, overrides.GetClusterAPIVersion())
```
mgianatagh, line 269: This logic only appears to be taking into account the ClusterAPIVersion from the overrides context.  The actual image name that results must also take into account registry, repository and URL overrides.

I feel there should be a set of test cases with clusterAPI overrides that confirm this logic is working as expected.

Abhijithang: The logic uses ClusterAPIVersion from overrides context only for creating an upgrade option for capi client. The full images paths from the params co.Image, c.coreProvider path are used to compare. But I see I am initializing the c.coreProvider image from the BOM file. As you mentioned, the fix is to initialize it from the overrides. 

However, I agree with the test cases. Sorry about that. Will add test cases and fix the bom initialization.

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_test.go
```diff
@@ -69,3 +73,102 @@ func TestCreateClusterctlYaml(t *testing.T) {
...
+		}).Build()
+
+	compContext := spi.NewFakeContext(fakeClient, vz, nil, false)
+	config.SetDefaultBomFilePath(testBomFilePath)
+	config.TestHelmConfigDir = TestHelmConfigDir
+	overrides, _ := createOverrides(compContext)
+	overridesContext := newOverridesContext(overrides)
+	podMatcher := &PodMatcherClusterAPI{}
+	applyUpgradeOptions, err := podMatcher.matchAndPrepareUpgradeOptions(compContext, overridesContext)
+	assert.True(t, isUpgradeOptionsNotEmpty(applyUpgradeOptions))
```
mgianatagh, line 172: could there be additional asserts here to verify the upgrade options only contain the images you expected to change?

Abhijithang: Sure, will do the change. That would be helpful.

Abhijithang: Done. Added more assertions and refactored the tests with additional comments.

mgianatagh (approved):

mgianatagh (approved):

PR #6521, opened 2023-07-17
VZ-10265 Check enabled istio deployments for readiness check

davisgomes:
**Changes**
- The is ready check uses label selectors to make sure the deployments for Istio are ready

**Testing**
- Deploy a cluster with egress gateway disabled and verify the readiness check works
- Deploy a cluster with all enabled and verify the readiness check works


platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -443,8 +455,70 @@ func (i istioComponent) IsReady(context spi.ComponentContext) bool {
...
+	})
+	err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: istioOperatorName, Namespace: IstioNamespace}, &istioOp)
+	if kerrs.IsNotFound(err) || meta.IsNoMatchError(err) {
+		return istioDeployments, nil
+	}
+	if err != nil {
+		return istioDeploy, ctx.Log().ErrorfNewErr("Failed to get the Istio Operator object from the cluster, %v", err)
+	}
+
+	if parseIstioOperatorJSONAsBool(istioOp.Object, "spec", "components", "egressGateways", 0, "enabled") {
```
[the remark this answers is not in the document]
davisgomes: Migrating to label selectors to help address some of these edge case concerns and to reduce complexity.

desagar (dismissed):

platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -411,7 +396,7 @@ func (i istioComponent) Upgrade(context spi.ComponentContext) error {
 }
 
 func (i istioComponent) IsAvailable(ctx spi.ComponentContext) (reason string, available vzapi.ComponentAvailability) {
-	return (&ready.AvailabilityObjects{DeploymentNames: istioDeployments}).IsAvailable(ctx.Log(), ctx.Client())
+	return (&ready.AvailabilityObjects{DeploymentSelectors: []clipkg.ListOption{&istioLabelSelector}}).IsAvailable(ctx.Log(), ctx.Client())
```
pfmackin, line 399: what happens if this is called before any deployments are created, meaning the selector finds nothing.

davisgomes: The availability check handles an empty selector list like this:

platform-operator/controllers/verrazzano/component/istio/istio_component.go
```diff
@@ -443,8 +428,9 @@ func (i istioComponent) IsReady(context spi.ComponentContext) bool {
 	return true
 }
 
-func areIstioDeploymentsReady(context spi.ComponentContext, prefix string) bool {
-	return ready.DeploymentsAreReady(context.Log(), context.Client(), istioDeployments, 1, prefix)
+// areIstioDeploymentsReady verifies that the enabled deployments in the Istio Operator are ready
```
pfmackin, line 431: This should be IstioOperator CR, not Istio Operator

platform-operator/controllers/verrazzano/component/istio/istio_component.go
pfmackin: So this assumes that all deployments exist before any of them are ready.  That is probably the case but it is timing sensitive.

michael-cico: I would think that by the time `helm` finishes the `deployment` resources should exist right?

platform-operator/controllers/verrazzano/component/istio/istio_component.go
davisgomes: Thanks fixed.

platform-operator/controllers/verrazzano/component/istio/istio_component.go
davisgomes: Istioctl controls the Istio lifecycle, but I think your assumption is true at least a great amount of the time.

anders-swanson (approved):

pfmackin (approved):

platform-operator/controllers/verrazzano/component/istio/istio_component.go
michael-cico: yeah... I forgot about Istio.  But, I think `istioctl` blocks until it finishes at least creating the resources (although we run it in the background), so by the time we get to the ready checks it should be there.

But yes, it's still a potential race condition.

PR #6529, opened 2023-07-18
VZ-8082: Check whether Keycloak pod is ready before connecting to the running container

bhatpmk:
This PR adds a check to find out whether Keycloak pod is ready, before attempting to get inside the running container. With this, the probability of getting a failure while exec into the pod, the only case I can consider is when the password in the keycloak-http secret is different that the keycloak admin password in the underlying database.

verrazzano-cluster-operator/templates/clusterrole.yaml is updated to add cluster role list pods with this change, as function LoginKeycloak is called to de-register a managed cluster.


platform-operator/controllers/verrazzano/component/keycloak/keycloak.go
```diff
@@ -881,6 +870,12 @@ func configureKeycloakRealms(ctx spi.ComponentContext) error {
 		// to resolve the condition.
 		err = LoginKeycloak(ctx, cfg, cli)
 		if err != nil {
+			pod := keycloakPod()
```
lake-of-dreams, line 873: don't think we need to do it again here since already done in LoginKeycloak

platform-operator/controllers/verrazzano/reconcile/install_test.go
```diff
@@ -700,3 +701,21 @@ func TestCheckGenerationUpdated(t *testing.T) {
 		})
 	}
 }
+
+// getKeycloakPod constructs and returns Keycloak pod
+func getKeycloakPod() *corev1.Pod {
```
lake-of-dreams, line 706: can we move it somewhere so that both keycloak_test.go and install_test.go can use this function?

platform-operator/controllers/verrazzano/component/keycloak/keycloak_test.go
```diff
@@ -877,6 +877,21 @@ func TestLoginKeycloak(t *testing.T) {
 	cfg, restclient, _ := fakeRESTConfig()
 	k8sutil.NewPodExecutor = k8sutilfake.NewPodExecutor
 
+	keycloakPod := &v1.Pod{
```
lake-of-dreams, line 880: could it use same getKeycloakPod() defined below?

pfmackin (approved):

platform-operator/controllers/verrazzano/reconcile/install_test.go
bhatpmk: Done

lake-of-dreams (approved):
lgtm

pfmackin (approved):

PR #6531, opened 2023-07-18
VZ8548 2

saketmht:


platform-operator/controllers/verrazzano/component/opensearchoperator/legacyOS.go
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
anders-swanson, line 1: filename should be legacy_os.go

platform-operator/controllers/verrazzano/component/opensearchoperator/overrides.go
```diff
@@ -0,0 +1,457 @@
...
+	AdditionalConfig          map[string]string                 `json:"additionalConfig,omitempty"`
+	Labels                    map[string]string                 `json:"labels,omitempty"`
+	Annotations               map[string]string                 `json:"annotations,omitempty"`
+	Env                       []corev1.EnvVar                   `json:"env,omitempty"`
+	PriorityClassName         string                            `json:"priorityClassName,omitempty"`
+}
+
+// PersistenceConfig defines options for data persistence
+type PersistenceConfig struct {
+	PersistenceSource `json:","`
```
adiforluls, line 57: `json:",inline"`?

platform-operator/controllers/verrazzano/component/opensearchoperator/overrides.go
```diff
@@ -0,0 +1,457 @@
...
+
+type PVCSource struct {
+	StorageClassName string                              `json:"storageClass,omitempty"`
+	AccessModes      []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty"`
+}
+
+// GetOverrides gets the install overrides
+// Since nodePools are a list and not a map they are replaced when user overrides via the CR
+// So merge the nodePools and append the effective node pool overrides at the top so, it has the highest precedence
+func GetOverrides(object runtime.Object) interface{} {
```
adiforluls, line 74: I absolutely hate list based overrides because they can completely overhaul defaults. But as a user, if I understand the behaviour of list based overrides, I'd want or expect my node pool to overhaul the default node pool. So this kinda goes against the default behaviour of helm. 

Not saying we shouldn't do this but want other people's opinion on whether it makes sense to do this kind of merging. If this was already agreed upon by others then this is good.

PR #6544, opened 2023-07-18
VZ-10127 - Adds pre-uninstall to VMO component to wait for cleanup of VMI pods/ingresses, and to external-dns component to wait for external-dns annotated ingresses and services to be cleaned up, so that DNS records can be deleted.

desagar:
**Note to Reviewers**
There are minor refactoring changes, and making some constants public etc, but the main changes are in the following files:

vmo_component.go
externaldns.go

Some smaller relevant changes are in registry.go and kiali_component.go.

**Description:**
Adds pre-uninstall to VMO component to wait for cleanup of VMI pods/ingresses.
Adds pre-uninstall to external-dns component to wait for external-dns annotated ingresses and services to be cleaned up, so that DNS records can be deleted before external-dns is uninstalled.
Modifies the Kiali component to delete the ingress in PostUninstall, which it created in PostInstall.

I’ve manually verified numerous times that these changes result in the clean up of DNS records. However, this change does not absolutely confirm that the records are deleted, so there is a small possibility that it may still leave a bit of cruft behind (e.g. if external-dns has network trouble when deleting the records and has to retry over a period of time, before which we might uninstall external-dns).

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/vmi.go
+++ b/platform-operator/controllers/verrazzano/component/common/vmi.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 
 	vmov1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
+	vmoconst "github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
 	globalconst "github.com/verrazzano/verrazzano/pkg/constants"
 	ctrlerrors "github.com/verrazzano/verrazzano/pkg/controller/errors"
 	"github.com/verrazzano/verrazzano/pkg/k8s/ready"
@@ -27,6 +28,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -34,11 +36,13 @@ import (
 )
 
 const (
-	system                = "system"
-	vmoComponentName      = "verrazzano-monitoring-operator"
-	vmoComponentNamespace = constants.VerrazzanoSystemNamespace
+	VMIName               = "system"
+	VMOComponentName      = "verrazzano-monitoring-operator"
+	VMOComponentNamespace = constants.VerrazzanoSystemNamespace
 	defaultStorageSize    = "50Gi"
 )
+
+var vmoLabelSelector = labels.SelectorFromSet(labels.Set{vmoconst.VMOLabel: VMIName})
 
 // ResourceRequestValues defines the storage information that will be passed to VMI instance
 type ResourceRequestValues struct {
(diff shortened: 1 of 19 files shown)


platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -435,16 +453,72 @@ func ReassociateVMOResources(ctx spi.ComponentContext) error {
...
 		{Obj: &rbacv1.ClusterRoleBinding{}, NamespacedName: types.NamespacedName{Name: "verrazzano-monitoring-operator-cluster-role-binding"}},
 		{Obj: &rbacv1.ClusterRoleBinding{}, NamespacedName: types.NamespacedName{Name: "verrazzano-monitoring-operator-cluster-role-default-binding"}},
 		{Obj: &rbacv1.ClusterRoleBinding{}, NamespacedName: types.NamespacedName{Name: "verrazzano-monitoring-operator-get-nodes"}},
-		{Obj: &netv1.NetworkPolicy{}, NamespacedName: types.NamespacedName{Name: vmoComponentName, Namespace: vmoComponentNamespace}},
+		{Obj: &netv1.NetworkPolicy{}, NamespacedName: types.NamespacedName{Name: VMOComponentName, Namespace: VMOComponentNamespace}},
+	}
+}
+
+// GetVMOManagedDeployments returns the list of deployments managed by VMO
+func GetVMOManagedDeployments(ctx spi.ComponentContext) (*appsv1.DeploymentList, error) {
```
michael-cico, line 471: we probably need some minimal unit tests to for code coverage, unless these are called from existing unit tests indirectly.

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
```diff
@@ -95,6 +103,64 @@ func ResolveExernalDNSNamespace() string {
...
 }
 
+// preUninstall checks to make sure that all ingresses/services in the Verrazzano namespaces which
+// have the external DNS target containing the word verrazzano have already been uninstalled, and
+// also that the VMO has been uninstalled.
+// If not, they may require external-dns to clean up the corresponding
+// DNS records, so we cannot start external-dns uninstall yet.
+func preUninstall(log vzlog.VerrazzanoLogger, cli client.Client) error {
+	log.Progressf("Checking for leftover ingresses in Verrazzano component namespaces before uninstalling %s", ComponentName)
+	namespaceList := []string{constants.VerrazzanoSystemNamespace, constants.KeycloakNamespace, vzconst.RancherSystemNamespace, vzconst.VerrazzanoMonitoringNamespace}
```
michael-cico, line 113: Should we look for VZ-owned namespaces using a label?  Like `verrazzano.io/namespace`?  I think that's on all the namespaces we own.

It's possible it may be too broad of a net, but it also might be more resilient to new namespaces/ingresses.

platform-operator/controllers/verrazzano/component/common/vmi.go
desagar: these should be exercised from the vmo unit test cases I added for PreUninstall.

platform-operator/controllers/verrazzano/component/registry/registry.go
```diff
@@ -123,6 +123,20 @@ func GetComponents() []spi.Component {
 	return getComponentsFn()
 }
 
+func GetComponentsForUninstall() []spi.Component {
```
michael-cico, line 126: Does the order matter, if the pre/post uninstalls are looking for the right things?

michael-cico: Main reason I'm asking is that as we break things out to be more independent, we want to have less ordered depedencies between things.

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
desagar: yeah I debated with myself about that. Currently these are the only locations we have ingresses tied to external-dns, but you're right that it could change - looking for verrazzano.io/namespace might not be a bad idea. I'll change that and push.

platform-operator/controllers/verrazzano/component/registry/registry.go
desagar: Not as much as it would otherwise. I did this for 2 reasons:
1. To avoid the amount of churn external-dns goes through waiting for pre-uninstall conditions.
2. Because this approach is not completely guaranteed to work, it just felt better putting this _after_ the VMO uninstall.

platform-operator/controllers/verrazzano/component/common/vmi.go
desagar: added coverage for these individually.

platform-operator/controllers/verrazzano/component/externaldns/external_dns.go
desagar: implemented.

platform-operator/controllers/verrazzano/component/registry/registry.go
michael-cico: ok, understood.  I think it's another kind of "global" thing to consider during a full uninstall as we move to modules.

I think we'll have similar things to consider as we move in that direction.

pfmackin: Uninstall is in parallel so the order of this list should not matter

michael-cico (approved):

PR #6601, opened 2023-07-24
Gfaeillo/update bom1.6.3

George-Aeillo:
Update image versioning


stevshap (commented):
Should opensearch also get bumped to 1.6.3?

prasad-shirodkar:
Should this  https://github.com/verrazzano/verrazzano/blob/release-1.6/platform-operator/helm_config/charts/verrazzano-cluster-agent/Chart.yaml#L6 be bumped to 1.6.3?

PR #6622, opened 2023-07-26
VZ-10373 Add e2e tests for the Thanos Ruler

davisgomes:


tests/e2e/verify-install/thanos/thanos_test.go
```diff
@@ -64,7 +65,11 @@ var beforeSuite = t.BeforeSuiteFunc(func() {
 		// check if compactor component is enabled
 		isCompactorEnabled, err = isThanosComponentEnabledInOverrides(inClusterVZ.Spec.Components.Thanos.InstallOverrides.ValueOverrides, "compactor")
 		if err != nil {
-			AbortSuite(fmt.Sprintf("Failed to process VZ CR Thanos overrides: %v", err))
+			AbortSuite(fmt.Sprintf("Failed to process VZ CR Thanos overrides for the Store Gateway: %v", err))
```
desagar, line 68: typo - this should be Compactor. Storegateway is the one above on line 63.

tests/e2e/config/scripts/configure_thanos_storegateway_install.sh
```diff
@@ -11,6 +11,11 @@ if [[ $THANOS_ENABLED != "true" ]]; then
   exit
```
desagar, line 11: minor nit, but we should probably rename this file to something like configure_thanos_install_storage.sh or something like that.

desagar (approved):

PR #6632, opened 2023-07-26
VZ-10309 Integrate VPO with modules - core changes

pfmackin:
First pass at full lifecycle implementation of VPO module integration

1. This is still controlled by a feature flag (default off)
2. Install, Update, Uninstall, Upgrade tested with Coherence
3. Only Upgrade from Module to Module works. 
4. Need module-operator fix for non-module to module upgrade (change coming in follow-up PR)
5. Only Helm overrides is copied to Module - structured config is a follow up
6. Original Verrazzano controller runs concurrently with new Verrazzano controller
7. There is some synchronization done between controllers, enough for Coherence to work
8. Module handler shim updates both Verrazzano CR component status and Module status


platform-operator/constants/constants.go
```diff
@@ -271,13 +274,19 @@ const ThanosInternalUserName = "verrazzano-thanos-internal"
 const VerrazzanoPlatformOperatorHelmName = "verrazzano-platform-operator"
 
 // VerrazzanoCRNameAnnotation is the annotation for the verrazzano CR name
-const VerrazzanoCRNameAnnotation = "verrazzano.io/vz-cr-name"
+const VerrazzanoCRNameAnnotation = "module.verrazzano.io/vz-cr-name"
 
 // VerrazzanoCRNamespaceAnnotation is the annotation for the verrazzano CR namespace
-const VerrazzanoCRNamespaceAnnotation = "verrazzano.io/vz-cr-namespace"
+const VerrazzanoCRNamespaceAnnotation = "module.verrazzano.io/vz-cr-namespace"
```
michael-cico, line 280: This stuff was all inroduced in the firts pass right?  These are just changes to new annotations?

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -96,6 +96,11 @@ func (c clusterAPIComponent) ShouldInstallBeforeUpgrade() bool {
 	return false
 }
 
+// ShouldUseModule returns true if component is implemented using a Module
+func (c clusterAPIComponent) ShouldUseModule() bool {
```
michael-cico, line 100: Nit, maybe use `IsModule` or `UsesModuleInstall`.  But not a hill I'm going to die on.

platform-operator/controllers/verrazzano/reconcile/upgrade_component.go
```diff
@@ -53,6 +53,14 @@ func (r *Reconciler) upgradeComponents(log vzlog.VerrazzanoLogger, cr *installv1
 	// Loop through all of the Verrazzano components and upgrade each one.
 	// Don't move to the next component until the current one has been succcessfully upgraded
 	for _, comp := range registry.GetComponents() {
+		if comp.ShouldUseModule() {
+			// Ignore if this component is being handled by a Module
+			continue
+		}
+		if !comp.IsEnabled(spiCtx.EffectiveCR()) {
```
michael-cico, line 60: Is this a bug in general?  Upgrading a component that isn't enabled?

pfmackin: we skip it if it is disabled.

platform-operator/constants/constants.go
pfmackin: yes, we will need an API review when we revisit the module api

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
pfmackin: it is more that install, it is lifecycle.   I can change this in the future if we decide on a better name.

platform-operator/controllers/verrazzano/reconcile/upgrade_component.go
pfmackin: To answer your question, yes i think it was a latent bug.  We should never upgrade a component that is disable, unless you can think of a reason why.

michael-cico (approved):

PR #6638, opened 2023-07-27
VZ-10143 Add the Prometheus rules to the Thanos Ruler

davisgomes:
**Changes**
- Add the prometheus config reloader to the Thanos Ruler
- Mount the Prometheus Rule configmap
- Add e2e test to verify that the rules are properly loaded into Thanos Ruler

**Testing**
- Install by minimally enabling the Thanos ruler
- Verify the rules are populated in the Ruler UI.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/thanos/thanos.go
+++ b/platform-operator/controllers/verrazzano/component/thanos/thanos.go
@@ -33,6 +33,8 @@ const (
 	// Thanos Ruler ingress constants
 	rulerHostName        = "thanos-ruler"
 	rulerCertificateName = "system-tls-thanos-ruler"
+
+	configReloaderSubcomponentName = "prometheus-config-reloader"
 )
 
 // GetOverrides gets the install overrides for the Thanos component
@@ -60,11 +62,16 @@ func AppendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
 
 	image, err := bomFile.BuildImageOverrides(ComponentName)
 	if err != nil {
-		return kvs, ctx.Log().ErrorfNewErr("Failed to build Thanos image overrides from the Verrazzano BOM: %d", err)
+		return kvs, ctx.Log().ErrorfNewErr("Failed to build Thanos image overrides from the Verrazzano BOM: %v", err)
 	}
 	kvs = append(kvs, image...)
 
 	kvs = appendVerrazzanoOverrides(ctx, kvs)
+
+	kvs, err = appendReloaderSidecarOverrides(ctx, kvs, bomFile)
+	if err != nil {
+		return kvs, ctx.Log().ErrorfNewErr("Failed to build Thanos sidecar overrides from the Verrazzano BOM: %v", err)
+	}
 
 	return appendIngressOverrides(ctx, kvs)
 }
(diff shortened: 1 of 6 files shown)


platform-operator/controllers/verrazzano/component/thanos/thanos.go
```diff
@@ -92,6 +99,62 @@ func preInstallUpgrade(ctx spi.ComponentContext) error {
...
+	}
+
+	image := subcomponent.Images[0]
+	imageString := constructFullImageString(bomFile, *subcomponent, image)
+	sidecarPrefix := "ruler.sidecars[0]"
+	kvs = append(kvs, []bom.KeyValue{
+		{Key: fmt.Sprintf("%s.image", sidecarPrefix), Value: imageString},
+		{Key: fmt.Sprintf("%s.name", sidecarPrefix), Value: configReloaderSubcomponentName},
+		{Key: fmt.Sprintf("%s.args[0]", sidecarPrefix), Value: "--listen-address=:8080"},
+		{Key: fmt.Sprintf("%s.args[1]", sidecarPrefix), Value: "--reload-url=http://127.0.0.1:10902/-/reload"},
```
desagar, line 123: what is 10902?

platform-operator/controllers/verrazzano/component/thanos/thanos.go
```diff
@@ -92,6 +99,62 @@ func preInstallUpgrade(ctx spi.ComponentContext) error {
...
+		// Security setup for the sidecar
+		{Key: fmt.Sprintf("%s.securityContext.privileged", sidecarPrefix), Value: "false"},
+		{Key: fmt.Sprintf("%s.securityContext.capabilities.drop[0]", sidecarPrefix), Value: "ALL"},
+	}...)
+
+	return kvs, nil
+}
+
+func constructFullImageString(bomFile bom.Bom, subcomponent bom.BomSubComponent, image bom.BomImage) string {
+	registry := func() string {
```
desagar, line 139: should we be using bomFile.ResolveRegistry and bomFile.ResolveRepo here? Those also account for env var overrides.

platform-operator/controllers/verrazzano/component/thanos/thanos.go
davisgomes: This is the HTTP port for the Ruler. It is exposed by the image as well as the container here: https://github.com/verrazzano/verrazzano/blob/master/platform-operator/thirdparty/charts/thanos/templates/ruler/statefulset.yaml#L168

platform-operator/controllers/verrazzano/component/thanos/thanos.go
davisgomes: Thanks. I was looking for something like this, but I did not see it.

tests/e2e/pkg/thanos.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		Log(Error, fmt.Sprintf("Failed to get an OK response from %s, response: %d", url, resp.StatusCode))
+		return nil, err
+	}
+
+	type ruleData struct {
+		Data interface{} `json:"data"`
```
desagar, line 41: What ends up being the actual data type of this Data field? I'm wondering if we should assert more than should not be nil in the test - that can be a follow up though.

davisgomes: It is a `groups` field with a list of groups that each contain a list of rules. I'm just checking that it is not empty, which gives us good confidence the rules are loaded since there is no other rule source specified in the Ruler spec.

desagar (approved):

PR #6640, opened 2023-07-27
VZ-10398 install an AlertmanagerConfig

mbenmclean:
Install an AlertmanagerConfig by default and configure labels so that it gets wired up to Alertmanager so receiver automatically get configured.


platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -777,3 +786,42 @@ func reconcileAlertmanagerIngress(ctx spi.ComponentContext) error {
...
 	return nil
 }
+
+func createOrUpdateAlertmanagerConfig(ctx spi.ComponentContext) error {
+	alertmanagerEnabled, err := vzcr.IsAlertmanagerEnabled(ctx.EffectiveCR(), ctx)
+	if err != nil {
+		return err
+	}
+
+	if alertmanagerEnabled {
```
pfmackin, line 796: better to return immediatelyl if !alertmanageEnable so the following block can be shifted right.

platform-operator/controllers/verrazzano/component/prometheus/operator/promoperator.go
```diff
@@ -777,3 +786,42 @@ func reconcileAlertmanagerIngress(ctx spi.ComponentContext) error {
...
+				Labels: map[string]string{
+					alertmanagerConfigLabelName: alertmanagerConfigLabelValue,
+				},
+			},
+		}
+		_, err = controllerruntime.CreateOrUpdate(context.TODO(), ctx.Client(), alertmanagerConfig, func() error {
+			if alertmanagerConfig.Spec.Receivers == nil || len(alertmanagerConfig.Spec.Receivers) == 0 {
+				alertmanagerConfig.Spec.Receivers = []promoperapiv1alpha1.Receiver{
+					{
+						Name: nullAlertmanagerConfigReceiver,
```
lake-of-dreams, line 810: not sure why will we create a default `null-receiver` ?

PR #6651, opened 2023-07-28
VZ-10373 E2E tests for the Thanos Ruler with MC Fix

davisgomes:
**Changes**
- Enable the Ruler in the end to end tests
- Verify that the query and ruler ingresses are accessible (now only on admin clusters)


tests/e2e/verify-install/thanos/thanos_test.go
```diff
@@ -126,11 +135,42 @@ var _ = t.Describe("Thanos", Label("f:platform-lcm.install"), func() {
...
 			}, waitTimeout, pollingInterval).Should(BeTrue(), "Expected Thanos Ingresses should exist")
 		})
+
+		// GIVEN the Thanos is installed
+		// WHEN the ingresses exist
+		// THEN they should be accessible
+		WhenThanosInstalledIt("Thanos ingresses should be accessible", func() {
+			if pkg.IsManagedClusterProfile() {
+				Skip("Skip verifying ingress accessibility for managed clusters")
+			}
```
davisgomes, line 156: After a discussion with Deva, We've opted to disable the ingress check for the managed clusters. The test was getting the managed cluster Verrazzano password and failing. It is not likely that users will be logging into the consoles on managed clusters for Thanos, so we opted for simplicity for this fix for now.

pfmackin (approved):

PR #6659, opened 2023-07-28
Tweak the Ruler integration to fix communication issues

davisgomes:
**Changes**
- add the reload command to the prometheus config reloader so that it picks up rule changes
- alertmanager is outside of the mesh, so send http requests to the service endpoint
- Allow query to access the storeAPI of Thanos Ruler
- Allow Prometheus to scrape metrics from Thanos Ruler

**Testing**
- Install Verrazzano from this branch
- Change a rule and make sure it gets updated in Ruler
- Check the Ruler and Query pod logs to make sure no communication errors are happening
- Check the Ruler and Query consoles to make sure all rules/targets are available

### DIFF
--- a/platform-operator/helm_config/overrides/thanos-values.yaml
+++ b/platform-operator/helm_config/overrides/thanos-values.yaml
@@ -108,7 +108,7 @@ ruler:
     # Turn off hostname to disable auto-generated backend
     hostname: ""
   alertmanagers:
-    - https://prometheus-operator-kube-p-alertmanager:9093
+    - http://prometheus-operator-kube-p-alertmanager:9093
 
 receive:
   podSecurityContext:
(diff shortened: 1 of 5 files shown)


platform-operator/helm_config/overrides/thanos-values.yaml
```diff
@@ -108,7 +108,7 @@ ruler:
     # Turn off hostname to disable auto-generated backend
     hostname: ""
   alertmanagers:
-    - https://prometheus-operator-kube-p-alertmanager:9093
+    - http://prometheus-operator-kube-p-alertmanager:9093
```
desagar, line 111: because alertmanager is deployed without envoy?

desagar (approved):

platform-operator/helm_config/overrides/thanos-values.yaml
davisgomes: Yea exactly. We're communicating directly to the http server on the pod.

PR #6678, opened 2023-07-31
vz-9480 validating webhook rejects updates

AbeHern:
Introduced changes to verify Verrazzano Platform Operator is singleton


anders-swanson (approved):
lgtm

PR #6691, opened 2023-08-01
VZ-10395:CLI enhancement-dynamic progress bar

Mrunal-Rote:
Dynamic progress bar and table of components and their status.


tools/vz/cmd/install/state.go
```diff
@@ -0,0 +1,118 @@
+package install
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
```
adiforluls, line 7: Reorganize your imports and group the packages together, based on whether they are internal or external. This is kind of a good practice to follow.

 Example:
```
import (
        "fmt"
        "os"
        "strings"

        "github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
        "github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
              
        "sigs.k8s.io/controller-runtime/pkg/client"
)
```
In the example above I grouped go's standard packages, our packages and thirdparty packages separately.

tools/vz/cmd/install/state.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/helpers"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sort"
+	"strings"
+	"time"
+)
+
+// FindState Checks state of components until all of them are Ready or specified timeout is reached
+func FindState(cmd *cobra.Command, vzHelper helpers.VZHelper, duration time.Duration) error {
```
michael-cico, line 17: Just a general comment, can we add unit tests for new functionality?

michael-cico: Might require being able to override and capture stdout and be able to compare outputs, etc.

fredtibbitts, line 17: The name `FindState` doesn't seem to reflect what this function is really doing. Maybe `displayInstallationProgress` or something along those lines? Also the function should not be exported (see my other comment).

tools/vz/cmd/install/state.go
```diff
@@ -0,0 +1,132 @@
...
+	client, err := vzHelper.GetClient(cmd)
+	if err != nil {
+		fmt.Println("Error in client", err)
+		return err
+	}
+	time.Sleep(time.Second * 10)
+	fmt.Print("\033[2J\033[H") // Clear the console before entering the loop
```
fredtibbitts, line 24: If you look at the other CLI code that prints to stdout and stderr, you'll see code like this: `fmt.Fprintf(vzHelper.GetOutputStream()...`. All of these `Print`, `Fprint`, etc. should be changed to follow that pattern. That will allow writing unit test code that asserts on the stream content.

tools/vz/cmd/root/root.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/verrazzano/verrazzano/tools/vz/cmd/analyze"
 	"github.com/verrazzano/verrazzano/tools/vz/cmd/bugreport"
 	cmdhelpers "github.com/verrazzano/verrazzano/tools/vz/cmd/helpers"
-	"github.com/verrazzano/verrazzano/tools/vz/cmd/install"
+	install "github.com/verrazzano/verrazzano/tools/vz/cmd/install"
```
fredtibbitts, line 11: This change seems unnecessary.

tools/vz/cmd/install/state.go
```diff
@@ -0,0 +1,132 @@
...
+	readyCount := 0
+	totalCount := len(statusMap)
+	for index, status := range statusMap {
+		if status == "Ready" {
+			readyCount++
+		} else {
+			sync = append(sync, index)
+		}
+	}
+	return float64(readyCount) / float64(totalCount), sync
```
fredtibbitts, line 113: Is `totalCount` guaranteed to be non-zero?

fredtibbitts:
Is there a VZ ticket associated with this work? If so, please update the PR title to include the ticket so that automation will track the PR and commit. For example: "VZ-1234: CLI enhancement-dynamic progress bar".

fredtibbitts:
Please update `CHANGES.md` and add a change entry for this new feature.

tools/vz/cmd/install/state.go
Mrunal-Rote: Due to the use of ANSI code the Print statement it used

tools/vz/cmd/install/state.go
neeraj-n-singh: Added UT for coverage

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 Component version updates:
 
 - Kiali v1.66.1
+- Added progress flag in vz cli, which shows components progress in real time.
```
fredtibbitts, line 5: This is not a component update, it's a new feature. Please put this entry under the `Features` section for 2.0.0.

tools/vz/cmd/install/install.go
```diff
@@ -226,17 +227,20 @@ func runCmdInstall(cmd *cobra.Command, args []string, vzHelper helpers.VZHelper)
...
 		if err != nil {
 			return bugreport.AutoBugReport(cmd, vzHelper, err)
 		}
 		vzNamespace = vz.GetNamespace()
 		vzName = vz.GetName()
 	}
-
-	// Wait for the Verrazzano install to complete
-	err = waitForInstallToComplete(client, kubeClient, vzHelper, types.NamespacedName{Namespace: vzNamespace, Name: vzName}, timeout, vpoTimeout, logFormat)
+	progressFlag, err := cmd.PersistentFlags().GetBool(constants.ProgressFlag)
```
fredtibbitts, line 237: You should check `err` here and return appropriately. If for some reason it's not reasonable to check the error, replace `err` with underscore so it's clear that you're ignoring the return value.

tools/vz/cmd/install/install_test.go
```diff
@@ -937,4 +938,7 @@ func TestInstallFromFilename(t *testing.T) {
 	os.Args = append(os.Args, testFilenamePath)
 	cmd.Execute()
 	assert.Contains(t, errBuf.String(), "Error: invalid arguments specified:")
+	//Clean the resource args for
```
fredtibbitts, line 941: This comment seems to be truncated?

tools/vz/cmd/install/state.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1alpha1"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/helpers"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sort"
+	"strings"
+	"time"
```
fredtibbitts, line 16: These imports aren't formatted correctly. Use `goimports` or a similar tool to correctly format your imports. You can configure your IDE to automatically do this for you on file save.

tools/vz/cmd/install/state_test.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package install
+
+import (
```
fredtibbitts, line 6: Imports not formatted correctly.

tools/vz/cmd/install/install_test.go
neeraj-n-singh: Updated.

tools/vz/cmd/install/state_test.go
neeraj-n-singh: Updated.

tools/vz/cmd/install/state.go
neeraj-n-singh: updated.

CHANGES.md
neeraj-n-singh: Done

tools/vz/cmd/install/state.go
fredtibbitts: This was marked as resolved without answering my question.

tools/vz/cmd/install/state.go
fredtibbitts: Not really. The implementation of `fmt.Print` just calls `fmt.Fprint(os.Stdout, ...)`, so I'm not sure what ANSI codes have to do with it. I still think the calls should be changed to explicitly pass a stream, but at this point let's just get the code merged and that can be cleaned up later.

tools/vz/cmd/install/state.go
neeraj-n-singh: Yes, the total count be always non-zero in the current functional flow. as it will length of the component status map. Although we already have a checkpoint if len is zero then return an error.

fredtibbitts: It would probably be better to have the length check inside this function, even if the check already exists elsewhere. Otherwise it's not obvious from looking at this function that the check is already done, and if this function gets called from a different path in the future, it will be protected from a divide by zero error.

fredtibbitts (approved):

IshaGirdhar (approved):

PR #6719, opened 2023-08-03
VZ-10259 Updating OS file used

stevshap:
Using correct .txt file for trigger


agv426 (approved):
We don't need to check against the last clean commit here, but it won't hurt and we can clean that out in a followup task when we look at integrating the URL triggering back in later with the cron as a fallback

PR #6721, opened 2023-08-03
Add End-to-End Testing For ArgoCD Rancher Token Refresh Bug Fix

George-Aeillo:
In this PR, I edited the multi-cluster end-to-end related to ArgoCD to test that when an upgrade is triggered on a secret contains an ArgoCD token that talks with Rancher, the most recent token in the cluster is assigned to it. This test also checks the edge case where if no valid tokens are available in the cluster during an upgrade, a new token is created. I also created additional functionality in the Rancher end-to-end package, as there is a function that can add tokens for a given logged- in user and cluster ID and time-to-live. Also, I created a function for a given logged-in user and cluster ID that will find and delete all tokens in Rancher.


tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -36,6 +39,9 @@ var expectedPodsHelloHelidon = []string{"helidon-config-deployment"}
 var managedClusterName = os.Getenv("MANAGED_CLUSTER_NAME")
 var adminKubeconfig = os.Getenv("ADMIN_KUBECONFIG")
 var managedKubeconfig = os.Getenv("MANAGED_KUBECONFIG")
+var argoCDUsernameForRancher = "vz-argoCD-reg"
+var ttl = "240"
+var createdTimeStampForNewTokenCreated = ""
```
davisgomes, line 44: You don't have to assign empty strings to var declarations: 
```
var createdTimeStampForNewTokenCreated string
```
This defaults to an empty string

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+		// Checks that the secret corresponding to the managed-secret in the cluster has both createdAt and ExpiredAt annotations
+		t.It("The expected secret currently contains both the createdAt and ExpiredAt annotations", func() {
+			secretName := fmt.Sprintf("%s-argocd-cluster-secret", managedClusterName)
+			Eventually(func() error {
+				result, err := verifyCreatedAtAndExpiresAtTimestampsExist(argoCDNamespace, secretName)
+				if result != true {
+					pkg.Log(pkg.Error, fmt.Sprintf("Failed to get an ExpiredAt or Created Annotation in secret %s in namespace %s with error: %v", secretName, argoCDNamespace, err))
+					return err
+				}
+				return err
```
davisgomes, line 136: If result is untrue and the err is nil, this condition could pass when it should not. Instead, you can return both the result and the error and just check the boolean case:
```
secretName := fmt.Sprintf("%s-argocd-cluster-secret", managedClusterName)
			Eventually(func() error {
				return verifyCreatedAtAndExpiresAtTimestampsExist(argoCDNamespace, secretName)
			}, waitTimeout, pollingInterval).Should(BeTrue(), "Expected to find both Created and Expired At Annotations "+secretName)
```
ginkgo will treat a non-nil error as a failure, and the Eventually will not pass.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+				if clusterID == "" {
+					pkg.Log(pkg.Error, "The managed cluster does not have a clusterID value")
+					err := fmt.Errorf("ClusterID value is not yet populated for the managed cluster")
+					return err
+				}
+				httpClientForRancher, err := pkg.GetVerrazzanoHTTPClient(adminKubeconfig)
+				if err != nil {
+					pkg.Log(pkg.Error, "Error getting the Verrazzano http client")
+					return err
+				}
```
davisgomes, line 263: It looks like you use this code section multiple times. Can you extrapolate it out to a function call?

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+				if err != nil {
+					pkg.Log(pkg.Error, "Error creating New Token")
+					return err
+				}
+				return err
+
+			}, waitTimeout, pollingInterval).ShouldNot(HaveOccurred(), "Expected to Be Able To Create a Token through the API")
+		})
+		// Tests that a secret can be triggered to begin an update
+		t.It("The secret can be successfully altered to trigger an update locally and sent through the client", func() {
```
davisgomes, line 182: What use case is this testing? We only want the code to be updating the secret.

davisgomes: If it is a set up for another test, it doesn't need to be in its own "It" block

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+					pkg.Log(pkg.Error, "The managed cluster does not have a clusterID value")
+					err := fmt.Errorf("ClusterID value is not yet populated for the managed cluster")
+					return err
+				}
+				httpClientForRancher, err := pkg.GetVerrazzanoHTTPClient(adminKubeconfig)
+				if err != nil {
+					pkg.Log(pkg.Error, "Error getting the Verrazzano http client")
+					return err
+				}
+				_, err = pkg.GetAndDeleteTokenNamesForLoggedInUserBasedOnClusterID(httpClientForRancher, adminKubeconfig, clusterID, rancherConfigForArgoCD.APIAccessToken, *t.Logs)
```
davisgomes, line 264: Same comment about the Eventually check.

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: This code is the set up for another test, as it updates the secret. If it is not an "It" block, but there is still and a delay and I check that it is actually expired, I still would like it an eventually block. How do you suggest I change the ginkgo file to reflect this?

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
davisgomes: You can do two eventually blocks in the same It block

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+				if err != nil {
+					pkg.Log(pkg.Error, "Error editing the secret through the client")
+					return err
+				}
+				_, ok := editedSecret.Annotations["verrazzano.io/expires-at-timestamp"]
+				if ok {
+					pkg.Log(pkg.Error, "The client was successful, but the secret was not successful edited")
+					return fmt.Errorf("The client was successful, but the secret was not successful edited")
+				}
+				return err
```
davisgomes, line 205: Same comment here. You should check if the annotation exists at a minimum in the Eventually check

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -4,20 +4,23 @@
 package argocd_test
 
 import (
+	"context"
 	b64 "encoding/base64"
 	"fmt"
+	"github.com/verrazzano/verrazzano/tests/e2e/multicluster/examples"
+	dump "github.com/verrazzano/verrazzano/tests/e2e/pkg/test/clusterdump"
```
davisgomes, line 11: If you get the chance, can you clean up these imports?

tests/e2e/multicluster/verify-argocd/argocd_test.go
davisgomes: There's a few others I hadn't commented on, but this still applies so please check on those as well.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -188,3 +401,37 @@ func findServerName(namespace, name string) (bool, error) {
...
+	if err != nil {
+		pkg.Log(pkg.Error, fmt.Sprintf("Failed to get secret %s in namespace %s with error: %v", name, namespace, err))
+		return false, err
+	}
+	annotationMap := s.GetAnnotations()
+	createdValue, ok := annotationMap["verrazzano.io/create-timestamp"]
+	if !ok || createdValue == "" {
+		return false, fmt.Errorf("Created Annotation Value Not Found")
+	}
+	expiresAtValue, ok := annotationMap["verrazzano.io/expires-at-timestamp"]
```
davisgomes, line 417: Can you create constants for these flags. They are duplicated quite often

tests/e2e/pkg/rancher.go
```diff
@@ -123,13 +216,138 @@ func getRancherUserToken(log *zap.SugaredLogger, httpClient *retryablehttp.Clien
...
+	listOfTokens := listOfTokenOutputFromRancher.Data
+	var listOfTokensToReturn = make([]string, 0)
+
+	for _, token := range listOfTokens {
+		//Check if token is valid, if it is token that we want than append it
+		if token.ClusterID == clusterID {
+			listOfTokensToReturn = append(listOfTokensToReturn, token.Name)
+		}
+
+	}
```
davisgomes, line 325: I don't think you need to filter this twice. You can combine the for loops and make this the first condition and just have a continue.

tests/e2e/pkg/rancher.go
```diff
@@ -123,13 +216,138 @@ func getRancherUserToken(log *zap.SugaredLogger, httpClient *retryablehttp.Clien
...
+	fmt.Println(response.Body)
+	responseBody, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	err = httputil.ValidateResponseCode(response, http.StatusOK)
+	if err != nil {
+		return nil, err
+	}
```
davisgomes, line 310: It seems like there is a lot of shared code here. Can you extrapolate this to a "token request" function or something similar?

tests/e2e/pkg/rancher.go
davisgomes: Then you can append the ones that meet this condition to the "listOfTokensToReturn" slice.

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +120,214 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+					pkg.Log(pkg.Error, "Failed to add an expires-at-timestamp to the secret based on a new token that is created")
+					return fmt.Errorf("The secret was not successfully edited, as it does not have an expired timestamp")
+				}
+				createdTimestampCurrentlyOnUpdatedSecret, ok := updatedSecret.Annotations["verrazzano.io/create-timestamp"]
+				if createdTimestampCurrentlyOnUpdatedSecret != createdTimeStampForNewTokenCreated || !ok {
+					pkg.Log(pkg.Error, "Failed to successfully update the secret with the created timestamp of the most recent token created")
+					return fmt.Errorf("The created-at timestamp of the secret was not correctly updated with the created timestamp of the most recently created token")
+				}
+				return err
+			}, waitTimeout, pollingInterval).ShouldNot(HaveOccurred(), "Expected to have the secret reflect the created timestamp of the new token that was created")
```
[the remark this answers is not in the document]
George-Aeillo: Thanks!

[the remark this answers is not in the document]
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks! I refactored this function, as it returned an error that was ignored and not checked, so it really should be returning an error.

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks! I just did.

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks!

tests/e2e/pkg/rancher.go
```diff
@@ -19,11 +16,107 @@ import (
...
+
+type TokenPostResponse struct {
+	Token   string `json:"token"`
+	Created string `json:"created"`
+}
+
+type ListOfTokenOutputFromRancher struct {
+	Type  string `json:"type"`
+	Links struct {
+		Self string `json:"self"`
```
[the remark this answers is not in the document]
George-Aeillo: Thanks!

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks! I refactored it.

stevshap (approved):

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -135,13 +240,13 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
-})
-
-var _ = AfterSuite(afterSuite)
+//var afterSuite = t.AfterSuiteFunc(func() {
+//	if failed || !beforeSuitePassed {
+//		dump.ExecuteBugReport(testNamespace)
+//	}
+//})
+//
+//var _ = AfterSuite(afterSuite)
```
davisgomes, line 249: Should this be deleted? If so, remove it instead of commenting it out.

tests/e2e/pkg/rancher.go
```diff
@@ -19,11 +16,50 @@ import (
...
+		ExpiresAt      string      `json:"expiresAt"`
+		ID             string      `json:"id"`
+		IsDerived      bool        `json:"isDerived"`
+		LastUpdateTime string      `json:"lastUpdateTime"`
+		Name           string      `json:"name"`
+		TTL            int         `json:"ttl"`
+		Type           string      `json:"type"`
+		UserID         string      `json:"userId"`
+		UserPrincipal  string      `json:"userPrincipal"`
+		UUID           string      `json:"uuid"`
```
davisgomes, line 59: I think there are still a lot of unused fields here. Remove any that do not get used in the code.

tests/e2e/pkg/rancher.go
```diff
@@ -330,3 +420,70 @@ func GetClusterKubeconfig(log *zap.SugaredLogger, httpClient *retryablehttp.Clie
...
+		log.Errorf("API Endpoint not successfully received based on KubeConfig Path")
+		return nil, "", err
+	}
+	rancherURL, err := GetURLForIngress(&log, api, "cattle-system", "rancher", "https")
+	if err != nil {
+		log.Errorf("URL For Rancher not successfully found")
+		return nil, "", err
+	}
+	reqURL := rancherURL + "/v3/tokens"
+	if requestType == "POST" {
```
davisgomes, line 438: This `if requestType == "POST"` and `if requestType == "GET"` strike me as odd. Maybe instead of an if statement, you can write wrapper functions for Get and Post that both call the same underlying request function.

davisgomes: Then you can also support a 'DELETE' request and simplify the `GetAndDeleteTokenNamesForLoggedInUserBasedOnClusterID` function

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +123,103 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+		t.It("The expected secret currently contains both the createdAt and ExpiredAt annotations", func() {
+			secretName := fmt.Sprintf("%s-argocd-cluster-secret", managedClusterName)
+			Eventually(func() (bool, error) {
+				return verifyCreatedAtAndExpiresAtTimestampsExist(argoCDNamespace, secretName)
+			}, waitTimeout, pollingInterval).Should(BeTrue(), "Expected to find both Created and Expired At Annotations "+secretName)
+		})
+		// Tests that a new ArgoCD token is able to be created
+		t.It("A new ArgoCD token is able to be created through the Rancher API", func() {
+			Eventually(func() error {
+				httpClientForRancher, clusterID, APIAccessToken, err := getRequiredInfoToPreformTokenOperationsInRancherForArgoCD()
```
davisgomes, line 140: These values are essential to a few of the pkg functions, and a lot of variables get passed into the subsequent function calls. You could consider moving `getRequiredInfoToPreformTokenOperationsInRancherForArgoCD()`. To the pkg package, and just call it inside of these functions. i.e.
```
func AddAccessTokenToRancherForLoggedInUser(...) (string, err) {
    httpClientForRancher, clusterID, APIAccessToken, err := getRequiredInfoToPreformTokenOperationsInRancherForArgoCD(...)
    ...
}
```
Then you don't have to deal with passing this information as a parameter, and it might make the initial function calls simpler.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -114,7 +123,103 @@ var _ = t.Describe("Multi Cluster Argo CD Validation", Label("f:platform-lcm.ins
...
+					return err
+				}
+				createdTimeStampForNewTokenCreated, err = pkg.AddAccessTokenToRancherForLoggedInUser(httpClientForRancher, adminKubeconfig, clusterID, ttl, APIAccessToken, *t.Logs)
+				if err != nil {
+					pkg.Log(pkg.Error, "Error creating New Token")
+					return err
+				}
+				return err
+
+			}, waitTimeout, pollingInterval).ShouldNot(HaveOccurred(), "Expected to Be Able To Create a Token through the API")
```
davisgomes, line 152: I think you can change this to a `ShouldNot(BeEmpty())` and return the `createdTimeStampForNewTokenCreated`

davisgomes: Typically for these tests, we only want to return the error if it is the only information we have. Otherwise, we can look at a variable and return the error as the second parameter. This gives us more information on test failures.

tests/e2e/pkg/rancher.go
```diff
@@ -330,3 +420,70 @@ func GetClusterKubeconfig(log *zap.SugaredLogger, httpClient *retryablehttp.Clie
...
+		}
+		responseBody, err := io.ReadAll(response.Body)
+		if err != nil {
+			return nil, reqURL, err
+		}
+		err = httputil.ValidateResponseCode(response, http.StatusOK)
+		if err != nil {
+			return nil, reqURL, err
+		}
+		return responseBody, reqURL, err
```
davisgomes, line 486: If you choose to incorporate delete into the underlying functions, then I don't think you'll need to return the reqURL. I would make a comment about separating out this functionality, but it might not be necessary if you make that change.

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks! That was a good catch!

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks!

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks! I will make that change, so returning the reqURL is not necessary.

tests/e2e/multicluster/verify-argocd/argocd_test.go
```diff
@@ -188,3 +293,97 @@ func findServerName(namespace, name string) (bool, error) {
...
+	}
+	editedSecret, err := clientSetForCluster.CoreV1().Secrets(argoCDNamespace).Update(context.TODO(), secretToTriggerUpdate, metav1.UpdateOptions{})
+	if err != nil {
+		pkg.Log(pkg.Error, "Error editing the secret through the client")
+		return err
+	}
+	_, ok := editedSecret.Annotations[expiresAtTimeStamp]
+	if ok {
+		pkg.Log(pkg.Error, "The client was successful, but the secret was not successful edited")
+		return fmt.Errorf("The client was successful, but the secret was not successful edited")
```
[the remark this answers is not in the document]
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks!

tests/e2e/multicluster/verify-argocd/argocd_test.go
George-Aeillo: Thanks!

tests/e2e/pkg/rancher.go
George-Aeillo: Thanks!

davisgomes (approved):

PR #6734, opened 2023-08-07
VZ:10491:Updated error handling noise in effective config map

neeraj-n-singh:


platform-operator/controllers/configmaps/overrides/controller.go
```diff
@@ -78,8 +78,14 @@ func (r *OverridesConfigMapsReconciler) reconcileInstallOverrideConfigMap(ctx co
 
 	// Get the ConfigMap present in the Verrazzano CR namespace
 	configMap := &corev1.ConfigMap{}
-	if err := controllers.CreateOrUpdateEffectiveConfigCM(ctx, r.Client, vz, r.log); err != nil {
-		return newRequeueWithDelay(), nil
+	var currentCondition installv1alpha1.ConditionType
+	if len(vz.Status.Conditions) > 0 {
+		currentCondition = vz.Status.Conditions[len(vz.Status.Conditions)-1].Type
+	}
+	if currentCondition == installv1alpha1.CondInstallComplete || currentCondition == installv1alpha1.CondUpgradeComplete {
```
prasad-shirodkar, line 85: Any reason to switch from `r.log` to `vzlog.DefaultLogger()`?

adiforluls: What's the logic behind configMap controller attempting to createOrUpdate this configMap?

neeraj-n-singh: In case of cm deletion by some manual action, cm reconciliation will create the same. VZ reconciliation will not be recreated until there is any operation changes.

I removed the logging as it was not necessary as per discussion. We are already handling errors.

adiforluls (approved):

PR #6751, opened 2023-08-08
VZ-10451 watch secrets and configmaps for VZ config

pfmackin:
Watch secrets and configmaps for VZ config
Use latest version of modules with SPI changes


platform-operator/experimental/module-integration/controllers/verrazzano/watcher.go
```diff
@@ -0,0 +1,57 @@
...
+		{
+			WatchedResourceKind: source.Kind{Type: &corev1.ConfigMap{}},
+			FuncShouldReconcile: r.ShouldConfigMapTriggerReconcile,
+		},
+	}
+}
+
+// ShouldSecretTriggerReconcile returns true if reconcile should be done in response to a Secret lifecycle event
+func (r Reconciler) ShouldSecretTriggerReconcile(vzNSN types.NamespacedName, secret client.Object, _ controllerspi.WatchEventType) bool {
+	if secret.GetNamespace() != vzNSN.Namespace {
```
mbenmclean, line 33: Do we need this limitation of secrets and configmaps have to be in the VZ CR namespace? It makes more sense to have the secret and configmap in the same namespace as the module.

mbenmclean: Or is this just a holdover while we still are specifying overrides in the VZ CR?

mbenmclean (approved):

michael-cico (approved):

PR #6754, opened 2023-08-08
OCNE Cluster Driver Tests - Single Cluster Creation

marcogelle:
This removes all cluster creation test cases except for 1 bare minimum case. This is a temporary measure to help fix the recent test failures.

### DIFF
--- a/tests/e2e/clusterapi/ocne-driver/ocne_cluster_driver_test.go
+++ b/tests/e2e/clusterapi/ocne-driver/ocne_cluster_driver_test.go
@@ -23,6 +23,7 @@ const (
 	waitTimeout            = 45 * time.Minute
 	pollingInterval        = 30 * time.Second
 	skipOCNEUpgradeMessage = "Skipping test since the kubernetes version is same for install and update operations for OCNE cluster upgrade"
+	skipRunAllTestsMessage = "Skipping test since the runAllTests flag was set to false"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


tests/e2e/clusterapi/ocne-driver/ocne_cluster_driver_test.go
```diff
@@ -161,150 +155,4 @@ var _ = t.Describe("OCNE Cluster Driver", Label("f:rancher-capi:ocne-cluster-dri
 			}, shortWaitTimeout, shortPollingInterval).Should(BeNil(), fmt.Sprintf("could not verify cluster %s", clusterNameSingleNode))
 		})
 	})
-
-	// Cluster 2. Create with a node pool, then perform an update.
-	t.Context("OCNE cluster creation with node pools", Ordered, func() {
```
prasad-shirodkar, line 166: instead of removing the test cases, can we instead they be run only if some bool flag is set to true?

marcogelle: I like that idea. I'll try implementing this now

prasad-shirodkar: Say, add a bool flag minimal (set to true). Then using ginkgo's Skip function, skip the tests that are not minimal

```
t.It("....", func() {
  if minimal {
     Skip("Skipping test since it is not minimal")
 }
.....
.....
}
```
Not sure if Skip function works at container nodes like `Context` etc. Feel free to give it a try.

marcogelle: Looking at the Ginkgo docs under "Skipping Specs" -- https://onsi.github.io/ginkgo/ -- you can't call `Skip` in container nodes. However, I will try putting the Skips in `BeforeAll` nodes to avoid lots of code duplication.

theabmitra (approved):
lgtm

tests/e2e/clusterapi/ocne-driver/ocne_cluster_driver_test.go
```diff
@@ -266,9 +286,6 @@ var _ = t.Describe("OCNE Cluster Driver", Label("f:rancher-capi:ocne-cluster-dri
 		})
 
 		t.It("check OCNE cluster is active with the minimum OCNE supported Kubernetes version and related info", func() {
-			if !ocneMetadataItemToInstall.KubernetesVersion.LessThan(ocneMetadataItemToUpgrade.KubernetesVersion) {
```
prasad-shirodkar, line 269: Please do not remove these. These should be retained. They should also be preceded by the new if condition block.

marcogelle: I did not remove them, rather just moved them into the `BeforeAll` block. Instead of repeating this check in every `It` node, it is functionally equivalent to do this check once in the `BeforeAll`. Does this look okay to you?

prasad-shirodkar (approved):

PR #6755, opened 2023-08-08
VZ-10433.  Pass component VZ CR config to module as structured values

michael-cico:
Modules integration, this change adds the relevant portions of the VZ CR to a module's `Module` CR values section.  This is to set enough in the Values overrides to 

* Identify module configuration from the VZ CR
* Ensure that the module only reconciles when the Module CR changes, based on relevant configuration

The generated Helm values are not currently used by the modules, but may be later on so the module management can be driven solely from the Module CR and not from the computed "effective" CR.

The changes 

* Introduces a new method to the `ModuleIntegration` interface, `GetModuleConfigAsHelmValues`.  
* This new method is called by the new main controller to populate the `Module.Spec.Values` field with a well-known values prefix, `verrazzano.module.spec`, to store relevant configuration information as Helm values.  
* These values will be ignored by the Helm charts, but will provide enough context so that if those values change a reconcile of the Module CR will occur.
* Each component implements the method to provide as much context as needed
* The returned values are merged with any `InstallOverrides.Values` fields defined in the component's Verrazzano CR configuration
* The resulting merged values set are stored in the `Module.Spec.Values` field

For example, the following Verrazzano CR for CertManager installation

```
apiVersion: install.verrazzano.io/v1beta1
kind: Verrazzano
metadata:
  name: my-verrazzano
spec:
  profile: none
  components:
    certManager:
      enabled: true
```

Will result in the following module CR for CertManager, with a Helm values snippet under the well-known `verrazzano.module.spec` stanza:

```
apiVersion: platform.verrazzano.io/v1alpha1
kind: Module
metadata:
  annotations:
    module.verrazzano.io/vz-cr-name: my-verrazzano
    module.verrazzano.io/vz-cr-namespace: default
  name: cert-manager
  namespace: verrazzano-install
spec:
  moduleName: cert-manager
  targetNamespace: cert-manager
  version: 2.0.0-local+80a4759a
  values:
    verrazzano:
      module:
        spec:
          clusterResourceNamespace: cert-manager
    webhook:
      replicaCount: 1
      affinity:
        podAntiAffinity:
        ....
```


platform-operator/controllers/verrazzano/component/spi/component.go
```diff
@@ -127,6 +128,8 @@ type ModuleIntegration interface {
 	ShouldUseModule() bool
 	// GetWatchDescriptors returns the list of WatchDescriptor for objects being watched by the component
 	GetWatchDescriptors() []controllerspi.WatchDescriptor
+	// GetModuleSpec returns an unstructured JSON snippet representing the portion of the Verrazzano CR that corresponds to the module
+	GetModuleSpec(effectiveCR *v1alpha1.Verrazzano) (*apiextensionsv1.JSON, error)
```
michael-cico, line 132: name is confusing, options?

* ModuleSpecValues
* ModuleValue

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager_component.go
```diff
@@ -227,3 +228,8 @@ func checkExistingCertManager(_ runtime.Object) error {
 
 	return nil
 }
+
+// GetModuleSpec returns an unstructured JSON snippet representing the portion of the Verrazzano CR that corresponds to the module
+func (c certManagerComponent) GetModuleSpec(effectiveCR *v1alpha1.Verrazzano) (*apiextensionsv1.JSON, error) {
+	return spi.NewModuleSpecValue(effectiveCR.Spec.Components.CertManager)
```
michael-cico, line 234: check nil and dereference the struct

PR #6760, opened 2023-08-09
Add certificate to the Rancher requests

davisgomes:
Updates the source of truth for additional CA certs for the multicluster Rancher proxy requests.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/common/rancher.go
+++ b/platform-operator/controllers/verrazzano/component/common/rancher.go
@@ -76,19 +76,19 @@ func GetRootCA(c client.Reader) ([]byte, error) {
 	return secret.Data[RancherCACert], nil
 }
 
-// GetAdditionalCA fetches the Rancher tls-cA secret
-// returns empty byte array of the secret tls-ca is not found
+// GetAdditionalCA fetches the verrazzano-tls-ca secret
+// returns empty byte array of the secret verrazzano-tls-ca is not found
 func GetAdditionalCA(c client.Reader) []byte {
 	secret := &corev1.Secret{}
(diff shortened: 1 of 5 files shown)


platform-operator/controllers/verrazzano/component/common/rancher.go
```diff
@@ -62,18 +64,18 @@ func GetAdminSecret(c client.Reader) (string, error) {
 	return string(secret.Data["password"]), nil
 }
 
-// GetRootCA gets the root CA certificate from the Rancher TLS secret. If the secret does not exist, we
-// return a nil slice.
-func GetRootCA(c client.Reader) ([]byte, error) {
+// GetRootCAAndCertChain gets the root CA certificate, the certificate chain, and the certificate key
+// from the Rancher TLS secret. If the secret does not exist, we return a nil slice.
```
michael-cico, line 68: So, this will return nil, which will cause the trusted cert chain inside the pod/container to be used, correct?

You should pull the private CA bundles from {{cattle-system/tls-ca}} if it exists, or {{verrazzano-system/verrazzano-tls-ca}}.  The latter was recently created to be a source of truth for private CA configurations.  The `ca.crt` field is not populated in CertManager cases where ACME issuers are being used (e.g., Let's Encrypt staging env).

michael-cico: Also, I don't think you need to provide the private key and leaf cert from the ingress.  Those are provided, really it's just making sure that we provide the right CA bundle for private CAs, or nil for certs issued by trusted CAs.

fredtibbitts (approved):

PR #6765, opened 2023-08-09
VZ-10472:  Cluster Operator controller for processing CAPI cluster registration/de-registration with Rancher

jmaron99:
The controller reconciles CAPI clusters and performs the operations necessary to register or de-register CAPI clusters for creation/deletion, respectively.  The controller will also ignore Rancher created clusters since they are already handled by the OCNE container driver.


cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,407 @@
...
+
+func SetClusterRegistrationFunction(f ClusterRegistrationFnType) {
+	clusterRegistrationFn = f
+}
+
+func SetDefaultClusterRegistrationFunction() {
+	clusterRegistrationFn = ensureRancherRegistration
+}
+
+var gvk = schema.GroupVersionKind{
```
anders-swanson, line 68: We have `sigs.k8s.io/cluster-api` in the go mod. is it necessary to use unstructured?

anders-swanson: i guess it doesn't really matter

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,407 @@
...
+				return ctrl.Result{}, err
+			}
+		}
+
+		if err := r.removeFinalizer(cluster); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// delete the cluster id secret
+		clusterRegistrationStatusSecret, err := r.getClusterRegistrationStatusSecret(ctx, cluster)
```
anders-swanson, line 123: don't we want to do this before removing the finalizers?

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	return nil
+}
+
+// ensureRancherRegistration ensures that the CAPI cluster is registered with Rancher.
+func ensureRancherRegistration(r *CAPIClusterReconciler, ctx context.Context, cluster *unstructured.Unstructured) (ctrl.Result, error) {
+	kubeconfig, err := r.getWorkloadClusterKubeconfig(ctx, cluster)
+	if err != nil {
+		// requeue since we're waiting for cluster
+		return vzctrl.NewRequeueWithDelay(2, 3, time.Second), err
```
anders-swanson, line 177: can you use a helper for requeue, since we are doing this so often

cluster-operator/controllers/capi/capi_cluster_controller.go
theabmitra: if we use the client, do we have to make a code change when CAPI moves to `v1beta2` ?

jmaron99: Is there a response required from me to resolve this?  Not clear...

theabmitra: i think its okay, because i realized the client wont just auto work with `v1beta2` unless that itself is updated in the go.mod.  But we need to remember to update this when that happens.

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: Tried that and got hung up on not finding the cluster id secret for a situation in which the cluster deregistration re-queued.  The finalizer would therefore not clear (wasn't reached).  Had to move it here.  There may be another solution but this one seemed straightforward.

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: Not sure what you mean - this is a helper, isn't it?

cluster-operator/controllers/capi/capi_cluster_controller.go
anders-swanson: you should handle apierrors.Isnotfound in that case, we don't need to delete the secret if it doesn't exist.

cluster-operator/controllers/capi/capi_cluster_controller.go
anders-swanson: i mean its called a lot with the same arguments `2, 3, time.Second`

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,416 @@
...
+		Contexts:       contexts,
+		CurrentContext: contextName,
+		AuthInfos:      authInfos,
+	}, nil
+}
+
+// GetRancherAPIResources returns the set of resources required for interacting with Rancher
+func (r *CAPIClusterReconciler) GetRancherAPIResources(cluster *unstructured.Unstructured) (*rancherutil.RancherConfig, vzlog.VerrazzanoLogger, error) {
+	// Get the resource logger needed to log message using 'progress' and 'once' methods
+	log, err := vzlog.EnsureResourceLogger(&vzlog.ResourceConfig{
```
anders-swanson, line 377: can we create the logger when we get the cluster?

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,416 @@
...
+		// no cluster id found
+		return fmt.Errorf("No cluster ID found for cluster %s", cluster.GetName())
+	}
+	rc, log, err := r.GetRancherAPIResources(cluster)
+	if err != nil {
+		return err
+	}
+	_, err = vmc.DeleteClusterFromRancher(rc, clusterId, log)
+	if err != nil {
+		log.Errorf("Unable to unregister cluster %s from Rancher: %v", cluster.GetName(), err)
```
anders-swanson, line 411: mixes logging and error returning here

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		r.Log.Errorf("Unable to persist status for cluster %s: %v", cluster.GetName(), err)
+		return err
+	}
+
+	return nil
+}
+
+// getWorkloadClusterKubeconfig returns a kubeconfig for accessing the workload cluster
+func (r *CAPIClusterReconciler) getWorkloadClusterKubeconfig(ctx context.Context, cluster *unstructured.Unstructured) ([]byte, error) {
```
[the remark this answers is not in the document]
anders-swanson: See equivalent method in the kontainerdriver:
https://github.com/verrazzano/kontainer-engine-driver-ociocne/blob/main/pkg/variables/variables.go#L378

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: Ah

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: I guess I wasn't clear.  When I moved the deletion of the secret to where you suggest I ran into a situation in which the secret had been deleted but the code was attempting to retrieve the cluster id.  Moving it here resolved that scenario.

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: this is being created specifically since the registration method requires it.  It's a VerrazzanoLogger and I didn't identify a method for wrapping the existing reconciler logger or the like.

cluster-operator/controllers/capi/capi_cluster_controller.go
jmaron99: not sure I understand what you mean.  Do you want the log method removed?

cluster-operator/controllers/capi/capi_cluster_controller.go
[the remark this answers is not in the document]
jmaron99: I tried exactly that initially.  For some reason the kubeconfig from the secret could never establish a connection with the workload cluster api server.  I debugged with @theabmitra for a while and eventually he suggested this mechanism.  Once I implemented this mechanism the communication issues resolved.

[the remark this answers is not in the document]
jmaron99: Re-implemented the "kubeconfig from secret" logic and it did work this time - perhaps the issue I saw previously had more to do with tenancy/cluster resources. Anyhow - pushed the changes.

anders-swanson (approved):

PR #6776, opened 2023-08-10
VZ-10488 Add Check To VZ Job to See If VZ Helper Was Ran

George-Aeillo:
This ticket adds an additional stage to the verrazzano job, which checks that the current verrazzano development version does not match a release version. If this check fails on master, or a release branch, it sends a slack message to the release owners, but does not fail the test. This ticket also adds an additional stage to the release job to the upgrade resiliency job. On this job, it is the same logic, as the Verrazzano job, except that the job will not only send a slack message to the release owners if the check fails, the job will also fail outright.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -233,7 +233,7 @@ pipeline {
                          script {
                             VERRAZZANO_DEV_VERSION_CHECK = sh (script: "git ls-remote --tags origin | grep -F ${VERRAZZANO_DEV_VERSION}", returnStatus: true )
                             if (VERRAZZANO_DEV_VERSION_CHECK == 0) {
-                                if (env.JOB_NAME == "verrazzano/master" || env.JOB_NAME ==~ "verrazzano/release-*") {
+                                if (env.JOB_NAME ==~ "verrazzano/release-*") {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -226,6 +227,26 @@ pipeline {
                 }
             }
         }
+       stage('Verrazzano development version check') {
```
agv426, line 230: I don't think we want to fail the entire job here as that could block other folks from getting their work in until the dev version gets updated. So in this Jenkinsfile, we don't want it to "exit 1", but we do want it to send the alert so someone knows that the dev version needs to get updated.

Having said that, the upgrade resiliency tests that run in the periodic testing are likely to fail due to this, so those tests should also get this check and in that pipeline I think the test should fail before wasting lots of time and resources trying to test something that will fail in an obscure way. So I think that what you have here can be added into the ci/chaos/JenkinsfileResiliencyTrigger so fail describing the issue and send the alert

George-Aeillo: Thanks!

ci/chaos/JenkinsfileResiliencyTrigger
```diff
@@ -122,6 +123,26 @@ pipeline {
...
+                               sh """
+                           [[ -z \$(git ls-remote --tags origin | grep -F ${VERRAZZANO_DEV_VERSION}) ]] || exit 1
+
+
+                       """
+                           }
+                           post {
+                               failure {
+                                   script {
+                                       SKIP_TRIGGERED_TESTS = true
```
agv426, line 137: Can delete this line, this Jenkinsfile doesn't use SKIP_TRIGGERED_TESTS

George-Aeillo: Thanks!

agv426 (dismissed):

agv426 (approved):

PR #6790, opened 2023-08-11
Fix verrazzano-cluster-operator VMC reconcile error metric counting

fredtibbitts:
The VMC reconcile error metric was not being incremented correctly. Any errors during reconciliation were being ignored and the success metric was incremented in all cases.

### DIFF
--- a/cluster-operator/controllers/vmc/vmc_controller.go
+++ b/cluster-operator/controllers/vmc/vmc_controller.go
@@ -108,13 +108,13 @@ func (r *VerrazzanoManagedClusterReconciler) Reconcile(ctx context.Context, req 
 	r.log = log
 	log.Oncef("Reconciling Verrazzano resource %v", req.NamespacedName)
 	res, err := r.doReconcile(ctx, log, cr)
-	if err != nil || vzctrl.ShouldRequeue(res) {
-		if err != nil {
-			// Never return an error since it has already been logged and we don't want the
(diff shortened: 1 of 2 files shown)


davisgomes (dismissed):

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -108,18 +108,17 @@ func (r *VerrazzanoManagedClusterReconciler) Reconcile(ctx context.Context, req
...
 	log.Oncef("Reconciling Verrazzano resource %v", req.NamespacedName)
 	res, err := r.doReconcile(ctx, log, cr)
-	if vzctrl.ShouldRequeue(res) {
+	if err != nil || vzctrl.ShouldRequeue(res) {
+		if err != nil {
+			// Never return an error since it has already been logged and we don't want the
+			// controller runtime to log again (with stack trace).  Just re-queue if there is an error.
+			reconcileErrorCount.Inc()
+			return newRequeueWithDelay(), nil
+		}
```
davisgomes, line 117: Is there a reason these are nested instead of `if err != nil ... return` then `if  vzctrl.ShouldRequeue(res) ... return`

fredtibbitts: No good reason, I had moved the code around and I'm not sure what I was thinking. Fixed it, thanks for pointing it out.

davisgomes (approved):

PR #6796, opened 2023-08-14
Tvlatas/1.6 use jao

agv426:
This switches the release-1.6 branch over to use the JAO for provisioning instances (from the OCI plugin).

### DIFF
--- a/ci/private-registry/Jenkinsfile
+++ b/ci/private-registry/Jenkinsfile
@@ -2,7 +2,7 @@
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 def DOCKER_IMAGE_TAG
-def agentLabel = params.DISTRIBUTION_VARIANT == "Lite" ? "large_1_6" : "fralarge_1_6"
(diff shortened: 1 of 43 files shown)


ci/private-registry/Jenkinsfile
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 def DOCKER_IMAGE_TAG
-def agentLabel = params.DISTRIBUTION_VARIANT == "Lite" ? "large_1_6" : "fralarge_1_6"
+def agentLabel = params.DISTRIBUTION_VARIANT == "Lite" ? "1.6-large" : "airgap-1.6-large"
```
George-Aeillo, line 5: In the 1.5 file, it is 1.5-large-phx, it it supposed to be 1.6-large-phx in this case?

agv426: We are moving away from locking the release pipelines to PHX only, it is intentional (1.5 can move away from PHX as well actually)

George-Aeillo (approved):

PR #6798, opened 2023-08-14
VZ-10407 add module catalog

mbenmclean:


platform-operator/internal/config/config.go
```diff
@@ -31,9 +31,11 @@ const (
 	helmOamChartsDirSuffix       = "/platform-operator/thirdparty/charts/oam-kubernetes-runtime"
 	helmOverridesDirSuffix       = "/platform-operator/helm_config/overrides"
 	integrationChartsDirSuffix   = "/platform-operator/experimental/manifests/integration-charts"
+	modulesIntegrationDir        = "/platform-operator/experimental/catalog"
```
pfmackin, line 34: need to follow the naming convention with suffix.  Can you just use catalogDirSuffix

platform-operator/internal/config/config.go
```diff
@@ -247,6 +252,13 @@ func GetIntegrationChartsDir() string {
 	return filepath.Join(instance.VerrazzanoRootDir, integrationChartsDirSuffix)
 }
 
+func GetModuleIntegrationCatalog() string {
```
pfmackin, line 255: Please rename to GetCatalogDir.  Not sure we need the ModuleIntegration part, maybe GetModuleCatalogDir

pfmackin (approved):

PR #6803, opened 2023-08-14
VZ-6812: Helper functions for converting the VPO Reconciler to v1beta1

marcogelle:
This PR introduces helper functions that use Verrazzano version `v1beta1` to talk to the Kubernetes API server.

These functions will be used to replace instances of talking to the Kubernetes API server in the VPO's Verrazzano controller. These allow the internal VPO code to still make use of `v1alpha1`.

Unit tests were also written for these new functions.

The actual usage of these functions will come in a later PR.


bobdonat:
Code looks fine to me.  The question I have is why do we need conversion routines for the VPO reconciler?  Can we just use v1beta1

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package v1alpha1
```
anders-swanson, line 4: I'd put this in pkg/k8s/verrazzano

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+
+	"github.com/verrazzano/verrazzano/platform-operator/apis/verrazzano/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetVerrazzanoV1Alpha1 returns a v1alpha1 Verrazzano struct.
+// This function internally uses v1beta1 Verrazzano to ask Kubernetes API server for the VZ resource.
+func GetVerrazzanoV1Alpha1(ctx context.Context, client client.Client, name types.NamespacedName) (*Verrazzano, error) {
```
anders-swanson, line 16: If in pkg/k8s/verrazzano, would rename functions like GetV1Alpha1(), UpdateV1Alpha1(), and so forth

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if err := client.Update(ctx, vzV1Beta1); err != nil {
+		return err
+	}
+	return nil
+}
+
+// UpdateVerrazzanoV1Alpha1Status takes in a v1alpha1 Verrazzano struct and sends an update request to the K8s API server
+// for that resource's status. This function internally converts the Verrazzano to v1beta1 before sending the update request.
+func UpdateVerrazzanoV1Alpha1Status(ctx context.Context, statusWriter client.StatusWriter, vzV1Alpha1 *Verrazzano) error {
```
anders-swanson, line 71: Where is this function used?

marcogelle: These are the places I was intending so far:
- https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/healthcheck/status_updater.go#L106
- https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/reconcile/controller.go#L827
- https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/reconcile/status.go#L353

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
marcogelle: Sounds good

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
marcogelle: Sounds good

marcogelle:
> Code looks fine to me. The question I have is why do we need conversion routines for the VPO reconciler? Can we just use v1beta1

To my understanding, changing the VPO reconciler to use the newer Verrazzano version serves the purpose of getting rid of the "Verrazzano is deprecated" logs that comes from the internal use of `v1alpha1`. The cause of these logs is when we use the deprecated version specifically when talking to the Kubernetes API server.

So, the purpose of these conversions is to allow us to still use `v1alpha1` internally while avoiding the deprecation logs. The way this benefits us, over just using `v1beta1` directly, is that it makes it easier to replace instances of `v1beta1` with an even newer version down the line. With these conversions, we would only have to replace the few instances of `v1beta1` inside these helper functions.

platform-operator/apis/verrazzano/v1alpha1/talk_to_k8s_api.go
anders-swanson: Can you check with Paul why he did not use the status updater in those places?

marcogelle: Sure

marcogelle: I think for now, it makes sense to keep this function to accommodate for https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/healthcheck/status_updater.go#L106.

Knowing the usage of the status updater vs calling `Update` directly I think is controller-specific, and can be discussed when changing the controller itself.

bobdonat (approved):

anders-swanson (approved):

PR #6810, opened 2023-08-15
VZ-1045 pre-install must requeue if error

pfmackin:
Fix the module component handler to always requeue if there is a pre-install error.  This was causing Jaeger install to fail.

### DIFF
--- a/platform-operator/experimental/controllers/module/component-handler/installupdate/install_update_handler.go
+++ b/platform-operator/experimental/controllers/module/component-handler/installupdate/install_update_handler.go
@@ -94,8 +94,10 @@ func (h ComponentHandler) PreWork(ctx handlerspi.HandlerContext) result.Result {
 	}
 
 	// Do the pre-install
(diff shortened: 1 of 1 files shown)


mbenmclean (approved):

platform-operator/experimental/controllers/module/component-handler/installupdate/install_update_handler.go
```diff
@@ -94,8 +94,10 @@ func (h ComponentHandler) PreWork(ctx handlerspi.HandlerContext) result.Result {
 	}
 
 	// Do the pre-install
-	if err := comp.PreInstall(compCtx); err != nil && !vzerrors.IsRetryableError(err) {
-		h.updateReadyConditionStartedOrFailed(ctx, err.Error(), true)
+	if err := comp.PreInstall(compCtx); err != nil {
```
prasad-shirodkar, line 97: Does the `DoWork` method needs to be fixed as well?

PR #6815, opened 2023-08-16
VZ-10544 use module watches to re-reconcile when dependent modules are ready

pfmackin:
Add module watches so that a module (component) can watch other modules and re-reconcile if the watch triggers by going to a ready state.  This allows the following scenario, where cert-manager has a FluentBitConfig resource in the chart, but only created if the FluentBitConfig exists.

1. Install cert-manager only with none profile.  No FluentBitConfig is created because the CRD does not exist
2. Update vz CR to enable fluentoperator
3. Once fluentoperator module is ready, it triggers a cert-manager re-reconcile and cert-manager creates a FluentBitConfig since the CRD exists.

Also remove the calls to create an integration event since the module watchers replaces this functionality


platform-operator/controllers/verrazzano/component/common/watch/watch.go
```diff
@@ -0,0 +1,60 @@
...
+		FuncShouldReconcile: func(cli client.Client, wev controllerspi.WatchEvent) bool {
+			// Return false if this is not one of the modules that is being watched
+			var newModule = wev.NewWatchedObject.(*moduleapi.Module)
+			_, ok := moduleNameSet[newModule.Spec.ModuleName]
+			if !ok {
+				return false
+			}
+
+			// Get new module Ready condition and return false if not ready
+			newCond := status.GetReadyCondition(newModule)
```
mbenmclean, line 33: I'm working on a Module Ready function that checks the ready condition, that the generations match, and that the versions match. We should probably call that to check if it's ready. You can leave this for now and I'll update as part of my PR

mbenmclean (approved):

platform-operator/controllers/verrazzano/component/common/watch/watch.go
pfmackin: ok

PR #6824, opened 2023-08-18
VZ-10494: Add annotation to authrpoxy deployment to automatically pick configuration

karan56625:
Add annotation to authrpoxy deployment to automatically pick AuthProxy config.

### DIFF
--- a/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
+++ b/platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
@@ -51,6 +51,7 @@ spec:
         {{ if .Values.managedClusterRegistered }}
         verrazzano.io/managedClusterRegistered: "true"
         {{ end }}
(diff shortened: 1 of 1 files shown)


saketmht (approved):

platform-operator/helm_config/charts/verrazzano-authproxy/templates/verrazzano-authproxy.yaml
```diff
@@ -51,6 +51,7 @@ spec:
         {{ if .Values.managedClusterRegistered }}
         verrazzano.io/managedClusterRegistered: "true"
         {{ end }}
+        verrazzano.io/cmChecksum: {{ include (print $.Template.BasePath "/verrazzano-authproxy-configmap.yaml") . | sha256sum }}
```
mbenmclean, line 54: Will this result in the auth proxy pod bouncing every time there's an upgrade even if there's no change to cm? Can you test to make sure that the auth proxy pow only bounced when the cm actually changes please.

karan56625: Nope. Only if there is any change in AuthProxy config. I have tested this scenario.

PR #6827, opened 2023-08-18
VZ-10273 If CloudCredential is updated, cluster state is not updated with new credential

knelson7:


platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -72,6 +76,28 @@ type Reconciler struct {
 	StatusUpdater     vzstatus.Updater
 }
 
+// Struct to unmarshall the ocne clusters list
+type ocneCluster struct {
```
anders-swanson, line 80: This isn't an OCNE Cluster, it's a Rancher management cluster

platform-operator/controllers/verrazzano/reconcile/controller.go
anders-swanson, line 1: Can we move this code to a new package, the watches should be invoked from the controller, but 99% of the code should live in another package.

anders-swanson: ideally I would like to see code similar to this in the controller, only:
```
if r.WatchCloudCredentials {
  cc.Watch(...)
}
```

anders-swanson: another package or another file

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -966,6 +996,129 @@ func (r *Reconciler) watchRancherGlobalDataNamespace(namespace string, name stri
...
+				return watchCloudCredForUpdate(e, r)
+			},
+		},
+	)
+}
+
+func watchCloudCredForUpdate(e event.UpdateEvent, r *Reconciler) bool {
+	if r.isOCNECloudCredential(e.ObjectNew) {
+		// IS THIS RIGHT? OR SHOULD IT BE KEYCLOAK? KEYCLOAK AND RANCHER?
+		r.AddWatch(rancher.ComponentJSONName)
```
anders-swanson, line 1015: We don't need to add a component watch here, component watches let us reconcile components.

platform-operator/controllers/verrazzano/reconcile/controller.go
```diff
@@ -966,6 +996,129 @@ func (r *Reconciler) watchRancherGlobalDataNamespace(namespace string, name stri
...
+		//get ocne clusters and then update copy of cloud credential if necessary
+		dynClient, _ := getDynamicClient()
```
anders-swanson, line 1017: you need to handle the error here

anders-swanson: If a dynamic client is required, suggest creating and passing it in earlier. This way it is available for the lifetime of the process

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
```
anders-swanson, line 1: (c) 2023 since it's a new file

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,151 @@
...
+		GenericEngineConfig struct {
+			CloudCredentialID string `json:"cloudCredentialId"`
+		} `json:"genericEngineConfig"`
+	} `json:"spec"`
+}
+
+func (r *VerrazzanoSecretsReconciler) watchCloudCredForUpdate(updatedSecret *corev1.Secret) error {
+	if isOCNECloudCredential(updatedSecret) {
+		//get ocne clusters and then update copy of cloud credential if necessary
+		dynClient, err := getDynamicClient()
```
anders-swanson, line 46: Please inject dynamic client into the VerrazzanoSecretsReconciler, rather than recreating a client during each cycle

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,151 @@
...
+		Name string `json:"name"`
+	} `json:"metadata"`
+	Spec struct {
+		GenericEngineConfig struct {
+			CloudCredentialID string `json:"cloudCredentialId"`
+		} `json:"genericEngineConfig"`
+	} `json:"spec"`
+}
+
+func (r *VerrazzanoSecretsReconciler) watchCloudCredForUpdate(updatedSecret *corev1.Secret) error {
```
anders-swanson, line 43: not sure if this method name is accurate given what the method is doing

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,151 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// createOrUpdateCAPISecret updates CAPI based on the updated credentials
+func (r *VerrazzanoSecretsReconciler) updateCAPISecret(updatedSecret *corev1.Secret, clusterCredential *corev1.Secret) error {
+	data := map[string][]byte{
+		ociTenancyField:              updatedSecret.Data[ociTenancyField],
```
anders-swanson, line 60: This is the OCI CAPI secret, right? The fields used by OCI CAPI are not the same as the Rancher CC secret fields.

platform-operator/controllers/secrets/cloud_credential.go
mgianatagh: this still looks unresolved to me.  The "2022," should be removed if this is a new file.

// Copyright (c) 2023, Oracle and/or its affiliates.

platform-operator/controllers/secrets/cloud_credential.go
anders-swanson: These are the constants for OCI CAPI:
```
const (
	ociTenancyField              = "tenancy"
	ociUserField                 = "user"
	ociFingerprintField          = "fingerprint"
	ociRegionField               = "region"
	ociPassphraseField           = "passphrase"
	ociKeyField                  = "key"
	ociUseInstancePrincipalField = "useInstancePrincipal"
)
```

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,151 @@
...
+			return err
+		}
+		// if the cluster is an OCNE cluster
+		if rancherMgmtCluster.Spec.GenericEngineConfig.CloudCredentialID != "" {
+			// extract cloud credential name from CloudCredentialID field
+			cloudCredentialStringSplit := strings.Split(rancherMgmtCluster.Spec.GenericEngineConfig.CloudCredentialID, ":")
+			// if cloud credential name matches updatedSecret name, get and update the cc copy held by the cluster
+			if cloudCredentialStringSplit[1] == updatedSecret.Name {
+				secretName := fmt.Sprintf("%s-principal", rancherMgmtCluster.Metadata.Name)
+				clusterCredential := &corev1.Secret{}
```
anders-swanson, line 97: Can we call this capiCredential

anders-swanson (commented):
See comments

platform-operator/controllers/secrets/cloud_credential.go
knelson7: is there a rancher CC secret field that's equivalent to ociUseInstancePrincipalField? I don't see one

anders-swanson: No, all rancher ccs are user principle.

anders-swanson (dismissed):

platform-operator/controllers/secrets/cloud_credential.go
```diff
@@ -0,0 +1,134 @@
...
+			// if the cluster is an OCNE cluster
+			if rancherMgmtCluster.Spec.GenericEngineConfig.CloudCredentialID != "" {
+				// extract cloud credential name from CloudCredentialID field
+				capiCredential := strings.Split(rancherMgmtCluster.Spec.GenericEngineConfig.CloudCredentialID, ":")
+				// if cloud credential name matches updatedSecret name, get and update the cc copy held by the cluster
+				if len(capiCredential) >= 2 {
+					if capiCredential[1] == updatedSecret.Name {
+						secretName := fmt.Sprintf("%s-principal", rancherMgmtCluster.Metadata.Name)
+						clusterCredential := &corev1.Secret{}
+						if err = r.Client.Get(context.TODO(), client.ObjectKey{Namespace: rancherMgmtCluster.Metadata.Name, Name: secretName}, clusterCredential); err != nil {
```
anders-swanson, line 90: suggest to use CreateOrUpdate rather than Get, Update. Of course, this works too

anders-swanson (approved):

platform-operator/controllers/secrets/secrets_controller.go
```diff
@@ -88,6 +91,20 @@ func (r *VerrazzanoSecretsReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return r.reconcileVerrazzanoCABundleCopies()
 	}
 
+	// updates capi credential if an ocne cloud credential has been updated
+	if config.Get().CloudCredentialWatchEnabled {
```
anders-swanson, line 95: Suggest to follow the existing idiom of checking the secret type, and conditionally doing something. Your code does this, it's just not styled in the same pattern.

PR #6833, opened 2023-08-21
VZ-10576: ingress-nginx-controller image size too large

Raaghavcodes:
Ingress-nginx-controller img tag bom update

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -28,7 +28,7 @@
           "images": [
             {
               "image": "nginx-ingress-controller",
-              "tag": "v1.7.1-20230712031907-3ae8a8da4",
+              "tag": "v1.7.1-20230817215350-c8315d559",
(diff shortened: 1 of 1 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -28,7 +28,7 @@
           "images": [
             {
               "image": "nginx-ingress-controller",
-              "tag": "v1.7.1-20230712031907-3ae8a8da4",
+              "tag": "v1.7.1-20230817215350-c8315d559",
```
bhatpmk, line 31: There are couple of more places in this file, you have to update the image.
The BFS pipeline builds 2 more images - "nginx-ingress-default-backend" and "kube-webhook-certgen". Please check whether the optimization can be done for those images as well. Also if there are no changes to "nginx-ingress-default-backend" and "kube-webhook-certgen" also, it is better if you update those images built with the latest commit.

bobdonat (approved):

PR #6855, opened 2023-08-24
vz 10235 empty dir mysql returns error

AbeHern:
Catching error when emptyDir is used as a DefaultVolumeSource.


platform-operator/apis/verrazzano/validators/validate.go
```diff
@@ -454,3 +454,14 @@ func VerifyPlatformOperatorSingleton(runtimeClient client.Client) error {
 	}
 	return nil
 }
+
+// ValidateMySQL Verifies that the defaultVolumeSource emptyDir is not left blank. Verrazzano does not support this behavior.
+func ValidateMySQL(defaultVolumeSource *corev1.VolumeSource) error {
```
anders-swanson, line 459: Can you rename this function, `ValidateDefaultVolumeSource` seems more appropriate.

platform-operator/controllers/verrazzano/component/mysql/mysql_component.go
```diff
@@ -155,6 +156,10 @@ func (c mysqlComponent) ValidateUpdate(old *vzapi.Verrazzano, new *vzapi.Verrazz
 	if err := common.CompareInstallArgs(c.getInstallArgs(old), c.getInstallArgs(new)); err != nil {
 		return fmt.Errorf("Updates to mysqlInstallArgs not allowed for %s", ComponentJSONName)
 	}
+	// Validate the DefaultVolumeSource of mySQL
+	if err := validators.ValidateMySQL(convertedNewVZ.Spec.DefaultVolumeSource); err != nil {
```
anders-swanson, line 160: You need to make sure this is called by the v1beta1 validator as well. Because we have two supported API versions, there's two different validations.

platform-operator/apis/verrazzano/validators/validate.go
```diff
@@ -454,3 +454,14 @@ func VerifyPlatformOperatorSingleton(runtimeClient client.Client) error {
...
+	if defaultVolumeSource == nil {
+		return nil
+	}
+	if defaultVolumeSource.EmptyDir != nil {
```
mbenmclean, line 463: Check to make sure mysql is enabled too

platform-operator/controllers/verrazzano/component/mysql/mysql_component.go
AbeHern: Your question confuses me since the definition of the function is in `platform-operator/controllers/verrazzano/component/mysql/mysql_component.go` and not in either v1alpha1 or v1beta1

platform-operator/apis/verrazzano/validators/validate.go
AbeHern: Adding the check now

mbenmclean (approved):

PR #6860, opened 2023-08-24
VZ-10616.  New controller must not reconcile if upgrade is required but not active

michael-cico:
Short-circuit reconcile of new controller if an upgrade is required but not yet started.  This is the case where

* The VPO has been updated
* The VZ CR spec has not been set to the version in the updated BOM

An upgrade is deemed required when
* the Spec.Version field is NOT empty and is less than the BOM version
* the Spec.Version field is empty, and the Status.Version is less than the BOM version

### DIFF
--- a/platform-operator/experimental/controllers/verrazzano/reconciler.go
+++ b/platform-operator/experimental/controllers/verrazzano/reconciler.go
@@ -132,11 +132,11 @@ func (r Reconciler) mutateModule(log vzlog.VerrazzanoLogger, effectiveCR *vzapi.
 	return r.setModuleValues(log, effectiveCR, module, comp)
 }
 
-// isUpgradeRequired Returns true if the Spec version is set but doesn't match what's in the BOM.
-//   - if spec version is empty then it's either an initial install or an update within the same version
-//   - if the Spec version IS NOT empty and matches the BOM, we're either in an upgrade or updating within the same version
-//   - if the Spec version IS empty and does NOT match the BOM, then we have updated the VPO but not yet upgraded, so we should
-//     skip reconciling
+// isUpgradeRequired Returns true if we detect that an upgrade is required but not (at least) in progress:
+//   - if the Spec version IS NOT empty is less than the BOM version, an upgrade is required
+//   - if the Spec version IS empty the Status version is less than the BOM, then an upgrade is required (upgrade of initial install scenario)
+//
+// If we return true here, it means we should stop reconciling until an upgrade has been requested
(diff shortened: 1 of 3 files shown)


platform-operator/experimental/controllers/verrazzano/reconciler.go
```diff
@@ -120,3 +130,24 @@ func (r Reconciler) mutateModule(log vzlog.VerrazzanoLogger, effectiveCR *vzapi.
...
+//     skip reconciling
+func (r Reconciler) isUpgradePending(actualCR *vzapi.Verrazzano) (bool, error) {
+	if actualCR == nil {
+		return false, fmt.Errorf("no Verrazzano CR provided")
+	}
+	newBom, err := bom.NewBom(config.GetDefaultBOMFilePath())
+	if err != nil {
+		return false, err
+	}
+	bomVersion := strings.TrimSpace(newBom.GetVersion())
```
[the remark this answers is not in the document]
michael-cico: I thought about that, but not sure it was needed.  But I'll look into it.

pfmackin (approved):

platform-operator/experimental/controllers/verrazzano/reconciler.go
```diff
@@ -120,3 +131,35 @@ func (r Reconciler) mutateModule(log vzlog.VerrazzanoLogger, effectiveCR *vzapi.
...
+	if bomVersion, err = semver.NewSemVersion(newBom.GetVersion()); err != nil {
+		return false, err
+	}
+	if len(actualCR.Spec.Version) > 0 {
+		if specVersion, err = semver.NewSemVersion(actualCR.Spec.Version); err != nil {
+			return false, err
+		}
+		return specVersion.IsLessThan(bomVersion), nil
+	}
+	if len(actualCR.Status.Version) > 0 {
```
pfmackin, line 158: This code should not determine if upgrade is required, that is done in the module reonciler.  This code just decides if it reconcile should be allowed.  Trying to understand why we need to look at the status version

mbenmclean (approved):

pfmackin (approved):

platform-operator/experimental/controllers/verrazzano/reconciler.go
michael-cico: It’s the case where we’re upgrading an existing install for the first time, where the version field wasn’t set yet.

Subsequent upgrades should have the spec version set.

PR #6870, opened 2023-08-25
VZ-10647 create boilerplate files for the Authproxy

davisgomes:
**Changes**
- Add new directory for the Authproxy
- Create boilerplate files for the initial setup of the Authproxy

### DIFF
--- a/authproxy/Makefile
+++ b/authproxy/Makefile
@@ -1,0 +1,140 @@
+# Copyright (C) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+include ../make/quality.mk
+include ../make/generate.mk
+include ../make/retry.mk
+
+SCRIPT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))/../build
+TOOLS_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))/../tools
+
+NAME:=verrazzano-authproxy
+REPO_NAME:=verrazzano-authproxy
+
+ifdef KUBECONFIG
+  KUBECONFIG ?= ${KUBECONFIG}
+else
+  KUBECONFIG ?= ${HOME}/.kube/config
+endif
+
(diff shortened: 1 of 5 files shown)


authproxy/Makefile
```diff
@@ -0,0 +1,140 @@
...
+GO_LDFLAGS ?= -extldflags -static -X main.buildVersion=${BUILDVERSION} -X main.buildDate=${BUILDDATE}
+
+.PHONY: build
+build: go-fmt go-vet
+	go build -o bin/manager main.go
+
+# Run against the configured Kubernetes cluster in ~/.kube/config
+.PHONY: run
+run:
+	# To enable webhooks: $(GO) run main.go --kubeconfig=${KUBECONFIG} --enable-webhooks=true --metrics-addr=localhost:0 --cert-dir=build/webhook-certs
```
fredtibbitts, line 56: Since Auth Proxy doesn't have webhooks this should be removed.

authproxy/Makefile
```diff
@@ -0,0 +1,140 @@
...
+	RELEASE_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
+endif
+
+DIST_DIR:=dist
+GO ?= CGO_ENABLED=0 GO111MODULE=on GOPRIVATE=github.com/verrazzano go
+GO_LDFLAGS ?= -extldflags -static -X main.buildVersion=${BUILDVERSION} -X main.buildDate=${BUILDDATE}
+
+.PHONY: build
+build: go-fmt go-vet
+	go build -o bin/manager main.go
```
fredtibbitts, line 51: I know this was just copied as all of the other Makefiles have this, but I wonder what this is even used for? There's a `go-build` target that seems to be more robust, I wonder if we use this anywhere?

fredtibbitts (approved):

authproxy/Makefile
davisgomes: I'm not sure. this does the go fmt check and the go-vet (whatever that means). Not sure if it's totally necessary.

authproxy/Makefile
davisgomes: Thanks. I missed this one.

PR #6872, opened 2023-08-25
VZ-10404 module upgrade pre post controller sync

pfmackin:
This is Ben's branch with unit tests fixes.
From Ben:
This change creates global atomics that synchronize the new experimental verrazzano controller and the legacy verrazzano controller to facilitate module install and upgrade.


platform-operator/controllers/verrazzano/reconcile/status.go
```diff
@@ -366,3 +369,35 @@ func (r *Reconciler) setUninstallCondition(log vzlog.VerrazzanoLogger, vz *insta
 	}
 	return r.updateStatus(log, vz, msg, newCondition, nil)
 }
+
+func (r *Reconciler) modulesReady(ctx spi.ComponentContext) (bool, error) {
```
michael-cico, line 373: even though it's (hopefully) temporary, maybe add a comment here about what it's doing.  Next pass.

platform-operator/experimental/catalog/catalog.yaml
```diff
@@ -35,9 +35,9 @@ modules:
   - name: verrazzano-application-operator
     version: 2.0.0
   - name: weblogic-operator
-    version: 4.1.0
+    version: 4.1.0-1
```
michael-cico, line 38: why are we bumping these here?

platform-operator/experimental/controllers/module/component-handler/migration/migration.go
```diff
@@ -62,5 +62,5 @@ func (h migrationHandler) UpdateStatusIfAlreadyInstalled(ctx handlerspi.HandlerC
 	}
 
 	// Set the module status condition, installed generation and installed version
-	return modulestatus.UpdateModuleStatusToInstalled(ctx, module, vzcr.Status.Version, compStatus.LastReconciledGeneration)
+	return modulestatus.UpdateModuleStatusToInstalled(ctx, module, vzcr.Status.Version, 0)
```
michael-cico, line 65: So we want to force a reconcile here?

michael-cico (dismissed):

platform-operator/controllers/verrazzano/reconcile/status.go
pfmackin: fixed

platform-operator/experimental/controllers/module/component-handler/migration/migration.go
pfmackin: fixed

michael-cico (approved):

PR #6874, opened 2023-08-25
VZ-10626 add the authproxy v2 container to the deployment

davisgomes:


platform-operator/controllers/verrazzano/component/appoper/app_operator.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2021, 2023, Oracle and/or its affiliates.
```
desagar, line 1: no changes but copyright updated?

platform-operator/controllers/verrazzano/component/authproxy/authproxy_test.go
```diff
@@ -153,6 +153,15 @@ func TestAppendOverrides(t *testing.T) {
 	defer func() {
 		config.SetDefaultBomFilePath("")
 	}()
+
+	testImageVar := "test-image"
+	err := os.Setenv(vpoconst.VerrazzanoAuthProxyImageEnvVar, testImageVar)
```
desagar, line 158: is this known to work? If the tests run in parallel in the same process, it might not.

platform-operator/helm_config/charts/verrazzano-authproxy/values.yaml
```diff
@@ -40,3 +40,7 @@ config:
 dns:
   wildcard:
     domain:
+
+v2:
+  image:
```
desagar, line 45: is the empty override for image needed?

desagar: oh this is values.yaml. Sorry thought it was the overrides file for install. Please ignore.

platform-operator/controllers/verrazzano/component/authproxy/authproxy_test.go
davisgomes: We do for other tests, and if it is not set, it will not be checked because I check the size of the key values before evaluating.

platform-operator/controllers/verrazzano/component/appoper/app_operator.go
davisgomes: Yea there were changes, and the copyright flags them even if they are backed out.

authproxy/src/proxy/proxy.go
```diff
@@ -218,10 +232,7 @@ func obfuscateRequestData(req *http.Request) *http.Request {
 	hiddenReq := req.Clone(context.TODO())
 	authKey := "Authorization"
 	for i := range hiddenReq.Header[authKey] {
-		// List of authorization schemes compiled from
-		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes
-		authRegEx := "(Bearer|Basic|Digest|HOBA|Mutual|Negotiate|NTLM|VAPID|SCRAM|AWS4-HMAC-SHA256)"
-		hiddenReq.Header[authKey][i] = vzpassword.MaskFunction(authRegEx)(hiddenReq.Header[authKey][i])
+		hiddenReq.Header[authKey][i] = vzpassword.MaskFunction("")(hiddenReq.Header[authKey][i])
```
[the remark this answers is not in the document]
davisgomes: It does split it by word nicely, so that is one benefit. I'm not sure that it makes too big a difference.

fredtibbitts:
Do we also need to update network policies? For the ingress?

davisgomes:
> Do we also need to update network policies? For the ingress?

Yes, but I'm deferring the Ingress setup to another subtask, so I did not spend too much time on it for this.

fredtibbitts (approved):

desagar (approved):

PR #6877, opened 2023-08-26
VZ-10627 Create a local API proxy

davisgomes:
**Changes**
- Implement a local API proxy that will forward API requests with the leading path `/clusters/local`

**Testing**
- build image and use as a deployment in a Kubernetes cluster
- execute localhost requests on the pod with a bearer token and make sure they get properly forwarded to the API server

### DIFF
--- a/authproxy/Makefile
+++ b/authproxy/Makefile
@@ -42,6 +42,8 @@ ifndef RELEASE_BRANCH
 	RELEASE_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 endif
 
+VZ_BASE_IMAGE ?= ghcr.io/verrazzano/verrazzano-base:v1.0.0-20230327155846-4653b27
+
 DIST_DIR:=dist
 GO ?= CGO_ENABLED=0 GO111MODULE=on GOPRIVATE=github.com/verrazzano go
 GO_LDFLAGS ?= -extldflags -static -X main.buildVersion=${BUILDVERSION} -X main.buildDate=${BUILDDATE}
--- a/authproxy/src/proxy/proxy.go
+++ b/authproxy/src/proxy/proxy.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 )
@@ -28,6 +29,7 @@ type AuthProxy struct {
 }
 
 type Handler struct {
+	URL    string
 	Client *http.Client
 	Log    *zap.SugaredLogger
 }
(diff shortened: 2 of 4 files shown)


authproxy/Makefile
```diff
@@ -42,6 +42,8 @@ ifndef RELEASE_BRANCH
 	RELEASE_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 endif
 
+VZ_BASE_IMAGE ?= ghcr.io/verrazzano/verrazzano-base:v1.0.0-20230327155846-4653b27
```
desagar, line 45: where is this used?

authproxy/src/proxy/proxy.go
```diff
@@ -0,0 +1,147 @@
...
+	Log    *zap.SugaredLogger
+}
+
+var _ http.Handler = Handler{}
+
+// InitializeProxy returns a configured AuthProxy instance
+func InitializeProxy() *AuthProxy {
+	return &AuthProxy{
+		Server: http.Server{
+			Addr:         ":8777",
```
desagar, line 46: should we make this configurable in the helm chart?

authproxy/src/proxy/proxy_test.go
```diff
@@ -0,0 +1,111 @@
...
+	r := httptest.NewRequest(http.MethodPost, url, strings.NewReader(testBody))
+	w := httptest.NewRecorder()
+	handler.ServeHTTP(w, r)
+}
+
+// TestReformatAPIRequest tests the reformatting of the request to be sent to the API server
+// GIVEN a request to the Auth proxy server
+// WHEN  the request is formatted correctly
+// THEN  the request is properly formatted to be sent to the API server
+func TestReformatAPIRequest(t *testing.T) {
```
desagar, line 70: we should probably add some non-happy path tests here.

desagar (dismissed):

authproxy/src/proxy/proxy.go
davisgomes: Yes. I can patch this over with the component Helm chart changes that should go in soon.

authproxy/src/proxy/proxy_test.go
davisgomes: I left those out because I was assuming that the validation function would filter out all "non-happy path" malformed requests, but it couldn't hurt to add some.

authproxy/src/proxy/proxy.go
```diff
@@ -0,0 +1,147 @@
...
+			Timeout:   time.Minute,
+			Transport: transport,
+		},
+		Log: log,
+	}
+	return nil
+}
+
+// ServeHTTP accepts an incoming server request and forwards it to the Kubernetes API server
+func (h Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
```
desagar, line 78: This should probably be called ServeRequest because it processes a single request. The name makes it sound like this is starting the http server, which is misleading.

fredtibbitts (approved):

authproxy/src/proxy/proxy.go
desagar: I'm not sure if we should expose it via helm or not. We should check if we do that for other components.

authproxy/src/proxy/proxy.go
davisgomes: It implements an interface that uses the serveHTTP function, so it has to be named this way.

fredtibbitts (approved):

PR #6915, opened 2023-08-30
VZ-10788: Add OIDC config to Auth Proxy v2

fredtibbitts:
This PR adds OIDC configuration to Auth Proxy v2. Configuration is stored in a secret and mounted into the Auth Proxy pod. A watcher checks for updates to the secret data and automatically reloads the values when changes are detected.

The configuration currently allows for an issuer URL and a client id. We can extend this as needed.

### DIFF
--- a/authproxy/src/config/config.go
+++ b/authproxy/src/config/config.go
@@ -1,0 +1,147 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package config
+
+import (
+	"os"
+	"sync"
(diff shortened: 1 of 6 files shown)


authproxy/src/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	issuerURL = value
+	issuerURLFileModTime = *modTime
+	return nil
+}
+
+// loadClientID loads the client ID from a file and stores the file modification time
+func loadClientID() error {
+	mutex.Lock()
```
desagar, line 65: should we be doing this lock in the loadConfigValue function instead?

desagar (approved):

davisgomes (approved):

PR #6921, opened 2023-08-30
Add File To Bug Report Output To Indicate Ca.Crt Issues

George-Aeillo:
In this PR, I edited the analyzeCertificates function to now look at the ca.crt field in the referenced secret from the certificate object, if it exists. For each namespace, if a ca.crt field exists in any of the referenced secrets in that namespace, a caCrtInfo.json file is created for that namespace that lists the referenced secrets and if the ca.crt field is expired. The analysis tooling is also edited to detect this file and parses it. The analysis tool reports an issue if the ca.crt field is expired.


tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -46,6 +49,11 @@ var isVerbose bool
 var multiWriterOut io.Writer
 var multiWriterErr io.Writer
 
+type CaCrtInfo struct {
+	NameofSecret string `json:"nameofSecret"`
```
michael-cico, line 53: Maybe should be name of certificate?  Or just Name?

George-Aeillo: Ok!

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -308,6 +316,52 @@ func captureCertificates(client clipkg.Client, namespace, captureDir string, vzH
...
+		captureCaCrtExpirationInfo(client, certificateList, namespace, captureDir, vzHelper)
+	}
+	return nil
+}
+
+// captureCaCrtExpirationInfo is a helper function of the captureCertificates function that loops through the certificates in a particular namespace and outputs a file containing information about the ca.crt of each certificate
+func captureCaCrtExpirationInfo(client clipkg.Client, certificateList v1.CertificateList, namespace string, captureDir string, vzHelper VZHelper) error {
+	caCrtList := []CaCrtInfo{}
+	for _, cert := range certificateList.Items {
+		correspondingSecretName := cert.Spec.SecretName
```
michael-cico, line 328: nit-pick, for readability might be good to wrap the code that decodes and checks the cert into a method like `isCACertExpired(cert)`.  It also reduces the complexity of this method.

So the body of this would be

```
if isCACertExpired(cert) {
   // append to list of expired certs
}
```

George-Aeillo: Thanks! I have made these changes.

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -796,3 +829,35 @@ func removePods(podList []corev1.Pod, pods []string) []corev1.Pod {
...
+		return nil, false, fmt.Errorf("Failure to PEM Decode Certificate")
+	}
+	certificate, err := x509.ParseCertificate(caCrtDataPemDecoded.Bytes)
+	if err != nil {
+		return nil, false, err
+	}
+	caCrtInfoForCert := CaCrtInfo{Name: correspondingSecretName, Expired: false}
+	expirationDateOfCert := certificate.NotAfter
+
+	if time.Now().Unix() > expirationDateOfCert.Unix() {
```
michael-cico, line 858: not for this pass, but I wonder if it would also be helpful to identify if a cert is expiring soon (say, within a month).

Perhaps something to consider for another enh... maybe around the leaf cert checks.

michael-cico (approved):

tools/vz/pkg/helpers/vzcapture.go
George-Aeillo: Ok! Thanks!

michael-cico (approved):

PR #6922, opened 2023-08-30
VZ-10404 Delete CAPI RBAC resources before upgrade

mbenmclean:
This changes fixes a problem where Rancher finalizer will prevent CAPI RBAC resources from deleting before they get recreated during upgrade. This had the effect of the CAPI RBAC resources that got upgraded being permanently removed from the cluster during upgrade.

### DIFF
--- a/platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
+++ b/platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
@@ -7,16 +7,22 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"os"
-	"strings"
-	"text/template"
-
+	"github.com/verrazzano/verrazzano/pkg/constants"
 	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
 	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/spi"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"os"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/api/v1alpha3"
 	clusterapi "sigs.k8s.io/cluster-api/cmd/clusterctl/client"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+	"text/template"
 )
+
+const rbacGroup = "rbac.authorization.k8s.io"
 
 const clusterctlYamlTemplate = `
 {{- if .IncludeImagesHeader }}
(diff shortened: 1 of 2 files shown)


platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -316,6 +316,19 @@ func (c clusterAPIComponent) Upgrade(ctx spi.ComponentContext) error {
...
+		// get all the resource that will be deleted and recreated
+		components, err := getComponentsToUpgrade(capiClient, applyUpgradeOptions)
+		if err != nil {
+			ctx.Log().Errorf("Error generating CAPI components to be upgraded")
+			return err
+		}
+
+		// delete the RBAC resources that Rancher puts finalizers on and keep requeuing until they're gone
+		if err = deleteRBACComponents(ctx, components); err != nil {
+			return err
```
michael-cico, line 328: is this going to spit out a lot of messages/stack-traces in the log?  I think the `RetryableError` may be of help here, or @pfmackin also created some helpers I think along those lines.

pfmackin: yes, i would use a log.Once here

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -316,6 +316,19 @@ func (c clusterAPIComponent) Upgrade(ctx spi.ComponentContext) error {
 		return err
 	}
 	if isUpgradeOptionsNotEmpty(applyUpgradeOptions) {
+		// get all the resource that will be deleted and recreated
+		components, err := getComponentsToUpgrade(capiClient, applyUpgradeOptions)
+		if err != nil {
+			ctx.Log().Errorf("Error generating CAPI components to be upgraded")
```
pfmackin, line 322: can you use throttledError

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
```diff
@@ -257,3 +265,72 @@ func isUpgradeOptionsNotEmpty(upgradeOptions clusterapi.ApplyUpgradeOptions) boo
...
+	}
+
+	for i := range components {
+		component := components[i]
+		if component.GroupVersionKind().Group == rbacGroup {
+			err := ctx.Client().Get(context.TODO(),
+				types.NamespacedName{Name: component.GetName(), Namespace: component.GetNamespace()}, &component)
+			if errors.IsNotFound(err) {
+				continue
+			} else if err != nil {
```
michael-cico, line 326: nit, since each branch of this conditional returns or continues, they can just be separate if's... I think that's more in line with Go conventions:

```
if errors.IsNotFound(...) {
  continue
}
if err != nil {
   return err
}
return fmt.Errorf("Waiting for...")
```

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
```diff
@@ -257,3 +265,72 @@ func isUpgradeOptionsNotEmpty(upgradeOptions clusterapi.ApplyUpgradeOptions) boo
...
+				ctx.Log().Errorf("Unexpected error getting %s %s: %v", component.GetKind(), component.GetName(), err)
+				return err
+			} else {
+				ctx.Log().Progress("Waiting for CAPI RBAC resources to be deleted before upgrade")
+				return err
```
michael-cico, line 331: `err` will be nil here right?  The goal is for this to return a non-nil err to requeue correct?

perhaps `return ctx.Log().ErrorfThrottledNewErr(...)` here.

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
mbenmclean: The logging is already done in the function call

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
mbenmclean: fixed

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi.go
```diff
@@ -257,3 +265,71 @@ func isUpgradeOptionsNotEmpty(upgradeOptions clusterapi.ApplyUpgradeOptions) boo
...
+			err := ctx.Client().Get(context.TODO(),
+				types.NamespacedName{Name: component.GetName(), Namespace: component.GetNamespace()}, &component)
+			if errors.IsNotFound(err) {
+				continue
+			}
+			if err != nil {
+				ctx.Log().Errorf("Unexpected error getting %s %s: %v", component.GetKind(), component.GetName(), err)
+				return err
+			}
+			return ctx.Log().ErrorfThrottledNewErr("Waiting for cluster-api RBAC resources to be deleted before upgrade")
```
pfmackin, line 331: this needs to be progress or info

mbenmclean: Ok I'll make progress

pfmackin (approved):

PR #6947, opened 2023-09-05
VZ-10825: Add opensearchOperator as a high level component

saketmht:
List of changes:

<h2>CRD Changes</h2>

- Added opensearchOperator as a high level component
- It will now managed the existing opensearch and opensearchDashboards component
- opensearchOperator component doesn't have any overrides

<h2>OpenSearchOperator Install</h2><h3>PreInstall</h3><ol><li>Create the <code>verrazzano-logging</code> namespace</li><li>Handle legacyOS<ol><li>Set existing PVs to retain if they exist</li><li>Disable exiting OS and OSD via VMI</li><li>Delete the master nodes PVCs as they are created by sts so not deleted by default</li><li>Create new PVCs and set them to claim the existing PVs</li><li>Check if the new PVCs are bound to the existing PVs</li></ol></li></ol><p>Note:- We are not maintaining any state for the upgrade/pv-migration. The functions were designed to be re-entrant so that they can be run again in-case of errors and continue from where it left. These functions will be invoked during fresh installs as well but will have no effect as there will be no exiting PVs</p><h3>Install</h3><ol><li>Install the opensearch-operator helm chart in <code>verrazzano-logging</code> namespace</li><li>This will create the opensearch cluster as well</li></ol><h3>PostInstall</h3><ol><li>Set the reclaim policy of the PVs to what it was before</li></ol><p><br /></p><h2>
<h2>AuthProxy Changes</h2><p>Since the OpenSearch service will be changing with the new OpenSearch, the backed service in the authproxy needs to changed as well. The new services are passed thorough value overrides in the helm. These values are newly added.</p>

```
  opensearch:
    protocol: https
    namespace: verrazzano-logging
    service: opensearch
    osdService: opensearch-dashboards 
```


cluster-operator/controllers/vmc/jaeger.go
```diff
@@ -70,21 +70,10 @@ func (r *VerrazzanoManagedClusterReconciler) getJaegerOpenSearchConfig(vzList *v
...
-	// If the Jaeger OpenSearch URL is the default operator based URL, use operator based OpenSearch ingress URL.
+	// If the Jaeger OpenSearch URL is the default URL, use VZ OpenSearch ingress URL.
 	// If the Jaeger OpenSearch URL  is not the default, meaning it is a custom OpenSearch, use the external OpenSearch URL.
 	if jsc.OSURL == vzconstants.DefaultJaegerOSURL {
-		jc.URL, err = r.getESURL(*vzList, vmiIngest)
-		if err != nil {
-			return jc, err
-		}
-		// Get the CA bundle needed to connect to the admin keycloak
-		jc.CA, err = r.getAdminCaBundle()
```
pfmackin, line 82: why did you remove this?

platform-operator/controllers/verrazzano/component/common/namespace.go
```diff
@@ -140,3 +141,11 @@ func CheckExistingNamespace(ns []corev1.Namespace, includeNamespace func(*corev1
 	}
 	return nil
 }
+
+func DoesNamespaceExists(client clipkg.Client, name string) bool {
+	ns := corev1.Namespace{}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: name}, &ns); err != nil {
+		return false
```
pfmackin, line 148: need to check for isNotFound error

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
```diff
@@ -466,6 +470,13 @@ type ComponentSpec struct {
 	WebLogicOperator *WebLogicOperatorComponent `json:"weblogicOperator,omitempty"`
 }
 
+// OpenSearchOperatorComponent specifies the OpenSearchOperator configuration
+type OpenSearchOperatorComponent struct {
+	// If true, then OpenSearchOperator will be installed
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
michael-cico, line 477: If this is a Helm chart, it should support `InstallOverrides` right?

cluster-operator/controllers/vmc/jaeger.go
saketmht: Its not removed. I actually deleted the else if block below. But since they had almost identical code, it is showing this part as deleted.

platform-operator/controllers/verrazzano/component/common/namespace.go
saketmht: Even if it is not found, this would have to return false

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
saketmht: We decided against it. As per discussions with Mike/Kevin, we might want to move towards a more restrictive API in the future (VerrazzanoLogging CR). So if we support overrides now, it might be more difficult to move towards that later.

platform-operator/controllers/verrazzano/component/opensearchoperator/legacyOS.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return nil
+}
+
+// vmiExists returns true if the VMI kind exists
+func vmiExists(ctx spi.ComponentContext) bool {
+	vmiList := vmov1.VerrazzanoMonitoringInstanceList{}
+	err := ctx.Client().List(context.TODO(), &vmiList)
+
+	if _, ok := err.(*meta.NoKindMatchError); ok {
```
anders-swanson, line 74: suggest to use `meta.IsNoMatchError`

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+	if cr != nil && cr.Spec.Components.Elasticsearch != nil {
+		for _, node := range cr.Spec.Components.Elasticsearch.Nodes {
+			nodeMap[node.Name] = node
+		}
+	}
+	return nodeMap
+}
+
+func getMasterNode(ctx spi.ComponentContext) string {
+	for _, node := range ctx.EffectiveCR().Spec.Components.Elasticsearch.Nodes {
```
anders-swanson, line 305: Is it necessary to consider the edge case where there are multiple master node groups in this function?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+
+// AppendOverrides appends the additional overrides for install
+func AppendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	// TODO: Image overrides once the BFS images are done
+
+	// Bootstrap pod overrides
+	if IsUpgrade(ctx) || IsSingleMasterNodeCluster(ctx) {
+		kvs = append(kvs, bom.KeyValue{
+			Key:   `opensearchCluster.bootstrap.additionalConfig.cluster\.initial_master_nodes`,
+			Value: fmt.Sprintf("%s-%s-0", clusterName, getMasterNode(ctx)),
```
anders-swanson, line 81: What if this returns an empty string?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+		ingressTarget := fmt.Sprintf("verrazzano-ingress.%s", dnsSubDomain)
+
+		ingressAnnotations := make(map[string]string)
+		ingressAnnotations[`cert-manager\.io/cluster-issuer`] = constants.VerrazzanoClusterIssuerName
+		if vzcr.IsExternalDNSEnabled(ctx.EffectiveCR()) {
+			ingressAnnotations[`external-dns\.alpha\.kubernetes\.io/target`] = ingressTarget
+			ingressAnnotations[`external-dns\.alpha\.kubernetes\.io/ttl`] = "60"
+		}
+
+		kvs, _ = appendOSIngressOverrides(ingressAnnotations, dnsSubDomain, ingressClassName, kvs)
```
anders-swanson, line 336: why are the errors being ignored?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+			Key:   "ingress.opensearchDashboards.enable",
+			Value: "false",
+		})
+	}
+
+	return kvs, nil
+}
+
+// appendOSDIngressOverrides appends the additional overrides for OpenSearchDashboards ingress
+func appendOSDIngressOverrides(ingressAnnotations map[string]string, dnsSubDomain, ingressClassName string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
```
anders-swanson, line 354: this function doesn't appear to return an error

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+	kvs = append(kvs, bom.KeyValue{
+		Key:   "ingress.opensearchDashboards.tls[0].hosts[0]",
+		Value: osdHostName,
+	})
+
+	return kvs, nil
+}
+
+// appendOSIngressOverrides appends the additional overrides for OpenSearch ingress
+func appendOSIngressOverrides(ingressAnnotations map[string]string, dnsSubDomain, ingressClassName string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
```
anders-swanson, line 390: This function also doesn't appear to return an error. It's also very similar to the prior function, are you able to make it more generic?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+	return kvs, nil
+}
+
+// isReady checks if all the sts and deployments for OpenSearch are ready or not
+func (o opensearchOperatorComponent) isReady(ctx spi.ComponentContext) bool {
+	deployments := getEnabledDeployments(ctx)
+	return ready.DeploymentsAreReady(ctx.Log(), ctx.Client(), deployments, 1, getPrefix(ctx))
+}
+
+// IsSingleMasterNodeCluster returns true if the cluster has a single mater node
```
anders-swanson, line 431: typo in comment

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+func IsSingleMasterNodeCluster(ctx spi.ComponentContext) bool {
+	var replicas int32
+	for _, node := range ctx.EffectiveCR().Spec.Components.Elasticsearch.Nodes {
+		for _, role := range node.Roles {
+			if role == "master" && node.Replicas != nil {
+				replicas += *node.Replicas
+			}
+		}
+	}
+	return replicas <= 1
```
anders-swanson, line 441: Is it a single master cluster if there are zero master nodes? This is some invalid topology which may come from user error

platform-operator/controllers/verrazzano/component/opensearchoperator/pvc.go
```diff
@@ -0,0 +1,269 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	c "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// getOSPersistentVolumes returns the list of older PersistentVolumes used by VMO OpenSearch
+func getOSPersistentVolumes(ctx spi.ComponentContext, nodes []vzapi.OpenSearchNode) ([]v1.PersistentVolume, error) {
```
anders-swanson, line 26: Can you use the VMI to get the PVCs?

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -87,13 +94,19 @@ func isOSNodeReady(ctx spi.ComponentContext, node vzapi.OpenSearchNode, prefix s
...
+	isLegacyOS := common.IsLegacyOS(ctx)
+	ns := ComponentNamespace
+	if !isLegacyOS {
+		ns = constants.VerrazzanoLoggingNamespace
+	}
+	nodeControllerName := getNodeControllerName(node, isLegacyOS)
 
 	// If a node has the master role, it is a statefulset
-	if hasRole(node.Roles, vmov1.MasterRole) {
+	// If the operator is managing OS, then all nodes are statefulset
```
michael-cico, line 105: nit, might want to distinguish between opster-managed and the VMO in these comments.  They're both operators.

pkg/vzcr/enabled.go
```diff
@@ -697,3 +697,17 @@ func IsAlertmanagerEnabled(cr runtime.Object, ctx spi.ComponentContext) (bool, e
 
 	return false, nil
 }
+
+// IsOpenSearchOperatorEnabled returns false only if OpenSearchOperator is explicitly disabled in the CR
```
michael-cico, line 701: Do we need to change the `IsVMOEnabled` func to not check for Opensearch?  Or do we need to keep having it return true if Opensearch is enabled so that it continues managing things until the upgrade/swap is done with Opster?

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards.go
```diff
@@ -15,11 +16,20 @@ import (
...
+				Name:      kibanaDeployment,
+				Namespace: ComponentNamespace,
+			},
+		}
+	}
 	return []types.NamespacedName{
 		{
-			Name:      kibanaDeployment,
-			Namespace: ComponentNamespace,
+			Name:      "opensearch-dashboards",
```
michael-cico, line 31: nit, should define a constant for this.

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator_component.go
```diff
@@ -0,0 +1,134 @@
...
+			ReleaseName:               ComponentName,
+			JSONName:                  ComponentJSONName,
+			ChartDir:                  filepath.Join(config.GetThirdPartyDir(), ComponentName),
+			ChartNamespace:            ComponentNamespace,
+			ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "opensearch-operator-values.yaml"),
+			IgnoreNamespaceOverride:   true,
+			SupportsOperatorInstall:   true,
+			SupportsOperatorUninstall: true,
+			Dependencies:              []string{networkpolicies.ComponentName, certmanager.ComponentName, nginx.ComponentName},
+			AppendOverridesFunc:       AppendOverrides,
```
michael-cico, line 56: `AppendOverrides` func probably doesn't need to be exported

platform-operator/controllers/verrazzano/component/opensearchoperator/pvc.go
```diff
@@ -0,0 +1,269 @@
...
+		pv := pvList[i]
+		oldReclaimPolicy := pv.Spec.PersistentVolumeReclaimPolicy
+		pv.Spec.PersistentVolumeReclaimPolicy = v1.PersistentVolumeReclaimRetain
+
+		if pv.Labels == nil {
+			pv.Labels = make(map[string]string)
+		}
+		// If old reclaim policy is not already set, set it
+		// So that it can be changed back to its original value later
+		_, ok := pv.Labels[constants.OldReclaimPolicyLabel]
```
michael-cico, line 91: if you used `controllerruntime.CreateOrUpdate()` you could just set values and that func will only update the CR if anything's changed.

platform-operator/internal/config/config.go
```diff
@@ -182,6 +183,13 @@ func GetHelmClusterOpChartsDir() string {
 	return filepath.Join(instance.VerrazzanoRootDir, helmClusterOpChartsDirSuffix)
 }
 
+func GetHelmOpenSearchOpChartsDir() string {
```
michael-cico, line 186: Didn't know we had all these component-chart helpers... but should we use this to set the chart location when the OS operator component is instantiated?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+		Resource: "opensearchuserrolebindings",
+		Version:  "v1",
+	}
+
+	gvrList = []schema.GroupVersionResource{clusterGVR, roleGVR, rolesMappingGVR}
+)
+
+// AppendOverrides appends the additional overrides for install
+func AppendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	// TODO: Image overrides once the BFS images are done
```
michael-cico, line 75: aren't the image overrides handled by the `HelmComponent` code?  Or are there other images that can't be handled by that?

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,498 @@
...
+					Key:   fmt.Sprintf("opensearchCluster.general.pluginsList[%d]", i),
+					Value: plugin,
+				})
+			}
+		}
+	}
+
+	if osd != nil {
+		osdPlugins := osd.Plugins
+		if osdPlugins.Enabled && len(osdPlugins.InstallList) > 0 {
```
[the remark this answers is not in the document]
michael-cico: might be good to wrap this in an `IsXXXXEnabled` func internally to handle the nil-pointer case.

platform-operator/apis/verrazzano/v1alpha1/verrazzano_types.go
adiforluls: Are we also going to prohibit the use of overrides for the "operator" as well?

adiforluls: I think for configs like pod/container security, will have to externalize those things in the CRD since we're not going to make helm values configurable.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +78,36 @@ func DeleteVMI(ctx spi.ComponentContext) error {
...
+		systemVMI := vmov1.VerrazzanoMonitoringInstance{}
+		err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: VMIName, Namespace: globalconst.VerrazzanoSystemNamespace}, &systemVMI)
+		if err != nil && errors.IsNotFound(err) {
+			return false
+		}
+		if systemVMI.Spec.Opensearch.Enabled {
+			return true
+		}
+	}
+	return false
```
adiforluls, line 93: If the error is `IsNotFound`, I think this should return false, but what about other type of errors? Would you want to retry in that case?
Maybe the func should return `(bool, error)`, that way the caller can inspect the error value also along with bool value.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +78,36 @@ func DeleteVMI(ctx spi.ComponentContext) error {
...
+		systemVMI := vmov1.VerrazzanoMonitoringInstance{}
+		err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: VMIName, Namespace: globalconst.VerrazzanoSystemNamespace}, &systemVMI)
+		if err != nil && errors.IsNotFound(err) {
+			return false
+		}
+		if systemVMI.Spec.OpensearchDashboards.Enabled {
+			return true
+		}
+	}
+	return false
```
adiforluls, line 108: Same question/comment as above.

platform-operator/controllers/verrazzano/component/opensearchoperator/pvc.go
adiforluls: +1

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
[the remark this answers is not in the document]
saketmht: Plugins.Enabled is not a pointer actually

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +78,36 @@ func DeleteVMI(ctx spi.ComponentContext) error {
 	return nil
 }
 
+// isLegacyOS returns true if the OS that is running is managed by VMO
+func IsLegacyOS(ctx spi.ComponentContext) bool {
+	if vzcr.IsOpenSearchEnabled(ctx.EffectiveCR()) {
+		systemVMI := vmov1.VerrazzanoMonitoringInstance{}
+		err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: VMIName, Namespace: globalconst.VerrazzanoSystemNamespace}, &systemVMI)
+		if err != nil && errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
saketmht: Probably just check for newer opensearch sts in logging ns since that is the default going forward

platform-operator/controllers/verrazzano/component/opensearchoperator/legacyOS.go
saketmht: Updated

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: Any one master node will do

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: That would be an invalid configuration. I will add a check and return error if no master nodes are present

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: It was not really returning any error since we I was just appending values to kvs

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: Updated

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: Updated

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: If there are no master nodes, it is an invalid configuration but I don't think it makes sense to check here

pkg/vzcr/enabled.go
saketmht: I am testing a few scenarios but for now I think we can leave it as it is.

platform-operator/controllers/verrazzano/component/opensearchoperator/pvc.go
saketmht: The thing here is that these functions can be called multiple times. So oldReclaimPolicy should only be set when it PV is set to RETAIN the first time. SInce the next time, its policy would be retain(instead of delete) and we don't want to set the oldReclaimPolicy to RETAIN as well.

platform-operator/internal/config/config.go
saketmht: I am not sure what you mean by this. Every component has it this way only.

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
saketmht: The image for opensearch-operator will be handled by HelmComponent itself. But since we are managing OS and OSD via this component now as well, we would need to append images for OS and OSD now here

platform-operator/controllers/verrazzano/component/common/vmi.go
saketmht: This is called from the Ready checks which don't return an error. But I have added an error log

saketmht:
I have refactored the code based on comments and to add overrides for the opensearchOperator component. For this, I removed the cluster chart from the operator helm and added a yaml manifest file for the cluster to which we pass some dynamic values and apply that to create the OS cluster

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +79,42 @@ func DeleteVMI(ctx spi.ComponentContext) error {
...
 
+// IsLegacyOS returns true if the OS that is running is managed by VMO
+func IsLegacyOS(ctx spi.ComponentContext) bool {
+	if vzcr.IsOpenSearchEnabled(ctx.EffectiveCR()) {
+		systemVMI := vmov1.VerrazzanoMonitoringInstance{}
+		if err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: VMIName, Namespace: globalconst.VerrazzanoSystemNamespace}, &systemVMI); err != nil {
+			if meta.IsNoMatchError(err) || errors.IsNotFound(err) {
+				return false
+			}
+			ctx.Log().Errorf("Failed to get VMI, considering legacy OS to be disabled")
```
michael-cico, line 90: You should use `ErrorfThrottled` so the message is throttled.

michael-cico: Can be done in a second pass if you prefer.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +79,42 @@ func DeleteVMI(ctx spi.ComponentContext) error {
...
+
+// IsLegacyOSD returns true if the OSD that is running is managed by VMO
+func IsLegacyOSD(ctx spi.ComponentContext) bool {
+	if vzcr.IsOpenSearchDashboardsEnabled(ctx.EffectiveCR()) {
+		systemVMI := vmov1.VerrazzanoMonitoringInstance{}
+		if err := ctx.Client().Get(context.TODO(), types.NamespacedName{Name: VMIName, Namespace: globalconst.VerrazzanoSystemNamespace}, &systemVMI); err != nil {
+			if meta.IsNoMatchError(err) || errors.IsNotFound(err) {
+				return false
+			}
+			ctx.Log().Errorf("Failed to get VMI, considering legacy OSD to be disabled")
```
michael-cico, line 108: Use `ErrorfThrottled`

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator_component.go
```diff
@@ -114,6 +117,31 @@ func (o opensearchOperatorComponent) PreInstall(ctx spi.ComponentContext) error
...
+		if err != nil {
+			return err
+		}
+		// substitute template values to all files in the directory and apply the resulting YAML
+		filePath := path.Join(config.GetThirdPartyManifestsDir(), "opensearch-operator/opensearch_cluster_cr.yaml")
+		yamlApplier := k8sutil.NewYAMLApplier(ctx.Client(), "")
+		err = yamlApplier.ApplyFT(filePath, args)
+
+		if err != nil {
+			return ctx.Log().ErrorfNewErr("Failed to substitute template values for OpenSearchCluster CR: %v", err)
```
michael-cico, line 138: Use `ErrorfThrottledNewErr`

pkg/k8sutil/apply_yaml.go
```diff
@@ -478,3 +479,15 @@ func nindent(indent int, s string) string {
 
 	return sb.String()
 }
+
+func multiLineIndent(indentNum int, aff string) string {
+	var b = make([]byte, indentNum)
+	for i := 0; i < indentNum; i++ {
```
anders-swanson, line 485: see [strings.Repeat](https://pkg.go.dev/strings#Repeat) for a more idiomatic way of creating strings like this. It also is not left up to the reader to know that `32` here is used as the ascii code for a space

anders-swanson: i still don't get it, we use nindent defined above for multi line indents, the input is split on newline/linefeed characters

saketmht: nindent doesn't work for multi-line inputs. I tried it and didn't get the desired result so used this instead. If I want to format a yaml at 4 space and I use nindent, the result will be
```
    - component: es-master         // First line indented
replicas: 3
roles:
- ...
- component: ...                  // Rest of line not indented
```
I didn't really write. I just copied it over from https://github.com/verrazzano/verrazzano/blob/2ae22d8487738ad9840b4742f25c09f93c102a00/platform-operator/apis/verrazzano/v1alpha1/convert_istio_to.go#L212 which is something you wrote actually. 
Your comment
> / Define the IstioOperator template which is used to insert the generated YAML values.
//
// NOTE: The go template rendering doesn't properly indent the multi-line YAML value
// For example, the template fragment only indents the fist line of values

platform-operator/controllers/verrazzano/component/common/namespace.go
fredtibbitts: Per Paul's comment, this function should return `(bool, error)` and return an error if there's an actual error fetching the namespace (ignoring not found).

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -6,6 +6,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/meta"
```
fredtibbitts, line 9: Imports not formatted correctly.

platform-operator/controllers/verrazzano/component/common/vmi.go
```diff
@@ -78,6 +79,42 @@ func DeleteVMI(ctx spi.ComponentContext) error {
 	return nil
 }
 
+// IsLegacyOS returns true if the OS that is running is managed by VMO
+func IsLegacyOS(ctx spi.ComponentContext) bool {
```
fredtibbitts, line 83: Return `(bool, error)`, and let the caller decide whether it wants to ignore errors.

platform-operator/controllers/verrazzano/component/common/vmi.go
fredtibbitts: Agree with @adiforluls, these functions should not eat errors. Return the error and the caller can decide to ignore or not.

platform-operator/controllers/verrazzano/component/opensearch/opensearch.go
```diff
@@ -5,6 +5,7 @@ package opensearch
 
 import (
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
```
fredtibbitts, line 8: Import formatting.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -5,6 +5,7 @@ package opensearch
 
 import (
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/controllers/verrazzano/component/opensearchoperator"
```
fredtibbitts, line 8: Import formatting.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
```diff
@@ -103,24 +103,12 @@ func NewComponent() spi.Component {
...
-	}
-	ctx.Log().Debug("OpenSearch pre-install")
-	if err := common.CreateAndLabelVMINamespaces(ctx); err != nil {
-		return ctx.Log().ErrorfNewErr("Failed creating/labeling namespace %s for OpenSearch : %v", ComponentNamespace, err)
-	}
 	return nil
 }
 
 // Install OpenSearch component install processing
 func (o opensearchComponent) Install(ctx spi.ComponentContext) error {
```
fredtibbitts, line 110: I think you can just delete these functions since they just return `nil`.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_test.go
```diff
@@ -5,6 +5,7 @@ package opensearch
 
 import (
 	"fmt"
+	globalconst "github.com/verrazzano/verrazzano/pkg/constants"
```
fredtibbitts, line 8: Import formatting.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards.go
```diff
@@ -1,10 +1,11 @@
-// Copyright (c) 2022, Oracle and/or its affiliates.
+// Copyright (c) 2022, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package opensearchdashboards
 
 import (
 	"fmt"
+	"github.com/verrazzano/verrazzano/platform-operator/constants"
```
fredtibbitts, line 8: Import formatting.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
```diff
@@ -105,24 +107,12 @@ func NewComponent() spi.Component {
...
-	}
-	ctx.Log().Debug("OpenSearch-Dashboards pre-install")
-	if err := common.CreateAndLabelVMINamespaces(ctx); err != nil {
-		return ctx.Log().ErrorfNewErr("Failed creating/labeling namespace %s for OpenSearch-Dashboards : %v", ComponentNamespace, err)
-	}
 	return nil
 }
 
 // Install OpenSearch-Dashboards component install processing
 func (d opensearchDashboardsComponent) Install(ctx spi.ComponentContext) error {
```
fredtibbitts, line 114: Delete these functions since they just return `nil`.

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component_test.go
```diff
@@ -4,6 +4,8 @@ package opensearchdashboards
 
 import (
 	"context"
+	vmov1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
```
fredtibbitts, line 7: Import formatting.

fredtibbitts:
Please review the code smells. At least a few of those should be addressed.

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
saketmht: Since OS and OSD component do not implement Helm component, these need to have some implementation defined

platform-operator/controllers/verrazzano/component/opensearchdashboards/opensearchdashboards_component.go
saketmht: This doesn't extend Helm Component so these need to have implementations defined

platform-operator/controllers/verrazzano/component/common/vmi.go
saketmht: Updated

platform-operator/controllers/verrazzano/component/opensearch/opensearch_component.go
fredtibbitts: Oh, right. I didn't look to see that it wasn't a Helm component. Thanks.

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator_component.go
```diff
@@ -0,0 +1,173 @@
...
+		HelmComponent: helm.HelmComponent{
+			ReleaseName:               ComponentName,
+			JSONName:                  ComponentJSONName,
+			ChartDir:                  filepath.Join(config.GetThirdPartyDir(), ComponentName),
+			ChartNamespace:            ComponentNamespace,
+			ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "opensearch-operator-values.yaml"),
+			IgnoreNamespaceOverride:   true,
+			SupportsOperatorInstall:   true,
+			SupportsOperatorUninstall: true,
+			Dependencies:              []string{networkpolicies.ComponentName, certmanager.ComponentName, nginx.ComponentName},
```
michael-cico, line 57: You may want to add a dependecy on the `clusterIssuer` component instead of `certManager`.  the `clusterIssuer` configuration controls the configuration of the `clusterResouceNamespace` we install certManager with.

michael-cico: Or, just add `clusterIssuer` instead of replacing `certManager`

platform-operator/controllers/obsolete/reconcile/controller.go
```diff
@@ -911,7 +911,36 @@ func (r *Reconciler) watchResources(namespace string, name string, log vzlog.Ver
 	}
 
 	log.Debugf("Watching the Rancher global data namespace to create admin cloud credential")
-	return r.watchRancherGlobalDataNamespace(namespace, name)
+	if err := r.watchRancherGlobalDataNamespace(namespace, name); err != nil {
```
pfmackin, line 914: This code is obsolete and not executeed. Any changes you make here have no effect.

platform-operator/controllers/verrazzano/component/common/opensearch_operator.go
```diff
@@ -0,0 +1,567 @@
...
+				if role == "data" && node.Replicas != nil {
+					replicas += *node.Replicas
+				}
+			}
+		}
+	}
+	return replicas <= 1
+}
+
+// IsUpgrade returns true if we are upgrading from <=1.6.x to 2.x
```
pfmackin, line 156: comment is wrong, shoudld <= 1.6.x to > 1.6.x

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,260 @@
...
+		Value: clusterResourceNamespace,
+	})
+	return kvs, nil
+}
+
+// deleteRelatedResource deletes the resources handled by the opensearchOperator
+// Like OpenSearchRoles, OpenSearchUserRolesBindings
+// Since the operator adds a finalizer to these resources, they need to deleted before the operator is uninstalled
+func (o opensearchOperatorComponent) deleteRelatedResource() error {
+	client, err := k8sutil.GetDynamicClient()
```
pfmackin, line 104: Change this to do it the same way that cert-manager does it so that the finalizer is removed.  See https://github.com/verrazzano/verrazzano/blob/master/platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager_component.go#L186

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -0,0 +1,260 @@
...
+			return fmt.Errorf("waiting for all %s to be deleted", gvr.String())
+		}
+	}
+	return nil
+}
+
+// buildIngressOverrides builds the overrides required for the OpenSearch and OpenSearchDashboards ingresses
+func buildIngressOverrides(ctx spi.ComponentContext, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	if vzcr.IsNGINXEnabled(ctx.EffectiveCR()) {
+
```
pfmackin, line 149: remove line 149

platform-operator/helm_config/overrides/opensearch-operator-values.yaml
```diff
@@ -0,0 +1,60 @@
...
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+        - ALL
+    privileged: false
+    runAsGroup: 65532
+    runAsNonRoot: true
+    runAsUser: 65532
+  image:
+    repository: public.ecr.aws/opsterio/opensearch-operator
```
pfmackin, line 19: why is in the overrides.  We get this image from the BOM don't we?

tests/e2e/security/netpol/network_policy_test.go
```diff
@@ -478,11 +482,11 @@ var _ = t.Describe("Test Network Policies", Label("f:security.netpol"), func() {
 				err := testAccess(metav1.LabelSelector{MatchLabels: map[string]string{"app": "netpol-test"}}, "netpol-test", metav1.LabelSelector{MatchLabels: map[string]string{"k8s-app": "verrazzano-monitoring-operator"}}, vzconst.VerrazzanoSystemNamespace, 8000, false, false)
 				Expect(err).To(BeNil(), fmt.Sprintf("FAIL: Negative test verrazzano-monitoring-operator ingress rules failed: reason = %s", err))
 			},
-			func() {
-				t.Logs.Info("Negative test vmi-system-es-master ingress rules")
-				err := testAccess(metav1.LabelSelector{MatchLabels: map[string]string{"app": "netpol-test"}}, "netpol-test", metav1.LabelSelector{MatchLabels: map[string]string{"app": "system-es-master"}}, vzconst.VerrazzanoSystemNamespace, 9200, false, false)
-				Expect(err).To(BeNil(), fmt.Sprintf("FAIL: Negative test vmi-system-es-master ingress rules failed: reason = %s", err))
-			},
+			//func() {
```
pfmackin, line 485: you should do a follow up JIRA to remove these commented out tests to get them to work.  If for some reason they are no longer valid then remove them.

tests/e2e/pkg/elasticsearch.go
```diff
@@ -344,37 +343,22 @@ func GetSystemOpenSearchIngressURL(kubeconfigPath string) string {
 		Log(Error, fmt.Sprintf("Failed to get clientset for cluster %v", err))
 		return ""
 	}
-	// Return the os ingress host from the verrazzano-logging namespace if it exists
+	// Return the newer OS ingress if >= 2.0.0
```
pfmackin, line 346: 2.0.0 is wrong.

tests/e2e/pkg/elasticsearch.go
```diff
@@ -344,37 +343,22 @@ func GetSystemOpenSearchIngressURL(kubeconfigPath string) string {
...
 	// Else return the vmi ingress as usual
-	useloggingNSIngress, _ := osIngressInLoggingNSExists()
+	useNewOSIngress, _ := IsVerrazzanoMinVersion("2.0.0", kubeconfigPath)
```
pfmackin, line 348: 2.0.0 is wrong.

tests/e2e/pkg/opensearch_dashboards.go
```diff
@@ -133,14 +133,15 @@ func PostOpensearchDashboards(path string, body string, additionalHeaders ...str
 func getOpenSearchDashboardsURL(kubeconfigPath string) string {
 	clientset, err := GetKubernetesClientsetForCluster(kubeconfigPath)
 	isMinversion150, _ := IsVerrazzanoMinVersion("1.5.0", kubeconfigPath)
+	isMinversion200, _ := IsVerrazzanoMinVersion("2.0.0", kubeconfigPath)
```
pfmackin, line 136: 2.0.0 is wrong.

tests/e2e/uninstall/verifycrds/verify_crds_suite_test.go
```diff
@@ -10,12 +10,10 @@ import (
...
 )
 
-var mySQLOperatorEnabled bool
-var vmoEnabled bool
+var isManagedCluster bool
 
 func init() {
-	flag.BoolVar(&mySQLOperatorEnabled, "mySQLOperatorEnabled", true, "mySQLOperatorEnabled indicates whether the mySQLOperator component is enabled")
-	flag.BoolVar(&vmoEnabled, "vmoEnabled", true, "vmoEnabled indicates whether the Verrazzano monitoring operator is enabled")
+	flag.BoolVar(&isManagedCluster, "isManagedCluster", false, "isManagedCluster indicates if it is a managed-cluster profile or not")
```
pfmackin, line 16: we should test based on a profile.  We should check the CR for disabled/enabled components

tests/e2e/uninstall/verifycrds/verify_crds_test.go
```diff
@@ -114,7 +123,7 @@ var _ = t.Describe("Verify CRDs after uninstall.", Label("f:platform-lcm.unnstal
 	if err != nil {
 		Fail(err.Error())
 	}
-	if vmoEnabled {
+	if !isManagedCluster {
```
pfmackin, line 126: we should test based on a profile.  We should check the CR for disabled/enabled components

tests/e2e/uninstall/verifycrds/verify_crds_test.go
```diff
@@ -151,11 +160,14 @@ var _ = t.Describe("Verify CRDs after uninstall.", Label("f:platform-lcm.unnstal
 		checkCrds(crds, map[string]bool{"coherence.coherence.oracle.com": false}, "coherence.coherence.oracle.com")
 	})
 
-	if mySQLOperatorEnabled {
+	if !isManagedCluster {
```
pfmackin, line 163: we should test based on a profile.  We should check the CR for disabled/enabled components

tests/e2e/upgrade/post-upgrade/verify/verify_restart_test.go
```diff
@@ -296,12 +328,44 @@ var _ = t.Describe("Checking if Verrazzano system components are ready, post-upg
...
 			t.Entry("Checking StatefulSet mysql", mysql.ComponentNamespace, mysql.ComponentName, "mysql"),
 		)
 	})
 
+	Context("Checking optional StatefulSets for post-upgrade", func() {
+		t.DescribeTable("StatefulSet should be ready post-upgrade",
+			func(namespace string, componentName string, stsName string) {
+				// Currently we have no way of determining if some components are installed by looking at the status (grafana)
+				// Because of this, make this test non-managed cluster only.
+				if vzcr.Spec.Profile == vzapi.ManagedCluster {
```
pfmackin, line 343: we should test based on a profile.  We should check the CR for disabled/enabled components

tests/e2e/verify-infra/vmi/vmi_test.go
```diff
@@ -154,11 +155,13 @@ var beforeSuite = t.BeforeSuiteFunc(func() {
 		return err
 	}, waitTimeout, pollingInterval).ShouldNot(HaveOccurred())
 
-	elastic = vmi.GetOpensearch("system")
+	ok, _ := pkg.IsVerrazzanoMinVersion("2.0.0", kubeconfigPath)
```
pfmackin, line 158: 2.0.0 is wrong

tests/e2e/upgrade/post-upgrade/verify/verify_restart_test.go
```diff
@@ -188,6 +191,19 @@ var _ = t.Describe("Checking if Verrazzano system components are ready, post-upg
 						pkg.Log(pkg.Error, fmt.Sprintf("failed to find the verrazzano version: %v", err))
 						return false
 					}
+					isVersionAbove2_0_0, err := pkg.IsVerrazzanoMinVersion("2.0.0", kubeconfigPath)
```
pfmackin, line 194: Version 2.0 is wrong

saketmht: Currently the version is 2.0 in master as well, will update once it is set to 1.7 in msater

tests/e2e/upgrade/post-upgrade/verify/verify_restart_test.go
saketmht: All the other tests are written in this fashion. Changing one would likely mean to change all the other existing tests as well.

tests/e2e/uninstall/verifycrds/verify_crds_suite_test.go
saketmht: This parameter is passed through the jenkins file where we are checking if it is a managed cluster or not

tests/e2e/security/netpol/network_policy_test.go
saketmht: Sure

platform-operator/helm_config/overrides/opensearch-operator-values.yaml
saketmht: removed

michael-cico (dismissed):

IshaGirdhar (approved):

PR #6954, opened 2023-09-05
VZ-10836: Updates to Auth Proxy v2 configuration

fredtibbitts:
This PR includes:
1. Adding a new `oidcExternalURL` configuration field to the Auth Proxy v2 secret and Helm chart. The value is the externally accessible OIDC provider URL.
2. Renamed the `oidcIssuerURL` configuration field to `oidcServiceURL`. The value is the in-cluster URL to access the OIDC provider.
3. Implemented Helm overrides in the authproxy component that set `oidcServiceURL` and `oidcExternalURL` appropriately.
4. Refactoring to use a constant for the Verrazzano OIDC system realm.

After installing in a cluster, the OIDC config secret has these values for the URLs:
```
oidcExternalURL: https://keycloak.default.172.18.0.230.nip.io/auth/realms/verrazzano-system
oidcServiceURL: http://keycloak-http.keycloak.svc.cluster.local/auth/realms/verrazzano-system
```

### DIFF
--- a/authproxy/src/proxy/proxy.go
+++ b/authproxy/src/proxy/proxy.go
@@ -154,7 +154,7 @@ func (h Handler) handleAPIRequest(rw http.ResponseWriter, req *http.Request) {
 	}
 
 	oidcConfig := auth.OIDCConfiguration{
-		IssuerURL:   config.GetIssuerURL(),
+		IssuerURL:   config.GetServiceURL(),
 		ClientID:    config.GetClientID(),
 		CallbackURL: fmt.Sprintf("https://%s%s", ingressHost, callbackPath),
 	}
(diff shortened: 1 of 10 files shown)


davisgomes (approved):

authproxy/src/proxy/proxy.go
```diff
@@ -154,7 +154,7 @@ func (h Handler) handleAPIRequest(rw http.ResponseWriter, req *http.Request) {
 	}
 
 	oidcConfig := auth.OIDCConfiguration{
-		IssuerURL:   config.GetIssuerURL(),
+		IssuerURL:   config.GetServiceURL(),
```
desagar, line 157: we are going to need both the service url and the external url here.

desagar (approved):

PR #6970, opened 2023-09-06
VZ-10879 Update new Verrazzano controller so that it can be used without the other legacy controller.

pfmackin:
Update new Verrazzano controller so that it can be used without the other legacy controller.

1. implement full install, update with pre and post logic
2. implement full upgrade with pre and post logic
3. implement full uninstall with pre and post logic 
4. update VZ conditions and state as needed
5. remove module logic from legacy controller
6. remove controller sync code
7. either start new controller or the legacy controller, not both
8. move custom resource or component code to a custom package


platform-operator/experimental/controllers/verrazzano/custom/istio.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
```
michael-cico, line 1: nit, not sure I like the `custom` package name, but I don't have anything better.  Is this all global pre/post handling ?

pfmackin: it is all custom code for components, multicluster.  We can rename after if you have a better name.  Ideally it goes away and pushed to components.

michael-cico: nah, I'm good ;-)

michael-cico (approved):

platform-operator/experimental/controllers/verrazzano/module.go
```diff
@@ -8,19 +8,84 @@ import (
...
+		log.ErrorfThrottled("Error loading module catalog: %v", err)
+		return result.NewResultShortRequeueDelayWithError(err)
+	}
+
+	// Create or Update a Module for each enabled resource
+	for _, comp := range registry.GetComponents() {
+		if !comp.IsEnabled(effectiveCR) {
+			continue
+		}
+		if !comp.ShouldUseModule() {
```
mbenmclean, line 37: This can be removed now I think

mbenmclean (approved):

platform-operator/experimental/controllers/verrazzano/module.go
pfmackin: yes, i will do it post merge

PR #6974, opened 2023-09-06
VZ 10558: OCNE OCI Single Node support

anders-swanson:
Allow quick create OCI OCNE clusters to be created with only control plane nodes.


pkg/k8s/node/node.go
```diff
@@ -20,3 +25,30 @@ func GetK8sNodeList(k8sClient client.Client) (*v1.NodeList, error) {
 	)
 	return nodeList, err
 }
+
```
jmaron99, line 28: any additions to node_test.go that can be added here?

anders-swanson: Added test case

jmaron99 (dismissed):

jmaron99 (approved):

PR #7006, opened 2023-09-09
VZ-10783 Auth proxy v2 login redirect and token verification

davisgomes:
**Changes**
- Add login redirect flow
- Add token verification for API server requests
- Unit tests for new and existing code


authproxy/src/auth/auth_types.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
```
desagar, line 2: very minor nit - I'm not sure I like the name auth_types.go - with that name, I think of types of authentication, not data types related to authentication. An alternative I prefer (still not ideal) would be types.go since it's in the auth package already.

authproxy/src/proxy/proxy.go
```diff
@@ -124,105 +102,70 @@ func (h Handler) findPathHandler(req *http.Request) handlerFuncType {
...
 
 	handlerFunc := h.findPathHandler(req)
 	handlerFunc(rw, req)
 }
 
 // handleAuthCallback is the http handler for authentication callback
-func (h Handler) handleAuthCallback(rw http.ResponseWriter, req *http.Request) {
-
+func (h *Handler) handleAuthCallback(rw http.ResponseWriter, req *http.Request) {
+	// ingressHost := getIngressHost(req)
```
desagar, line 121: we should probably remove this commented out code

authproxy/src/apiserver/apiserver.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, err
+	}
+
+	a.Authenticator.SetCallbackURL(fmt.Sprintf("https://%s%s", ingressHost, a.CallbackPath))
+	continueProcessing, err := a.Authenticator.AuthenticateRequest(req, rw)
+	if err != nil {
+		http.Error(rw, err.Error(), http.StatusUnauthorized)
+		return nil, err
+	}
+	if !continueProcessing {
```
desagar, line 77: This is not necessarily an error condition. We may return false if we have performed a redirect already and there is no further processing needed.

davisgomes: Ah okay, thanks. Does it make sense to just return here in that case?

fredtibbitts (approved):

PR #7007, opened 2023-09-11
VZ-10605: Dex component in Verrazzano

bhatpmk:
This change introduces a new component Dex in Verrazzano
1. Dex is not enabled by default, will be done in a follow-up change as part of Keycloak integration with Dex
2. When Verrazzano is installed with Dex enabled, the Dex configuration is constructed to include a static admin user (staticPassword in Dex terminology) and a static PKCE client. Auth Proxy is enhanced to use Dex as the OIDC provider for console, when enabled. So when the Verrazzano console URL is accessed, it redirects to the Dex console for the authentication.


pkg/vzcr/enabled.go
```diff
@@ -697,3 +697,17 @@ func IsAlertmanagerEnabled(cr runtime.Object, ctx spi.ComponentContext) (bool, e
 
 	return false, nil
 }
+
+// IsDexEnabled returns true only if Dex is explicitly enabled in the CR
+func IsDexEnabled(cr runtime.Object) bool {
```
michael-cico, line 702: should probably add a unit test for this

michael-cico: can be on a second pass

platform-operator/controllers/verrazzano/component/dex/dex_component.go
```diff
@@ -0,0 +1,200 @@
...
+	secret := &corev1.Secret{}
+	err := ctx.Client().Get(context.TODO(), client.ObjectKey{
+		Namespace: constants.VerrazzanoSystemNamespace,
+		Name:      constants.Verrazzano,
+	}, secret)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			ctx.Log().Debugf("Component Dex creating Verrazzano secret %s/%s",
+				constants.VerrazzanoSystemNamespace, constants.Verrazzano)
+			if err := common.EnsureVMISecret(ctx.Client()); err != nil {
```
michael-cico, line 128: can you just call `EnsureVMISecret()` without doing a `Get` first?  Assuming it does a CreateOrUpdate.

michael-cico: Never mind, I see the Get is on the Verrazzano secret.

platform-operator/controllers/verrazzano/component/dex/dex.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// AppendDexOverrides appends the default overrides for the Dex component
+func AppendDexOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs []bom.KeyValue) ([]bom.KeyValue, error) {
+	bomFile, err := bom.NewBom(config.GetDefaultBOMFilePath())
+	if err != nil {
+		return kvs, err
+	}
+
+	image, err := bomFile.BuildImageOverrides(ComponentName)
```
michael-cico, line 157: we should get image overrides for free from `HelmComponent` ?

platform-operator/controllers/verrazzano/component/dex/dex.go
```diff
@@ -0,0 +1,496 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// getDNSDomain returns the DNS Domain
+func getDNSDomain(c client.Client, vz *vzapi.Verrazzano) (string, error) {
```
michael-cico, line 269: I think there's shared code for this in the `internal` package perhaps

pkg/vzcr/enabled.go
bhatpmk: Sure, will add that

platform-operator/constants/constants.go
```diff
@@ -293,3 +296,12 @@ const VerrazzanoModuleEventLabel = "module.verrazzano.io/event"
 
 // AlertmanagerIngress is the name of the ingress for Alertmanager
 const AlertmanagerIngress = "alertmanager"
+
+// DexNamespace is the Dex namespace name
+const DexNamespace = "dex"
```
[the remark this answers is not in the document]
bhatpmk: The namespace is verrazano-auth now

platform-operator/controllers/verrazzano/component/dex/dex.go
bhatpmk: Removed unnecessary code

platform-operator/controllers/verrazzano/component/dex/dex.go
bhatpmk: I could not find one, and found similar usage in other components. Probably we can write a function and cleanup.

platform-operator/controllers/verrazzano/component/authproxy/authproxy.go
```diff
@@ -344,6 +359,24 @@ func appendAuthProxyImageOverrides(kvs []bom.KeyValue) []bom.KeyValue {
...
+		errMsg := fmt.Sprintf("Unable to retrieve %s secret from %s namespace, %v", adminClusterOidcID, constants.DexNamespace, err)
+		ctx.Log().Once(errMsg)
+		return nil, fmt.Errorf(errMsg)
+	}
+
+	if data, ok := secret.Data["clientSecret"]; ok {
+		return data, nil
+	}
+
+	return nil, fmt.Errorf("client secret not present in %s/%s secret", constants.DexNamespace, adminClusterOidcID)
```
pfmackin, line 377: use the throttledError

bhatpmk: Updated

pkg/vzcr/enabled.go
bhatpmk: Done

michael-cico (approved):

PR #7012, opened 2023-09-11
Fix to Add Additional Labels To Managed Cluster ArgoCD Secret

George-Aeillo:
In this PR, additional labels to to a managed cluster can be added to an ArgoCD Secret and not be overwritten. A unit test related to this PR is also committed and it has clean runs in the pipeline.

### DIFF
--- a/cluster-operator/controllers/vmc/argocd.go
+++ b/cluster-operator/controllers/vmc/argocd.go
@@ -236,8 +236,9 @@ func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *c
 		secret.StringData = make(map[string]string)
 	}
 	secret.Type = corev1.SecretTypeOpaque
-	secret.ObjectMeta.Labels = map[string]string{"argocd.argoproj.io/secret-type": "cluster"}
-
+	if secret.ObjectMeta.Labels == nil {
+		secret.ObjectMeta.Labels = map[string]string{"argocd.argoproj.io/secret-type": "cluster"}
+	}
 	secret.StringData["name"] = clusterName
 	secret.StringData["server"] = rancherURL
 
--- a/cluster-operator/controllers/vmc/argocd.go
+++ b/cluster-operator/controllers/vmc/argocd.go
@@ -237,14 +237,11 @@ func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *c
 	}
 	secret.Type = corev1.SecretTypeOpaque
 	if secret.ObjectMeta.Labels == nil {
-		secret.ObjectMeta.Labels = map[string]string{"argocd.argoproj.io/secret-type": "cluster"}
+		secret.ObjectMeta.Labels = map[string]string{}
 	}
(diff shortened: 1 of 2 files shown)


cluster-operator/controllers/vmc/argocd.go
```diff
@@ -236,10 +236,15 @@ func (r *VerrazzanoManagedClusterReconciler) mutateArgoCDClusterSecret(secret *c
...
 	}
 	secret.Type = corev1.SecretTypeOpaque
-	secret.ObjectMeta.Labels = map[string]string{"argocd.argoproj.io/secret-type": "cluster"}
-
+	if secret.ObjectMeta.Labels == nil {
+		secret.ObjectMeta.Labels = map[string]string{"argocd.argoproj.io/secret-type": "cluster"}
+	}
 	secret.StringData["name"] = clusterName
 	secret.StringData["server"] = rancherURL
+	valueToIndicateIfClusterIsRegistered, ok := secret.ObjectMeta.Labels["argocd.argoproj.io/secret-type"]
```
desagar, line 244: there's no need to check whether that label exists. You can simply set it. If it's already there and set, it will remain the same. If it's not, it will be set. 
i.e. inside the nil check if statement
secret.ObjectMeta.Labels = map[string]string{}

and outside it:
secret.ObjectMeta.Labels["argocd.argoproj.io/secret-type"] = "cluster"

cluster-operator/controllers/vmc/argocd_test.go
```diff
@@ -483,3 +483,68 @@ func TestUpdateArgoCDClusterRoleBindingTemplate(t *testing.T) {
...
+		Client: cli,
+		log:    vzlog.DefaultLogger(),
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "demo" + "-" + clusterSecretName,
+			Namespace:   constants.ArgoCDNamespace,
+			Annotations: map[string]string{createTimestamp: time.Now().Add(-10 * time.Hour).Format(time.RFC3339), expiresAtTimestamp: time.Now().Format(time.RFC3339)},
+			Labels:      map[string]string{"testlabel": "shouldnotbedeleted"},
```
desagar, line 530: It would be nice to have multiple test cases - 
1. A secret with no labels
2. A secret with an existing argocd cluster label and an additional label.
3. A secret with existing argocd cluster label and no other labels.

In all cases, the argocd cluster label should exist, and any additional labels initially present in the secret should also still exist.

If you want to get this in quickly for Ali, that can be a follow up PR.

cluster-operator/controllers/vmc/argocd.go
George-Aeillo: Thanks!

cluster-operator/controllers/vmc/argocd_test.go
George-Aeillo: Thanks! I will reach out further to see if whether additional tests should be added.

desagar (approved):

PR #7040, opened 2023-09-13
VZ-10925 Register CAPI clusters with Verrazzano when Rancher is not enabled/installed

jmaron99:
- Refactored CAPI cluster controller to process either Rancher or VZ registration
- Created VZ registration code
- Some modifications to existing VMC/MC code

### DIFF
--- a/cluster-operator/controllers/capi/rancher.go
+++ b/cluster-operator/controllers/capi/rancher.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2023, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package capi
+
+import (
+	"context"
+	"fmt"
+	"github.com/verrazzano/verrazzano/cluster-operator/controllers/vmc"
+	"github.com/verrazzano/verrazzano/pkg/constants"
+	vzctrl "github.com/verrazzano/verrazzano/pkg/controller"
+	"github.com/verrazzano/verrazzano/pkg/k8s/ready"
+	"github.com/verrazzano/verrazzano/pkg/k8sutil"
+	"github.com/verrazzano/verrazzano/pkg/log/vzlog"
+	"github.com/verrazzano/verrazzano/pkg/rancherutil"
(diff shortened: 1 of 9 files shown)


cluster-operator/controllers/capi/rancher.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		log.Errorf("Unable to unregister cluster %s from Rancher: %v", cluster.GetName(), err)
+		return err
+	}
+
+	return nil
+}
+
+// ensureRancherRegistration ensures that the CAPI cluster is registered with Rancher.
+func ensureRancherRegistration(ctx context.Context, r *RancherRegistration, cluster *unstructured.Unstructured) (ctrl.Result, error) {
```
mgianatagh, line 120: is this logic something that could be shared at some point with the existing sequence of steps for registering a cluster with Rancher?

cluster-operator/controllers/capi/verrazzano.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	} else {
+		// need to create a CA secret in admin cluster
+		// get the workload cluster API CA cert
+		caCrt, err := v.getWorkloadClusterCACert(workloadClient)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		// persist the workload API certificate on the admin cluster
+		adminWorkloadCertSecret := &v1.Secret{ObjectMeta: metav1.ObjectMeta{
```
mgianatagh, line 64: do you know if we handle the expiration of this certificate?

mgianatagh (approved):

cluster-operator/controllers/capi/verrazzano.go
jmaron99: Perhaps there's merit to continuously reconcile on relatively moderate cadence - that would pickup and update the cert secret if it changes

jmaron99: Apparently the VMC controller does handle workload cluster cert refresh

cluster-operator/controllers/capi/rancher.go
mgianatagh: oh ok, I didn't realize it was already refactored

PR #7049, opened 2023-09-15
VZ-10969: Multi-cluster with Dex as OIDC provider

lake-of-dreams:
Added a field `dex-url` to the registration secret pointing to Dex in admin cluster and also a field `oidc-provider` to specify whether to use dex or keycloak for oidc provider. Also made the presence of namespace `verrazzano-monitoring` and the secret `verrazzano-es-internal` secret optional so that Multi-cluster can be enabled for cases where Prometheus/Opensearch/Keycloak are not enabled.


cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -380,6 +380,12 @@ func (r *VerrazzanoManagedClusterReconciler) syncMultiClusterCASecret(ctx contex
 
 // mutateManagedClusterCACertsSecret adds and removes managed cluster CA certs to/from the managed cluster CA certs secret
 func (r *VerrazzanoManagedClusterReconciler) mutateManagedClusterCACertsSecret(ctx context.Context, vmc *clustersv1alpha1.VerrazzanoManagedCluster, cacrtSecret *corev1.Secret) error {
+	ns := &corev1.Namespace{}
+	err := r.Client.Get(ctx, types.NamespacedName{Namespace: "default", Name: constants.VerrazzanoMonitoringNamespace}, ns)
```
michael-cico, line 384: Why is `Namespace` set to `default`?  I don't think it's needed when getting a namespace.

michael-cico (approved):

cluster-operator/controllers/vmc/vmc_controller.go
lake-of-dreams: ok, I will remove in a subsequent PR

PR #7052, opened 2023-09-15
vz 10382 set flag in upgrade

AbeHern:
Refactoring functions from install to introduce the --set flag in upgrade


tools/vz/cmd/helpers/command.go
```diff
@@ -240,3 +242,68 @@ func AddManifestsFlags(cmd *cobra.Command) {
...
+	for _, setFlag := range setFlags {
+		pv := strings.Split(setFlag, "=")
+		if len(pv) != 2 {
+			fmt.Fprintf(vzHelper.GetErrorStream(), fmt.Sprintf("Invalid set flag \"%s\" specified. Flag must be specified in the format path=value\n", setFlag))
+			invalidFlag = true
+			continue
+		}
+		if !invalidFlag {
+			path, value := strings.TrimSpace(pv[0]), strings.TrimSpace(pv[1])
+			if !strings.HasPrefix(path, "spec.") {
```
anders-swanson, line 264: why do we add spec if not present?

tools/vz/cmd/helpers/command.go
```diff
@@ -240,3 +242,68 @@ func AddManifestsFlags(cmd *cobra.Command) {
...
+
+	invalidFlag := false
+	for _, setFlag := range setFlags {
+		pv := strings.Split(setFlag, "=")
+		if len(pv) != 2 {
+			fmt.Fprintf(vzHelper.GetErrorStream(), fmt.Sprintf("Invalid set flag \"%s\" specified. Flag must be specified in the format path=value\n", setFlag))
+			invalidFlag = true
+			continue
+		}
+		if !invalidFlag {
```
anders-swanson, line 262: should use either invalidFlag or continue, not sure i see the need for both

tools/vz/cmd/helpers/command.go
AbeHern: These piece were moved from install.go so on a clean install this section would use this.

tools/vz/cmd/helpers/command.go
AbeHern: Artifacts from refactoring from install.go

anders-swanson: What do you mean, why is it an artifact?

anders-swanson: I get that you are creating a common code for both upgrade and install to use set flags

bobdonat: this function was just moved from install.go as is.

tools/vz/cmd/helpers/command.go
bobdonat: Allow a set argument without prefixing with spec.  In the end I think spec. is needed for all set arguments.

tools/vz/cmd/upgrade/upgrade.go
```diff
@@ -282,3 +296,43 @@ func validateCmd(cmd *cobra.Command) error {
 	}
 	return nil
 }
+
+// mergeSetFlagsIntoVerrazzanoResource - determines is any --set flags are used and merges them into the existing Verrazzano resource to be applied at upgrade
```
bobdonat, line 300: typo "is any"

tools/vz/cmd/helpers/overlay.go
```diff
@@ -65,10 +66,21 @@ func MergeYAMLFiles(filenames []string, stdinReader io.Reader) (*unstructured.Un
 
 // MergeSetFlags merges yaml representing a set flag passed on the command line with a
 // verrazano install resource.  A merged verrazzano install resource is returned.
-func MergeSetFlags(gv schema.GroupVersion, vz clipkg.Object, overlayYAML string) (clipkg.Object, *unstructured.Unstructured, error) {
-	baseYAML, err := yaml.Marshal(vz)
-	if err != nil {
-		return vz, nil, err
+func MergeSetFlags(gv schema.GroupVersion, vz clipkg.Object, v1beta1vz *v1beta1.Verrazzano, overlayYAML string, isVerrazzanoResource bool) (clipkg.Object, *unstructured.Unstructured, error) {
```
bobdonat, line 69: This function seems to be overloaded.  I think it would be better to have two different functions instead of passing a flag to determine which args to look at.

AbeHern: I thought about it too, was just trying to avoid duplicating code since a new function would perform the same action with a different parameter (i.e. the function would be the same except it would be using a v1beta1 Verrazzano)

tools/vz/cmd/upgrade/upgrade.go
AbeHern: Fixing now

tools/vz/cmd/helpers/command.go
AbeHern: GetSetArguments() and GenerateYAMLForSetFlags() were both moved from install.go to command.go as is.

tools/vz/cmd/helpers/overlay.go
AbeHern: Added a new function with the v1beta1 being passed @bobdonat @anders-swanson

bobdonat (approved):

PR #7057, opened 2023-09-18
VZ-10850 Build automation that notifies when image sizes have increased

Raaghavcodes:
Fails pipeline and sends alert when an image has increased in size.
Put every new generated image-sizes.txt file into object store.


ci/scripts/check_image_sizes.sh
```diff
@@ -0,0 +1,81 @@
...
+    # Extract image size
+    IMAGESIZE=$(echo "$line" | cut -d ',' -f 2 | cut -d ',' -f 2)
+    # Convert imagesize to an integer & remove spaces
+    IMAGESIZE_INT=$(echo "$IMAGESIZE" | tr -d ' ')
+    # Store the image name and size in separate arrays
+    IMAGENAME_ARRAY+=("$IMAGENAME")
+    IMAGESIZE_ARRAY+=("$IMAGESIZE_INT")
+  done < "$IMAGE_FILE"
+
+  for ((i = 0; i < ${#IMAGENAME_ARRAY[@]}; i++)); do
```
bobdonat, line 26: I think the echoing of each image name and size is not necessary

ci/scripts/check_image_sizes.sh
```diff
@@ -0,0 +1,81 @@
...
+    echo "${IMAGENAME_ARRAY[$i]}:${IMAGESIZE_ARRAY[$i]}"
+  done
+}
+
+oci --region us-phoenix-1 os object get --namespace ${OCI_OS_NAMESPACE} -bn ${OCI_OS_BUCKET} --name ${CLEAN_BRANCH_NAME}/image-list --file ${WORKSPACE}/image-sizes-objectstore.txt
+if [ $? -ne  0 ] ; then
+  echo "image-sizes.txt not found"
```
bobdonat, line 33: I would change to 
echo "${CLEAN_BRANCH_NAME}/image-list not found in object store"

ci/periodic/JenkinsfileDistributions
```diff
@@ -91,6 +92,12 @@ pipeline {
         stage('Build Release Distributions') {
             steps {
                 script {
+                   def commitList = getCommitList()
```
bobdonat, line 95: This should be moved to before the slackSend

Raaghavcodes: I moved this, but it did not work when I moved it, so it is still in the original spot.

ci/periodic/JenkinsfileDistributions
```diff
@@ -91,6 +93,12 @@ pipeline {
         stage('Build Release Distributions') {
             steps {
                 script {
+def commitList =        getCommitList()
```
bobdonat, line 96: indentation - this can fixed when you add the parameter

bobdonat (approved):

PR #7088, opened 2023-09-20
VZ-10805.  Fix CM on uninstall with modules enabled

michael-cico:
- Have Cert-Manager component clean up leader-election resources in kube-system on uninstall and wait for them to be deleted successfully
- Have the new delete handler use Exists() instead of IsInstalled() as trigger to do delete work

### DIFF
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
@@ -254,7 +254,7 @@ func cleanupLeaderElectionResources(compContext spi.ComponentContext) error {
 	return nil
 }
 
-// checkOrphanedResources Checks that the leader-election resources in kube-system are successfully deleted; returns
+// verifyLeaderElectionResourcesDeleted Checks that the leader-election resources in kube-system are successfully deleted; returns
 // a retryable error if the resource still exists
 func verifyLeaderElectionResourcesDeleted(ctx spi.ComponentContext) error {
 	for _, resource := range leaderElectionSystemResources {
--- a/platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
+++ b/platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
@@ -6,10 +6,16 @@ package certmanager
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
+	spi2 "github.com/verrazzano/verrazzano/pkg/controller/errors"
 	"io"
+	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"os"
 	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
 
 	"github.com/verrazzano/verrazzano/pkg/bom"
@@ -41,8 +47,10 @@ const (
 	acmeSolverArg = "--acme-http01-solver-image="
 
 	// Uninstall resources
-	controllerConfigMap = "cert-manager-controller"
-	caInjectorConfigMap = "cert-manager-cainjector-leader-election"
+	controllerConfigMap          = "cert-manager-controller"
+	caInjectorConfigMap          = "cert-manager-cainjector-leader-election"
+	caInjectorLeaderElectionRole = "cert-manager-cainjector:leaderelection"
+	controllerLeaderElectionRole = "cert-manager:leaderelection"
 )
 
 const snippetSubstring = "rfc2136:\n"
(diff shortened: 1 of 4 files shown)


platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
```diff
@@ -214,35 +237,55 @@ func createSnippetWithPadding(padding string) []byte {
...
+	for _, resource := range leaderElectionSystemResources {
+		err := vzresource.Resource{
+			Name:      resource.Name,
+			Namespace: resource.Namespace,
+			Client:    compContext.Client(),
+			Object:    resource.obj,
+			Log:       compContext.Log(),
+		}.RemoveFinalizersAndDelete()
+		if err != nil {
+			return err
```
mbenmclean, line 252: You may want to log this error as it might be an unexpected error and I'm not sure if it gets logged somewhere else. Make sure it's logged further up the call stack and if not log it here.

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
```diff
@@ -214,35 +237,55 @@ func createSnippetWithPadding(padding string) []byte {
...
 
+// resourceExists checks if the specified Object still exists in the cluster; returns true/nil if the object exists,
+// false/nil if not
+func resourceExists(cli client.Client, key client.ObjectKey, obj client.Object) (bool, error) {
+	err := cli.Get(context.TODO(), key, obj)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
```
mbenmclean, line 284: You may want to log this error as it might be an unexpected error and I'm not sure if it gets logged somewhere else. Make sure it's logged further up the call stack and if not log it here.

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
```diff
@@ -214,35 +237,55 @@ func createSnippetWithPadding(padding string) []byte {
...
-	err = vzresource.Resource{
-		Name:      caInjectorConfigMap,
-		Namespace: constants.KubeSystem,
-		Client:    compContext.Client(),
-		Object:    &v1.ConfigMap{},
-		Log:       compContext.Log(),
-	}.Delete()
-	if err != nil {
-		return err
+// checkOrphanedResources Checks that the leader-election resources in kube-system are successfully deleted; returns
```
pfmackin, line 258: name is wrong

pfmackin (dismissed):

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
```diff
@@ -41,8 +47,10 @@ const (
 	acmeSolverArg = "--acme-http01-solver-image="
 
 	// Uninstall resources
-	controllerConfigMap = "cert-manager-controller"
-	caInjectorConfigMap = "cert-manager-cainjector-leader-election"
+	controllerConfigMap          = "cert-manager-controller"
+	caInjectorConfigMap          = "cert-manager-cainjector-leader-election"
+	caInjectorLeaderElectionRole = "cert-manager-cainjector:leaderelection"
```
pfmackin, line 52: should we do the role bindings also?

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
```diff
@@ -76,6 +84,21 @@ var ociDNSSnippet = strings.Split(`ocidns:
...
 
+// Defines a set of leader election resources that Cert-Manager creates that sometimes gets orphaned on uninstall,
+// possibly due to Rancher finalizers
+var leaderElectionSystemResources = []struct {
+	client.ObjectKey
+	obj client.Object
+}{
+	{types.NamespacedName{Name: controllerConfigMap, Namespace: constants.KubeSystem}, &v1.ConfigMap{}},
+	{types.NamespacedName{Name: caInjectorConfigMap, Namespace: constants.KubeSystem}, &v1.ConfigMap{}},
+	{types.NamespacedName{Name: caInjectorLeaderElectionRole, Namespace: constants.KubeSystem}, &rbac.RoleBinding{}},
```
pfmackin, line 95: is this a dup

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
michael-cico: we do, the role bindings have the same name, but they're called separately

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
michael-cico: Fixed

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
michael-cico: Will address in a follow-up

platform-operator/controllers/verrazzano/component/certmanager/certmanager/certmanager.go
michael-cico: The controller will spit out a stack trace, but will address this in a follow-up as a retryable error

mbenmclean (approved):

PR #7105, opened 2023-09-21
VZ-11001: Create ClusterIssuer to issue the opensearch self signed certificates

karan56625:
Earlier, we were using verrazzano-cluster-issuer to issue the certificates required for the internal communication among Opensearch Nodes. As verrazzano-cluster-issuer ClusterIssuer could be CA or ACME depending on the Verrazzano CR, we need a separate Issuer for Opensearch Certificates. The reason being, ClusterIssuer type ACME will not accept the Certificate Request for Opensearch certificates as in those Cert request, k8s service DNS are being mentioned as DNS. So we need an other issuer that will be always CA type to sign the Opensearch certs.

Changes:

opensearch-self-signed-issuer Issuer will be created in cert-manager namespace to bootstrap the CA.
- opensearch-ca-certCA Certitifcate in cert-manager namespace that will be used to create CA Issuer.
opensearch-root-issuer CA ClusterIssuer that will sign the Opensearch certificates. ClusterIssuer is being created in this PR instead of Issuer, because it needs to sign the certificates from verrazzano-logging and verrazzano-monitoring namespace.


platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
```diff
@@ -2,7 +2,40 @@
...
 # Required for the security plugin in OpenSearch
 
-# Certificate name and secret name are kept same and will be generated based on opensearchCluster.name variable
+# Create a selfsigned Issuer in order to create a root CA certificate for
+# signing certificates required for the security plugin in OpenSearch.
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: opensearch-self-signed-issuer
+  namespace: cert-manager
```
michael-cico, line 11: We shouldn't be putting things in the cert-manager namespace as it may not be ours.  We now support customer-managed cert-manager instances.

Shouldn't it be in the verrazzano-system namespace?

platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
```diff
@@ -2,7 +2,40 @@
...
+  duration: 2160h0m0s
+  renewBefore: 360h0m0s
+  issuerRef:
+    name: opensearch-self-signed-issuer
+  commonName: "ca.opensearch"
+  isCA: true
+---
+# Create an ClusterIssuer that uses the above generated CA certificate to issue certs
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
```
michael-cico, line 32: Do you need a ClusterIssuer?  Is an Issuer good enough?

michael-cico: You can use a ClusterIssuer if it will need to manage certs across the whole cluster, if the certs are going to be local to a namespace it should be local to the namespace.

platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
michael-cico: You can probably put all this stuff in the operator namespace, if verrazzano-system is not appropriate.

platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
karan56625: Actual we need to generate certificate in two namespace. One is Verrazzano-logging and the other one is Verrazzano-monitoring.  Namespace issuer will not be able to do that?? or is there any way?

platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
karan56625: To create the Cluster Issue, we need the CA certificate to be in the `Cluster Resource Namespace` and by default that is the same namespace where we install cert-manager.  I guess, if we put these in some other namespace , ClusterIssuer will not gonna use that and could fail.

platform-operator/thirdparty/charts/opensearch-operator/templates/verrazzano/opensearch-certificate.yaml
michael-cico: Then parameterize it here and pass it in when we invoke Helm on this chart.

michael-cico: Since this PR is just the chart changes, probably you can just parameterize the template here and add a default to `values.yaml`, and do the `clusterIssuer` overrides work in the other PR.

platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -214,7 +214,14 @@ func appendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
 	if err != nil {
 		return kvs, ctx.Log().ErrorfNewErr("Failed to build ingress overrides: %v", err)
 	}
-
+	clusterResourceNamespace := constants.CertManagerNamespace
```
michael-cico, line 217: Looks good 👍

gaurav-05 (approved):

PR #7115, opened 2023-09-21
VZ-10840 change internal module versions to match Verrazzano version

mbenmclean:
This change changes the internal modules versions in the catalog to match the Verrazzano version. The catalog is generated  at build time similar to the how the bom is generated. This change also moved the catalog into the manifests directory.

### DIFF
--- a/platform-operator/Makefile
+++ b/platform-operator/Makefile
@@ -128,10 +128,10 @@ docker-clean:
 	rm -rf ${DIST_DIR}
 
 .PHONY: docker-build
-docker-build: generate-bom go-build-linux docker-build-common
+docker-build: generate-bom generate-catalog go-build-linux docker-build-common
 
 .PHONY: docker-build-debug
-docker-build-debug: generate-bom go-build-linux-debug docker-build-common
+docker-build-debug: generate-bom generate-catalog go-build-linux-debug docker-build-common
 
 .PHONY: docker-build-common
 docker-build-common:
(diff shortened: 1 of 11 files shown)


platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -9,7 +9,7 @@ modules:
   - name: cert-manager-webhook-oci
     version: 0.1.0
   - name: cluster-issuer
-    version: 2.0.0
+    version: VERRAZZANO_VERSION
```
pfmackin, line 12: Where are you evaluating this?

tools/scripts/generate_catalog.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
michael-cico, line 1: Could we combine this with the generate_bom.sh script?  i.e. consolidate the processing into one place?

Don't have to for this pass.

platform-operator/manifests/catalog/catalog.yaml
michael-cico: or provide a path-to-catalog argument to the operator like we do with the BOM.  Then you can point that to a locally generated catalog also.

platform-operator/Makefile
```diff
@@ -225,3 +225,8 @@ generate-bom:
 	mkdir out || true
 	../tools/scripts/generate_bom.sh verrazzano-bom.json ${VERRAZZANO_VERSION} ${VERRAZZANO_APPLICATION_OPERATOR_IMAGE} ${VERRAZZANO_CLUSTER_OPERATOR_IMAGE} ${VERRAZZANO_AUTHPROXY_IMAGE} ${DOCKER_IMAGE_NAME} ${DOCKER_IMAGE_TAG} out/generated-verrazzano-bom.json
 
+.PHONY: generate-catalog
```
michael-cico, line 228: if we can consolidate this with the BOM script/target and have one that would be good.  Can be a follow-up

platform-operator/manifests/catalog/catalog.yaml
mbenmclean: Hmm, well the VPO code don't know what the version is. I could some code to check if the version is VERRAZZANO_VERSION and if so use the version from the BOM. That will allow you to set the VZ_INSTALL_VERSION in the debugger.

tools/scripts/generate_catalog.sh
mbenmclean: I thought about them having them together but decided I preferred separate. It allows for more flexibility.

michael-cico (approved):

PR #7116, opened 2023-09-21
VZ-11008

marcogelle:


application-operator/mcagent/mcagent_syncer.go
```diff
@@ -177,10 +195,52 @@ func (s *Syncer) updateVMCStatus() error {
 	// If Thanos is disabled, we want to empty the host so Prometheus federation returns
 	vmc.Status.ThanosQueryStore = thanosAPIHost
 
+	// Update the Kubernetes version on the VMC
+	k8sVersion, err := s.getWorkloadK8sVersion()
+	if err != nil {
+		return fmt.Errorf("Failed to get Kubernetes information to update VMC %s: %v", vmcName, err)
```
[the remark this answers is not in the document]
marcogelle: The surrounding code in this `updateVMCStatus` function also just returns reformatted errors, so I just followed that pattern.
If you or others think I should change it to log an error here still, I can change it though.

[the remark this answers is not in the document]
davisgomes: I guess that's okay as long as these errors get logged somewhere else. I don't want them to get lost.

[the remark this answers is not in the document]
marcogelle: Yup, `updateVMCStatus` is only called in one place, which logs the returned errors here: https://github.com/verrazzano/verrazzano/blob/master/application-operator/mcagent/mcagent.go#L166.

davisgomes (approved):

application-operator/mcagent/mcagent.go
```diff
@@ -189,10 +190,18 @@ func (r *Reconciler) doReconcile(ctx context.Context, agentSecret corev1.Secret)
 // updateMCAgentStateConfigMap updates the managed cluster name and cattle agent hash in the
 // agent state config map if those have changed from what was there before
 func (r *Reconciler) updateMCAgentStateConfigMap(ctx context.Context, managedClusterName string, cattleAgentHashValue string) error {
+	// create the ConfigMap's namespace if it doesn't already exist
+	mcAgentStateNamespace := corev1.Namespace{}
```
anders-swanson, line 194: why are we needing to add this?

michael-cico (approved):

application-operator/mcagent/mcagent_syncer.go
```diff
@@ -49,6 +52,21 @@ var (
...
+
+func defaultGetDiscoveryClientFunc() (discovery.DiscoveryInterface, error) {
+	config, err := k8sutil.GetConfigFromController()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Kubeconfig for this workload cluster: %v", err)
+	}
+	return discovery.NewDiscoveryClientForConfig(config)
+}
+
+func setDiscoveryClientFunc(f func() (discovery.DiscoveryInterface, error)) {
```
anders-swanson, line 66: is there a need for this function? why can't you do `getDiscoveryClientFunc = myfunc`?

application-operator/mcagent/mcagent_syncer.go
```diff
@@ -177,10 +195,52 @@ func (s *Syncer) updateVMCStatus() error {
...
+	}
+	vmc.Status.Verrazzano.Version = vzVersion
+
 	// update status of VMC
 	return s.AdminClient.Status().Update(s.Context, &vmc)
 }
 
+// getWorkloadK8sVersion retrieves the current Kubernetes version on this managed cluster
+func (s *Syncer) getWorkloadK8sVersion() (string, error) {
+	discoveryClient, err := getDiscoveryClientFunc()
```
anders-swanson, line 218: is there a reason we can't initialize the `Syncer` with an instance of the discovery client?

application-operator/mcagent/mcagent.go
marcogelle: I guess this could have been a separate PR, but there was a bug with if you install a minimal VZ profile (https://verrazzano.io/latest/docs/setup/mc-install/multicluster/#minimal-managed-cluster-configurations) on the managed cluster.

In this case, the `verrazzano-mc` namespace would never be created, and the verrazzano agent code gets stuck here, never updating the VMC.

application-operator/mcagent/mcagent_syncer.go
```diff
@@ -177,10 +195,52 @@ func (s *Syncer) updateVMCStatus() error {
...
+}
+
+// getWorkloadVZVersion retrieves the current VZ version on this managed cluster,
+// as reported by the VZ CR's status
+func (s *Syncer) getWorkloadVZVersion() (string, error) {
+	vzList := &v1beta1.VerrazzanoList{}
+	if err := s.LocalClient.List(s.Context, vzList, &client.ListOptions{}); err != nil {
+		return "", fmt.Errorf("error listing Verrazzanos: %v", err)
+	}
+	if len(vzList.Items) != 1 {
```
anders-swanson, line 237: Is it ever valid for there to be 0 verrazzanos on the managed cluster?

application-operator/mcagent/mcagent.go
anders-swanson: OK, thanks! it's fine to be a part of this PR

application-operator/mcagent/mcagent_syncer.go
marcogelle: I think that would also work. I can change it if you'd like.

My reasoning behind this function was copying a similar pattern I saw elsewhere in the repo, such as https://github.com/verrazzano/verrazzano/blob/mmgellec/VZ-11008/cluster-operator/controllers/vmc/sync_agent_secret.go#L27.

application-operator/mcagent/mcagent_syncer.go
anders-swanson: it seems like if there were, for some reason, the verrazzano version would be nil

application-operator/mcagent/mcagent_syncer.go
anders-swanson: your choice, it seemed unnecessary to me

application-operator/mcagent/mcagent_syncer.go
marcogelle: No reason, I just hadn't thought of that.

I suppose initializing the `Syncer` with that would reduce the amount of repeated work with creating discover clients, so it seems like a good change. I could use some confirmation if you think I should work on testing that out now though.

anders-swanson: it doesn't need to block this pr

application-operator/mcagent/mcagent_syncer.go
marcogelle: I see. I'll plan on leaving it then.

application-operator/mcagent/mcagent_syncer.go
marcogelle: It should be possible for there to be 0 verrazzanos on a managed cluster in general. However, I don't think this agent code would run in that case, unless it is possible for the verrazzano-cluster-agent to run without a Verrazzano installation on that cluster.

davisgomes: At this point, it is (nearly) impossible for that to be the case.

application-operator/mcagent/mcagent_syncer.go
marcogelle: Ok, I want to look into this as a follow-up PR if I can get this one merged soon

application-operator/mcagent/mcagent_syncer.go
marcogelle: What does "nearly" mean? I suppose that means I should probably at least take the 0 verrazzanos case into account here?

anders-swanson: My question was somewhat rhetoric, sorry. I mean in general, is it better to return an error here, or nil/empty string if this happens? What would be more defensible

marcogelle: I would want to return nil in the 0 Verrazzanos case.
The reasoning is that returning an error would prevent the VMC on the cluster from actually being updated with all the previous changes. See this line of code: https://github.com/verrazzano/verrazzano/blob/mmgellec/VZ-11008/application-operator/mcagent/mcagent_syncer.go#L213.

marcogelle: Perhaps since the 0 Verrazzano case is practically impossible, I can plan to include this change in a follow-up, along with the changes about the `Syncer` initialization that were mentioned above.

anders-swanson: for sure, i'm just being nit-picky

marcogelle: I appreciate it. Nit-picky is better than not reviewing the code well

anders-swanson (approved):

PR #7121, opened 2023-09-21
PR to add support for internal module versions running in the debugger

mbenmclean:
This change adds support for the VZ_INSTALL_VERSION envvar when running from the debugger. Internal modules will use that version when running in the debugger.


platform-operator/apis/verrazzano/validators/validate.go
```diff
@@ -121,7 +121,7 @@ func GetCurrentBomVersion() (*semver.SemVersion, error) {
 		if len(os.Getenv("VZ_INSTALL_VERSION")) > 0 {
 			v = os.Getenv("VZ_INSTALL_VERSION")
 		} else {
-			v = "1.0.1"
+			v = "0.0.0"
```
michael-cico, line 124: what's this used for?  I see it was there before but seems weird.

mbenmclean: It's like a fall back if you run in the debugger and don't set the envvar

mbenmclean: Just though 0.0.0 was more appropriate

michael-cico (dismissed):

pfmackin (approved):

PR #7131, opened 2023-09-22
Make helm logging changes to reduce Rancher Helm Install/Upgrade Mess…

George-Aeillo:
…age Noise

### DIFF
--- a/pkg/helm/helmcli.go
+++ b/pkg/helm/helmcli.go
@@ -153,7 +153,7 @@ func Upgrade(log vzlog.VerrazzanoLogger, releaseName string, namespace string, c
 
 		rel, err = client.Run(chart, vals)
 		if err != nil {
-			log.ErrorfThrottledNewErr("Failed running Helm command for release %s: %v",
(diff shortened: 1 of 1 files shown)


pkg/helm/helmcli.go
```diff
@@ -153,13 +153,13 @@ func Upgrade(log vzlog.VerrazzanoLogger, releaseName string, namespace string, c
 
 		rel, err = client.Run(chart, vals)
 		if err != nil {
-			log.Errorf("Failed running Helm command for release %s: %v",
+			log.ErrorfThrottledNewErr("Failed running Helm command for release %s: %v",
```
michael-cico, line 156: I think this should be `ErrorfThrottled`, `ErrorfThrottledNewErr` returns a new error.  The first just records the error and throttles it.  We return the error on the next line.

George-Aeillo: Ok! Thanks!

pfmackin (approved):

PR #7135, opened 2023-09-22
VZ-11085.  Remove config mappings and replace with watches for DEX, ArgoCD, and Jaeger

michael-cico:
Remove config mappings and replace with watches where possible for DEX, ArgoCD, and Jaeger.
- change some dependencies to match module-installed watches
- replace some CertManager dependencies with ClusterIssuer dependencies


platform-operator/controllers/verrazzano/component/argocd/argocd_component.go
```diff
@@ -60,7 +60,7 @@ func NewComponent() spi.Component {
 			ValuesFile:                filepath.Join(config.GetHelmOverridesDir(), "argocd-values.yaml"),
 			AppendOverridesFunc:       AppendOverrides,
 			Certificates:              certificates,
-			Dependencies:              []string{networkpolicies.ComponentName, common.IstioComponentName, nginx.ComponentName, cmconstants.CertManagerComponentName, keycloak.ComponentName},
+			Dependencies:              []string{networkpolicies.ComponentName, common.IstioComponentName, nginx.ComponentName, cmconstants.ClusterIssuerComponentName, keycloak.ComponentName},
```
pfmackin, line 63: did you mean to change this?

pfmackin (approved):

platform-operator/controllers/verrazzano/component/argocd/argocd_component.go
michael-cico: Yes, I think that people really mean to wait for the issuer to be ready.  We may not install cert manager ourselves (customer-managed case).  The `clusterIssuer` already has a dependency on `certManager`.

I'm actually doubtful it needs this dependency anyways, but don't want to rock the boat too much.

michael-cico: I'll see how the testing goes

PR #7136, opened 2023-09-22
VZ-11119: Use clusterctl utility instead of the API's

mgianatagh:


platform-operator/Dockerfile
```diff
@@ -27,7 +27,10 @@ RUN microdnf install -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs ope
     && groupadd -r verrazzano \
     && useradd --no-log-init -r -m -d /verrazzano -g verrazzano -u 1000 verrazzano \
     && mkdir /home/verrazzano \
-    && chown -R 1000:verrazzano /home/verrazzano
+    && chown -R 1000:verrazzano /home/verrazzano \
+    && curl --proto "=https" -L https://github.com/verrazzano/cluster-api/releases/download/v1.4.6/clusterctl-linux-amd64 -o /usr/local/bin/clusterctl \
+    && chown verrazzano:verrazzano /usr/local/bin/clusterctl \
```
bobdonat, line 32: Do the owner and permissions need to be different than kubectl and istioctl?  Wondering if the chown and chmod are not needed.

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
```diff
@@ -261,18 +273,11 @@ func (c clusterAPIComponent) PreUninstall(_ spi.ComponentContext) error {
 }
 
 func (c clusterAPIComponent) Uninstall(ctx spi.ComponentContext) error {
-	capiClient, err := capiInitFunc("")
-	if err != nil {
-		return err
-	}
+	cmd := exec.Command("clusterctl", "delete", "--all", "--include-namespace")
+	// Temporarily ignore the return status until a clusterctl version issue is resolved
```
bobdonat, line 277: Would this be blocking the merge?

platform-operator/Dockerfile
mgianatagh: The chmod was needed to add execute.

I can remove the chown.  I was making it the same as the VPO image.

mgianatagh: removed, clusterctl is now the same as istioctl

platform-operator/controllers/verrazzano/component/clusterapi/clusterapi_component.go
mgianatagh: I didn't feel it would, because we will fix the build of clusterctl before this get released.

bobdonat: ok, i thought is was causing the delete to not happen.

bobdonat (approved):

PR #7139, opened 2023-09-22
VZ-11094 Delete CAPI RBAC resources before upgrade

mbenmclean:
This change deletes and removes the finalizer from CAPI RBAC resources before upgrade in order to prevent the Rancher finalizer from interfering with the delete.


platform-operator/controllers/verrazzano/component/clusterapi/fake_api_client_test.go
```diff
@@ -4,13 +4,19 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/cluster-api/cmd/clusterctl/api/v1alpha3"
 	"sigs.k8s.io/cluster-api/cmd/clusterctl/client"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
 	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/tree"
 )
 
 type FakeClusterAPIClient struct{}
 
+type FakeComponentCLient struct{}
```
michael-cico, line 18: Nit, typo in capitlization `CLient`.  Maybe another pass

michael-cico (approved):

PR #7140, opened 2023-09-22
VZ-11008: Follow-up to cleanup VMC extensions for K8s and VZ versions

marcogelle:
Follow-up to https://github.com/verrazzano/verrazzano/pull/7116.

Addresses the previous comments made:

- Initialize the `Syncer` to contain a `DiscoveryClient`
- Account for the case that there are 0 `verrazzanos` on the workload cluster

### DIFF
--- a/application-operator/mcagent/mcagent.go
+++ b/application-operator/mcagent/mcagent.go
@@ -113,14 +113,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (r *Reconciler) doReconcile(ctx context.Context, agentSecret corev1.Secret) error {
 	managedClusterName := string(agentSecret.Data[constants.ClusterNameData])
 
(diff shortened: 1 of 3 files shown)


application-operator/mcagent/mcagent.go
```diff
@@ -141,6 +141,13 @@ func (r *Reconciler) doReconcile(ctx context.Context, agentSecret corev1.Secret)
 	}
 	s.AdminClient = adminClient
 
+	// Create the discovery client for the managed cluster
```
anders-swanson, line 144: i would do this at time of Syncer initialization

marcogelle: sure, I'll change that

marcogelle: I made a commit to change this. Let me know if this isn't what you had in mind.

anders-swanson (approved):

PR #7144, opened 2023-09-25
VZ-10256: fluentbit oci logging analytics output component

adiforluls:


platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func checkAPISecret(overrides []string) error {
+	var apiSecret interface{}
+	var err error
+	client, err := getClient()
+	if err != nil {
+		return err
+	}
+	for _, overrideYAML := range overrides {
```
michael-cico, line 193: nit, could probably break this loop into a helper method to find the value, improves the readability.

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	return c.HelmComponent.PreInstall(ctx)
+}
+
+func (c fluentbitOCILoggingAnalyticsOutput) PreUpgrade(ctx spi.ComponentContext) error {
+	return c.HelmComponent.PreUpgrade(ctx)
+}
+
+func (c fluentbitOCILoggingAnalyticsOutput) Reconcile(ctx spi.ComponentContext) error {
+	installed, err := c.IsInstalled(ctx)
```
michael-cico, line 69: you probably need to implement `IsEnabled` to participate in the module lifecycle.

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+	if installed {
```
michael-cico, line 73: Is this just to fix up anything that may have drifted?  I'm not sure `Reconcile` is called anymore with Modules, I have to check.

platform-operator/controllers/verrazzano/component/common/fluentOperator_utils.go
```diff
@@ -48,3 +61,37 @@ func RenderTemplate(templatePath string, args map[string]interface{}, outputFile
...
+			return "", err
+		}
+		if apiSecretName != nil {
+			return apiSecretName.(string), nil
+		}
+	}
+	return "", nil
+}
+
+func IsAuthTypeUserPrincipal(ctx spi.ComponentContext) (bool, error) {
```
michael-cico, line 83: General question, is the auth secret when user principals are involved similar to what is done with other components that have OCI integration?

Also, are we following similar conventions for workload/instance auth as we are using elsewhere?   I think that there's been work to support workload auth in a few other areas as well.

adiforluls: The format of the auth secret is the same as fluentd oci logging integration's auth secret. The same secret can be applied/used here.

For oke workload auth, I think some work was done for thanos https://github.com/thanos-io/objstore/compare/main...fredtibbitts:objstore:oke-workload-identity. I think the convention is similar https://github.com/thanos-io/objstore/pull/64/files

Instance principal mode of auth is used in oci logging integration that we have today. For that, you don't need to pass in any additional auth config like in the case of user_principal and oke_workload_identity. That behaviour is maintained here.

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
adiforluls: Added `Reconcile` because something like this exists in most helm components. I can remove this if this is legacy code.

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
adiforluls: A general problem for `IsEnabled` is that the component doesn't have an `enabled` field in it's spec, since just setting it to true will result in nothing. 

The intention was to search for `system.logGroupId` and `application.logGroupId` in the overrides YAMLs. If both are absent, that should result in component being disabled. But the funcs that search and extract overrides for values can return errors, which need to be handled appropriated if the operation should be retried. Since `IsEnabled` can only return `true` or `false`, this type of behaviour might not fly. I might need to add an `enabled` flag for this component now I'm thinking.

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
michael-cico: I just wasn't sure if this behavior was being depended on for anything.  Only handful of components really implement it, and there's only a few I know of that are needed (keycloak and jaeger I think).

platform-operator/controllers/verrazzano/component/fluentbitocilaoutput/fluentbitocilaoutput_component.go
michael-cico: Yeah, I'm not sure how you can get around it.

We can have a zoom on it if needed.

PR #7145, opened 2023-09-25
VZ-9875: refresh component images

lake-of-dreams:
Update more component images in continuation of #7142


mgianatagh (dismissed):

platform-operator/verrazzano-bom.json
```diff
@@ -93,7 +93,7 @@
           "images": [
             {
               "image": "cert-manager-webhook-oci",
-              "tag": "v0.1.0-20230526160243-7a73e30",
+              "tag": "v0.1.0-20230925072151-7a73e30",
```
michael-cico, line 96: I think we'll need to update the catalog to add `-1` to each of the affected components/modules, if the `m.m.p` version hasn't changed.

lake-of-dreams: ok will do in a subsequent PR

bhatpmk (approved):

PR #7180, opened 2023-09-26
VZ-11052 Compare branch with master to ensure catalog is up to date

mbenmclean:
This change adds a unit test to the catalog to compare the target branch (master for now) with the feature branch. It gets the BOM and catalog from the feature branch and the target branch and checks that if the bom entry for a module has been updated, the catalog version has been updated as well.


platform-operator/Makefile
```diff
@@ -167,6 +167,8 @@ endif
 #
 .PHONY: unit-test
 unit-test: go-install
+#	env BRANCH_NAME=$(git branch --show-current)
```
michael-cico, line 170: Should this be removed?

platform-operator/controllers/verrazzano/catalog/catalog_test.go
```diff
@@ -4,19 +4,30 @@
...
-	"testing"
+
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/stretchr/testify/assert"
 )
 
 const catalogPath = "../../../manifests/catalog/catalog.yaml"
 const bomPath = "../../../verrazzano-bom.json"
 const testBOMPath = "../testdata/test_bom.json"
+const remoteBOMPath = "https://raw.githubusercontent.com/verrazzano/verrazzano/master/platform-operator/verrazzano-bom.json"
```
michael-cico, line 29: once we branch, maybe we should parameterize the target branch name through an env var or something.  It will likely have to be specific to each branch, e.g., `release-1.7`.

platform-operator/controllers/verrazzano/catalog/catalog_test.go
```diff
@@ -100,3 +124,107 @@ func TestGetVersionForAllRegistryComponents(t *testing.T) {
...
+		assert.NotNil(t, resp)
+		bodyRaw, err = io.ReadAll(resp.Body)
+		assert.NoError(t, err)
+		assert.NotEmpty(t, bodyRaw)
+		remoteCatalog, err := NewCatalogfromYAML(bodyRaw)
+		assert.NoError(t, err)
+		assert.NotNil(t, remoteCatalog)
+
+		for _, module := range localCatalog.Modules {
+			if vzString.SliceContainsString(modulesNotInBom, module.Name) ||
```
michael-cico, line 178: might be good to break this out into a separate fn for readability

platform-operator/controllers/verrazzano/catalog/catalog_test.go
```diff
@@ -100,3 +124,107 @@ func TestGetVersionForAllRegistryComponents(t *testing.T) {
...
+				assert.Truef(t, localVersion.IsGreatherThan(remoteVersion),
+					"BOM entry for module %s has been updated so local catalog version %s should be greater than remote catalog version %s",
+					module.Name, localVersion.ToString(), remoteVersion.ToString())
+			}
+		}
+	}
+}
+
+func checkBOMModifiedInBranch(t *testing.T) bool {
+	_, err := exec.Command("git", "config", "--add", "remote.origin.fetch", "+refs/heads/master:refs/remotes/origin/master").Output()
```
michael-cico, line 220: similar comment, probably wherever we refer to `master` should be parameterized or at least use a var that can be changed for each release branch name.

michael-cico (approved):

agv426 (approved):

platform-operator/Makefile
mbenmclean: Will remove in follow up

platform-operator/controllers/verrazzano/catalog/catalog_test.go
mbenmclean: Will be doing this

platform-operator/controllers/verrazzano/catalog/catalog_test.go
mbenmclean: Ok can break it out in follow-up

platform-operator/controllers/verrazzano/catalog/catalog_test.go
mbenmclean: Will do this in follow up

PR #7191, opened 2023-09-26
VZ-11009: Extend VMC states for CAPI Clusters

marcogelle:
Extends the `VerrazzanoManagedCluster` custom resource's status fields:

- VMC API Changes:
  - Adds a `clusterRef` field to refer to a CAPI cluster
  - Adds more possible values to the `state` field, in the case that a CAPI cluster exists for the VMC
- Adds to the VMC controller to reconcile the `state` field
- Moves the CAPI Cluster `GroupVersionKind` to a common package `cluster-operator/internal/capi` so it can be called in the VMC controller


cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -637,23 +682,43 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
...
+// getCAPIClusterPhase returns the phase reported by the CAPI Cluster CR which is referenced by clusterRef.
+func (r *VerrazzanoManagedClusterReconciler) getCAPIClusterPhase(clusterRef *clustersv1alpha1.ClusterReference) (clustersv1alpha1.StateType, error) {
+	// Get the CAPI Cluster CR
+	cluster := &unstructured.Unstructured{}
+	cluster.SetGroupVersionKind(capi.GVKCAPICluster)
+	clusterNamespacedName := types.NamespacedName{
+		Name:      clusterRef.Name,
+		Namespace: clusterRef.Namespace,
+	}
+	if err := r.Get(context.TODO(), clusterNamespacedName, cluster); err != nil {
```
jmaron99, line 696: this code should probably only return an error if it is not a NotFound error?  If it doesn't find a CAPI cluster it simply shouldn't return a state, unless a clusterRef implies the existence of a CAPI cluster?

marcogelle: I agree that if it doesn't find a CAPI cluster when expecting one, then it should return no state and return the `NotFound` error.
In the case that the error is something else besides `NotFound`, what are you suggesting? To return no state again, but with a `nil` error?

jmaron99: I'm saying that:
- if the error is a notfound error, return "", nil
- if the error is not a notfound error, return "", err

something like:

`	if err != nil {
		if errors.IsNotFound(err) {
			return "", nil
		}
		return "", err
	}
`

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -637,23 +682,43 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
...
+		return "", err
 	}
 
-	// Replace the existing status conditions and state with the conditions generated from this reconcile
-	for _, genCondition := range vmc.Status.Conditions {
-		r.setStatusCondition(existingVMC, genCondition, genCondition.Type == clustersv1alpha1.ConditionManifestPushed)
+	// Get the state
+	phase, found, err := unstructured.NestedString(cluster.Object, "status", "phase")
+	if !found {
+		err = fmt.Errorf("could not find field status.phase field inside cluster %s: %v", clusterNamespacedName, err)
```
anders-swanson, line 703: i'd log a progress message here and return an empty phase

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -637,23 +682,43 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
...
+		return "", err
+	}
+	if err != nil {
```
anders-swanson, line 706: same as above, log progress and return empty phase (no error)

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -623,6 +625,49 @@ func (r *VerrazzanoManagedClusterReconciler) setStatusCondition(vmc *clustersv1a
...
+		r.updateStateFromLastAgentConnectTime(vmc)
+		return nil
+	}
+
+	// If there is an underlying CAPI cluster, set the state field according to the phase of the CAPI cluster.
+	capiClusterPhase, err := r.getCAPIClusterPhase(vmc.Status.ClusterRef)
+	if err != nil {
+		return err
+	}
+	vmc.Status.State = capiClusterPhase
```
anders-swanson, line 664: what do we do if phase is empty string?

cluster-operator/controllers/vmc/vmc_controller.go
```diff
@@ -637,23 +682,43 @@ func (r *VerrazzanoManagedClusterReconciler) updateStatus(ctx context.Context, v
...
+	switch state := clustersv1alpha1.StateType(phase); state {
+	case clustersv1alpha1.StatePending,
+		clustersv1alpha1.StateProvisioning,
+		clustersv1alpha1.StateProvisioned,
+		clustersv1alpha1.StateDeleting,
+		clustersv1alpha1.StateUnknown,
+		clustersv1alpha1.StateFailed:
+		return state, nil
+	default:
+		return "", fmt.Errorf("retrieved an invalid ClusterAPI Cluster phase of %s", state)
```
anders-swanson, line 720: Consider defaulting to PhaseUnknown?

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: That makes sense, I can make that change.
With that change, I'm guessing I may have to make some changes to also address
> If it doesn't find a CAPI cluster it simply shouldn't return a state, unless a clusterRef implies the existence of a CAPI cluster.

jmaron99: I think if there is not CAPI cluster state returned you don't update the status

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: To confirm, that would look something like this?
```
r.log.Progress("message here")
return "", nil
```

anders-swanson: yeah exactly

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: That does make sense to me. I can change that

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: I see, I'll look into changing that

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: Right now, it sets the state to an empty string. I think I'll make a change to implement Jon's suggestion here: https://github.com/verrazzano/verrazzano/pull/7191/files#r1339086693.

cluster-operator/controllers/capi/capi_cluster_controller.go
```diff
@@ -45,15 +46,9 @@ type CAPIClusterReconciler struct {
...
-var gvk = schema.GroupVersionKind{
-	Group:   "cluster.x-k8s.io",
-	Version: "v1beta1",
-	Kind:    "Cluster",
-}
-
 func CAPIClusterClientObject() client.Object {
 	obj := &unstructured.Unstructured{}
-	obj.SetGroupVersionKind(gvk)
+	obj.SetGroupVersionKind(capi.GVKCAPICluster)
```
desagar, line 51: The GVKCAPICluster is in the internal/capi folder and has the same package name as this controller - is this an accepted pattern in Golang to have the same package name in multiple directories, one external and one internal?

anders-swanson: I'm not aware, it's a different package

marcogelle: https://go.dev/blog/package-names
According to this it is accepted, if you search for "packages in different directories can have the same name."
However, there is a disclaimer, if you search for "Avoid unnecessary package name collisions" in the same web page.

marcogelle: Especially since this internal/capi package already existed before my PR, I'm tempted to keep it unless you all think I should change it.

anders-swanson (dismissed):

cluster-operator/controllers/vmc/vmc_controller.go
marcogelle: I have changed this to default to `StateUnknown`, but upon manually testing this part of the code, I notice that returning an error results in hitting this line: https://github.com/verrazzano/verrazzano/pull/7191/files#diff-e43c1f5486c03e16914b7dcf76feb73c30b07dc75db40665e026de932d31cfc5R661, which leads to not updating the VMC state to `StateUnknown`.

I'm thinking of changing the order of `if` blocks in the part of code that I linked. Does anything pop out as wrong to you with this?
```
if capiClusterPhase != "" {
    vmc.Status.State = capiClusterPhase
}
if err != nil {
    return err
}
```

anders-swanson: you don't need to return an error

marcogelle: Do you think the default case maybe needs a Progress message then? Saying that an invalid CAPI phase was retrieved?

marcogelle: I'm thinking now to do something like this under the default case:
```
r.log.Progressf("retrieved an invalid ClusterAPI Cluster phase of %s", state)
return clustersv1alpha1.StateUnknown, nil
```

anders-swanson: sure

marcogelle: I pushed the commit for this change and manually verified that it works as expected.

anders-swanson (approved):

cluster-operator/controllers/capi/capi_cluster_controller.go
desagar: I do think it should be changed because it is confusing to the reader when you import the _capi_ package within the capi package. I leave it to you/Anders/Jon etc. to determine when and how to change it.

marcogelle: I see. I'd like to look into changing this in a future PR. I still at least one more coming having to do with VMC extensions, which will touch the CAPI controller code.

jmaron99: File a JIRA to assess/track that future change, please :)

marcogelle: Sure 👍🏽

marcogelle:
> LGTM. Let me know when you have completed testing (just in case there are some additional commits).

I'm done with my manual testing; I just need to get a clean run of the push-triggered tests.

kminder (approved):
Approved.

PR #7217, opened 2023-09-27
Fix Null Pointer Exception 

George-Aeillo:


pfmackin (approved):

platform-operator/controllers/verrazzano/controller/reconciler.go
```diff
@@ -89,7 +89,7 @@ func (r Reconciler) Reconcile(controllerCtx controllerspi.ReconcileContext, u *u
 	// If an upgrade is pending, do not reconcile; an upgrade is pending if the VPO has been upgraded, but the user
 	// has not modified the version in the Verrazzano CR to match the BOM.
 	if upgradePending, err := r.isUpgradeRequired(actualCR); upgradePending || err != nil {
-		if metricsexporter.IsMetricError(err) {
+		if err != nil && metricsexporter.IsMetricError(err) {
```
pfmackin, line 92: better to fix IsMetricError to check for nil, but merge this since it fixes issue.

pfmackin: Actually this metric should not be reported

PR #7246, opened 2023-09-28
VZ-11052 Compare charts with remote to ensure catalog up to date

mbenmclean:
This change adds optional chart and valuesOverrides fields to the catalog.
It also adds a test that generate file diffs from master and cross references it with all of the module chart and values overrides files from the catalog.
If there is an updated file that's in the catalog, the test ensures that the corresponding module version has also been updated.


platform-operator/helm_config/overrides/prometheus-node-exporter-values.yaml
```diff
@@ -18,7 +18,7 @@ prometheus:
       - action: replace
         sourceLabels:
         - job
-        targetLabel: job
+        targetLabel: jobfdsaf
```
michael-cico, line 21: typo?

mbenmclean: meant to take that out. Fixed

platform-operator/verrazzano-bom.json
```diff
@@ -468,7 +468,7 @@
     },
     {
       "name": "weblogic-operator",
-      "version": "4.1.2",
+      "version": "4.1.2-1",
```
mbenmclean, line 471: helm linter made me update the copyright on weblogic file so I have to rev this unfortunately

go.mod
```diff
@@ -187,7 +187,7 @@ require (
 	k8s.io/gengo v0.0.0-20220902162205-c0856e24416d // indirect
 	k8s.io/klog/v2 v2.100.1 // indirect
 	k8s.io/kube-aggregator v0.24.2 // indirect
-	k8s.io/utils v0.0.0-20230209194617-a36077c30491 // indirect
+	k8s.io/utils v0.0.0-20230209194617-a36077c30491
```
michael-cico, line 190: did we mean to change this?

pkg/semver/semver.go
```diff
@@ -122,6 +122,42 @@ func (v *SemVersion) CompareTo(from *SemVersion) int {
 	return result
 }
 
+// CompareToPrereleaseInts Compares the current version to another version treating the Prerelease field as an int
+// - if from > this, -1 is returned
+// - if from < this, 1 is returned
+// - if they are equal, 0 is returned
+// - if the prerelease field can not be converted to an int, it will return an error
+func (v *SemVersion) CompareToPrereleaseInts(from *SemVersion) (int, error) {
```
michael-cico, line 130: Feels like we need a customizatin/extension of SemVer since we're using pre-release in a specific way.

Not for this pass of course.

michael-cico (dismissed):

platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -4,76 +4,156 @@
 modules:
   - name: ingress-controller
     version: 1.7.1
+    chart: platform-operator/thirdparty/charts/ingress-nginx
```
pfmackin, line 7: are we going to change the code to use these new fields (chart / valuesFiles)

go.mod
mbenmclean: Yes I'm using that dependency directly now so I think that's why this happened.

pkg/semver/semver.go
mbenmclean: You don't think this is sufficient?

platform-operator/manifests/catalog/catalog.yaml
mbenmclean: yes eventually.

George-Aeillo (approved):

go.mod
kminder: This should have been moved to the first required section then.

pkg/semver/semver.go
kminder: Should probably add some unit tests to prove the behavior is as expected.
We might need to go through and add a -0 to everything with only one version.

kminder:
I imagine what is needed in addition is some form of test that confirms effectiveCRs for some set of actualCRs are unchanged if none of the module versions are changed.

pkg/semver/semver.go
michael-cico: I don't think what I proposed is required, just seems like we're imposing our own semantics on the SemVer fields.  so as opposed to a custom method for that comparison it seems like a case of subclassing/specialization.

But, superfluous to requirements I think.  Just my ramblings.

michael-cico: I think Ben did add some unit tests for these cases

PR #7250, opened 2023-09-29
VZ-11244: Fix opensearchOperator PV migration failure on upgrade

saketmht:


platform-operator/controllers/verrazzano/component/opensearchoperator/opensearchoperator.go
```diff
@@ -103,20 +103,20 @@ func appendOverrides(ctx spi.ComponentContext, _ string, _ string, _ string, kvs
 func (o opensearchOperatorComponent) deleteRelatedResource() error {
 	client, err := k8sutil.GetDynamicClient()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create dynamic client: %v", err)
```
michael-cico, line 106: You should make this return a throttled error from the VZ logger if you can, can be on a follow-up PR.

michael-cico (dismissed):

michael-cico (approved):

PR #7256, opened 2023-09-29
Use dry-run options for validators e2e tests

michael-cico:


tests/e2e/pkg/update/updater.go
```diff
@@ -6,6 +6,7 @@ package update
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
fredtibbitts, line 9: Nit, import formatting.

tests/e2e/pkg/update/updater.go
```diff
@@ -176,7 +178,12 @@ func UpdateCRV1beta1(m CRModifierV1beta1) error {
 		return err
 	}
 	vzClient := client.VerrazzanoV1beta1().Verrazzanos(cr.Namespace)
-	_, err = vzClient.Update(context.TODO(), cr, metav1.UpdateOptions{})
+
+	updateOpts := metav1.UpdateOptions{}
+	if len(opts) > 0 {
+		updateOpts = opts[0]
```
fredtibbitts, line 184: Why is this forcing the update options to the first option? Can't you just pass them through? If you end up leaving it like this a comment would be good because it's not clear why it's just taking the first option.

michael-cico: This method is using the strongly typed vzclient `Update`, which only takes a single `UpdateOptions`.  

The other one uses the `controllerruntime` client which does take a variable args list.  

I could modify this to follow the same `controllerruntime` pattern if you think it's worth it.

michael-cico: Made the change

fredtibbitts: I like this better, thanks for making that change.

fredtibbitts (approved):

PR #7257, opened 2023-10-02
VZ-10610 Add OKE CAPI cluster driver tests

ayaranga:
Add OKE CAPI cluster driver tests and invoke them from periodic pipeline.

### DIFF
--- a/ci/JenkinsfilePeriodicTests
+++ b/ci/JenkinsfilePeriodicTests
@@ -574,6 +574,28 @@ pipeline {
                         }
                     }
                 }
+                stage('OKE CAPI Cluster Driver') {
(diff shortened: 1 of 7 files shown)


anders-swanson (approved):

anders-swanson:
Approved, but is it possible to have as a parallel stage in the ocne tests? there's a lot of overlap with the ocne kontainerengine tests.

ayaranga:
> Approved, but is it possible to have as a parallel stage in the ocne tests? there's a lot of overlap with the ocne kontainerengine tests.

Yes, it should be possible. I will create a separate task to track this. Thanks.

PR #7262, opened 2023-10-02
VZ-11007, VZ-11210: VMC extensions for clusterRef, imported, and provider fields

marcogelle:
This PR adds code to set the `clusterRef`, `imported`, and `provider` fields in the VMC.
<br>
**API Changes to the `VerrazzanoManagedCluster` CRD:**
- adds the `.status.imported` and `.status.provider` fields

**Refactoring changes:**
- VZ-11210: use an import alias when the CAPI controller uses the `cluster-operator/internal/capi` package
- Created `update_status.go` and `update_status_test.go` under the VMC controller

**Added code:**
- Reconcile the new `imported` and `provider` fields in the VMC controller
- Set the `clusterRef` field of the VMC in the CAPI controller, at the time of VMC creation
- Added to the `cluster-operator/internal/capi` package
- Adds the ability to `get` `clusterclasses` in the `verrazzano-cluster-operator` ClusterRole
- Added unit tests


cluster-operator/internal/capi/cluster.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+var GVKCAPIClusterClass = schema.GroupVersionKind{
+	Group:   "cluster.x-k8s.io",
+	Version: "v1beta1",
+	Kind:    "ClusterClass",
+}
+
+// GetCluster returns the requested CAPI Cluster as an unstructured pointer.
+func GetCluster(ctx context.Context, cli clipkg.Client, clusterNamespacedName types.NamespacedName) (*unstructured.Unstructured, error) {
```
anders-swanson, line 36: just curious why are we using unstructured instead of type capi cluster?

cluster-operator/internal/capi/cluster.go
```diff
@@ -0,0 +1,69 @@
...
+	cluster := &unstructured.Unstructured{}
+	cluster.SetGroupVersionKind(GVKCAPICluster)
+	if err := cli.Get(context.TODO(), clusterNamespacedName, cluster); err != nil {
+		return nil, err
+	}
+	return cluster, nil
+}
+
+// GetClusterClass returns the requested CAPI ClusterClass as an unstructured pointer.
+func GetClusterClass(ctx context.Context, cli clipkg.Client, clusterClassNamespacedName types.NamespacedName) (*unstructured.Unstructured, error) {
```
anders-swanson, line 46: same for this, is there is a reason we have to use unstructured? typed is easier to work with

cluster-operator/controllers/vmc/update_status.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// This cluster does not use ClusterClass.
+	// Get infrastructure provider
+	infraProvider, found, err := unstructured.NestedString(capiCluster.Object, "spec", "infrastructureRef", "kind")
+	if !found {
+		r.log.Progressf("could not find spec.infrastructureRef.kind field inside cluster %s: %v", clusterNamespacedName, err)
+		return "", nil
+	}
+	if err != nil {
+		r.log.Progressf("error while looking for spec.infrastructureRef.kind field for cluster %s: %v", clusterNamespacedName, err)
```
jmaron99, line 120: So this error is expected to be repeatedly generated during reconciliation?

cluster-operator/internal/capi/cluster.go
marcogelle: Same thing here as with the Cluster

cluster-operator/controllers/vmc/update_status.go
marcogelle: My logic here was to do something similar as here -- https://github.com/verrazzano/verrazzano/pull/7191#discussion_r1339078576 -- from my previous VMC PR.

I don't expect this error to be repeatedly generated during reconciliation, since I believe these fields should already be set in the CAPI Cluster spec.

jmaron99: @anders-swanson does a progress log make sense in this context?  It would potentially make sense if there is an expectation of repeated errors that may not be "critical", but is that the case here?

anders-swanson: i would really prefer if this did not use unstructured, these kind of issues wouldn't exist

cluster-operator/internal/capi/cluster.go
anders-swanson: Please do, unstructured is difficult to work with

cluster-operator/controllers/vmc/update_status.go
marcogelle: Is there an okay solution for error reporting here for this PR, before I get to converting the use of unstructured?

cluster-operator/internal/capi/cluster.go
marcogelle: Actually, I'll just work on this change on this PR, and reach back out once I'm done with this.

marcogelle:
I made changes to remove the use of `unstructured` to represent CAPI `Clusters` and `ClusterClasses` in the VMC controller, so I'll mark the previous comments as resolved.

anders-swanson (approved):

PR #7271, opened 2023-10-02
Rancher OKE Node Image API

anders-swanson:
Fixes VZ 11251

### DIFF
--- a/platform-operator/manifests/catalog/catalog.yaml
+++ b/platform-operator/manifests/catalog/catalog.yaml
@@ -30,7 +30,7 @@ modules:
   - name: istio
     version: 1.17.2-1
   - name: rancher
-    version: 2.7.5-6
+    version: 2.7.6
     chart: platform-operator/thirdparty/charts/rancher
     valuesFiles:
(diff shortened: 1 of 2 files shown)


platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -30,7 +30,7 @@ modules:
   - name: istio
     version: 1.17.2-1
   - name: rancher
-    version: 2.7.5-5
+    version: 2.7.5-6
```
desagar, line 33: we are using Rancher v2.7.6 now - shouldn't this be incremented?

mbenmclean: Yes if we're using rancher 2.7.6 this should be changed. The bom component version should also be changed.

desagar (dismissed):

platform-operator/manifests/catalog/catalog.yaml
anders-swanson: resolved, can i get re-review

desagar (approved):

PR #7295, opened 2023-10-04
VZ-10800: Avoid restart of OAM application pods while adding metrics labels/annotations

lake-of-dreams:
The erstwhile implementations of [updateRelatedDeployment](https://github.com/verrazzano/verrazzano/blob/master/application-operator/controllers/metricstrait/workloads.go#L58) and [updateRelatedStatefulSet](https://github.com/verrazzano/verrazzano/blob/master/application-operator/controllers/metricstrait/workloads.go#L58) propagate changes to labels and annotations to underlying pods by modifying `*.Spec.Template.ObjectMeta` on the corresponding `deployment` or `statefulset` that will result in recycling of any pod that might already have been created . In this PR , the behaviour is changed to mutate the labels and annotations on the underlying pod (if exists) directly and not change the spec of parent `deployment` or `statefulset` that was causing the restart of pods.


desagar:
Not sure of I understand the reason for the change. If you update the pod labels/annotations directly, those will not be retained if the pod later goes down for some reason, right? It would be overwritten from the deployment/statefulset's pod spec.

lake-of-dreams:
> Not sure I understand the reason for the change. If you update the pod labels/annotations directly, those will not be retained if the pod later goes down for some reason, right? It would be overwritten from the deployment/statefulset's pod spec.

The reason for this change is because if we update the deployment or statefulset spec, it will result in termination of existing pods and new pods will be created with the updated labels and annotations. This approach updates the running pods directly. If the pods were to go down with any other reason, the reconciler again updates the labels and annotations on the new pods. For example when I delete the pod `nginx-workload-6d8fc65ffd-rbnks` deliberately, the pod  `nginx-workload-6d8fc65ffd-74vl5 ` is spawned and the reconciler updates the annotations and labels on new pod.

```
bash-3.2$ kubectl -n ngx-try2 get pods -w
NAME                              READY   STATUS    RESTARTS   AGE
nginx-workload-6d8fc65ffd-rbnks   1/1     Running   0          17h
nginx-workload-6d8fc65ffd-rbnks   1/1     Terminating   0          17h
nginx-workload-6d8fc65ffd-74vl5   0/1     Pending       0          0s
nginx-workload-6d8fc65ffd-74vl5   0/1     Pending       0          0s
nginx-workload-6d8fc65ffd-74vl5   0/1     ContainerCreating   0          0s
nginx-workload-6d8fc65ffd-rbnks   0/1     Terminating         0          17h
nginx-workload-6d8fc65ffd-rbnks   0/1     Terminating         0          17h
nginx-workload-6d8fc65ffd-rbnks   0/1     Terminating         0          17h
nginx-workload-6d8fc65ffd-rbnks   0/1     Terminating         0          17h
nginx-workload-6d8fc65ffd-74vl5   1/1     Running             0          1s
nginx-workload-6d8fc65ffd-74vl5   1/1     Running             0          8s
^Cbash-3.2$ kubectl -n ngx-try2 get pod nginx-workload-6d8fc65ffd-74vl5 -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    verrazzano.io/metricsEnabled: "true"
    verrazzano.io/metricsPath: /metrics
    verrazzano.io/metricsPort: "8080"
  creationTimestamp: "2023-10-05T02:16:12Z"
  generateName: nginx-workload-6d8fc65ffd-
  labels:
    app: nginx
    app.oam.dev/component: ngx-component
    app.oam.dev/name: ngx-appconf
    app.oam.dev/resourceType: WORKLOAD
    app.oam.dev/revision: ""
    containerizedworkload.oam.crossplane.io: 0b11a2f8-8c7d-4a8e-82db-95c31aac1d32
    pod-template-hash: 6d8fc65ffd
    version: v1

```

In the existing implementation, even without any deliberate deletion of pod
1. `nginx-workload-6d8fc65ffd-rbnks` would have started.
2. The reconciler will have updated the labels and annotations in deployment spec.
3. The pod `nginx-workload-6d8fc65ffd-rbnks` will have terminated 
4. And a new pod will have started

desagar (approved):

PR #7304, opened 2023-10-04
VZ-11268.  Fix scenario where VPO is updated to v1.7.0 and uninstalled before upgrading

michael-cico:
Move isUpgradeRequired() check to the module handlers so that `Module` instances are always created after the VPO is installed in an existing pre-1.7.0 install before the upgrade is issued.

This fixes an edge case where an uninstall is started after VPO is updated from a non-modules (pre-1.7.0) release, but before the system is upgraded.  

Previously Modules for existing installed components pre-1.7.0 were only created once the upgrade was started.  When an uninstall was issued, no `Module` instances existed and uninstall would be broken and/or incomplete.


platform-operator/controllers/verrazzano/controller/reconciler.go
```diff
@@ -109,20 +109,17 @@ func (r Reconciler) doReconcile(log vzlog.VerrazzanoLogger, controllerCtx contro
 		return res
 	}
 
-	// If an upgrade is pending, do not reconcile; an upgrade is pending if the VPO has been upgraded, but the user
```
pfmackin, line 112: Did you mean to remove the comment

platform-operator/controllers/module/component-handler/migration/migration.go
```diff
@@ -35,6 +38,14 @@ func (h migrationHandler) UpdateStatusIfAlreadyInstalled(ctx handlerspi.HandlerC
 	if err != nil {
 		return result.NewResultShortRequeueDelayWithError(err)
 	}
+
+	if upgradeRequired, err := vzctrlcommon.IsUpgradeRequired(vzcr); upgradeRequired || err != nil {
+		if upgradeRequired {
+			ctx.Log.Oncef("Upgrade required before reconciling module %s", client.ObjectKeyFromObject(module))
+		}
```
pfmackin, line 45: need to return NewSHortRequeue if no error and upgrade required.  Might be cleaner to remove the 2 ifs

upgradeRequires, err := ...
if err != nil {
  return result.NewResultShortRequeueDelayWithError(err)
}
if upgradeRequired {
  log...
 return result.NewRequeueWithDelay
}

platform-operator/controllers/verrazzano/controller/reconciler.go
```diff
@@ -109,20 +109,17 @@ func (r Reconciler) doReconcile(log vzlog.VerrazzanoLogger, controllerCtx contro
...
-	// has not modified the version in the Verrazzano CR to match the BOM.
-	if upgradePending, err := r.isUpgradeRequired(actualCR); upgradePending || err != nil {
-		controllerCtx.Log.Oncef("Upgrade required before reconciling modules")
-		return result.NewResultShortRequeueDelayIfError(err)
+	if upgradeRequired, err := vzctrlcommon.IsUpgradeRequired(actualCR); upgradeRequired || err != nil {
```
pfmackin, line 112: same comment as in controller_utils.go.  Need to return NetShortRequeueWithDelay if no error and upgradRequired

platform-operator/controllers/verrazzano/controller/finalizer.go
```diff
@@ -58,6 +59,21 @@ func (r Reconciler) PreRemoveFinalizer(controllerCtx controllerspi.ReconcileCont
...
+	// user has deleted the VZ CR before initiating an upgrade
+	if !r.doesUninstallConditionExist(actualCR) {
+		effectiveCR, err := r.createEffectiveCR(actualCR)
+		if err != nil {
+			result.NewResultShortRequeueDelayWithError(err)
+		}
+		if result := r.createOrUpdateModules(log, actualCR, effectiveCR); result.ShouldRequeue() {
+			return result
+		}
+		return result.NewResultShortRequeueDelay()
```
pfmackin, line 73: Don't requeue here, just drop through

platform-operator/controllers/verrazzano/controller/reconciler.go
michael-cico: no

platform-operator/controllers/module/component-handler/migration/migration.go
michael-cico: done

platform-operator/controllers/verrazzano/controller/finalizer.go
michael-cico: reverted this

platform-operator/controllers/verrazzano/controller/reconciler.go
michael-cico: fixed

platform-operator/controllers/verrazzano/controller/reconciler.go
michael-cico: fixed

pfmackin (approved):

PR #7307, opened 2023-10-04
Move catalog unit test into a separate make target

mbenmclean:
This change moves the catalog into a separate make target and calls it separately because it requires connection to the internet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -306,6 +306,7 @@ pipeline {
                     make unit-test-coverage-ratcheting
                     echo "Checking versions..."
                     release/scripts/check_versions.sh ${VERRAZZANO_DEV_VERSION}
+                    make unit-test-catalog
                 """
                     }
                     post {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -306,6 +306,7 @@ pipeline {
                     make unit-test-coverage-ratcheting
                     echo "Checking versions..."
                     release/scripts/check_versions.sh ${VERRAZZANO_DEV_VERSION}
+                    make unit-test-catalog
```
michael-cico, line 309: should it just be a dependent target of `precommit`?  e.g.,

```
.PHONY: precommit
precommit: precommit-check precommit-build unit-test-coverage unit-test-catalog ## run all precommit checks
```

mbenmclean: No because the whole point is it's separate from the other targets that people use to check locally so it doesn't fail when they're not on internet

mbenmclean: Or I thought that was what we decided

PR #7354, opened 2023-10-10
Raagwadh/VZ-11265: Update ImageSize alert and Baseline settings

Raaghavcodes:
require a COMMIT is specified and FAIL if it was NONE
Skip running all the tests, we don't want to burn up resources and time doing that, we just want to put in a new baseline into object storage
Fetch the image sizes from that per-commit location in object storage
Stuff the new image size into object storage as the baseline
Update the displayname of the job so we know it only updated the baseline


ci/periodic/JenkinsfileDistributions
```diff
@@ -133,12 +138,22 @@ pipeline {
             }
             steps {
                 script {
+                    if (params.GIT_COMMIT_TO_USE == "NONE") {
+                        error("Specific GIT commit was not specified. GIT_COMMIT_TO_USE is required when BASELINE_IMAGE_SIZES_FILE_UPDATE is set to true.")
+                    }
+                   currentBuild.displayName = "${currentBuild.displayName} : Update Baseline image-sizes.txt"
```
bobdonat, line 144: This should probably go before the above if statement

bobdonat (dismissed):
Approved but one small comment.

ci/periodic/JenkinsfileDistributions
Raaghavcodes: Done

bobdonat (approved):

bobdonat (approved):

PR #7387, opened 2023-10-17
VZ-11326 Move all cluster management to VMC controller

jmaron99:
Cluster operator controllers have been refactored as follows:
- CAPI controller is mostly responsible for creating a VMC when a CAPI cluster is created and removing it when it is deleted
- The VMC controller now handles the management and pushing of cluster related resources via Rancher or CAPI mechanisms depending on whether CAPI cluster creation, Rancher is enabled, VZ is installed etc.

Testing done:
- VZ, CAPI, and MC pipelines built
- Manual scenarios:
1.  Check VMC and Rancher status for a CAPI cluster creation, followed by installation of VZ, followed by enabling Rancher
2. Check VMC and Rancher status for CAPI cluster creation with Rancher enabled, followed by installation of VZ
3. Rancher UI generated cluster, w/o VZ followed by enabling VZ (in progress)


mgianatagh (dismissed):

cluster-operator/controllers/vmc/push_manifest_objects.go
```diff
@@ -24,51 +53,128 @@ func (r *VerrazzanoManagedClusterReconciler) pushManifestObjects(vmc *clusterapi
...
+	return false, nil
+}
+
+func (r *VerrazzanoManagedClusterReconciler) pushViaCAPIClient(ctx context.Context, vmc *clusterapi.VerrazzanoManagedCluster, rancherEnabled bool) (bool, error) {
+	r.log.Debugf("Pushing via CAPI client, status: %s", vmc.Status.RancherRegistration.Status)
+	if vmc.Status.RancherRegistration.Status != clusterapi.RegistrationApplied {
+		cluster := &unstructured.Unstructured{}
+		cluster.SetGroupVersionKind(internalcapi.GVKCAPICluster)
+		err := r.Get(ctx, types.NamespacedName{Namespace: vmc.Status.ClusterRef.Namespace, Name: vmc.Status.ClusterRef.Name}, cluster)
+		if err != nil && !errors.IsNotFound(err) {
```
desagar, line 107: if the we get IsNotFound, it looks like we are proceeding further. I presume getClusterManifest will then return an error, and that will be returned. Not sure if that's the behavior you want.

fredtibbitts (commented):
This generally seems fine to me. It looks like there's an API change for the new "Applied" registration status. Did this go through API review?

cluster-operator/controllers/vmc/sync_manifest_secret.go
```diff
@@ -33,82 +39,97 @@ const (
...
 // Verrazzano and the Rancher cluster ID has not yet been set in the status. This allows the
 // calling function to requeue and process the VMC again before marking the VMC ready.
-func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Context, vmc *clusterapi.VerrazzanoManagedCluster) (bool, error) {
+func (r *VerrazzanoManagedClusterReconciler) syncManifestSecret(ctx context.Context, rancherEnabled bool, vmc *clusterapi.VerrazzanoManagedCluster) (bool, error) {
 	// Builder used to build up the full YAML
 	// For each secret, generate the YAML and append to the full YAML which contains multiple resources
 	var sb = strings.Builder{}
 
-	// add agent secret YAML
-	agentYaml, err := r.getSecretAsYaml(GetAgentSecretName(vmc.Name), vmc.Namespace,
```
desagar, line 42: I'm having a hard time following all the if else cases, but it _seems_ as though the agentYaml is only being written to the manifest in the CAPI case, and not in the Rancher case. That would be incorrect. Of course, it's possible I am mis-reading this.

cluster-operator/controllers/vmc/push_manifest_objects.go
jmaron99: As it stands - yes, the following method will error and the result will be similar.  The code could be corrected but I find that each change, if I want to be safe, requires an entire day of testing.

cluster-operator/controllers/vmc/rancher.go
```diff
@@ -344,10 +349,34 @@ func GetCACertFromManagedClusterSecret(rc *rancherutil.RancherConfig, clusterID,
 }
 
 // isNamespaceCreated attempts to ascertain whether the given namesapce is created in the cluster
-func isNamespaceCreated(client client.Client, ingressHost, clusterID, namespace string, log vzlog.VerrazzanoLogger) (bool, error) {
+func isNamespaceCreated(vmc *v1alpha1.VerrazzanoManagedCluster, r *VerrazzanoManagedClusterReconciler, clusterID, namespace string) (bool, error) {
+	if vmc.Status.ClusterRef != nil {
+		cluster := &unstructured.Unstructured{}
+		cluster.SetGroupVersionKind(internalcapi.GVKCAPICluster)
+		err := r.Get(context.TODO(), types.NamespacedName{Namespace: vmc.Status.ClusterRef.Namespace, Name: vmc.Status.ClusterRef.Name}, cluster)
+		if err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
jmaron99: 1.  It's unlikely that the CAPI cluster doesn't exist if the cluster Ref exists (which is the only way to get into this code)
2. If it somehow did get here the following method would fail anyway

desagar (approved):

PR #7426, opened 2023-11-06
VZ-11406: Go 1.20 support

lake-of-dreams:
- Go 1.20 required some changes to `go.mod` and `go.sum` to get `go mod tidy` and `controller-gen` both working
- `golangci-lint 1.49` hangs with go 1.20, upgraded to `1.55.2`
- Also fixed some lint errors per `golangci-lint 1.55.2` but some checks are excluded which will be fixed after fixing the linter erros in a subsequent PR


Jenkinsfile
```diff
@@ -24,7 +24,7 @@ pipeline {
 
     agent {
        docker {
-            image "${RUNNER_DOCKER_IMAGE}"
+            image "${EXPERIMENTAL_RUNNER_DOCKER_IMAGE}"
```
agv426, line 27: You missed removing this

go.mod
```diff
@@ -3,15 +3,17 @@
 
 module github.com/verrazzano/verrazzano
 
-go 1.19
+go 1.20
```
bhatpmk, line 6: Is this change to increment version in go.mod really required ?

lake-of-dreams: by having a lower version in go.mod we are saying that it is ok to compile/build the code with a lower version because thats the minimum version required but in practice we ourselves don't build it using that version , I think that is not consistent

Jenkinsfile
lake-of-dreams: changed to `GOLANG20_RUNNER_DOCKER_IMAGE` , will be changed to original after I would have verified other pipelines and will do in a subsequent pipeline

michael-cico (approved):

JeffTancill (approved):

PR #7438, opened 2023-11-08
VZ-11427: Update VPO Dockerfile to use distroless base image

fredtibbitts:
Distroless base includes glibc and openssl but is smaller and has a reduced CVE surface area compared to OL 8-slim. The VPO does require a shell and several utilities to function, including curl, and their associated shared libs.

### DIFF
--- a/platform-operator/Dockerfile
+++ b/platform-operator/Dockerfile
@@ -86,7 +86,6 @@ COPY --from=build_base /home/ /home/
 # copy users/groups added in build_base stage
 COPY --from=build_base /etc/passwd /etc/passwd
 COPY --from=build_base /etc/group /etc/group
-COPY --from=build_base /bin/bash /bin/bash
 
 # copy the operator binary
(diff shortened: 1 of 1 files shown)


platform-operator/Dockerfile
```diff
@@ -39,16 +33,74 @@ RUN microdnf install -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs ope
...
+RUN mkdir -p /tmp/stage/verrazzano && \
+    chmod 700 /tmp/stage/verrazzano
+
+# istioctl image to copy the istioctl binary to the final image
+FROM $ISTIO_ISTIOCTL_IMAGE AS istio_istioctl
+
+# Create the verrazzano-platform-operator image
+FROM $FINAL_IMAGE AS final
+
+COPY --from=istio_istioctl /usr/local/bin/istioctl /usr/local/bin/istioctl
```
desagar, line 49: not sure how this works. This is the istio_istioctl stage, isn't it?

davisgomes (approved):

PR #7443, opened 2023-11-09
VZ-11441 - Switch the Verrazzano application operator image to use the ol8-static image as its base image.

desagar:


davisgomes (approved):

platform-operator/helm_config/charts/verrazzano-application-operator/values.yaml
```diff
@@ -17,5 +17,7 @@ requestMemory: 72Mi
 webhook:
   replicas: 1
 
+ol8SlimImage: ghcr.io/oracle/oraclelinux:8-slim
```
fredtibbitts, line 20: I think we should name this something like `webhookWaitImage` or similar, that way if we change it to use OL 9 for example, this key doesn't need to change.

fredtibbitts: Also this needs to go in the BOM otherwise it won't work in the private registry case. Images not in the BOM won't have their registry patched for private registry.

desagar: did both, running tests including private reg tests.

fredtibbitts (approved):

PR #7512, opened 2023-11-29
Karak/update opesnearch operator image

karan56625:


platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -155,7 +155,7 @@ modules:
     version: VERRAZZANO_VERSION
     chart: platform-operator/helm_config/charts/verrazzano-network-policies
   - name: opensearch-operator
-    version: 2.4.0-1
+    version: 2.4.0-3
```
bhatpmk, line 158: This should be 2, although it won't fail. Same in verrazzano-bom.json

bhatpmk (approved):

PR #7556, opened 2023-12-11
VZ-11554: New version of fluent-bit

neeraj-n-singh:

### DIFF
--- a/platform-operator/manifests/catalog/catalog.yaml
+++ b/platform-operator/manifests/catalog/catalog.yaml
@@ -78,7 +78,7 @@ modules:
     valuesFiles:
       - platform-operator/helm_config/overrides/fluent-operator-values.yaml
   - name: fluentbit-opensearch-output
-    version: 2.0.11
+    version: 2.0.12
     chart: platform-operator/helm_config/charts/fluentbit-opensearch-output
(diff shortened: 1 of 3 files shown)


karan56625 (dismissed):

platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -73,7 +73,7 @@ modules:
     valuesFiles:
       - platform-operator/helm_config/overrides/coherence-values.yaml
   - name: fluent-operator
-    version: 2.2.0-2
+    version: 2.2.0-3
```
mgianatagh, line 76: shouldn't the image tag in the BOM also be getting updated?

platform-operator/verrazzano-bom.json
```diff
@@ -535,7 +535,7 @@
     },
     {
       "name": "fluent-operator",
-      "version": "2.2.0-2",
+      "version": "2.2.0-3",
```
mgianatagh, line 538: This still doesn't seem correct to me.  These changes do not change the actual image being used.  Isn't there a new tag for line 546?

adiforluls (approved):

PR #7558, opened 2023-12-11
VZ-11554: New version of fluent-bit

neeraj-n-singh:

### DIFF
--- a/platform-operator/manifests/catalog/catalog.yaml
+++ b/platform-operator/manifests/catalog/catalog.yaml
@@ -78,7 +78,7 @@ modules:
     valuesFiles:
       - platform-operator/helm_config/overrides/fluent-operator-values.yaml
   - name: fluentbit-opensearch-output
-    version: 2.0.11
+    version: 2.0.12
     chart: platform-operator/helm_config/charts/fluentbit-opensearch-output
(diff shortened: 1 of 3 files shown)


platform-operator/manifests/catalog/catalog.yaml
```diff
@@ -73,12 +73,12 @@ modules:
...
       - platform-operator/helm_config/overrides/coherence-values.yaml
   - name: fluent-operator
-    version: 2.2.0-1
+    version: 2.2.0-2
     chart: platform-operator/thirdparty/charts/fluent-operator
     valuesFiles:
       - platform-operator/helm_config/overrides/fluent-operator-values.yaml
   - name: fluentbit-opensearch-output
-    version: 2.0.11
+    version: 2.0.12
```
mgianatagh, line 81: Is there an associated image tag change for this?  I notice the build failed.

mgianatagh (dismissed):

platform-operator/manifests/catalog/catalog.yaml
neeraj-n-singh: This is fixed now, there is no entry for the same in bom file.

adiforluls (approved):

PR #7636, opened 2024-01-08
First pass of adding version of CLI output to commands

George-Aeillo:
This commit adds the CLI version as first output to all of the commands in the cmd folder, except for the version command.

### DIFF
--- a/tools/vz/cmd/version/version.go
+++ b/tools/vz/cmd/version/version.go
@@ -5,7 +5,6 @@ package version
 
 import (
 	"fmt"
-
 	"github.com/spf13/cobra"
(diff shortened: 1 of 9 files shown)


tools/vz/cmd/version/version.go
```diff
@@ -74,3 +72,7 @@ func GetEffectiveDocsVersion() string {
 func GetCLIVersion() string {
 	return cliVersion
 }
+
+func GetVZCLIVersionMessageString() string {
+	return "This command was run with VZ CLI Version " + GetCLIVersion() + "\n"
```
bobdonat, line 77: Should VZ be lower case

agv426 (approved):

PR #7642, opened 2024-01-10
VZ-11415: Update prometheus stack images

neeraj-n-singh:

### DIFF
--- a/platform-operator/verrazzano-bom.json
+++ b/platform-operator/verrazzano-bom.json
@@ -718,8 +718,8 @@
           "name": "prometheus",
           "images": [
             {
-              "image": "prometheus-jenkins",
-              "tag": "v2.44.0-20240103123723-092952cc",
+              "image": "prometheus",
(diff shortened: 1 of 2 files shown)


platform-operator/verrazzano-bom.json
```diff
@@ -767,15 +767,15 @@
...
       "subcomponents": [
         {
           "repository": "verrazzano",
           "name": "prometheus-pushgateway",
           "images": [
             {
-              "image": "prometheus-pushgateway",
-              "tag": "v1.6.2-20230925134721-9944fc39",
+              "image": "pushgateway",
+              "tag": "v1.6.2-1-20240104075814-f66f60f0",
```
bhatpmk, line 778: May I know the reason to have "-1" in the image tag ? The other images built from source do not follow this pattern.

neeraj-n-singh: That image was built using olcne pipelines and already committed previously.

bhatpmk (approved):

PR #7656, opened 2024-01-12
VZ-11650: Log standard error when granting XA_RECOVER_ADMIN fails during upgrade

bhatpmk:
This change logs standard error when granting XA_RECOVER_ADMIN fails during upgrade. Also a check is added when the root password obtained from the secret is nil.


mgianatagh (approved):

platform-operator/controllers/verrazzano/component/mysql/mysql.go
```diff
@@ -720,6 +720,9 @@ func grantXARecoverAdmin(ctx spi.ComponentContext) error {
 	if err != nil {
 		return err
 	}
+	if rootPassword == nil {
+		ctx.Log().Progressf("%s is empty in secret %s", mySQLRootKey, rootSec)
+	}
```
pfmackin, line 725: probably should return an error

bhatpmk: Done

pfmackin (approved):

PR #7693, opened 2024-01-22
vz-11649 Have --include-logs flag grab pod logs without needing --include-namespaces

AbeHern:
Changes grab pod logs as expected with --include-logs flag without the need of explicitly using --include-namespaces for individual namespaces

### DIFF
--- a/tools/vz/cmd/bugreport/bugreport.go
+++ b/tools/vz/cmd/bugreport/bugreport.go
@@ -33,6 +33,9 @@ When --report-file is not provided, the command creates bug-report.tar.gz in the
 # Create a bug report file, bugreport.tar.gz, including the additional namespace ns1 from the cluster:
 vz bug-report --report-file bugreport.tgz --include-namespaces ns1
 
+# Use the --include-logs flag to collect logs from Verrazzano that are already being captured
+vz bug-report --report-file bugreport.tgz --include-logs
+
 # The flag --include-namespaces accepts comma-separated values and can be specified multiple times. For example, the following commands create a bug report by including additional namespaces ns1, ns2, and ns3:
    a. vz bug-report --report-file bugreport.tgz --include-namespaces ns1,ns2,ns3
    b. vz bug-report --report-file bugreport.tgz --include-namespaces ns1,ns2 --include-namespaces ns3
--- a/tools/vz/cmd/bugreport/bugreport.go
+++ b/tools/vz/cmd/bugreport/bugreport.go
@@ -33,7 +33,7 @@ When --report-file is not provided, the command creates bug-report.tar.gz in the
 # Create a bug report file, bugreport.tar.gz, including the additional namespace ns1 from the cluster:
 vz bug-report --report-file bugreport.tgz --include-namespaces ns1
 
-# Use the --include-logs flag to collect logs from Verrazzano that are already being captured
+# Use the --include-logs flag to collect logs from Verrazzano namespaces that are already being captured
 vz bug-report --report-file bugreport.tgz --include-logs
 
 # The flag --include-namespaces accepts comma-separated values and can be specified multiple times. For example, the following commands create a bug report by including additional namespaces ns1, ns2, and ns3:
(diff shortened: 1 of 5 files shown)


tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -154,6 +154,8 @@ func captureResources(client clipkg.Client, kubeClient kubernetes.Interface, dyn
 	for _, ns := range namespaces {
 		go captureK8SResources(wg, ecr, client, kubeClient, dynamicClient, ns, bugReportDir, vzHelper)
 	}
+	// captures pod logs of resources in namespaces if --include-logs flag is enabled
+	capturePodLogs(client, kubeClient, vzHelper, bugReportDir, namespaces, podLogs)
```
bhatpmk, line 158: May I know the reason for this ? This along with the call to capturePodLogs in line 361, calls captureAdditionalLogs twice.

AbeHern: I wanted to break out the logic for determining when to capture pod logs based solely on the --include-logs flag and not on --include-namespaces and --include-logs. I thought to do this because on line 158 is where I actually care about the call to capture pod logs.

George-Aeillo (commented):
It looks good to me, but I know other folks have yet to review it.

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -33,6 +33,9 @@ When --report-file is not provided, the command creates bug-report.tar.gz in the
 # Create a bug report file, bugreport.tar.gz, including the additional namespace ns1 from the cluster:
 vz bug-report --report-file bugreport.tgz --include-namespaces ns1
 
+# Use the --include-logs flag to collect logs from Verrazzano that are already being captured
```
bhatpmk, line 36: Help requires a change here and in line no. 45 "# Use the --include-logs flag to collect the log ..."

With this change, the tool will collect the logs from all the pods in the namespaces, where Verrazzano installs components.

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -155,6 +155,9 @@ func captureResources(client clipkg.Client, kubeClient kubernetes.Interface, dyn
 		go captureK8SResources(wg, ecr, client, kubeClient, dynamicClient, ns, bugReportDir, vzHelper)
 	}
 
+	// captures pod logs of vz component namespaces if --include-logs flag is enabled
+	captureAdditionalResources(client, kubeClient, dynamicClient, vzHelper, bugReportDir, namespaces, podLogs)
```
bhatpmk, line 159: Can you investigate capturing the logs by iterating the pods in the component namespaces, using goroutine ? I am referring have a loop in place of the code from line no.s 145 to 149

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -496,15 +595,67 @@ func TestBugReportSuccessWithDuration(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+// TestBugReportLogsFromAllNamespaces
```
agv426, line 598: Name in comment doesn't match function name

AbeHern: Fixed

tools/vz/pkg/bugreport/reportgen.go
AbeHern: I am actually working on that on side, because having the explicit Podlists from 124 to 129. In terms of the scope of this ticket, the changes made do meet the expectations. I think a follow-up ticket to this one for the refactor of goroutines is necessary

tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -33,6 +33,9 @@ When --report-file is not provided, the command creates bug-report.tar.gz in the
 # Create a bug report file, bugreport.tar.gz, including the additional namespace ns1 from the cluster:
 vz bug-report --report-file bugreport.tgz --include-namespaces ns1
 
+# Use the --include-logs flag to collect logs from Verrazzano namespaces that are already being captured
```
agv426, line 36: # Use the --include-logs flag to collect logs from the namespaces being captured

agv426, line 36: I'm posting my comments on what we need to do here in the slack thread

agv426 (approved):

PR #7700, opened 2024-01-25
VZ-3151: Add Namespaces and metadata capture to analysis tool, along with detection of namespace termination issues 

George-Aeillo:
In this PR, I added support for a cluster-wide metadata.json file to be added that currently records the time of capture, along with a namespace.json file, that captures the namespace resource for a particular resource. Using these two new items in the cluster-snapshot, I have added the ability for the analysis tool to detect and report namespace termination issues that are being experienced by a cluster.


tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -126,9 +130,39 @@ func CaptureK8SResources(client clipkg.Client, kubeClient kubernetes.Interface,
...
+
+	LogMessage("Capturing Time In RFC 3339 Format  ...\n")
+	timeStructToWrite := Metadata{Time: timetoCaptureString}
+	vzJSON, err := json.MarshalIndent(timeStructToWrite, constants.JSONPrefix, constants.JSONIndent)
+
+	if err != nil {
+		LogError(fmt.Sprintf("An error occurred while creating JSON encoding of %s: %s\n", vzRes, err.Error()))
+		return err
+	}
+	_, err = f.WriteString(SanitizeString(string(vzJSON)))
```
fredtibbitts, line 157: This can be simplified by replacing `OpenFile` and `WriteString` with just `os.WriteFile`.

tools/vz/pkg/helpers/vzcapture_test.go
```diff
@@ -28,8 +23,12 @@ import (
 	fakedynamic "k8s.io/client-go/dynamic/fake"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	k8scheme "k8s.io/client-go/kubernetes/scheme"
+	"os"
```
fredtibbitts, line 26: Import formatting, these should go in the first grouping.

tools/vz/pkg/helpers/vzcapture_test.go
```diff
@@ -511,3 +516,54 @@ func TestRedactHostNamesForCertificates(t *testing.T) {
...
+	t.Log(captureDir)
+	defer cleanupTempDir(t, captureDir)
+	buf := new(bytes.Buffer)
+	errBuf := new(bytes.Buffer)
+	tempFile, err := os.CreateTemp(captureDir, "temporary-log-file-for-test")
+	assert.NoError(t, err)
+	SetMultiWriterOut(buf, tempFile)
+	rc := testhelpers.NewFakeRootCmdContext(genericclioptions.IOStreams{In: os.Stdin, Out: buf, ErrOut: errBuf})
+	err = captureNamespaces(client, "test-namespace", captureDir, rc)
+	assert.NoError(t, err)
```
fredtibbitts, line 549: Should this do more validation beyond checking for error? Maybe validate the contents of the captured file?

tools/vz/pkg/helpers/vzcapture_test.go
```diff
@@ -511,3 +516,54 @@ func TestRedactHostNamesForCertificates(t *testing.T) {
...
+	t.Log(captureDir)
+	defer cleanupTempDir(t, captureDir)
+	buf := new(bytes.Buffer)
+	errBuf := new(bytes.Buffer)
+	tempFile, err := os.CreateTemp(captureDir, "temporary-log-file-for-test")
+	assert.NoError(t, err)
+	SetMultiWriterOut(buf, tempFile)
+	SetMultiWriterErr(errBuf, tempFile)
+	err = CaptureMetadata(captureDir)
+	assert.NoError(t, err)
```
fredtibbitts, line 568: What about reading the file back and validating that the times match? Or is that already tested someplace else?

tools/vz/pkg/analysis/internal/util/files/files.go
```diff
@@ -148,3 +151,28 @@ func unmarshallFile(clusterPath string, object interface{}) error {
 
 	return nil
 }
+
+// GetTimeOfCapture parses the metadata.json file and converts the time string into a time.Time object to be used by other functions
+func GetTimeOfCapture(log *zap.SugaredLogger, clusterRoot string) (*time.Time, error) {
+	timeCaptureRegExp := regexp.MustCompile(constants.MetadataJSON)
+	timeCaptureFileList, err := GetMatchingFiles(log, clusterRoot, timeCaptureRegExp)
```
fredtibbitts, line 158: This is confusing, shouldn't there only be one metadata.json file? And if there really is more than one, why choose the first one in the list? It seems like there should only ever be one, but if there are multiple there should be comments here explaining why there might be multiple and why the first one is used.

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -126,9 +130,39 @@ func CaptureK8SResources(client clipkg.Client, kubeClient kubernetes.Interface,
 	if err := captureCertificates(client, namespace, captureDir, vzHelper); err != nil {
 		return err
 	}
+	if err := captureNamespaces(kubeClient, namespace, captureDir, vzHelper); err != nil {
+		return err
+	}
 	return nil
 }
 
+// captureTime gets the current time in UTC on the user's system and outputs it in ISO 8601 format to the user's system
```
fredtibbitts, line 139: Comment says 8601 but the code says 3339.

George-Aeillo: Ok! Thanks!

tools/vz/pkg/analysis/internal/util/files/files.go
George-Aeillo: There should be only be one metadata.json file. The reason that I select it from the first entry in the list is that the GetMatchingFiles helper function returns a list of filenames that match a particular format and I wanted to use the pre-existing helper functions when possible. I select the first entry in the list because it should be the only entry in the list. Do you think I should not use this helper function or that I should explicitly check that the list only has length of one?

tools/vz/pkg/analysis/internal/util/files/files.go
```diff
@@ -8,10 +8,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
```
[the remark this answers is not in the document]
George-Aeillo: Ok! Thanks!

tools/vz/pkg/helpers/vzcapture_test.go
George-Aeillo: Ok! I will add this functionality to the test!

tools/vz/pkg/helpers/vzcapture_test.go
George-Aeillo: Yes, I will add that functionality to the test.

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+// getNamespaceResource returns the namespace object that is in the namespace file
+func getNamespaceResource(log *zap.SugaredLogger, path string) (namespaceObject *corev1.Namespace, err error) {
+	namespaceResource := &corev1.Namespace{}
+	file, err := os.Open(path)
+	if err != nil {
+		log.Debug("file %s not found", path)
+		return nil, nil
```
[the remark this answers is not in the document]
George-Aeillo: Yes, this will occur when the cluster dump is corrupt, or it may be a prior dump that doesn't have the namespace resource captured, since this commit adds namespace resource capture to the bug-report output. If we were to throw an error here, this would make the analyze tool not compatible on past cluster dumps. This pattern seems to happen with other resources in the code as well, such as certificates. How do you suggest that I proceed further with this?

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	timeOfCapture, err := files.GetTimeOfCapture(log, clusterRoot)
+	if err != nil {
+		return err
+	}
+	if timeOfCapture == nil {
+		return nil
+	}
+	for _, namespace := range allNamespacesFound {
+		namespaceFile := files.FindFileInNamespace(clusterRoot, namespace, constants.NamespaceJSON)
```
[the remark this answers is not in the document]
George-Aeillo: This helper function call doesn't return an empty string or search for the file, but returns a string that would be the path to the resource in that namespace. If that file does not exist, in the next call, it will return nil, as it may be an old cluster dump that doesn't have that resource. Is this the intended behavior?

tools/vz/pkg/helpers/vzcapture.go
George-Aeillo: Ok! Thanks!

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
```diff
@@ -0,0 +1,111 @@
...
+	var listOfMessagesFromRelevantConditions = []string{}
+	if namespaceObject.Status.Phase != corev1.NamespaceTerminating {
+		return false, listOfMessagesFromRelevantConditions
+	}
+	if namespaceObject.DeletionTimestamp != nil {
+		timeOfCaptureUnixSeconds := timeOfCapture.Unix()
+		deletionTimestampUnixSeconds := namespaceObject.DeletionTimestamp.Unix()
+		timePassedInSecondsBetween := timeOfCaptureUnixSeconds - deletionTimestampUnixSeconds
+		minutesSpentDeleting := timePassedInSecondsBetween / 60
+		secondsRemainingSpentDeleting := timePassedInSecondsBetween % 60
```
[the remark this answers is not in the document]
George-Aeillo: Ok! Thanks!

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	var issueReporter = report.IssueReporter{
+		PendingIssues: make(map[string]report.Issue),
+	}
+	timeOfCapture, err := files.GetTimeOfCapture(log, clusterRoot)
+	if err != nil {
+		return err
+	}
+	if timeOfCapture == nil {
+		return nil
```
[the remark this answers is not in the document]
George-Aeillo: This metadata.json file is only being generated in this commit, so if analyze is being run on an old dump, I did not want it to return an error, if it did not have this resource. Do you think this is intended behavior? How do you suggest that I resolve this?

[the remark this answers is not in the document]
agv426: We could have 2 variants of the detected issue, when we have the capture time we can report with more confidence including the time range, and when we do not have the capture time we can report with less confidence and say that we did not have a time range

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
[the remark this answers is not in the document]
agv426: The FindFileInNamespace really is just doing an sprintf to form a string path, so I think this is ok, the getNamespaceResource will 

If it is confusing, we could change the name of the existing helper function as it really is not doing a "find" it is more just "forming" a file path there (there are some others there as well named similarly)

tools/vz/pkg/analysis/internal/util/files/files.go
agv426: If it is in the root of the cluster dump, you can use FindFileInClusterRoot

George-Aeillo: Ok! Thanks!

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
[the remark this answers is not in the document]
George-Aeillo: Ok!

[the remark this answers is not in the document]
George-Aeillo: Ok! Thanks! I added this

agv426 (dismissed):

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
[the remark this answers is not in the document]
fredtibbitts: Yeah the function name is misleading. Maybe just follow up with another PR that fixes some of the naming?

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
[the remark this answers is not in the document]
fredtibbitts: I don't like the fact that we're potentially masking real errors, but since the tool needs to work with older cluster dumps, I guess there's not much we can do, so ok to leave it as is.

fredtibbitts (approved):

tools/vz/pkg/analysis/internal/util/cluster/namespaces.go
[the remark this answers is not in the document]
George-Aeillo: Ok!

fredtibbitts (approved):

PR #7716, opened 2024-01-30
VZ-11671 Get previous pod logs for problematic pods

AbeHern:
Alerts after vz analyze to run bug-report to gather previous logs if problematic pods were found on the cluster

### DIFF
--- a/tools/vz/cmd/analyze/analyze.go
+++ b/tools/vz/cmd/analyze/analyze.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
-	"github.com/verrazzano/verrazzano/pkg/log"
 	cmdhelpers "github.com/verrazzano/verrazzano/tools/vz/cmd/helpers"
 	"github.com/verrazzano/verrazzano/tools/vz/pkg/analysis"
 	vzbugreport "github.com/verrazzano/verrazzano/tools/vz/pkg/bugreport"
@@ -93,7 +92,7 @@ func analyzeLiveCluster(cmd *cobra.Command, vzHelper helpers.VZHelper, directory
 	helpers.SetIsLiveCluster()
 
 	// Capture cluster snapshot
-	podLogs := vzbugreport.PodLogs{
+	podLogs := helpers.PodLogs{
 		IsPodLog: true,
 		Duration: int64(0),
 	}
@@ -122,15 +121,6 @@ func RunCmdAnalyze(cmd *cobra.Command, vzHelper helpers.VZHelper, printReportToC
 			if err := analyzeLiveCluster(cmd, vzHelper, validatedStruct.directory); err != nil {
 				return err
 			}
-
-			problematicPods, err := findProblematicPodFiles(validatedStruct.directory, cmd)
-			if err != nil {
-				return err
-			}
-			if len(problematicPods) != 0 {
-				fmt.Printf("Problematic pods found, run bug-report with --include-logs flag and --previous flag to capture pod logs")
-			}
-
 		} else {
 			//This is the case where only the tar string is specified
 			file, err := os.Open(validatedStruct.tarFile)
@@ -214,13 +204,4 @@ func getReportFormat(cmd *cobra.Command) string {
 		return constants.SummaryReport
 	}
 	return reportFormat.Value.String()
-}
-
-func findProblematicPodFiles(bugReportDir string, cmd *cobra.Command) (problematicPods []string, err error) {
-	logger := log.GetDebugEnabledLogger()
-	errorPodFiles, _ := analysis.FindProblematicPodFiles(logger, bugReportDir)
-	if len(errorPodFiles) == 0 {
-		return nil, fmt.Errorf("an error occurred while reading values for the flag --previous: %s", err.Error())
-	}
-	return errorPodFiles, nil
 }
--- a/tools/vz/cmd/analyze/analyze.go
+++ b/tools/vz/cmd/analyze/analyze.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
+	"github.com/verrazzano/verrazzano/pkg/log"
 	cmdhelpers "github.com/verrazzano/verrazzano/tools/vz/cmd/helpers"
 	"github.com/verrazzano/verrazzano/tools/vz/pkg/analysis"
 	vzbugreport "github.com/verrazzano/verrazzano/tools/vz/pkg/bugreport"
@@ -121,6 +122,15 @@ func RunCmdAnalyze(cmd *cobra.Command, vzHelper helpers.VZHelper, printReportToC
 			if err := analyzeLiveCluster(cmd, vzHelper, validatedStruct.directory); err != nil {
 				return err
 			}
+
+			problematicPods, err := findProblematicPodFiles(validatedStruct.directory, cmd)
+			if err != nil {
+				return err
+			}
+			if len(problematicPods) != 0 {
+				fmt.Printf("Problematic pods found, run bug-report with --include-logs flag and --previous flag to capture pod logs")
+			}
+
 		} else {
 			//This is the case where only the tar string is specified
 			file, err := os.Open(validatedStruct.tarFile)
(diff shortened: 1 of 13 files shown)


tools/vz/cmd/analyze/analyze.go
```diff
@@ -121,6 +122,15 @@ func RunCmdAnalyze(cmd *cobra.Command, vzHelper helpers.VZHelper, printReportToC
 			if err := analyzeLiveCluster(cmd, vzHelper, validatedStruct.directory); err != nil {
 				return err
 			}
+
+			problematicPods, err := findProblematicPodFiles(validatedStruct.directory, cmd)
+			if err != nil {
+				return err
+			}
+			if len(problematicPods) != 0 {
+				fmt.Printf("Problematic pods found, run bug-report with --include-logs flag and --previous flag to capture pod logs")
```
agv426, line 131: A few things:

- I didn't think we were adding a flag to control the previous behaviour, we wanted it to be automatic
- This appears to unconditionally output that message which may not match the scenario
-- They may be only running an analyze here, this may not be a bug-report command
-- Is it going to output this message even if they had already specified the --include-logs already?

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -1020,3 +1062,12 @@ func isCaExpired(client clipkg.Client, cert v1.Certificate, namespace string) (*
 	}
 	return &caCrtInfoForCert, true, nil
 }
+
+func FindProblematicPods(bugReportDir string) (namespaces []string, problematicPods []string, err error) {
+	logger := log.GetDebugEnabledLogger()
```
bhatpmk, line 1067: Please do not leave the debug log enabled.

AbeHern: Resolved now

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -361,6 +385,51 @@ func getClientWithVZWatch() client.WithWatch {
 	return fake.NewClientBuilder().WithScheme(pkghelper.NewScheme()).WithObjects(getVpoObjects()...).Build()
 }
 
+// getKubeClient returns a
```
agv426, line 388: nit incomplete sentence

AbeHern: fixed

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -1020,3 +1061,16 @@ func isCaExpired(client clipkg.Client, cert v1.Certificate, namespace string) (*
 	}
 	return &caCrtInfoForCert, true, nil
 }
+
+func FindProblematicPods(bugReportDir string) (namespaces []string, problematicPods []string, err error) {
+	namespaces, pods, err := FindProblematicPodFiles(bugReportDir)
+	if err != nil {
+		return nil, nil, fmt.Errorf("an error occurred while reading values for the flag --previous: %s", err.Error())
```
bhatpmk, line 1068: This is in helpers and a common function to find pods which are not in Running or Succeeded phase. Please update the error message by removing --previous from here.

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -97,7 +153,13 @@ func CaptureClusterSnapshot(kubeClient kubernetes.Interface, dynamicClient dynam
 		pkghelpers.LogError(fmt.Sprintf("There is an error with capturing the Verrazzano resources: %s", err.Error()))
 	}
 
-	// Capture logs from resources when the --include-logs flag is enabled
+	// set podLogs to capture previous pod logs
+	if len(problematicPodList) != 0 {
```
bhatpmk, line 157: Set podLogs.IsPrevious = true only for the pod having issue. In this case, when there is a pod having problems in namespace other than verrazzano-install and verrazzano-system, the CLI will set --previous to capture the logs for all the pods. So the command will create an empty previous-logs.txt for the pods which are in Running state as well. Also there will be a standard error indicating the previously terminated container not found for all such cases.

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -214,34 +275,24 @@ func captureAdditionalLogs(client clipkg.Client, kubeClient kubernetes.Interface
...
 // captureK8SResources captures Kubernetes workloads, pods, events, ingresses and services from the list of namespaces in parallel
-func captureK8SResources(wg *sync.WaitGroup, ec chan ErrorsChannel, client clipkg.Client, kubeClient kubernetes.Interface, dynamicClient dynamic.Interface, namespace, bugReportDir string, vzHelper pkghelpers.VZHelper) {
+func captureK8SResources(wg *sync.WaitGroup, ec chan pkghelpers.ErrorsChannel, client clipkg.Client, kubeClient kubernetes.Interface, dynamicClient dynamic.Interface, namespace, bugReportDir string, vzHelper pkghelpers.VZHelper) {
 	defer wg.Done()
 	if err := pkghelpers.CaptureK8SResources(client, kubeClient, dynamicClient, namespace, bugReportDir, vzHelper); err != nil {
-		ec <- ErrorsChannel{ErrorMessage: err.Error()}
+		ec <- pkghelpers.ErrorsChannel{ErrorMessage: err.Error()}
+	}
+	// find problematic pods from captured resources
+	problematicPodsNamespaces, pods, _ := pkghelpers.FindProblematicPods(bugReportDir)
```
bhatpmk, line 293: The function captureK8SResources is called for all the namespaces in a loop, and the change here attempts to find the problematic pods in the bugReportDir repeatedly for all the namespaces. Evaluating the pods which are not running should be one time activity.

tools/vz/pkg/helpers/podAnalysis.go
```diff
@@ -0,0 +1,129 @@
...
+	"sync"
+)
+
+// podListMap holds podLists which have been read in already
+var podListMap = make(map[string]*corev1.PodList)
+var podCacheMutex = &sync.Mutex{}
+
+// FindProblematicPodFiles looks at the pods.json files in the cluster and will give a list of files
+// if any have pods that are not Running or Succeeded.
+func FindProblematicPodFiles(clusterRoot string) (namespaces []string, podFiles []string, err error) {
```
bhatpmk, line 23: Most of the change is already there in pods.go, although it is internal right now.

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -178,6 +180,29 @@ func TestBugReportSuccess(t *testing.T) {
...
+	setUpGlobalFlags(cmd)
+	err := cmd.PersistentFlags().Set(constants.BugReportFileFlagName, bugRepFile)
+	assert.NoError(t, err)
+	err = cmd.PersistentFlags().Set(constants.VerboseFlag, "true")
+	assert.NoError(t, err)
+	err = cmd.Execute()
+	if err != nil {
+		assert.Error(t, err)
+	}
+	assert.NoError(t, err)
```
bhatpmk, line 203: The test is not validating what intends to do.

tools/vz/pkg/bugreport/reportgen.go
AbeHern: Yes you are correct, I've moved this call out and AFTER resources are captured

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -1062,15 +1066,13 @@ func isCaExpired(client clipkg.Client, cert v1.Certificate, namespace string) (*
 	return &caCrtInfoForCert, true, nil
 }
 
-func FindProblematicPods(bugReportDir string) (namespaces []string, problematicPods []string, err error) {
-	namespaces, pods, err := FindProblematicPodFiles(bugReportDir)
+func FindProblematicPods(bugReportDir string) (problematicPodNamespaces map[string][]corev1.Pod, err error) {
+	namespaces, err := FindProblematicPodFiles(bugReportDir)
```
bhatpmk, line 1070: This function and FindProblematicPodFiles in podAnalysis.go construct namespaces, better to name the functions accordingly.

Same holds true for the error message in line 1072

tools/vz/pkg/helpers/vzcapture.go
```diff
@@ -581,7 +581,11 @@ func CapturePodLog(kubeClient kubernetes.Interface, pod corev1.Pod, namespace, c
 			podLogOptions.InsecureSkipTLSVerifyBackend = true
 			podLog, err := kubeClient.CoreV1().Pods(namespace).GetLogs(podName, &podLogOptions).Stream(context.TODO())
 			if err != nil {
-				LogError(fmt.Sprintf("An error occurred while reading the logs from pod %s: %s\n", podName, err.Error()))
+				// ignore error if previous container could not be found
+				if !strings.Contains(err.Error(), "(BadRequest)") {
```
bhatpmk, line 585: An alternative is to check the standard error, by skipping the container and pod
For example, a sample error will be of this format -
Error from server (BadRequest): previous terminated container "<container name>" in pod "<pod name>" not found

tools/vz/cmd/bugreport/bugreport_test.go
bhatpmk: Please work on the test case in a follow-up change

bhatpmk (approved):
Please update the test case in a follow-up change.

PR #7751, opened 2024-02-13
vz-11696 flag missing sidecar containers in istio-injection namespaces

AbeHern:
Iterates through pod's containers in istio-injection namespaces and flags pods that are missing sidecar containers

### DIFF
--- a/tools/vz/cmd/bugreport/bugreport_test.go
+++ b/tools/vz/cmd/bugreport/bugreport_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/verrazzano/verrazzano/tools/vz/pkg/constants"
 	pkghelper "github.com/verrazzano/verrazzano/tools/vz/pkg/helpers"
 	"github.com/verrazzano/verrazzano/tools/vz/test/helpers"
+	"io/ioutil"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -416,6 +417,63 @@ func TestBugReportFailureUsingInvalidClient(t *testing.T) {
 	assert.FileExists(t, bugRepFile)
 }
 
+// TestIstioSidecarContainersExist
+// GIVEN a CLI bug-report command
+// WHEN I call cmd.Execute
+// THEN expect the command to verify the existence of istio sidecar containers in pods in istio-injection labeled namespaces
+func TestIstioSidecarContainersExist(t *testing.T) {
+	tests := []struct {
+		name    string
+		success bool
+	}{
+		{"Sidecar Containers found", true},
+		{"Sidecar Containers NOT found", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := getClientWithVZWatch()
+			if !tt.success {
+				c = getClientWithMissingSidecar()
+			}
+
+			var kubeClient *k8sfake.Clientset
+
+			// Verify the vz resource is as expected
+			vz := v1beta1.Verrazzano{}
+			err := c.Get(context.TODO(), types.NamespacedName{Namespace: "default", Name: "verrazzano"}, &vz)
+			assert.NoError(t, err)
+			stdoutFile, stderrFile := createStdTempFiles(t)
+
+			rc := setupFakeDynamicClient(c, genericclioptions.IOStreams{In: os.Stdin, Out: stdoutFile, ErrOut: stderrFile})
+			kubeClient = getKubeClient()
+			rc.SetClient(c)
+			rc.SetKubeClient(kubeClient)
+			cmd := NewCmdBugReport(rc)
+			assert.NotNil(t, cmd)
+
+			tmpDir, _ := os.MkdirTemp("", "bug-report")
+
+			bugRepFile := tmpDir + string(os.PathSeparator) + "bug-report.tgz"
+			setUpGlobalFlags(cmd)
+			err = cmd.PersistentFlags().Set(constants.BugReportFileFlagName, bugRepFile)
+			assert.NoError(t, err)
+			err = cmd.PersistentFlags().Set(constants.VerboseFlag, "true")
+			assert.NoError(t, err)
+			err = cmd.Execute()
+
+			if !tt.success {
+				stderrFileData, _ := ioutil.ReadFile(stderrFile.Name())
+				temp := string(stderrFileData)
+				err := strings.Contains(temp, "was not found for pod:")
+				assert.True(t, err)
+			}
+			defer os.Remove(stdoutFile.Name())
+			defer os.Remove(stderrFile.Name())
+			defer cleanupTempDir(t, tmpDir)
+		})
+	}
+}
+
 // getClientWithWatch returns a client containing all VPO objects
 func getClientWithWatch() client.WithWatch {
 	return fake.NewClientBuilder().WithScheme(pkghelper.NewScheme()).WithObjects(getVpoObjects()[1:]...).Build()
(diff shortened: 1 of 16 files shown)


tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -434,6 +443,55 @@ func captureMultiClusterResources(dynamicClient dynamic.Interface, captureDir st
...
 
+// flagMissingSidecarContainers identifies pods in namespaces with --label istio-injection=enabled that are missing sidecar containers
+func flagMissingSidecarContainers(client clipkg.Client, kubeClient kubernetes.Interface, namespaces []string) (error error) {
+	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{vzconstants.LabelIstioInjection: "enabled"}}
+	listOptions := metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()}
+	namespaceList, err := kubeClient.CoreV1().Namespaces().List(context.TODO(), listOptions)
+	if err != nil {
+		return err
+	}
+	if len(namespaceList.Items) != 0 {
```
fredtibbitts, line 454: This `if` check is not needed, `range` works on nil slices.

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -434,6 +443,55 @@ func captureMultiClusterResources(dynamicClient dynamic.Interface, captureDir st
...
+		for _, namespace := range namespaceList.Items {
+			//fmt.Printf("Namespace %s, has matching label %s\n", namespace.Name, vzconstants.LabelIstioInjection)
```
fredtibbitts, line 456: Remove commented out code.

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -104,6 +105,11 @@ var DefaultPodLog = pkghelpers.PodLogs{
 	Duration:   0,
 }
 
+const (
+	istioSidecarStatus  = "sidecar.istio.io/status"
+	istioInjectionLabel = "sidecar.istio.io/inject=false"
```
fredtibbitts, line 110: Is this used anywhere?

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -434,6 +443,55 @@ func captureMultiClusterResources(dynamicClient dynamic.Interface, captureDir st
...
+			err := json.Unmarshal([]byte(sidecarContainers), &obj)
+			if err != nil {
+				return err
+			}
+			sidecarContainersFromAnnotation := obj["containers"]
+			containersFromPod := pod.Spec.Containers
+
+			for _, sidecar := range sidecarContainersFromAnnotation.([]interface{}) {
+				for i, container := range containersFromPod {
+					if i+1 == len(containersFromPod) && sidecar != container.Name {
```
fredtibbitts, line 485: Is this assuming that the Istio sidecar is always the last container in the pod? I'm not sure that's a valid assumption, but if it is, why iterate over `containersFromPod` if you only care about the last one?

fredtibbitts (commented):
One general comment, there are other ways to inject sidecars, besides labeling the namespace. For example, even if the namespace isn't labeled (or explicitly disables sidecar injection), that can be overridden on a per-pod basis with annotations. This does not handle those cases. Is that what you expect, or does this need to be extended to account for other injection mechanisms?

tools/vz/pkg/bugreport/reportgen.go
```diff
@@ -434,6 +443,55 @@ func captureMultiClusterResources(dynamicClient dynamic.Interface, captureDir st
...
+			}
+		}
+	}
+	return nil
+}
+
+// findMissingSidecarContainers identifies pods that are missing sidecar containers
+func findMissingSidecarContainers(pods []corev1.Pod) (error error) {
+	for _, pod := range pods {
+		if pod.Annotations[istioSidecarStatus] != "" {
```
zabdulre, line 473: Nesting can be avoided here by doing

`if pod.Annotations[istioSidecarStatus == ""{
      continue
}`

AbeHern: Good idea, went ahead and changed it

tools/vz/pkg/bugreport/reportgen.go
AbeHern: istioInjectionLabel, is left over, it should be removed. But istioSidecarStatus I did use. Fixed.

tools/vz/pkg/bugreport/reportgen.go
AbeHern: Done

tools/vz/pkg/bugreport/reportgen.go
AbeHern: Fixed

tools/vz/pkg/bugreport/reportgen.go
AbeHern: This is a mistake. There a condition missing. It should be something along the lines of `if we've reached the end of the array and the value is still not found then output the error`

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -416,13 +417,75 @@ func TestBugReportFailureUsingInvalidClient(t *testing.T) {
...
+			bugRepFile := tmpDir + string(os.PathSeparator) + "bug-report.tgz"
+			setUpGlobalFlags(cmd)
+			err = cmd.PersistentFlags().Set(constants.BugReportFileFlagName, bugRepFile)
+			assert.NoError(t, err)
+			err = cmd.PersistentFlags().Set(constants.VerboseFlag, "true")
+			assert.NoError(t, err)
+			err = cmd.Execute()
+
+			if !tt.success {
+				stderrFileData, _ := ioutil.ReadFile(stderrFile.Name())
```
marcogelle, line 464: Replace this with `os.ReadFile`, since `ioutil.ReadFile` is deprecated: https://pkg.go.dev/io/ioutil#ReadFile.

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -416,13 +417,75 @@ func TestBugReportFailureUsingInvalidClient(t *testing.T) {
...
+				temp := string(stderrFileData)
+				err := strings.Contains(temp, "was not found for pod:")
+				assert.True(t, err)
+			}
+			assert.NoError(t, err)
+			defer os.Remove(stdoutFile.Name())
+			defer os.Remove(stderrFile.Name())
+			defer cleanupTempDir(t, tmpDir)
```
marcogelle, line 472: Better to move these `defer` statements as early in the code as possible, i.e. right after file/directory creation.

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -416,13 +416,76 @@ func TestBugReportFailureUsingInvalidClient(t *testing.T) {
...
 func getClientWithWatch() client.WithWatch {
 	return fake.NewClientBuilder().WithScheme(pkghelper.NewScheme()).WithObjects(getVpoObjects()[1:]...).Build()
 }
 
 // getClientWithVZWatch returns a client containing all VPO objects and the Verrazzano CR
 func getClientWithVZWatch() client.WithWatch {
+	return fake.NewClientBuilder().WithScheme(pkghelper.NewScheme()).WithObjects(getVpoObjects()[:len(getVpoObjects())-1]...).Build()
+}
+
+// getClientWithMissingSidecar returns a pod with a missing sidecar container
```
marcogelle, line 487: Revise comment to say something like `returns a client containing all VPO object, the Verrazzano CR, and a pod with a missing sidecar container`, since it doesn't only return a pod

Abhijithang:
> One general comment, there are other ways to inject sidecars, besides labeling the namespace. For example, even if the namespace isn't labeled (or explicitly disables sidecar injection), that can be overridden on a per-pod basis with annotations. This does not handle those cases. Is that what you expect, or does this need to be extended to account for other injection mechanisms?
I think expected outcome for the task is to check the pods specifically in labeled namespaces. But  I agree with you, it is necessary to include the above mentioned cases if not in this task but creating a new subsequent one.

Abhijithang:
> One general comment, there are other ways to inject sidecars, besides labeling the namespace. For example, even if the namespace isn't labeled (or explicitly disables sidecar injection), that can be overridden on a per-pod basis with annotations. This does not handle those cases. Is that what you expect, or does this need to be extended to account for other injection mechanisms?

I think expected outcome for the task is to check the pods specifically in labeled namespaces. But  I agree with you, it is necessary to include the above mentioned cases if not in this task but creating a new subsequent one.

Abhijithang (approved):
Approved. Please follow up with Fred for the changes mentioned in the comment. Thanks.

PR #7752, opened 2024-02-14
vz-11752 followup task to create testBugReportFailedPods scenario

AbeHern:
Expands unit test to check if file created while capturing previous pod logs is actually captured

### DIFF
--- a/tools/vz/cmd/bugreport/bugreport_test.go
+++ b/tools/vz/cmd/bugreport/bugreport_test.go
@@ -182,7 +182,7 @@ func TestBugReportSuccess(t *testing.T) {
 // TestBugReportGetPreviousLogs
 // GIVEN a CLI bug-report command with no flags, but failed pods on the cluster
 // WHEN I call cmd.Execute
-// THEN expect the command to try and capture previous pod logs if possible and write them out to 'previous-logs.txt'
+// THEN expect the command to try and capture previous pod logs create the file 'previous-logs.txt'
 func TestBugReportGetPreviousLogs(t *testing.T) {
 	cmd := setUpAndVerifyResources(t)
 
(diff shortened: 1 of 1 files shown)


tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -180,27 +179,36 @@ func TestBugReportSuccess(t *testing.T) {
...
+	previousLogTxt := "previous-logs.txt"
+	pkghelper.UntarArchive(tmpDir, file)
+	stat, err := os.Stat(tmpDir + istioSystemPath + "istio/" + previousLogTxt)
+	assert.NoError(t, err)
+	assert.NotNil(t, stat.Name())
+	stat, err = os.Stat(tmpDir + istioSystemPath + "third-istio-pod/" + previousLogTxt)
+	assert.NoError(t, err)
+	assert.NotNil(t, stat.Name())
+
+	defer file.Close()
```
agv426, line 210: Why are these defers not being done directly after the resources they are intended to cleanup?

AbeHern: This is a mistake. Fixing it now

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -197,10 +196,19 @@ func TestBugReportFailedPods(t *testing.T) {
...
 	assert.NoError(t, err)
 	err = cmd.Execute()
-	if err != nil {
-		assert.Error(t, err)
-	}
 	assert.NoError(t, err)
+	file, _ := os.Open(bugRepFile)
+	defer file.Close()
+
+	istioSystemPath := "/cluster-snapshot/istio-system/"
```
marcogelle, line 203: A little nitpicky, but maybe it'd be better to use `filepath.Join` to form file paths instead of spelling it out with slashes. Applies to the below `os.Stat` arguments too.

tools/vz/cmd/bugreport/bugreport_test.go
```diff
@@ -197,10 +196,19 @@ func TestBugReportFailedPods(t *testing.T) {
...
+	previousLogTxt := "previous-logs.txt"
+	pkghelper.UntarArchive(tmpDir, file)
+	stat, err := os.Stat(tmpDir + istioSystemPath + "istio/" + previousLogTxt)
+	assert.NoError(t, err)
+	assert.NotNil(t, stat.Name())
+	stat, err = os.Stat(tmpDir + istioSystemPath + "third-istio-pod/" + previousLogTxt)
```
marcogelle, line 209: So right now the test just checks for the existence of the `previous-logs.txt` files, but the comment suggests that the test checks that the file is written to as well.
I think we'd either want to revise the comment or add that to this test, depending on what we want this follow-up to accomplish.

AbeHern: Fair point. Went ahead and updated the description.

tools/vz/cmd/bugreport/bugreport_test.go
AbeHern: Made the fix.

bhatpmk (approved):

PR #7760, opened 2024-02-21
VZ-11703: Refactor CLI Test Code

George-Aeillo:
In this PR, I refactored the test code for the CLI to make it more standardized, switched from buffers to files, and fixed some race conditions in the CLI


tools/vz/cmd/analyze/analyze_test.go
```diff
@@ -26,12 +25,9 @@ const istioIPErr = "Verrazzano install failed as no IP found for service istio-i
...
 // THEN expect the command to do the analysis and generate report file into tmp dir
 func TestAnalyzeDefaultFromReadOnlyDir(t *testing.T) {
-	stdoutFile, stderrFile := createStdTempFiles(t)
-	defer func() {
-		os.Remove(stdoutFile.Name())
-		os.Remove(stderrFile.Name())
-	}()
-	rc := helpers.NewFakeRootCmdContext(genericclioptions.IOStreams{In: os.Stdin, Out: stdoutFile, ErrOut: stderrFile})
+	rc, err := helpers.NewFakeRootCmdContextWithFiles()
+	assert.Nil(t, err)
```
agv426, line 29: should the defer be done before the assert here

tools/vz/pkg/helpers/vzsanitize.go
```diff
@@ -60,9 +60,11 @@ func InitRegexToReplacementMap() {
 // The redactedValuesOverride parameter can be used to override the default redactedValues map for keeping track of
 // redacted strings.
 func SanitizeString(l string, redactedValuesOverride map[string]string) string {
+	redactedValuesMutex.Lock()
 	if len(regexToReplacementList) == 0 {
 		InitRegexToReplacementMap()
 	}
+	redactedValuesMutex.Unlock()
```
marcogelle, line 67: Not a critique, more of a question: What does the locking here do? Were you seeing something like concurrent read and write errors otherwise?

tools/vz/cmd/analyze/analyze_test.go
```diff
@@ -53,22 +49,18 @@ func TestAnalyzeDefaultFromReadOnlyDir(t *testing.T) {
...
 func TestAnalyzeCommandDetailedReport(t *testing.T) {
-	stdoutFile, stderrFile := createStdTempFiles(t)
-	defer func() {
-		os.Remove(stdoutFile.Name())
-		os.Remove(stderrFile.Name())
-	}()
-	rc := helpers.NewFakeRootCmdContext(genericclioptions.IOStreams{In: os.Stdin, Out: stdoutFile, ErrOut: stderrFile})
+	rc, err := helpers.NewFakeRootCmdContextWithFiles()
+	defer helpers.CleanUpNewFakeRootCmdContextWithFiles(rc)
+	assert.Nil(t, err)
```
agv426, line 54: ie: example spot where defer was first. Looks like more spots have the assert first

tools/vz/cmd/analyze/analyze_test.go
agv426: Or if we can pass "t" into the helper there and have it assert when it fails rather than returning an error at all. Then we can get rid of all of the "assert.Nil(t, err)"'s

tools/vz/pkg/helpers/vzsanitize.go
agv426: looking at it, the conditional is not atomic, so multiple threads could simultaneously see an empty list there 
and initialize the map. I'm assuming that is what the race checker flagged.

marcogelle: Thanks for the explanation.

tools/vz/cmd/analyze/analyze_test.go
George-Aeillo: I think that passing "t" into the helper would work to further is a great idea! I will work on implementing that!

agv426 (approved):

PR #7770, opened 2024-02-23
vz-11738 Display analyze report on autobugreport

AbeHern:
Makes the analysis prompt displayed after autobugreport is taken


tools/vz/cmd/bugreport/bugreport.go
```diff
@@ -225,7 +225,7 @@ func runCmdBugReport(cmd *cobra.Command, args []string, vzHelper helpers.VZHelpe
 		if err != nil {
 			return bugRepFile.Name(), err
 		}
-		analyzeErr := analyze.RunCmdAnalyze(newCmd, vzHelper, false)
+		analyzeErr := analyze.RunCmdAnalyze(newCmd, vzHelper, true)
```
marcogelle, line 228: I'll approve because this seems to solve the issue.

However, from a code style view, it seems that `analyze.RunCmdAnalyze` now always is called with the `printReportToConsole` parameter set to `true`. Would it be better to remove this parameter completely, since it is effectively not used?

marcogelle (dismissed):

marcogelle (approved):

PR #7835, opened 2024-07-29
Update release-1.6 release notes

pfmackin:
Update release-1.6 release notes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### 1.6.12
+Fixes:
+
+- Fixed an issue where console certificates were being recreated approximately every 10-18 hours. 
+  Now they only get recreated shortly after certificate.Status.renewalTime.
+
 ### 1.6.11
 Component version updates:
 


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+### 1.6.12
+Fixes:
+
+- Fixed an issue where console certificates were being recreated approximately every 10-18 hours. 
+  Now they only get recreated shortly after certificate.Status.renewalTime.
```
George-Aeillo, line 5: Do we want references to Go structs/fields in release notes?

pfmackin: that is a CR reference and it is a public API so it is ok.

George-Aeillo (approved):

CHANGES.md
pfmackin:
```
 k explain cert.status
GROUP:      cert-manager.io
KIND:       Certificate
VERSION:    v1

FIELD: status <Object>

DESCRIPTION:
    Status of the Certificate. This is set and managed automatically.
    
FIELDS:
  conditions	<[]Object>
    List of status conditions to indicate the status of certificates. Known
    condition types are `Ready` and `Issuing`.

...

  renewalTime	<string>
    RenewalTime is the time at which the certificate will be next renewed. If
    not set, no upcoming renewal is scheduled.
```

